PR #17, opened 2019-04-26
Reuse ClientConn in ipsec

tpantelis:
There are several funtions that obtain a ClientConn via getClient
locally which leads to multiple ClientConn instances created and
closed in a call chain. Optimize by creating a single ClientConn and
passing to the lower-level functions.


.gitignore
```diff
@@ -2,3 +2,5 @@
 submariner-engine
 .dapper
 bin
+.project
```
skitt: Perhaps this should go in a separate patch (dedicated to ignoring Eclipse files). You can also configure git globally to ignore files specific to your IDE ;-).

tpantelis:
I removed the changes to .gitignore - will push another patch/PR for that.

Oats87:
Thank you for the PR. We will review this and get this merged ASAP.

mangelajo:
Thank you @tpantelis the changes make sense, and it compiles, I'm merging it.

PR #29, opened 2019-05-09
Improve error handling/reporting in route.go

tpantelis:
- There were several call sites that did not check the error return value.
    Either propagate the error or at least log it.

  - The cleanXfrmPolicies and cleanRoutes functions return error but the callers do not
    check the error. It seems an error is not catastrophic to the callers so
    they do not need to handle it. Also each function has a loop in which an error could
    occur but the current behavior ignores it and continues. If we return the error then
    subsequent iterations of the loop would be elided. It seems it's safest to keep the
    current behavior but log the error. Also there doesn't seem to be any reason to
    return an error so that was removed

  - Added format args to some log messages to provide more context

  - Modified the wording in some log messages for consistency

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -132,15 +132,14 @@ func (r *RouteController) processNextCluster() bool {
 	err := func() error {
 		defer r.clusterWorkqueue.Done(obj)
 		klog.V(4).Infof("Processing cluster object: %v", obj)
-		ns, key, err := cache.SplitMetaNamespaceKey(obj.(string))
+		key := obj.(string)
+		ns, name, err := cache.SplitMetaNamespaceKey(key)
 		if err != nil {
-			klog.Errorf("error while splitting meta namespace key: %v", err)
-			return nil
+			return fmt.Errorf("Error while splitting meta namespace key %s: %v", key, err)
 		}
-		cluster, err := r.submarinerClientSet.SubmarinerV1().Clusters(ns).Get(key, metav1.GetOptions{})
+		cluster, err := r.submarinerClientSet.SubmarinerV1().Clusters(ns).Get(name, metav1.GetOptions{})
 		if err != nil {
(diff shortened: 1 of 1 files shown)


tpantelis:
No - I don't have the means yet to test it either.

pkg/routeagent/controllers/route/route.go
```diff
@@ -256,14 +253,12 @@ func (r *RouteController) handleRemovedEndpoint(obj interface{}) {
 	if object, ok = obj.(*v1.Endpoint); !ok {
 		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
 		if !ok {
-			utilruntime.HandleError(fmt.Errorf("error decoding object, invalid type"))
-			klog.V(4).Infof("problem decoding object")
+			klog.Errorf("Error decoding object %v, invalid type", obj)
```
mpeterson: To be in line with #28 

```suggestion
			klog.Errorf("Couldn't get object from tombstone %#v", obj)
```

pkg/routeagent/controllers/route/route.go
```diff
@@ -256,14 +253,12 @@ func (r *RouteController) handleRemovedEndpoint(obj interface{}) {
...
 			return
 		}
 		object, ok = tombstone.Obj.(*v1.Endpoint)
 		if !ok {
-			utilruntime.HandleError(fmt.Errorf("error decoding object tombstone, invalid type"))
-			klog.V(4).Infof("problem decoding object tombstone")
+			klog.Errorf("Error decoding object tombstone %v, invalid type", tombstone.Obj)
```
mpeterson: To be in line with #28 
```suggestion
			klog.Errorf("Could not convert object tombstone %v to an Endpoint", tombstone.Obj)
```

pkg/routeagent/controllers/route/route.go
tpantelis: yeah - we should be consistent. I'll change it.

mpeterson: good, thanks!

PR #30, opened 2019-05-14
Modifications to main.go

tpantelis:
- There was one call site that did not check the error return value.

- Ensure Fatalf call sites properly log the error.

- Modified util.GetLocalEndpoint to propagate errors to let main handle it.

- Changed some local var names to improved readability (eg ds -> databroker).

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,8 +52,8 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	var ss types.SubmarinerSpecification
-	err := envconfig.Process("submariner", &ss)
+	var submSpec types.SubmarinerSpecification
+	err := envconfig.Process("submariner", &submSpec)
 	if err != nil {
 		klog.Fatal(err)
 	}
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -73,85 +73,94 @@ func main() {
...
+	submarinerInformerFactory := submarinerInformers.NewSharedInformerFactoryWithOptions(submarinerClient, time.Second*30,
+		submarinerInformers.WithNamespace(submSpec.Namespace))
+
 	start := func(context.Context) {
-		var ce cableengine.CableEngine
-		localCluster, err := util.GetLocalCluster(ss)
+		localCluster, err := util.GetLocalCluster(submSpec)
 		if err != nil {
-			klog.Fatalf("Fatal error occurred while retrieving local cluster %v", localCluster)
+			klog.Fatalf("Fatal error occurred while retrieving local cluster %v", err)
```
skitt: Wouldn’t it be better to log both `localCluster` and `err`? How about using [fields](https://github.com/sirupsen/logrus#fields)?

main.go
```diff
@@ -73,85 +73,94 @@ func main() {
...
 		}
-		localEndpoint, err := util.GetLocalEndpoint(ss.ClusterId, "ipsec", nil, ss.NatEnabled, append(ss.ServiceCidr, ss.ClusterCidr...))
+		localEndpoint, err := util.GetLocalEndpoint(submSpec.ClusterId, "ipsec", nil, submSpec.NatEnabled,
+			append(submSpec.ServiceCidr, submSpec.ClusterCidr...))
 		if err != nil {
-			klog.Fatalf("Fatal error occurred while retrieving local endpoint %v", localCluster)
+			klog.Fatalf("Fatal error occurred while retrieving local endpoint %v", err)
```
skitt: Same here ;-).

pkg/util/util.go
```diff
@@ -100,7 +101,7 @@ func GetLocalEndpoint(clusterID string, backend string, backendConfig map[string
 	if natEnabled {
 		publicIP, err := ipify.GetIp()
 		if err != nil {
-			klog.Fatalf("could not determine public IP: %v", err)
+			return types.SubmarinerEndpoint{}, fmt.Errorf("Could not determine public I: %v", err)
```
skitt: Missing “P”?

main.go
mangelajo: I'd have a more broader discussion to use logrus or any other logging library (which of course may make sense IMO) with everybody before pulling that dep :)

Regardless of that, I'd log both :)

main.go
mangelajo: Same too :D

main.go
skitt: > I'd have a more broader discussion to use logrus or any other logging library (which of course may make sense IMO) with everybody before pulling that dep :)

Oh dear, I’m so used to having `logrus` that I didn’t check what Submariner is using...

mangelajo: Hehehe :-D I suspected, yes

tpantelis: If err is non-nil then localCluster is invalid, ie probably null

skitt: Oh, yes, duh; `submSpec` then...

pkg/util/util.go
tpantelis: yup :)

main.go
```diff
@@ -73,85 +73,94 @@ func main() {
...
+	submarinerInformerFactory := submarinerInformers.NewSharedInformerFactoryWithOptions(submarinerClient, time.Second*30,
+		submarinerInformers.WithNamespace(submSpec.Namespace))
+
 	start := func(context.Context) {
-		var ce cableengine.CableEngine
-		localCluster, err := util.GetLocalCluster(ss)
+		localCluster, err := util.GetLocalCluster(submSpec)
 		if err != nil {
-			klog.Fatalf("Fatal error occurred while retrieving local cluster %v", localCluster)
+			klog.Fatalf("Fatal error occurred while retrieving local cluster from %v: %v", submSpec, err)
```
mpeterson: Would not be better a %#v for submSpec?

main.go
```diff
@@ -73,85 +73,94 @@ func main() {
...
 		}
-		localEndpoint, err := util.GetLocalEndpoint(ss.ClusterId, "ipsec", nil, ss.NatEnabled, append(ss.ServiceCidr, ss.ClusterCidr...))
+		localEndpoint, err := util.GetLocalEndpoint(submSpec.ClusterId, "ipsec", nil, submSpec.NatEnabled,
+			append(submSpec.ServiceCidr, submSpec.ClusterCidr...))
 		if err != nil {
-			klog.Fatalf("Fatal error occurred while retrieving local endpoint %v", localCluster)
+			klog.Fatalf("Fatal error occurred while retrieving local endpoint from %v: %v", submSpec, err)
```
mpeterson: Would not be better a %#v for submSpec?

main.go
```diff
@@ -73,85 +73,94 @@ func main() {
...
-		dsSyncer := datastoresyncer.NewDatastoreSyncer(ss.ClusterId, ss.Namespace, kubeClient, submarinerClient, submarinerInformerFactory.Submariner().V1().Clusters(), submarinerInformerFactory.Submariner().V1().Endpoints(), ds, ss.ColorCodes, localCluster, localEndpoint)
+		dsSyncer := datastoresyncer.NewDatastoreSyncer(submSpec.ClusterId, submSpec.Namespace, kubeClient, submarinerClient,
+			submarinerInformerFactory.Submariner().V1().Clusters(), submarinerInformerFactory.Submariner().V1().Endpoints(), datastore,
+			submSpec.ColorCodes, localCluster, localEndpoint)
 
 		kubeInformerFactory.Start(stopCh)
 		submarinerInformerFactory.Start(stopCh)
 
 		klog.V(4).Infof("Starting controllers")
+
```
mangelajo: nit: unrelated line change. I know it's a nit, but I try to avoid those, now, not very important, but in the future when we consider backports/reverts it can become messy.

main.go
```diff
@@ -73,85 +73,94 @@ func main() {
...
 	startLeaderElection(leClient, recorder, start)
 	klog.Fatal("All controllers stopped or exited. Stopping main loop")
 }
 
 func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder record.EventRecorder, run func(ctx context.Context)) {
 	id, err := os.Hostname()
 	if err != nil {
-		klog.Fatalf("error getting hostname: %s", err.Error())
+		klog.Fatalf("error getting hostname: %v", err)
```
mangelajo: The initial version seemed fine, why the change here?

pkg/util/util.go
```diff
@@ -79,10 +79,11 @@ func GetLocalCluster(ss types.SubmarinerSpecification) (types.SubmarinerCluster,
 	localCluster.Spec.ColorCodes = ss.ColorCodes
 	return localCluster, nil
 }
+
```
mangelajo: nit: unrelated change

pkg/util/util.go
```diff
@@ -79,10 +79,11 @@ func GetLocalCluster(ss types.SubmarinerSpecification) (types.SubmarinerCluster,
...
 func GetLocalEndpoint(clusterID string, backend string, backendConfig map[string]string, natEnabled bool, subnets []string) (types.SubmarinerEndpoint, error) {
 	hostname, err := os.Hostname()
 	if err != nil {
-		klog.Fatalf("error getting hostname: %s", err.Error())
+		return types.SubmarinerEndpoint{}, fmt.Errorf("Error getting hostname: %v", err)
```
mangelajo: Another option could be returning a pointer to SubmarinerEndpoint, that let's you return nil in case of error.

pkg/util/util.go
```diff
@@ -100,7 +101,7 @@ func GetLocalEndpoint(clusterID string, backend string, backendConfig map[string
 	if natEnabled {
 		publicIP, err := ipify.GetIp()
 		if err != nil {
-			klog.Fatalf("could not determine public IP: %v", err)
+			return types.SubmarinerEndpoint{}, fmt.Errorf("Could not determine public IP: %v", err)
```
mangelajo: ditto.

main.go
tpantelis: I'm not exactly sure what's better - still learning Go. FWIU, Error() returns the string representation, %v prints the default format for whatever type it is. Elsewhere in the code base, it uses %v for error so I changed this too be consistent.

pkg/util/util.go
tpantelis: trying to improve readability - I like functions separated by new lines. I like what someone told me way back - "whitespace is your friend" :)

main.go
tpantelis: yeah it seems so - I wasn't aware of # - still learning Go ...

skitt: https://golang.org/pkg/fmt/ shows there are a number of interesting formatting directives in Go ;-).

tpantelis: yup - studying that ...

pkg/util/util.go
tpantelis: yeah that crossed my mind but would require changes elsewhere as well. I can look into that in a separate patch.

main.go
mangelajo: ack :)

PR #31, opened 2019-05-19
Add support for travisci and golangci-lint

dimaunx:
Add basic support for travisci and golangci-lint.


tpantelis:
The changes for travisci and golangci-lint are unrelated and probably technically should be in separate patches ...

dimaunx:
Totally agree, i will prepare separate PRs for both of the changes.

PR #35, opened 2019-05-27
Check for errors before deferring body closure

skitt:
On results from http, the body should be closed only if there was no
error. This patch moves all deferred closures after the error checks,
and adds a missing error check.

### DIFF
--- a/pkg/datastore/phpapi/phpapi.go
+++ b/pkg/datastore/phpapi/phpapi.go
@@ -66,11 +66,11 @@ func (p *PHPAPI) GetCluster(clusterId string) (types.SubmarinerCluster, error) {
 	requestUrl := fmt.Sprintf("%s://%s/clusters.php?plurality=false&identifier=%s&cluster_id=%s", p.Proto, p.Server, p.APIToken, clusterId)
 	klog.V(8).Infof("request url: %s", requestUrl)
 	clustersGetter, err := http.Get(requestUrl)
-	defer clustersGetter.Body.Close()
 	if err != nil {
 		klog.Errorf("Encountered error while trying to retrieve clusters from the apiserver")
 		return types.SubmarinerCluster{}, err
 	}
+	defer clustersGetter.Body.Close()
 	clustersRaw, err := ioutil.ReadAll(clustersGetter.Body)
 	klog.V(8).Infof("response body: %v", string(clustersRaw[:]))
 	var cluster types.SubmarinerCluster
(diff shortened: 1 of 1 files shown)


pkg/datastore/phpapi/phpapi.go
```diff
@@ -188,10 +188,14 @@ func (p *PHPAPI) SetEndpoint(local types.SubmarinerEndpoint) error {
 	}
 	formVal := url.Values{}
 	formVal.Set("action", "reconcile")
-	formVal.Add("endpoint",string(marshaledEndpoint))
+	formVal.Add("endpoint", string(marshaledEndpoint))
 	formedURL := fmt.Sprintf("%s://%s/endpoints.php?identifier=%s&cluster_id=%s", p.Proto, p.Server, p.APIToken, local.Spec.ClusterID)
 	klog.V(8).Infof("Formed URL was %s", formedURL)
 	poster, err := http.PostForm(formedURL, formVal)
+	if err != nil {
+		klog.Errorf("Error while setting endpoint %v", err)
```
tpantelis: Instead of logging the error here, how about returning a formatted error with the added context to the caller instead of just returning err, ie

  return fmt.Errorf("Error while setting endpoint %v", err)

 This would avoid duplicate logging, assuming the caller also logs it, as it probably should and not assume the implementation will log it. However I can see it the other way as well, ie should the implementation assume the caller will appropriately handle and log the returned error. I kind of prefer the former. Either way, we should be consistent - line 87 just returns the err without logging any additional context.

That said, cleaning all that up should probably be a separate patch.

skitt: I seem to remember having this discussion on ODL too ;-). I tend to favour a “log, or throw, not both” approach (adding useful information to errors as they go up the stack), which matches your preference here; however I also think it’s best to keep things consistent, so fix up the body handling in this patch, and then make logging consistent everywhere later on...

mpeterson: I agree with @skitt on this

skitt:
>     * handles the trivial patch of the space missing in L191.

Yes, this was supposed to be left out! I just pushed an updated patch but Tom beat me by merging the patch a few seconds earlier.

tpantelis:
> Yes, this was supposed to be left out! I just pushed an updated patch but Tom beat me by merging the patch a few seconds earlier.

Sorry :)

PR #38, opened 2019-05-28
Sample UT using Ginkgo

Jaanki:
It tests function "populateCidrBlockList" in pkg "route"  at https://github.com/rancher/submariner/blob/master/pkg/routeagent/controllers/route/route.go#L119.

This is a sample UT to demo Ginkgo usage for writing UTs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,13 +35,13 @@ import (
 )
 
 var (
-	localMasterUrl  string
+	localMasterURL  string
 	localKubeconfig string
 )
 
 func init() {
 	flag.StringVar(&localKubeconfig, "kubeconfig", "", "Path to kubeconfig of local cluster. Only required if out-of-cluster.")
-	flag.StringVar(&localMasterUrl, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	flag.StringVar(&localMasterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 }
 
 func main() {
@@ -58,7 +58,7 @@ func main() {
 		klog.Fatal(err)
 	}
 
-	cfg, err := clientcmd.BuildConfigFromFlags(localMasterUrl, localKubeconfig)
+	cfg, err := clientcmd.BuildConfigFromFlags(localMasterURL, localKubeconfig)
 	if err != nil {
 		klog.Exitf("Error building kubeconfig: %s", err.Error())
 	}
--- a/pkg/cableengine/ipsec/ipsec.go
+++ b/pkg/cableengine/ipsec/ipsec.go
@@ -48,15 +48,15 @@ type Engine struct {
 	LogFile string
 }
 
-type IPSecSpecification struct {
-	PSK string
-	Debug bool
+type Specification struct {
+	PSK     string
+	Debug   bool
 	LogFile string
 }
 
 func NewEngine(localSubnets []string, localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint) *Engine {
 
-	ipSecSpec := IPSecSpecification{}
+	ipSecSpec := Specification{}
 
 	err := envconfig.Process("ce_ipsec", &ipSecSpec)
 	if err != nil {
@@ -140,16 +140,16 @@ func (i *Engine) StopEngine() error {
 	return nil
 }
 
-func (i *Engine) SyncCables(clusterId string, endpoints []types.SubmarinerEndpoint) error {
+func (i *Engine) SyncCables(clusterID string, endpoints []types.SubmarinerEndpoint) error {
 	klog.V(2).Infof("Starting selective cable sync")
-	
+
 	client, err := getClient()
 	if err != nil {
 		return err
 	}
 	defer client.Close()
-	
-	activeConnections, err := i.getActiveConns(false, clusterId, client)
+
+	activeConnections, err := i.getActiveConns(false, clusterID, client)
 	if err != nil {
 		return err
 	}
@@ -355,41 +355,41 @@ func (i *Engine) installCableInternal(endpoint types.SubmarinerEndpoint, client 
 	return nil
 }
 
-func (i *Engine) RemoveCable(cableId string) error {
+func (i *Engine) RemoveCable(cableID string) error {
 	client, err := getClient()
 	if err != nil {
 		return err
 
 	}
 	defer client.Close()
-	
-	return i.removeCableInternal(cableId, client)
+
+	return i.removeCableInternal(cableID, client)
 }
 
-func (i *Engine) removeCableInternal(cableId string, client *goStrongswanVici.ClientConn) error {
+func (i *Engine) removeCableInternal(cableID string, client *goStrongswanVici.ClientConn) error {
 	i.Lock()
 	defer i.Unlock()
 
 	/*
 	err = client.Terminate(&goStrongswanVici.TerminateRequest{
-		Child: "submariner-child-" + cableId,
+		Child: "submariner-child-" + cableID,
 	})
 	if err != nil {
-		klog.Errorf("Error when terminating child connection %s : %v", cableId, err)
+		klog.Errorf("Error when terminating child connection %s : %v", cableID, err)
 	}
 
 	err = client.Terminate(&goStrongswanVici.TerminateRequest{
-		Ike: cableId,
+		Ike: cableID,
 	})
 	if err != nil {
-		klog.Errorf("Error when terminating ike connection %s : %v", cableId, err)
+		klog.Errorf("Error when terminating ike connection %s : %v", cableID, err)
 	} */
-	klog.Infof("Unloading connection %s", cableId)
+	klog.Infof("Unloading connection %s", cableID)
 	err := client.UnloadConn(&goStrongswanVici.UnloadConnRequest{
-		Name: cableId,
+		Name: cableID,
 	})
 	if err != nil {
-		klog.Errorf("Error when unloading connection %s : %v", cableId, err)
+		klog.Errorf("Error when unloading connection %s : %v", cableID, err)
 		return err
 	}
 
@@ -411,11 +411,11 @@ func (i *Engine) removeCableInternal(cableId string, client *goStrongswanVici.Cl
 		if count > 2 {
 			klog.Infof("Waited for connection terminate for 2 iterations, triggering a force delete of the IKE")
 			err = client.Terminate(&goStrongswanVici.TerminateRequest{
-				Ike: cableId,
+				Ike:   cableID,
 				Force: "yes",
 			})
 			if err != nil {
-				klog.Errorf("error when terminating ike connection %s : %v", cableId, err)
+				klog.Errorf("error when terminating ike connection %s : %v", cableID, err)
 			}
 		}
 		sas, err := client.ListSas("", "")
@@ -425,7 +425,7 @@ func (i *Engine) removeCableInternal(cableId string, client *goStrongswanVici.Cl
 			found := false
 			for _, samap := range sas {
 				klog.V(6).Infof("Found SA %v", samap)
-				sa, stillExists := samap[cableId]
+				sa, stillExists := samap[cableID]
 				if stillExists && ( sa.State == "DELETING" || sa.State == "CONNECTING" ){
 					found = true
 					break
@@ -448,7 +448,7 @@ func (i *Engine) removeCableInternal(cableId string, client *goStrongswanVici.Cl
 			}
 		}
 	}
-	klog.Infof("Removed connection %s", cableId)
+	klog.Infof("Removed connection %s", cableID)
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


Jaanki:
The output looks like this
`$ go test --args -ginkgo.v
Running Suite: Route Unit Test Suite
====================================
Random Seed: 1559045011
Will run 2 of 2 specs

Route Unit test for router.go Testing for function populateCidrBlockList 
  should append cidrblock to subnet if subnets doesnot contain cidrblock
  /home/jchhatba/go/src/github.com/rancher/submariner/pkg/routeagent/controllers/route/route_test.go:18
•
------------------------------
Route Unit test for router.go Testing for function populateCidrBlockList 
  Should not append if subnets contain cidrblock
  /home/jchhatba/go/src/github.com/rancher/submariner/pkg/routeagent/controllers/route/route_test.go:25
•
Ran 2 of 2 Specs in 0.000 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
ok      github.com/rancher/submariner/pkg/routeagent/controllers/route  0.011s`

Jaanki:
Ginkgo UTs are generally written in the same folder for example kubeedge [1]. Thats what I have observed. But I do think that they should be in different folder like

tests/e2e
tests/unit

[1]. https://github.com/kubeedge/kubeedge/tree/master/edge/pkg/devicetwin/dtmodule

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,35 @@
+package routecontroller
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Route", func() {
+	Describe("Unit test for router.go", func() {
```
tpantelis: Do we need 2 Describes here?

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,35 @@
...
+		var rc RouteController
+		var testinputcidrblocks []string
+
+		BeforeEach(func() {
+			testinputcidrblocks = []string{"10.10.10.0/24", "192.168.1.0/24"}
+		})
+
+		Context("Testing for function populateCidrBlockList", func() {
+			It("should append cidrblock to subnet if subnets doesnot contain cidrblock", func() {
```
tpantelis: "does not" - 2 words. I think cidrblock should be 2 words as well. How about this text: "should append cidr blocks to subnets that are not already present"

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,35 @@
...
+				rc.subnets = []string{"192.168.1.0/24"}
+				rc.populateCidrBlockList(testinputcidrblocks)
+				want := []string{"192.168.1.0/24", "10.10.10.0/24"}
+				Expect(rc.subnets).To(Equal(want))
+			})
+
+			It("Should not append if subnets contain cidrblock", func() {
```
tpantelis: How about "should not append cidr block that is already present"

mangelajo:
It's the go standard to keep unit tests on the same folder as the package.

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,35 @@
+package routecontroller
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Route", func() {
+	Describe("Unit test for router.go", func() {
+		var rc RouteController
```
tpantelis: Can you name this "routeController"

pkg/routeagent/controllers/route/route_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package routecontroller_test
```
mangelajo: I wonder if we can setup a a single test suite for the whole project, or if we need to do package by package.

pkg/routeagent/controllers/route/route_test.go
Jaanki: Agree for "doesnot". I kept "cidrblock" as single so that it resembles to the var name and is easy to understand/read and relate.

pkg/routeagent/controllers/route/route_test.go
Jaanki: Ya. We can work on fine tuning once we decide on the framework.

pkg/routeagent/controllers/route/route_test.go
Jaanki: It has to be of type "RouteController" to reflect the struct "RouteController" defined in route.go

pkg/routeagent/controllers/route/route_suite_test.go
Jaanki: This needs to explored.

tpantelis: I think it needs to be per package b/c it relies on package initializers to register the specs, ie all specs must be registered either via global vars or init() functions before RunSpecs is called.

pkg/routeagent/controllers/route/route_test.go
tpantelis: I meant rename the var 'rc' to routeController for improved readability.

pkg/routeagent/controllers/route/route_test.go
tpantelis: I think we've already decided that :)

pkg/routeagent/controllers/route/route_test.go
tpantelis: But remember in BDD, the text should read as natural language and not in terms of code constructs, at least as much as possible.

pkg/routeagent/controllers/route/route_test.go
Jaanki: Ahh!! I kept rc to be short and its quite easy to remember what it means.

tpantelis: To you maybe since you wrote it :) But as we add more and more test cases and more folks reading the code, I prefer added readability over cryptic var names.

pkg/routeagent/controllers/route/route_test.go
tpantelis: I think for proper BDD of the form given/when/then, the "if subnets does not contain cidrblock" clause is really the 'when' part and thus should be part of the Context clause. The Describe clause should indicate the 'given' part, in this case that it's testing the populateCidrBlockList function. So to put it all together, it becomes:

    Describe("Function populateCidrBlockList", func() {
        Context("When the existing subnets does not contain an input cidr block", func() {
            It("should append the cidr block to subnets", func() {
                routeController := RouteController{subnets: []string{"192.168.1.0/24"}}
                routeController.populateCidrBlockList([]string{"10.10.10.0/24", "192.168.1.0/24"})
                want := []string{"192.168.1.0/24", "10.10.10.0/24"}
                Expect(routeController.subnets).To(Equal(want))
            })
        })

        Context("When the existing subnets does contain an input cidr block", func() {
            It("should not append the cidr block to subnets", func() {
                ...
            })
        })
    })

I employed similar methodology in https://github.com/rancher/submariner/pull/40 and I think it makes sense (I'm still getting familiar with BDD).

Also notice that I eliminated the global vars and the need for BeforeEach in the code snippet above.  I think in general the use of these makes the code a bit messier and should be avoided unless there's a compelling reason to use them, eg if you need to do expensive or lengthy setup.

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,35 @@
+package routecontroller
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Route", func() {
```
tpantelis: "Route" -> "Routecontroller"

tpantelis:
> It's the go standard to keep unit tests on the same folder as the package.

+1. They should be logically grouped with the code to test. Also putting the tests in the same package enables one to test package-scoped functions/fields (although whether or not that's good practice is debatable).

pkg/routeagent/controllers/route/route_test.go
Jaanki: Done.

pkg/routeagent/controllers/route/route_test.go
Jaanki: Done

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,46 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("RouteController", func() {
+
+	Describe("Function populateCidrBlockList", func() {
+
+		Context("subnets does not contain CIDR blocks", func() {
+			It("should append CIDR block to subnets", func() {
```
tpantelis: Describe, Context and It should read as natural language together, how about:
 
    Context("When input CIDR blocks are not present in the existing subnets", func() {
        It("should append the CIDR blocks to subnets", func() {

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,46 @@
...
+				routeController := RouteController{subnets: []string{"192.168.1.0/24"}}
+				routeController.populateCidrBlockList([]string{"10.10.10.0/24", "192.168.1.0/24"})
+				want := []string{"192.168.1.0/24", "10.10.10.0/24"}
+				Expect(routeController.subnets).To(Equal(want))
+			})
+		})
+
+		Context("subnets contain CIDR block", func() {
+			It("Should not append CIDR block subnets", func() {
```
tpantelis: Similarly

    Context("When input CIDR blocks are already present in the existing subnets", func() {
        It("should not append the CIDR blocks to subnets", func() {

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,46 @@
...
+				want := []string{"10.10.10.0/24", "192.168.1.0/24", "192.168.0.0./24"}
+				Expect(routeController.subnets).To(Equal(want))
+			})
+
+		})
+	})
+
+	Describe("Function containsString", func() {
+		Context("Given array of strings contain specified string", func() {
+			It("Should return true", func() {
```
tpantelis: Should -> should for consistency

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,46 @@
...
+				routeController.populateCidrBlockList([]string{"10.10.10.0/24", "192.168.1.0/24"})
+				want := []string{"10.10.10.0/24", "192.168.1.0/24", "192.168.0.0./24"}
+				Expect(routeController.subnets).To(Equal(want))
+			})
+
+		})
+	})
+
+	Describe("Function containsString", func() {
+		Context("Given array of strings contain specified string", func() {
```
tpantelis: How about "When the given string array contains the given string"

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,46 @@
...
+			It("Should return true", func() {
+				Expect(containsString([]string{"unit", "test"}, "unit")).To(BeTrue())
+			})
+		})
+
+		Context("Given array of strings does not contain specified string", func() {
```
tpantelis: How about "When the given string array does not contain the given string"

pkg/routeagent/controllers/route/route_test.go
```diff
@@ -0,0 +1,46 @@
...
+			It("Should return false", func() {
```
tpantelis: Should -> should

pkg/routeagent/controllers/route/route_test.go
Jaanki: Keeping first letter capital everywhere.

pkg/routeagent/controllers/route/route_test.go
Jaanki: 2 times given seems confusing. "given" array of strings contains "specified" string is more clear as to what needs to be searched where. Meaning "specified" needs to be searched in "given".

pkg/routeagent/controllers/route/route_test.go
Jaanki: Same

pkg/routeagent/controllers/route/route_test.go
Jaanki: First letter capital everywhere.

pkg/routeagent/controllers/route/route_test.go
Jaanki: First letter capital everywhere.

pkg/routeagent/controllers/route/route_test.go
Jaanki: First letter capital everywhere.

pkg/routeagent/controllers/route/route_test.go
tpantelis: "given" and "specified" are synonymous in this case, ie both params are given or specified by the caller. I just used the same word  in both clauses for consistency but it's fine to mix them.

PR #46, opened 2019-06-04
Update Dockerfiles

skitt:
This upgrades and cleans up the Dockerfiles used for the engine and
router agent:

* upgrade from Ubuntu 16.04 to 18.04
* set DEBIAN_FRONTEND instead of the debconf selection
* reduce the set of explicitly installed packages
* avoid the extra copy on StrongSwan

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,12 +1,15 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 WORKDIR /var/submariner
 
 COPY strongswan*.tar.gz /var/tmp
 
-RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections && apt-get -y update && apt-get -y install curl iproute2 iptables libatm1 libgmp10 libmnl0 libnfnetlink0 libssl1.0.0
+ENV DEBIAN_FRONTEND=noninteractive
 
-RUN /bin/bash -c "cd / && cp /var/tmp/strongswan*.tar.gz . && tar -zxvf strongswan*.tar.gz && rm strongswan*.tar.gz"
+RUN apt-get -y update && \
+    apt-get -y install curl iproute2 iptables libatm1 libgmp10
+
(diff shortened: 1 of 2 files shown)


skitt:
@dimaunx you’ll probably find this interesting!

package/Dockerfile
```diff
@@ -1,12 +1,15 @@
...
 WORKDIR /var/submariner
 
 COPY strongswan*.tar.gz /var/tmp
 
-RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections && apt-get -y update && apt-get -y install curl iproute2 iptables libatm1 libgmp10 libmnl0 libnfnetlink0 libssl1.0.0
+ENV DEBIAN_FRONTEND=noninteractive
 
-RUN /bin/bash -c "cd / && cp /var/tmp/strongswan*.tar.gz . && tar -zxvf strongswan*.tar.gz && rm strongswan*.tar.gz"
+RUN apt-get -y update
+RUN apt-get -y install curl iproute2 iptables libatm1 libgmp10
```
skitt: I’m not sure why `curl` and `libgmp10` are needed, but I left them in to err on the safe side. I’ll revisit this once we switch to StrongSwan packages (since that will take care of those dependencies).

Jaanki: Can the 2 RUNs be combined?

package/Dockerfile.routeagent
```diff
@@ -1,8 +1,11 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 WORKDIR /var/submariner
 
-RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections && apt-get -y update && apt-get -y install iproute2 iptables libatm1 libgmp10 libmnl0 libnfnetlink0 libssl1.0.0
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get -y update
+RUN apt-get -y install iproute2 iptables libatm1 libgmp10
```
Jaanki: Ditto

skitt: Done!

PR #43, opened 2019-05-30
Golint corrections for pkg

Jaanki:
golint corrections for all files in pkg folder

Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

PR #47, opened 2019-06-04
golint clean-ups

skitt:
* Id → ID (e.g. ClusterId → ClusterID)
* Api → API (e.g. ApiKey → APIKey)
* Drop unused import and unnecessary type specification in main.go
* tunnel.TunnelController → tunner.Controller to avoid stuttering

### DIFF
--- a/pkg/controllers/tunnel/tunnel.go
+++ b/pkg/controllers/tunnel/tunnel.go
@@ -18,7 +18,7 @@ import (
 	"k8s.io/klog"
 )
 
-type TunnelController struct {
+type Controller struct {
 	ce                  cableengine.CableEngine
 	kubeClientSet       kubernetes.Interface
 	submarinerClientSet submarinerClientset.Interface
@@ -29,8 +29,8 @@ type TunnelController struct {
 	endpointWorkqueue workqueue.RateLimitingInterface
 }
 
-func NewTunnelController(objectNamespace string, ce cableengine.CableEngine, kubeClientSet kubernetes.Interface, submarinerClientSet submarinerClientset.Interface, endpointInformer submarinerInformers.EndpointInformer) *TunnelController {
-	tunnelController := &TunnelController{
+func NewController(objectNamespace string, ce cableengine.CableEngine, kubeClientSet kubernetes.Interface, submarinerClientSet submarinerClientset.Interface, endpointInformer submarinerInformers.EndpointInformer) *Controller {
+	tunnelController := &Controller{
 		ce:                  ce,
 		kubeClientSet:       kubeClientSet,
 		submarinerClientSet: submarinerClientSet,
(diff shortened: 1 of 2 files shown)


Jaanki:
Some of these changes are part of #43

skitt:
> Some of these changes are part of #43

Ah yes, thanks, I’ll rebase this once that’s merged.

pkg/datastore/kubernetes/kubernetes.go
```diff
@@ -31,8 +31,8 @@ type k8s struct {
 }
 
 type k8sSpecification struct {
```
tpantelis: I think this should just be Specification - golint doesn't know that k8s is shortcut for kubernetes :)

skitt: I fixed the two stuttering issues you mention in #42 ;-).

skitt: > I think this should just be Specification - golint doesn't know that k8s is shortcut for kubernetes :)

That would export it, which I’m not sure is appropriate. The same exporting question goes for `k8s` too in the same class, which makes `golint` complain because it’s used as the return type for an exported function...

tpantelis: yeah it doesn't need to be exported so we could rename to 'specification'. 

RE: k8s, NewK8sDatastore cannot return a *datastore.Datastore. It could return datastore.Datastore but then the method receivers would have to be values and not pointers. This seems like a strange aspect to Go - in Java we're used to a factory method being able to return an interface and hide the implementation class. You can do that in Go but it's not straightforward.  To eliminate that warning we could rename 'k8s' to 'Datastore' - that would export the type but none of the fields are exported.

skitt: > yeah it doesn't need to be exported so we could rename to 'specification'.
> 
> RE: k8s, NewK8sDatastore cannot return a *datastore.Datastore. It could return datastore.Datastore but then the method receivers would have to be values and not pointers. This seems like a strange aspect to Go - in Java we're used to a factory method being able to return an interface and hide the implementation class. You can do that in Go but it's not straightforward. To eliminate that warning we could rename 'k8s' to 'Datastore' - that would export the type but none of the fields are exported.

Yup, the latter is exactly what I did in #49. I’ll update it to unexport `Specification`.

pkg/controllers/tunnel/tunnel.go
```diff
@@ -28,8 +28,8 @@ type TunnelController struct {
 	endpointWorkqueue workqueue.RateLimitingInterface
 }
 
-func NewTunnelController (objectNamespace string, ce cableengine.CableEngine, kubeClientSet kubernetes.Interface, submarinerClientSet submarinerClientset.Interface, endpointInformer submarinerInformers.EndpointInformer) *TunnelController {
-	tunnelController := &TunnelController{
+func NewTunnelController(objectNamespace string, ce cableengine.CableEngine, kubeClientSet kubernetes.Interface, submarinerClientSet submarinerClientset.Interface, endpointInformer submarinerInformers.EndpointInformer) *Controller {
```
tpantelis: Similar to removing "Tunnel" from the struct name, we can avoid stutter here as well by renaming to NewController or just New since it returns a Controller.

PR #48, opened 2019-06-04
Simple e2e testing with kind

dimaunx:
1. Add support for e2e tests with kind. make e2e command
2. Add support for running go test.

image

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -2,29 +2,26 @@ FROM ubuntu:16.04
 # FROM arm=armhf/ubuntu:16.04
 
 ARG DAPPER_HOST_ARCH
-ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
+ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPER_ENV=TAG \
+    DAPPER_SOURCE=/go/src/github.com/rancher/submariner DAPPER_DOCKER_SOCKET=true \
+    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} LINT_VERSION=v1.16.0
 
-RUN apt-get update && \
-    apt-get install -y gcc ca-certificates git wget curl vim less file && \
-    rm -f /bin/sh && ln -s /bin/bash /bin/sh
+RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
-RUN wget -O - https://storage.googleapis.com/golang/go1.11.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
-    go get github.com/rancher/trash
+RUN apt-get -q update && \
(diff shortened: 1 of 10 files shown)


dfarrell07:
I tested this on my system and it worked perfectly out of the box. :+1:

Dockerfile.dapper
```diff
@@ -3,22 +3,31 @@ FROM ubuntu:16.04
...
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
-RUN wget -O - https://storage.googleapis.com/golang/go1.11.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
+RUN wget -q -O - https://storage.googleapis.com/golang/go1.11.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
     go get github.com/rancher/trash
 
-ENV DOCKER_URL_amd64=https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 \
-    DOCKER_URL_arm=https://github.com/rancher/docker/releases/download/v1.10.3-ros1/docker-1.10.3_arm \
-    DOCKER_URL=DOCKER_URL_${ARCH}
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && \
```
skitt: Wouldn’t it be better to use `wget` or `curl` throughout, not both?

scripts/lint
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+
+cd $(dirname $0)/..
+
+golangci-lint run -D megacheck -E unused,gosimple,staticcheck
```
skitt: I think this should be an edit to `scripts/validate` (which currently lints using `golint`), but as a separate PR.

Dockerfile.dapper
dimaunx: Good suggestion, curl now is used in all commands and RUN block was consolidated to keep intermediate less layers.

scripts/lint
dimaunx: Removed the lint script. Agree, we should do it in separate PR.

scripts/kind-e2e/cluster2-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+  metadata:
+    name: config
+  networking:
+    podSubnet: 10.245.0.0/16
+    serviceSubnet: 100.95.0.0/16
+    dnsDomain: cluster2.local
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
```
tpantelis: duplicate of line 15 - is this intended? ...

scripts/kind-e2e/cluster3-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+  metadata:
+    name: config
+  networking:
+    podSubnet: 10.246.0.0/16
+    serviceSubnet: 100.96.0.0/16
+    dnsDomain: cluster3.local
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
```
tpantelis: ... same here

scripts/kind-e2e/cluster2-config.yaml
skitt: Yes, we want two worker nodes.

scripts/kind-e2e/cluster3-config.yaml
dimaunx: Yes this this is kind configuration for 2 worker nodes.

dfarrell07:
I tested this again locally at 70fa657 and it still worked.

Imho these should be merged into one commit. I'd also love that commit to have a more detailed commit message, following the [pattern well-described here](https://chris.beams.io/posts/git-commit/).

Very good work! :+1:

PR #50, opened 2019-06-04
Refactor NewK8sDatastore function

tpantelis:
- Rename to NewDatastore to to avoid stuttering
- Return error instead of calling Fatalf
- Move instantiation of Datastore to the end return


tpantelis:
@skitt Stephen - requesting review.

PR #51, opened 2019-06-04
Refactor route controller to avoid stuttering

tpantelis:
- Renamed package from routecontroller to route as controller is already in
  the import path and makes it consistent with tunnel.go
- route.RouteController -> route.Controller
- route.NewRouteController -> route.NewController


tpantelis:
@skitt Stephen - requesting review.

PR #54, opened 2019-06-06
Enable golangci-lint support

dimaunx:
Currently golint and gofmt are used in shadow mode.
I would like to propose to use golangci-lint.
golangci-lint is a linters aggregator. Has number of advantages:
1. It's 5 times faster than gometalinter. 
2. It's easy to integrate and use, has nice output and has a minimum number of false positives. 
3. It supports go modules.
4. Long list of supported linters. https://github.com/golangci/golangci-lint#supported-linters
5. Good IDE integration support. https://github.com/golangci/golangci-lint#editor-integration

This PR will change the **make validate** command to run golangci-lint on demand. 

Test locally with **make validate** command.

### DIFF
--- a/scripts/validate
+++ b/scripts/validate
@@ -3,18 +3,4 @@ set -e
 
 cd $(dirname $0)/..
 
-echo Running validation
-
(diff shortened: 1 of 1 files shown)


scripts/e2e
```diff
@@ -3,6 +3,7 @@ set -e
 
 cd $(dirname $0)
 ./test
+#./validate
```
tpantelis: Why was this commented out - should we just remove the line?

dimaunx: Yes we should remove it.

PR #56, opened 2019-06-06
Format Go files

tpantelis:
Ran goimports to adhere to standard Go formatting style. goimports
is a superset of gofmt that also groups imports.


mangelajo:
@dimaunx It'd be cool if we could make golint check all this.

PR #59, opened 2019-06-07
Fix make test after E2E being merged

mangelajo:
test/e2e directory is being found by the test script, and
it will fail when go test tries to run it. This stops
test/e2e being found out by the discovery script.


mangelajo:
I merged it on myself to avoid blocking other people because the test script is broken and blocks most of the other make actions.

PR #61, opened 2019-06-07
Cleanup error handling in ipsec

tpantelis:
- Propagate errors instead of calling Fatalf
- Refactored runCharon to propagate the error if cmd.Start() fails.
  Note that prior, cmd.Wait() would've returned an error if cmd.Start()
  failed and and thus immediately exited via Fatalf but I think it's cleaner
  to propagate the error via StartEngine.

### DIFF
--- a/pkg/cableengine/ipsec/ipsec.go
+++ b/pkg/cableengine/ipsec/ipsec.go
@@ -55,13 +55,13 @@ type Specification struct {
 	LogFile string
 }
 
-func NewEngine(localSubnets []string, localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint) *Engine {
+func NewEngine(localSubnets []string, localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint) (*Engine, error) {
 
 	ipSecSpec := Specification{}
 
 	err := envconfig.Process("ce_ipsec", &ipSecSpec)
 	if err != nil {
-		klog.Fatal(err)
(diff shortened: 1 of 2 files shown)


tpantelis:
@skitt @sridhargaddam

pkg/cableengine/ipsec/ipsec.go
```diff
@@ -553,23 +571,34 @@ func runCharon(debug bool, logFile string) {
...
 	}
 
-	cmd.Start()
+	if err := cmd.Start(); err != nil {
+		// Note - Close handles nil receiver
+		outputFile.Close()
+		return fmt.Errorf("error starting the charon process wih args %v: %v", args, err)
+	}
+
+	go func() {
```
mangelajo: I'm not sure this is equivalent to the original.

In the original the whole function was being parallelized (including the Start()) method that seems blocking.

In this case we only paralellize the Fatalf output. 

Shouldn't we paralellize the Start also, may be use a channel to return any error?

mangelajo: That said It seems to be passing the E2E tests, but I'm not sure if this have some implication. Like: it will block the caller on Start ?

tpantelis: Start() does not block awaiting completion:

    Start starts the specified command but does not wait for it to complete.
    The Wait method will return the exit code and release associated resources once the command exits.

mangelajo: Oh right, I didn't see the Wait. That makes sense, you made the blocking part parallel again by use of the go + lambda function.

PR #62, opened 2019-06-10
Cleanup error handling in tunnel

tpantelis:
- Propagate errors from functions that return error
- Ensure all error objects are logged

### DIFF
--- a/pkg/controllers/tunnel/tunnel.go
+++ b/pkg/controllers/tunnel/tunnel.go
@@ -88,23 +88,20 @@ func (t *TunnelController) processNextEndpoint() bool {
 		klog.V(4).Infof("Processing endpoint object: %v", obj)
 		ns, key, err := cache.SplitMetaNamespaceKey(obj.(string))
 		if err != nil {
-			klog.Errorf("error while splitting meta namespace key: %v", err)
-			return nil
+			return fmt.Errorf("error splitting meta namespace key for endpoint %s: %v", obj, err)
 		}
 		endpoint, err := t.submarinerClientSet.SubmarinerV1().Endpoints(ns).Get(key, metav1.GetOptions{})
 		if err != nil {
-			klog.Errorf("Error while retrieving submariner endpoint object %s: %v", obj, err)
 			t.endpointWorkqueue.Forget(obj)
-			return nil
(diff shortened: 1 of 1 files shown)


pkg/controllers/tunnel/tunnel.go
```diff
@@ -137,19 +133,23 @@ func (t *TunnelController) handleRemovedEndpoint(obj interface{}) {
...
-	klog.V(4).Infof("Removed endpoint from cable engine %s", object.Name)
+
+	klog.V(4).Infof("Informed of removed endpoint for tunnel controller object: %#v", object)
+	if err := t.ce.RemoveCable(object.Spec.CableName); err != nil {
+		utilruntime.HandleError(fmt.Errorf("error removing endpoint cable %s from engine: %v",
+			object.Spec.CableName, err))
+		return
+	}
+
+	klog.V(4).Infof("Removed endpoint cable %s from engine", object.Spec.CableName)
```
mangelajo: true nitpicking: 
Can we keep the original info on that last log message (the cable engine) , although now we just have one?

klog.V(4).Infof("Removed endpoint cable %s from engine %s", object.Spec.CableName, object.Name)

tpantelis: The previous wording was misleading, ie it implied it "removed a cable from an engine named X". Of course what we really mean is it "removed a cable named X from the engine".  Also since we're calling RemoveCable with 'object.Spec.CableName', I think it makes sense to substitute object.Spec.CableName for X.

pkg/controllers/tunnel/tunnel.go
```diff
@@ -88,23 +88,20 @@ func (t *TunnelController) processNextEndpoint() bool {
 		klog.V(4).Infof("Processing endpoint object: %v", obj)
```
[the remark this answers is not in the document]
tpantelis: I've seen that pattern used in Go for defers. It could be a named function. Since the inline function makes up the bulk of the outer function, I don't have a strong preference either way.

pkg/controllers/tunnel/tunnel.go
mangelajo: Oh, sorry, that's an endpoint, when I was making the comment I thought it was the cable engine name.., you're right...

PR #63, opened 2019-06-11
Add test as a phony target to restore "make test"

skitt:
The presence of the test directory broke this.


mangelajo:
Thank you!

PR #65, opened 2019-06-11
E2E local environment 

dimaunx:
Add additional functionality to e2e script.
Currently **make e2e** runs and ephemeral Dapperised job.
Three new parameters are added to the **make e2e** command:
1.  **status** - Allows to specify the desired state for the environment. 
     **make e2e status=keep**  keeps the e2e environment running for further development and troubleshooting.
     **make e2e status=clean**  cleans up the local environment.
2. **version** - Allows to specify k8s version for e2e local environment.
    **make e2e version=1.14.1** 
3.  **logging** - Allows to create Elastic Search and Filebeat deployment on the e2e local clusters.
     **make e2e logging=true**

The commands can be chained together: **make e2e status=keep version=1.14.1 logging=true**
Each time **make e2e status=keep** command is executed, code is build, deployed on top of 3 kind k8s clusters and tested.


scripts/kind-e2e/e2e.sh
```diff
@@ -1,36 +1,56 @@
...
 function setup_broker() {
     kubectl config use-context cluster1
-    helm install submariner-latest/submariner-k8s-broker \
-         --name ${SUBMARINER_BROKER_NS} \
-         --namespace ${SUBMARINER_BROKER_NS}
+    if kubectl get crd clusters.submariner.io > /dev/null 2>&1; then
+        echo Submariner CRDs already exist, skipping broker creation...
+    else
+        echo Installing broker on cluster1.
+        helm install submariner-latest/submariner-k8s-broker --name ${SUBMARINER_BROKER_NS} --namespace ${SUBMARINER_BROKER_NS}
```
mangelajo: are there any pods to be updated on the broker?, or just the CRDs?

scripts/kind-e2e/e2e.sh
```diff
@@ -117,34 +147,95 @@ function test_connection() {
...
 function cleanup {
-  echo "Cleanup..."
-  for i in 1 2 3; do kind delete cluster --name=cluster${i}; done
+  for i in 1 2 3; do
+
+    if [[ $(kind get clusters | grep cluster${i} | wc -l) -gt 0  ]]; then
+        kind delete cluster --name=cluster${i};
+    fi
+    done
+
```
mangelajo: I'd leave some time between the deletion and cleaning up later. May be some containers need a few seconds to exit?

scripts/kind-e2e/e2e.sh
```diff
@@ -117,34 +147,95 @@ function test_connection() {
...
 fi
 
-if [[ $1 != -k ]]; then
+if [[ $1 != keep ]]; then
     trap cleanup EXIT
 fi
 
-helm init --client-only
-helm repo add submariner-latest https://releases.rancher.com/submariner-charts/latest
+echo Starting a job with status: $1, k8s version: $2, monitor: $3.
```
mangelajo: nit, I would omit "a job",  because in some cases it will be a devel env :-)

Makefile
```diff
@@ -17,6 +21,3 @@ trash-keep: .dapper
 	./.dapper -m bind trash -k
 
 .DEFAULT_GOAL := ci
-
```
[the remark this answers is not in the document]
mangelajo: It's not clear how github shows it, but you're removing the .phony target for test :)

Makefile
```diff
@@ -8,7 +12,7 @@ TARGETS := $(shell ls scripts)
 	@mv .dapper.tmp .dapper
 
 $(TARGETS): .dapper
-	./.dapper $@
+	./.dapper -m bind $@ $(status) $(version) $(logging)
```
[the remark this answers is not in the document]
dimaunx: This is exactly the thoughts i was having. For now maybe the best approach would be to use bind and look for a fix, as you mentioned. On other hand we could use cp mode and if we would like to keep the environment running we must have at least one run that passes. We could run the make e2e status=keep on master branch and master should pass always,switch to the development branch and run make e2e status=keep again. One downside to this is that logs won't be exported if the run fails.

Makefile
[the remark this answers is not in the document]
dimaunx: Will fix.

dfarrell07:
I tested this locally and it worked for me. :+1:

scripts/kind-e2e/e2e.sh
```diff
@@ -1,36 +1,56 @@
...
     for i in 1 2 3; do
-        kind create cluster --name=cluster${i} --wait=5m --config=./kind-e2e/cluster${i}-config.yaml
-        master_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster${i}-control-plane | head -n 1)
-        sed -i -- "s/server: .*/server: https:\/\/$master_ip:6443/g" $(kind get kubeconfig-path --name="cluster$i")
-        sed -i -- "s/user: kubernetes-admin/user: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
-        sed -i -- "s/name: kubernetes-admin.*/name: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
+        if [[ $(kind get clusters | grep cluster${i} | wc -l) -gt 0  ]]; then
+            echo Cluster clister${i} already exists, skipping cluster creation...
+        else
+            if [[ -n "$2" ]]; then
```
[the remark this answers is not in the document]
dimaunx: Makes total sense.

scripts/kind-e2e/e2e.sh
```diff
@@ -1,36 +1,56 @@
...
+                kind create cluster --image=kindest/node:v$2 --name=cluster${i} --wait=5m --config=./kind-e2e/cluster${i}-config.yaml
```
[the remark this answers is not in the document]
dimaunx: Makes total sense.

scripts/kind-e2e/e2e.sh
dimaunx: Broker just contains 2 CRDs.

scripts/kind-e2e/e2e.sh
```diff
@@ -1,36 +1,56 @@
...
+        fi
     done
 }
 
 function setup_broker() {
     kubectl config use-context cluster1
-    helm install submariner-latest/submariner-k8s-broker \
-         --name ${SUBMARINER_BROKER_NS} \
-         --namespace ${SUBMARINER_BROKER_NS}
+    if kubectl get crd clusters.submariner.io > /dev/null 2>&1; then
```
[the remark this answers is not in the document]
dimaunx: Yes totally make sense , great suggestion i will test this.

scripts/kind-e2e/e2e.sh
dimaunx: Yes good suggestion.

scripts/kind-e2e/e2e.sh
dimaunx: From my testing it was working really good because kind deletes the containers forcefully and the second stage is more for containers that exited not related to kind.

Makefile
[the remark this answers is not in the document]
mangelajo: Let's keep it like this for now, and let's look for a fix in dapper, or let's make the bind mode optional if anyone needs it eventually.

scripts/kind-e2e/e2e.sh
mangelajo: ack, thanks.

scripts/kind-e2e/e2e.sh
mangelajo: ahh, ok, if it does forcefull deletion, yes, it's instantaneus AFAIK. We don't need to fix what it's not broken.

scripts/kind-e2e/e2e.sh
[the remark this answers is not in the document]
skitt: > we will need to figure out at some point if the CRDs need to be updated.
> 
> Could it make sense to remove-create always?

If we need to update the CRDs, we presumably need to upgrade Submariner, don’t we? I was thinking about this too when reading through the code, and my thought then was that we should KISS for now — it’s easy enough to delete everything and re-deploy...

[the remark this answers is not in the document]
dimaunx: This will require a larger refactor. The CRDs are created with helm if we remove the CRDs and helm resources we will have to re install subm on all the gateway nodes, because the endpoints are inserted during subm helm installation on  gateway nodes, lets keep it the way it is now and add this feature later.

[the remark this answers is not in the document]
skitt: > Lets keep it the current way, and add this feature later?

Yes, that’s what I’m thinking too.

skitt:
This now works in SELinux enforcing mode, somehow \o/

skitt:
> This now works in SELinux enforcing mode, somehow \o/

Oh wait so does the master branch, so nothing to do with this ;-).

mangelajo:
lol
---
irc: ajo / mangelajo
Miguel Angel Ajo Pelayo
+34 636 52 25 69
skype: ajoajoajo

scripts/kind-e2e/e2e.sh
```diff
@@ -1,36 +1,58 @@
...
-        kind create cluster --name=cluster${i} --wait=5m --config=./kind-e2e/cluster${i}-config.yaml
-        master_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster${i}-control-plane | head -n 1)
-        sed -i -- "s/server: .*/server: https:\/\/$master_ip:6443/g" $(kind get kubeconfig-path --name="cluster$i")
-        sed -i -- "s/user: kubernetes-admin/user: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
-        sed -i -- "s/name: kubernetes-admin.*/name: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
+        if [[ $(kind get clusters | grep cluster${i} | wc -l) -gt 0  ]]; then
+            echo Cluster cluster${i} already exists, skipping cluster creation...
+        else
+            if [[ -n ${version} ]]; then
+                kind create cluster --image=kindest/node:v$2 --name=cluster${i} --wait=5m --config=./kind-e2e/cluster${i}-config.yaml
```
mangelajo: lol, but use ${version} here? :)

scripts/kind-e2e/e2e.sh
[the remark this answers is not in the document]
mangelajo: I'm fine with that, ack :) thanks for the explanation.

scripts/kind-e2e/e2e.sh
dimaunx: Yes missed that :)

dfarrell07:
I do see one minor issue, but it's not related to this patch. The very last conditional in the cleanup function tests as false, which with -e (fail on errors) seems to make the entire e2e.sh script exit 1.

```
++ docker ps -qf status=exited
++ wc -l
+ [[ 0 -gt 0 ]]
++ docker images -qf dangling=true
++ wc -l
+ [[ 0 -gt 0 ]]
++ docker volume ls -qf dangling=true
++ wc -l
+ [[ 0 -gt 0 ]]
FATA[0687] exit status 1                                
make: *** [Makefile:15: e2e] Error 1
```

scripts/kind-e2e/README.md
```diff
@@ -0,0 +1,135 @@
...
+<!--links-->
+[kind]: https://github.com/kubernetes-sigs/kind
+[docker]: https://docs.docker.com/install/
+[kubectl]: https://kubernetes.io/docs/tasks/tools/install-kubectl/
+[k9s]: https://github.com/derailed/k9s
+[kubetail]: https://github.com/johanhaleby/kubetail
+[kind release page]: https://github.com/kubernetes-sigs/kind/releases
+[Change SELinux to permissive]: https://www.ibm.com/support/knowledgecenter/en/POWER8/p8ef9/p8ef9_selinux_setup.htm
+[Add local user to docker group]: https://docs.docker.com/install/linux/linux-postinstall/
+
```
mangelajo: Oh!, documentation! :-)  we can probably refine it later or put it together to the main docs/devel docs, awesome! :)

dimaunx: The docs definitely need refinement :) before we put it on the main doc page.

dimaunx:
> I do see one minor issue, but it's not related to this patch. The very last conditional in the cleanup function tests as false, which with -e (fail on errors) seems to make the entire e2e.sh script exit 1.
> 
> ```
> ++ docker ps -qf status=exited
> ++ wc -l
> + [[ 0 -gt 0 ]]
> ++ docker images -qf dangling=true
> ++ wc -l
> + [[ 0 -gt 0 ]]
> ++ docker volume ls -qf dangling=true
> ++ wc -l
> + [[ 0 -gt 0 ]]
> FATA[0687] exit status 1                                
> make: *** [Makefile:15: e2e] Error 1
> ```

Interesting.. Works for me. If you execute the last command in terminal it returns 1?

dfarrell07:
Ah, actually that line is not where the error is coming from, it's coming from the Makefile itself because I'm passing the keep/clean param wrong. I thought I was being slick shortening it to `make e2e clean`, should have stuck with `make e2e status=clean`.

PR #66, opened 2019-06-11
Cleanup error handling in phpapi

tpantelis:
- Propagate or log errors instead of calling Fatalf
- Improved logging messages where appropriate
- Return error for unimplemented method

### DIFF
--- a/pkg/datastore/phpapi/phpapi.go
+++ b/pkg/datastore/phpapi/phpapi.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kelseyhightower/envconfig"
 	"github.com/rancher/submariner/pkg/types"
 	"github.com/rancher/submariner/pkg/util"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/klog"
 )
 
@@ -28,18 +29,19 @@ type Specification struct {
 	Server string
 }
 
-func NewPHPAPI(apitoken string) *PHPAPI {
+func NewPHPAPI(apitoken string) (*PHPAPI, error) {
 	var pais Specification
 	err := envconfig.Process("backend_phpapi", &pais)
 	if err != nil {
-		klog.Fatal(err)
+		return nil, fmt.Errorf("error processing environment config for backend_phpapi: %v", err)
 	}
+
 	klog.Infof("Instantiating PHPAPI Backend at %s://%s with APIToken %s", pais.Proto, pais.Server, apitoken)
 	return &PHPAPI{
 		Proto:    pais.Proto,
 		Server:   pais.Server,
 		APIToken: apitoken,
-	}
+	}, nil
 }
 
 func (p *PHPAPI) GetClusters(colorCodes []string) ([]types.SubmarinerCluster, error) {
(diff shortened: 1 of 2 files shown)


pkg/datastore/phpapi/phpapi.go
```diff
@@ -48,59 +50,76 @@ func (p *PHPAPI) GetClusters(colorCodes []string) ([]types.SubmarinerCluster, er
...
 }
+
 func (p *PHPAPI) GetCluster(clusterID string) (types.SubmarinerCluster, error) {
 	requestURL := fmt.Sprintf("%s://%s/clusters.php?plurality=false&identifier=%s&cluster_id=%s", p.Proto, p.Server, p.APIToken, clusterID)
 	klog.V(8).Infof("request url: %s", requestURL)
 	clustersGetter, err := http.Get(requestURL)
 	if err != nil {
-		klog.Errorf("Encountered error while trying to retrieve clusters from the apiserver")
-		return types.SubmarinerCluster{}, err
+		return types.SubmarinerCluster{}, fmt.Errorf("error retrieving clusters from %s: %v", requestURL, err)
```
skitt: I know this is what the code already did, but wouldn’t it be better to return `nil` with the error?

pkg/datastore/phpapi/phpapi.go
```diff
@@ -48,59 +50,76 @@ func (p *PHPAPI) GetClusters(colorCodes []string) ([]types.SubmarinerCluster, er
...
+	if err = json.Unmarshal(endpointsRaw, &endpoints); err != nil {
+		return nil, fmt.Errorf("error unmarshalling JSON %s: %v", string(endpointsRaw[:]), err)
 	}
 	return endpoints, nil
 }
+
 func (p *PHPAPI) GetEndpoint(clusterID string, cableName string) (types.SubmarinerEndpoint, error) {
-	klog.Errorf("GetEndpoint not implemented")
-	return types.SubmarinerEndpoint{}, nil
+	return types.SubmarinerEndpoint{}, fmt.Errorf("GetEndpoint is not implemented")
```
skitt: Same here...

pkg/datastore/phpapi/phpapi.go
```diff
@@ -111,26 +130,26 @@ func (p *PHPAPI) WatchClusters(ctx context.Context, selfClusterID string, colorC
...
 			time.Sleep(5 * time.Second)
 		}
 	}()
 	wg.Wait()
 	klog.Errorf("I shouldn't have exited")
 	return nil
 }
-func (p *PHPAPI) WatchEndpoints(ctx context.Context, selfClusterID string, colorCodes []string, onChange func(endpoint types.SubmarinerEndpoint, deleted bool) error) error {
 
+func (p *PHPAPI) WatchEndpoints(ctx context.Context, selfClusterID string, colorCodes []string, onChange func(endpoint types.SubmarinerEndpoint, deleted bool) error) error {
```
skitt: This no longer returns errors, does it?

pkg/datastore/phpapi/phpapi.go
```diff
@@ -48,59 +50,76 @@ func (p *PHPAPI) GetClusters(colorCodes []string) ([]types.SubmarinerCluster, er
...
 	return endpoints, nil
 }
+
 func (p *PHPAPI) GetEndpoint(clusterID string, cableName string) (types.SubmarinerEndpoint, error) {
-	klog.Errorf("GetEndpoint not implemented")
-	return types.SubmarinerEndpoint{}, nil
+	return types.SubmarinerEndpoint{}, fmt.Errorf("GetEndpoint is not implemented")
 }
+
 func (p *PHPAPI) WatchClusters(ctx context.Context, selfClusterID string, colorCodes []string, onChange func(cluster types.SubmarinerCluster, deleted bool) error) error {
```
skitt: This no longer returns errors, AFAICT.

pkg/datastore/phpapi/phpapi.go
tpantelis: yeah probably but that changes the interface and call sites and outside the scope of this patch.

pkg/datastore/phpapi/phpapi.go
tpantelis: It didn't before - the only return is nil at the end of the method.

pkg/datastore/phpapi/phpapi.go
tpantelis: Same here...

pkg/datastore/phpapi/phpapi.go
skitt: Right, what I was really thinking was that, as part of the error-handling cleanup, it would make sense to change the method declaration to indicate that it doesn’t return anything.

tpantelis: yeah but it's part of the interface contract although, as it turns out, neither implementation (the other is kubernetes) returns an error. Of course when defining an interface it's safer to allow for error as you really can't anticipate the behavior of every implementation. In this case, performing the watch is assumed to be an async process so not much reason to return error.  I'll push another PR for that.

tpantelis: @skitt I'd like to progress this PR soon as it cleans up the remaining golangci-lint warnings 😄 Are you OK the way it is?

skitt: Yes, let’s get this merged.

mangelajo:
It looks good to me. It seems like this part is something to integrate with some rancher specific API (written in PHP) they didn't publish.  I'm doing a second pass to make sure this is not introducing any behaviour change which could negatively affect them.

tpantelis:
> It looks good to me. It seems like this part is something to integrate with some rancher specific API (written in PHP) they didn't publish. I'm doing a second pass to make sure this is not introducing any behaviour change which could negatively affect them.

yeah I'm not sure of the purpose of this implementation or if it's used - we should ask the Rancher folks at some point whether or not this should remain long term in this project.

PR #67, opened 2019-06-12
Declare all script targets as phony

skitt:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,4 +22,5 @@ trash-keep: .dapper
 
 .DEFAULT_GOAL := ci
 
-.PHONY: test
+.PHONY: $(TARGETS)
+


Makefile
```diff
@@ -18,5 +18,6 @@ trash-keep: .dapper
 
 .DEFAULT_GOAL := ci
 
-.PHONY: test
+.SECONDEXPANSION:
+.PHONY: $$(TARGETS)
```
tpantelis: Forgive me - I haven't worked with Makefiles in over 10 years 😄 - just curious what is the reason for using secondary expansion here as opposed to just `.PHONY: $(TARGETS)`?

skitt: No reason, it turns out — I was mis-remembering when secondary expansion is necessary! It’s only really useful with automatic variables (or variables which are re-set after the target definition).

mangelajo:
@skitt there's some conflict now. Github's UI get's stuck when trying to resolve here. Can you fix it? :)

skitt:
> @skitt there's some conflict now. Github's UI get's stuck when trying to resolve here. Can you fix it? :)

@mangelajo done!

PR #68, opened 2019-06-12
Change Datastore methods to use pointers

tpantelis:
Several Datastore methods pass or return by value rather than pointers.
Change to pointers for efficiency. This also cascaded to a couple util
methods.


tpantelis:
@skitt

PR #69, opened 2019-06-13
Simple instructions when e2e keep has finished

mangelajo:


mangelajo:
@skitt @dimaunx  ^ :-)

dimaunx:
Very helpful!

scripts/kind-e2e/e2e.sh
```diff
@@ -240,3 +240,14 @@ setup_broker
 setup_cluster2_gateway
 setup_cluster3_gateway
 test_connection
+
+if [[ $1 = keep ]]; then
+    echo "your 3 virtual clusters are deployed and working properly with your local"
+    echo "submariner source code, and can be accessed with:"
+    echo ""
+    echo "export KUBECONFIG=\$(echo \$(pwd)/output/kind-config/local-dev/kind-config-cluster{1..3} | sed 's/ /:/g')"
```
skitt: It’s a shame we can’t easily calculate the real value of `KUBECONFIG` for the user...

dimaunx:
If it is possible , could you also update the docs?
There are two very long export lines in the docs. The export that you suggest is a lot more efficient.

mangelajo:
> If it is possible , could you also update the docs?
> There are two very long export lines in the docs. The export that you suggest is a lot more efficient.

Makes sense, will do! :)

mangelajo:
Done :-)

scripts/kind-e2e/e2e.sh
manosnoam: in case not in root dir, the KUBECONFIG won't setup correctly. Run this instead:
`export KUBECONFIG=$(echo $(git rev-parse --show-toplevel)/output/kind-config/local-dev/kind-config-cluster{1..3} | sed 's/ /:/g')`

mangelajo:
Good suggestion, although to run make e2e you need to be on the root directory. I will update tomorrow Monday :)

Enviado desde mi iPhone

scripts/kind-e2e/e2e.sh
mangelajo: Done, thanks! :)

PR #70, opened 2019-06-13
Cleanup error handling in datastoresyncer

tpantelis:
- Propagate or log errors instead of calling Fatalf
- Propagate errors for functions that return error
- Improved logging messages where appropriate

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/70)


tpantelis:
@skitt

pkg/controllers/datastoresyncer/datastoresyncer.go
```diff
@@ -74,33 +74,35 @@ func NewDatastoreSyncer(thisClusterID string, objectNamespace string, kubeClient
...
+		return fmt.Errorf("error retrieving endpoints %v", err)
 	}
 
 	for _, endpoint := range endpoints {
 		if !util.CompareEndpointSpec(endpoint.Spec, d.localEndpoint.Spec) {
 			endpointCrdName, err := util.GetEndpointCRDName(endpoint)
 			if err != nil {
-				klog.Errorf("Error while converting endpoint to CRD Name %s", endpoint.Spec.CableName)
-				break
+				klog.Errorf("error converting endpoint %#v to CRD Name: %v", endpoint, err)
```
skitt: What’s the reasoning here (and elsewhere in the loop) which leads to logging an error rather than returning the error?

tpantelis: Originally I changed these sites to return error as well but I'm unsure as to the original intent. These are questions went thru my mind: Why was the error above fatal but not this one and the 2 below it... Why does it break here and not below...  Are these errors catastrophic enough to propagate and thus exit the process...  When deleting, would it return an error if the object doesn't exist...    The latter 2 made me squeamish about changing the behavior.

skitt: @Oats87 what do you think?

mangelajo:
There are some conflicts now, sorry.

mangelajo:
@tpantelis I updated this one, I'm fine to merge it so we move it out of backlog at this point.

tpantelis:
> @tpantelis I updated this one, I'm fine to merge it so we move it out of backlog at this point.

Thanks - I forgot about this one.

PR #73, opened 2019-06-14
OSX unit test running compatibility and speedups

mangelajo:
Skip one test that calls an unsuported netlink function in OSX,
the test will be skipped when the error is "not implemented".

Modify the scripts/test to make the find call compatible with OSX
and linux, also speeding up execution by avoiding "vendor" and
other unnecessary directories being explored by find.

Remove clutter from ginkgo output making it much cleaner.

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -28,7 +28,6 @@ var _ = Describe("Util", func() {
 
 	Describe("Function CompareEndpointSpec", testCompareEndpointSpec)
 
-	Describe("Function GetDefaultGatewayInterface", testGetDefaultGatewayInterface)
 })
 
 func testParseSecure() {
@@ -250,11 +249,5 @@ func testCompareEndpointSpec() {
 					Hostname:  "host2",
 				})).To(BeFalse())
 		})
-	})
-}
-
-func testGetDefaultGatewayInterface() {
-	It("should find and return a non-nil Interface object", func() {
-		Expect(util.GetDefaultGatewayInterface()).NotTo(BeNil())
 	})
 }
--- a/scripts/test
+++ b/scripts/test
@@ -3,10 +3,12 @@ set -e
 
 cd $(dirname $0)/..
 
+source $(dirname $0)/lib/find_functions
+
 echo Looking for packages to test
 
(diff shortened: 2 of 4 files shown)


mangelajo:
@skitt  @dimaunx ^

scripts/test
```diff
@@ -5,8 +5,16 @@ cd $(dirname $0)/..
 
 echo Looking for packages to test
 
-PACKAGES=". $(find -name '*.go' | xargs -I{} dirname {} |  cut -f2 -d/ | sort -u | grep -Ev '(^\.$|.git|.trash-cache|vendor|bin|test)' | sed -e 's!^!./!' -e 's!$!/...!')"
+PACKAGES=". $(find . -path ./vendor -prune -o \
+                     -path ./test/e2e -prune -o \
```
skitt: If/when we have Go files under `test` outside `test/e2e`, this will break, because the matching files under `test` will result in `test/...` being added to `PACKAGES`, which includes `test/e2e` again. I think we need to exclude `test` entirely for now.

skitt: See #74 for an alternate proposal (untested on macOS).

pkg/util/util_test.go
```diff
@@ -255,6 +256,10 @@ func testCompareEndpointSpec() {
 
 func testGetDefaultGatewayInterface() {
 	It("should find and return a non-nil Interface object", func() {
-		Expect(util.GetDefaultGatewayInterface()).NotTo(BeNil())
+		gatewayInterface, err := util.GetDefaultGatewayInterface()
+		if err == netlink.ErrNotImplemented {
+			Skip("netlink module does not implement RouteList on this OS")
+		}
+		Expect(gatewayInterface).NotTo(BeNil())
```
tpantelis: I'd actually prefer just to get rid of this test case as it really isn't all that useful since all we can do is verify it returns non-nil. In fact I omitted this in the Testify version but neglected to remove it in the Gingko version prior to merging.

scripts/test
mangelajo: oh, right, you're correct.

pkg/util/util_test.go
mangelajo: Ack, I will remove it in this patch.

mangelajo:
I have converted also the functional changes of https://github.com/rancher/submariner/pull/74 into this patch.

PR #74, opened 2019-06-14
Simplify find, restore it in scripts/validate

skitt:
This reduces the number of processes and filters involved in building
the list of packages, significantly reducing the execution time.  We’d
removed the dynamic construction of the directory list in
scripts/validate, this restores it.

### DIFF
--- a/scripts/lib/find_functions
+++ b/scripts/lib/find_functions
@@ -1,23 +1,21 @@
 function find_go_pkg_dirs() {
 
+    local BASE EXCLUDED_PKG_DIRS FIND_EXCLUDE PACKAGE_DIRS
+
     BASE=${2:-.}
-    EXCLUDED_PKG_DIRS=${EXCLUDED_PKG_DIRS:-"vendor test .git .trash-cache bin"}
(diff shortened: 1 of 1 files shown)


skitt:
> $ find . -regex './(.git|.trash-cache|vendor|bin|test)' -prune -o \

The backslashes are required...

mangelajo:
Sorry, I was playing around to try to make it work, the original has the same issue also:

https://paste.fedoraproject.org/paste/Fie4lWibbVRiQFv0CRqkpA

skitt:
> Sorry, I was playing around to try to make it work, the original has the same issue also:
> 
> https://paste.fedoraproject.org/paste/Fie4lWibbVRiQFv0CRqkpA

Indeed, macOS `find` needs `-E` but that doesn’t exist on GNU `find`!

mangelajo:
I have added the functional changes (validate) to this one too:

Creating a common function (which honestly could be better designed, but at least it's common to both.)

https://github.com/rancher/submariner/pull/73

skitt:
@mangelajo there’s still some stuff left over here ;-).

PR #78, opened 2019-06-18
Remove unused packages from the Dapper container

skitt:
gcc, ca-certificates, vim, less, and file are all unused.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -12,7 +12,7 @@ ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${ARC
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
 RUN apt-get -q update && \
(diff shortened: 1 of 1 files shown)


mangelajo:
Makes sense, could we keep vim & less, I sometimes use them through: 
./.dapper -s -m bind

which will drop you in a shell inside the dapper container. May be we should document that.

skitt:
> Makes sense, could we keep vim & less, I sometimes use them through:
> ./.dapper -s -m bind
> 
> which will drop you in a shell inside the dapper container. May be we should document that.

You can install them when you need them, instead of requiring every build to download them. Would that be good enough?

mangelajo:
True!

dimaunx:
Would it be possible to keep git? I would like to do  something like:
export KUBECONFIG=$(echo $(git rev-parse --show-toplevel)/output/kind-config/local-dev/kind-config-cluster{1..3} | sed 's/ /:/g') 
This is really helpful to prevent directory structure calculations. This will help further work removing dapper.

skitt:
> Would it be possible to keep git? I would like to do something like:

Done!

PR #80, opened 2019-06-18
NetworkPolicy support use cases definition.

mangelajo:
Creates a basic document with the basic use cases to support with
NetworkPolicies.

Please note this document does not define the technical solution
yet but tries to define the scope of the problem to solve.

The rendered document can be observed here: https://github.com/mangelajo/submariner/blob/network-policy-use-cases/docs/design/network-policy-support.md

Co-Authored-By: Janki Chhatbar <jchhatba@redhat.com>
Co-Authored-By: Miguel Angel Ajo <majopela@redhat.com>


docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
+
+## Multicluster Network Policy support
+
+This document describes the Network Policy use cases that we propose
+supporting for a first iteration in Submariner. In this iteration, the
+intention is to support the ingress side of existing k8s Network Policies
+as is, while in the future we could explore with the kubernetes community
+options to extend Network Policies for multi cluster scenarios.
+
+> A Kubernetes, network policy is a specification of how pods are allowed
```
tpantelis: remove comma

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+> to communicate with each other and other network endpoints. It is like
+> Access Control Lists using which we can specify the traffic that would
```
tpantelis: "It is similar to Access Control Lists where one can specify..."

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+> be allowed/denied to the PODs. Network Policies are additive, so having
```
tpantelis: PODs -> pods. Elsewhere in the doc, "pod" is lower case - we should be consistent.

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+> two Network Policies that select the same Pods will result in allowing
```
tpantelis: Pods -> pods.

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+> both defined policies.
+
+
+> Network Policies are implemented by the networking plugin which is the
+> CNI for the K8s cluster. If the underlying network plugin does not
```
tpantelis: How about "by the underlying  networking CNI plugin for the k8s cluster"?

Also the first statement states Network Policies are implemented by the CNI but the second statement sort of contradicts that by indicating it may not. I think it would be clearer to say "Network Policies are assumed to be implemented by the...".

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+> implement the Network Policies, then creating a NetworkPolicy will have
+> no effect.
+
+## Diagram
+
+The following diagram describes the example resources which will be
+analized in the use cases. It's made of two clusters, and
```
tpantelis: analized -> analyzed (sic)

Also, how about "It depicts two clusters and two namespaces."

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+two namespaces.
+
+![Diagram](img/network-policies-base-diagram.png)
+
+The cluster labels for the namespaces are just an illustrative example of
```
tpantelis: example -> examples

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+label assignment. The admin could follow that pattern or a different
+pattern. In future, we could integrate with federation
+[ClusterID](https://docs.google.com/document/d/1F__vEKeI41P7PPUCMM9PVPYY34pyrvQI5rbTJVnS5c4/edit)
+which is being discussed in the k8s multicluster-sig
```
tpantelis: add '.' at end

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+
+For the initial implementation (still under design) assume that
+POD/Service CIDRs are non-overlapping, and that we preserve the
```
tpantelis: POD -> pod. Service -> service

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+
+![Diagram](img/network-policies-base-diagram.png)
+
+The cluster labels for the namespaces are just an illustrative example of
+label assignment. The admin could follow that pattern or a different
+pattern. In future, we could integrate with federation
+[ClusterID](https://docs.google.com/document/d/1F__vEKeI41P7PPUCMM9PVPYY34pyrvQI5rbTJVnS5c4/edit)
+which is being discussed in the k8s multicluster-sig
+
+For the initial implementation (still under design) assume that
```
tpantelis: "assume" -> "it is assumed"

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+POD/Service CIDRs are non-overlapping, and that we preserve the
+source IP address on the datapath. In parallel we are also considering
+technical solutions to be able to support overlapping CIDRs in the future.
+
+The examples shown below can be extrapolated to UDP/SCTP too.
+
+We are only trying to taclke Ingress policies for a first iteration, same
```
tpantelis: taclke -> tackle (sic)

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+design could apply to egress policy on later iterations.
```
tpantelis: on -> for

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+|    3     | Allow all the PODs from the same project in a remote cluster to access the backend POD on the local cluster. |
+|    4     | Allow frontend POD from a different project in a remote cluster to access the backend POD on the local cluster. |
+|    5     | Allow all the PODs from the same project which are either in the local cluster or remote cluster to access the logging POD on the local cluster. |
+
+
+### 1. Connected Clusters with No Network Policies
+Normally in a single Cluster when there are no network policies, it implies that
+every pod can connect to other pods without any restriction. When two (or more)
+clusters are connected via Submariner, we would preserve the same behavior across
+the clusters (i.e., PODs in one cluster can access PODs in remote clusters) until
```
tpantelis: PODs -> pods

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+| -------  | ------------------------------------------------------- |
+|    1     | Connected Clusters without any Network Policy           |
+|    2     | Allow frontend pods from a remote cluster in the same project to access the backend pods on the local cluster. |
+|    3     | Allow all the PODs from the same project in a remote cluster to access the backend POD on the local cluster. |
+|    4     | Allow frontend POD from a different project in a remote cluster to access the backend POD on the local cluster. |
+|    5     | Allow all the PODs from the same project which are either in the local cluster or remote cluster to access the logging POD on the local cluster. |
+
+
+### 1. Connected Clusters with No Network Policies
+Normally in a single Cluster when there are no network policies, it implies that
```
tpantelis: Cluster -> cluster

Also,  ", it implies" -> "which implies"

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+    matchLabels:
+      role: backend
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels: # this could be {} to select all
+          project: fancy
+	    cluster: west
```
tpantelis: remove tab?

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+    matchLabels:
+      role: backend
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          project: fancy
+	    cluster: west
```
tpantelis: tab?

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,247 @@
...
+    matchLabels:
+      role: logging
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchExpressions:
+	    - {key: cluster, operator: In, values: [east, west, north]}
+	  matchLabels:
```
tpantelis: tabs

tpantelis:
Looks good - just grammar/wording nits.

mangelajo:
Thanks Thomas!, fixing the grammar/wording now.

docs/design/network-policy-support.md
mangelajo: Yeah, we are trying to explain that some CNIs don't implement it. I'll try to rephrase.

docs/design/network-policy-support.md
mangelajo: oops :D

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,246 @@
...
+
+
+| Use-case |       Description                                       |
+| -------  | ------------------------------------------------------- |
+|    1     | Connected Clusters without any Network Policy           |
+|    2     | Allow frontend pods from a remote cluster in the same project to access the backend pods on the local cluster. |
+|    3     | Allow all the pods from the same project in a remote cluster to access the backend pod on the local cluster. |
+|    4     | Allow frontend pod from a different project in a remote cluster to access the backend pod on the local cluster. |
+|    5     | Allow all the pods from the same project which are either in the local cluster or remote cluster to access the logging pod on the local cluster. |
+
```
mangelajo: two extra use cases to define:

6) Handle when a new cluster is added to the mesh
7) Handle when a cluster is removed from the mesh

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,246 @@
...
+![Diagram](img/network-policies-base-diagram.png)
+
+The cluster labels for the namespaces are just illustrative examples of
+label assignment. The admin could follow that pattern or a different
+pattern. In future, we could integrate with federation
+[ClusterID](https://docs.google.com/document/d/1F__vEKeI41P7PPUCMM9PVPYY34pyrvQI5rbTJVnS5c4/edit)
+which is being discussed in the k8s multicluster-sig.
+
+For the initial implementation (still under design) it is assumed that
+pod/service CIDRs are non-overlapping, and that we preserve the
```
thockin: If we're talking about network policy, service CIDR should not be a factor at all?

docs/design/network-policy-support.md
```diff
@@ -0,0 +1,246 @@
...
+design could apply to egress policy for later iterations.
+
+## Use cases
+
+
+
+| Use-case |       Description                                       |
+| -------  | ------------------------------------------------------- |
+|    1     | Connected Clusters without any Network Policy           |
+|    2     | Allow frontend pods from a remote cluster in the same project to access the backend pods on the local cluster. |
```
thockin: What is a project?  That's not a kubernetes term of art?  It's not clearly a namespace, because a namespace subdivides a cluster, and you said "remote cluster in the same project".  So I can't evaluate this very well.

docs/design/network-policy-support.md
mangelajo: True, service CIDRs aren't a factor here.

mangelajo: > The TL;DR of this seems to be "it works as if they were one cluster", right?

That is correct, at least as a converstation starter, this is a very naive design where we need to look at scale, and security implications.

> A few thoughts after a read-thru:
> 
> The thing I worry about here is scale. How many implementations of NetworkPolicy are ready to 3x or 5x or 10x their scale?

Correct, and agreed. We need to figure out a way to define how policies will be handled from cluster to cluster, one possibility being adding the ability able to select remote clusters as sources/destination of traffic, that minimizes the scale of the problem and the amount of data to be exchanged between clusters to implement those policies.

> 
> We also need to think about the implementation mechanisms here - do we expect NetworkPolicy impl on each cluster to carry auth tokens for every other cluster, or do we think we need to pre-cook this data and store it in a central place.

I didn't want to dive into implementation mechanisms yet. Although I have ideas.


When you talk about tokens, are you referring to clusters don't trust each other, but still administrators want to connect them and let the applications communicate, and you let that happen based in some sort of token associated to pods/network policies? Can you elaborate the idea about auth tokens?
 
> If I install a policy in the east cluster, does it apply in the west cluster, or is it asymmetric (target is local cluster, source is across clusters) ? I suppose this is why you dodged "egress" for now?

I was proposing asymmetric. If you install the policy in one cluster, it would affect only the pods in such cluster.
In the other hand, if you used a FederatedNetworkPolicy, such one could be deployed down to all clusters, and would become symmetric.

> Do we expect that labelling namespaces with their cluster name is going to be sufficient? Or should that be more first-class? That's a bigger change.

We proposed that as a first step, to have something tangible on a first iteration, and I was considering a 2nd step that works with a more first-class change, by creating "MulticlusterNetworkPolicy", enhancing the existing "NetworkPolicy" to have a cluster selector, or... anything else we find more reasonable, that discussion needs to happen, as you said on a wider audience.


> 
> You seem to be introducing a concept of "cross-cluster namespace", which I am actually OK with, but I think it needs to be spelled out and its semantics documented VERY clearly.

I guess it's something that could even be managed from federation with the FederatedNamespace?

docs/design/network-policy-support.md
mangelajo: Right, the correct term here would be namespace.

docs/design/network-policy-support.md
thockin: > > The TL;DR of this seems to be "it works as if they were one cluster", right?
> 
> That is correct, at least as a converstation starter, this is a very naive design where we need to look at scale, and security implications.

It's a fine starting point :)  I think the conversation should happen within sig-multi-cluster rather than in the comments of a specific impl, though :)

> > A few thoughts after a read-thru:
> > The thing I worry about here is scale. How many implementations of NetworkPolicy are ready to 3x or 5x or 10x their scale?
> 
> Correct, and agreed. We need to figure out a way to define how policies will be handled from cluster to cluster, one possibility being adding the ability able to select remote clusters as sources/destination of traffic, that minimizes the scale of the problem and the amount of data to be exchanged between clusters to implement those policies.

I am very interested in a sort of bainstorm on these sorts of things.  I'd like to propose a sig-multicluster face-to-face meeting to get deep in on some of them.
 
> > We also need to think about the implementation mechanisms here - do we expect NetworkPolicy impl on each cluster to carry auth tokens for every other cluster, or do we think we need to pre-cook this data and store it in a central place.
> 
> I didn't want to dive into implementation mechanisms yet. Although I have ideas.

Imp thoughts can help shape ideas.  They can also be self-limiting, I admit, but I personally can't talk about an idea too far without thinking about hand-waving ways it might work.

> When you talk about tokens, are you referring to clusters don't trust each other, but still administrators want to connect them and let the applications communicate, and you let that happen based in some sort of token associated to pods/network policies? Can you elaborate the idea about auth tokens?

Yeah, specifically when I add a cluster, do I need to go to N other clusters, make a service account, download the token, and share it with the new cluster, and vice-versa?  2 clusters is easy, teach each about the other.  3 clusters is a bit worse - each cluster needs to SAs and needs to carry 2 secrets for the other clusters.  4 is starting to smell bad.  5 is silly...
 
> > If I install a policy in the east cluster, does it apply in the west cluster, or is it asymmetric (target is local cluster, source is across clusters) ? I suppose this is why you dodged "egress" for now?
> 
> I was proposing asymmetric. If you install the policy in one cluster, it would affect only the pods in such cluster.
> In the other hand, if you used a FederatedNetworkPolicy, such one could be deployed down to all clusters, and would become symmetric.

If that's a model we want to pursue.  There are distinct tradeoffs of synchronization vs non.

> > Do we expect that labelling namespaces with their cluster name is going to be sufficient? Or should that be more first-class? That's a bigger change.
> 
> We proposed that as a first step, to have something tangible on a first iteration, and I was considering a 2nd step that works with a more first-class change, by creating "MulticlusterNetworkPolicy", enhancing the existing "NetworkPolicy" to have a cluster selector, or... anything else we find more reasonable, that discussion needs to happen, as you said on a wider audience.

Yes, thanks.

> > You seem to be introducing a concept of "cross-cluster namespace", which I am actually OK with, but I think it needs to be spelled out and its semantics documented VERY clearly.
> 
> I guess it's something that could even be managed from federation with the FederatedNamespace?

It could be, again, if federation is a pattern to follow.  I think our real duty here is to find common patterns without locking users into one implementation.  E.g. we can assert that a cross-cluster-namespace is a thing, and what that means semantically, without defining how those namespaces came to exist.

But I have some lingering thoughts about even this primitive topic.  I am still working on a doc.  Need more hours in the day...

mangelajo: > > > The TL;DR of this seems to be "it works as if they were one cluster", right?
> > 
> > 
> > That is correct, at least as a converstation starter, this is a very naive design where we need to look at scale, and security implications.
> 
> It's a fine starting point :) I think the conversation should happen within sig-multi-cluster rather than in the comments of a specific impl, though :)

Of course, this was more result of timing, also github for discussing this is becoming horrible...
I see that normally the discussion is done over google docs, so I'll move it into a google doc.

Can you also bring the people you mentioned you were discussing about this topic to the discussion?

> 
> > > A few thoughts after a read-thru:
> > > The thing I worry about here is scale. How many implementations of NetworkPolicy are ready to 3x or 5x or 10x their scale?
> > 
> > 
> > Correct, and agreed. We need to figure out a way to define how policies will be handled from cluster to cluster, one possibility being adding the ability able to select remote clusters as sources/destination of traffic, that minimizes the scale of the problem and the amount of data to be exchanged between clusters to implement those policies.
> 
> I am very interested in a sort of bainstorm on these sorts of things. I'd like to propose a sig-multicluster face-to-face meeting to get deep in on some of them.

That would be indeed very interesting, f2f generally works faster than video conference/shared docs, a good old whiteboard always works better. I will circulate that idea. If we do a F2F I think it's more worth the investment if we can also discuss around DNS propagation and service discovery.

> 
> > > We also need to think about the implementation mechanisms here - do we expect NetworkPolicy impl on each cluster to carry auth tokens for every other cluster, or do we think we need to pre-cook this data and store it in a central place.
> > 
> > 
> > I didn't want to dive into implementation mechanisms yet. Although I have ideas.
> 
> Imp thoughts can help shape ideas. They can also be self-limiting, I admit, but I personally can't talk about an idea too far without thinking about hand-waving ways it might work.

It's true that an idealised behaviour that doesn't scale or that it's not feasibly implemented does not help.. let's try to shape the implementation ideas in the briefest possible manner then.

> 
> > When you talk about tokens, are you referring to clusters don't trust each other, but still administrators want to connect them and let the applications communicate, and you let that happen based in some sort of token associated to pods/network policies? Can you elaborate the idea about auth tokens?
> 
> Yeah, specifically when I add a cluster, do I need to go to N other clusters, make a service account, download the token, and share it with the new cluster, and vice-versa? 2 clusters is easy, teach each about the other. 3 clusters is a bit worse - each cluster needs to SAs and needs to carry 2 secrets for the other clusters. 4 is starting to smell bad. 5 is silly...

Configuration needs to be easy, and at the same time allow for security. This area needs exploration, yes...

> 
> > > If I install a policy in the east cluster, does it apply in the west cluster, or is it asymmetric (target is local cluster, source is across clusters) ? I suppose this is why you dodged "egress" for now?
> > 
> > 
> > I was proposing asymmetric. If you install the policy in one cluster, it would affect only the pods in such cluster.
> > In the other hand, if you used a FederatedNetworkPolicy, such one could be deployed down to all clusters, and would become symmetric.
> 
> If that's a model we want to pursue. There are distinct tradeoffs of synchronization vs non.

What do you mean vs non?

We will always need to synchronize labels->pod ip address lists between clusters, so that policies can be properly interpreted on the clusters where the policies are applied.

> 
> > > Do we expect that labelling namespaces with their cluster name is going to be sufficient? Or should that be more first-class? That's a bigger change.
> > 
> > 
> > We proposed that as a first step, to have something tangible on a first iteration, and I was considering a 2nd step that works with a more first-class change, by creating "MulticlusterNetworkPolicy", enhancing the existing "NetworkPolicy" to have a cluster selector, or... anything else we find more reasonable, that discussion needs to happen, as you said on a wider audience.
> 
> Yes, thanks.
> 
> > > You seem to be introducing a concept of "cross-cluster namespace", which I am actually OK with, but I think it needs to be spelled out and its semantics documented VERY clearly.
> > 
> > 
> > I guess it's something that could even be managed from federation with the FederatedNamespace?
> 
> It could be, again, if federation is a pattern to follow. I think our real duty here is to find common patterns without locking users into one implementation. E.g. we can assert that a cross-cluster-namespace is a thing, and what that means semantically, without defining how those namespaces came to exist.

Correct.

> 
> But I have some lingering thoughts about even this primitive topic. I am still working on a doc. Need more hours in the day...

heh, I know what you mean, let's open that discussion on the sig too.

mangelajo:
Ok, I'm moving this discussion here: https://docs.google.com/document/d/1_QzuasJPiQ-4t8tUGODoRS2E-Q3KLJajqMyCvLPH2tY/edit?usp=sharing and I will announce it on the multicluster sig to get more eyes on it.

mangelajo:
Closing as the discussion is now moved the the google doc.

PR #81, opened 2019-06-19
Add kubefed installation to E2E flow as an option

dimaunx:
This PR is adding kubefed=true command to E2E flow.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,7 +4,11 @@ FROM ubuntu:16.04
 ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPER_ENV=TAG \
     DAPPER_SOURCE=/go/src/github.com/rancher/submariner DAPPER_DOCKER_SOCKET=true \
-    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} LINT_VERSION=v1.16.0 DAPPER_OUTPUT=output
+    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} DAPPER_OUTPUT=output \
+    LINT_VERSION=v1.16.0 \
+    HELM_VERSION=v2.14.0 \
+    KIND_VERSION=v0.3.0 \
+    KUBEFED_VERSION=0.1.0-rc2
 
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 
@@ -18,11 +22,13 @@ RUN apt-get -q update && \
     curl -Lo /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${ARCH}/kubectl && \
     chmod a+x /usr/bin/kubectl && \
     curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin ${LINT_VERSION} && \
-    curl https://storage.googleapis.com/kubernetes-helm/helm-v2.14.0-linux-amd64.tar.gz | tar -xzf - && \
-    cp linux-${ARCH}/helm /usr/bin/ && \
-    curl -Lo /usr/bin/kind https://github.com/kubernetes-sigs/kind/releases/download/v0.3.0/kind-linux-${ARCH} && chmod a+x /usr/bin/kind && \
+    curl "https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-${ARCH}.tar.gz" | tar -xzf - && \
+    cp linux-${ARCH}/helm /usr/bin/ && chmod a+x /usr/bin/helm && \
+    curl -LO "https://github.com/kubernetes-sigs/kubefed/releases/download/v${KUBEFED_VERSION}/kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz" && \
+    tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
+    curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
     go get -v github.com/onsi/ginkgo/ginkgo && go get -v github.com/onsi/gomega/... && \
-    go get golang.org/x/tools/cmd/goimports
+    go get -v golang.org/x/tools/cmd/goimports
 
 WORKDIR ${DAPPER_SOURCE}
 
--- a/scripts/kind-e2e/README.md
+++ b/scripts/kind-e2e/README.md
@@ -103,6 +103,23 @@ the following lucene query example can be used to query the logs.
 kubernetes.namespace:"submariner" AND kubernetes.node.name: (cluster2* OR cluster3*) AND kubernetes.labels.app: submariner*
 ```
 
+#### Federation V2
+Providing **kubefed=true** parameter to **make e2e** command will setup federation.
+Federation control plane will be created on cluster1 and clusters 2/3 will be added as members. 
+
+```bash
+make e2e status=keep kubefed=true
+```
+
+To get the status of federated clusters:
+
+```bash
+kubectl -n kube-federation-system get kubefedclusters
+``` 
+
+Federated deployment example resides in **scripts/kind-e2e/nginx-federated.sh**. 
+To federate resources across the clusters [kubefedctl] tool must be installed on the local system.
+
 #### Cleanup
 At any time you can run a cleanup command that will remove kind resources.
 
@@ -120,7 +137,7 @@ docker system prune --all
 #### Full example
 
 ```bash
-make e2e status=keep version=1.14.1 logging=true
+make e2e status=keep version=1.14.1 logging=true kubefed=true
 ```
 
 <!--links-->
@@ -129,7 +146,7 @@ make e2e status=keep version=1.14.1 logging=true
 [kubectl]: https://kubernetes.io/docs/tasks/tools/install-kubectl/
 [k9s]: https://github.com/derailed/k9s
 [kubetail]: https://github.com/johanhaleby/kubetail
+[kubefedctl]: https://github.com/kubernetes-sigs/kubefed/releases
 [kind release page]: https://github.com/kubernetes-sigs/kind/releases
 [Change SELinux to permissive]: https://www.ibm.com/support/knowledgecenter/en/POWER8/p8ef9/p8ef9_selinux_setup.htm
 [Add local user to docker group]: https://docs.docker.com/install/linux/linux-postinstall/
-
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -65,6 +65,7 @@ function setup_cluster2_gateway() {
             echo Submariner already installed, skipping submariner helm installation...
             update_subm_pods
         else
+            echo Installing submariner on cluster2...
             worker_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster2-worker | head -n 1)
             kubectl label node cluster2-worker "submariner.io/gateway=true" --overwrite
             helm install submariner-latest/submariner \
@@ -101,6 +102,7 @@ function setup_cluster3_gateway() {
             echo Submariner already installed, skipping submariner helm installation...
             update_subm_pods
         else
+            echo Installing submariner on cluster3...
             worker_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster3-worker | head -n 1)
             kubectl label node cluster3-worker "submariner.io/gateway=true" --overwrite
             helm install submariner-latest/submariner \
@@ -164,10 +166,10 @@ function test_connection() {
 
 function update_subm_pods() {
     echo Removing submariner engine pods...
-    kubectl get pods -n submariner -l app=submariner-engine | awk 'FNR > 1 {print $1}' | xargs kubectl delete pods -n submariner
+    kubectl delete pods -n submariner -l app=submariner-engine
     kubectl wait --for=condition=Ready pods -l app=submariner-engine -n submariner --timeout=60s
     echo Removing submariner route agent pods...
-    kubectl get pods -n submariner -l app=submariner-routeagent | awk 'FNR > 1 {print $1}' | xargs kubectl delete pods -n submariner
+    kubectl delete pods -n submariner -l app=submariner-routeagent
     kubectl wait --for=condition=Ready pods -l app=submariner-routeagent -n submariner --timeout=60s
 }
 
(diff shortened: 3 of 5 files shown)


Dockerfile.dapper
```diff
@@ -18,11 +19,13 @@ RUN apt-get -q update && \
...
-    cp linux-${ARCH}/helm /usr/bin/ && \
-    curl -Lo /usr/bin/kind https://github.com/kubernetes-sigs/kind/releases/download/v0.3.0/kind-linux-${ARCH} && chmod a+x /usr/bin/kind && \
+    curl "https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-${ARCH}.tar.gz" | tar -xzf - && \
+    cp linux-${ARCH}/helm /usr/bin/ && chmod a+x /usr/bin/helm && \
+    curl -LO "https://github.com/kubernetes-sigs/kubefed/releases/download/v${KUBEFED_VERSION}/kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz" && \
+    tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
+    curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
     go get -v github.com/onsi/ginkgo/ginkgo && go get -v github.com/onsi/gomega/... && \
-    go get golang.org/x/tools/cmd/goimports
+    go get -v golang.org/x/tools/cmd/goimports
```
mangelajo: Just a question out of curiosity (not related to this patch, and no action required):

is there any big advantage in doing it all in one big run with && .. && .. && 
beyond avoiding several layers? Is the layer overhead too big?

It's a little bit annoying to wait for the whole container being rebuilt for a simple change :) :-)

scripts/kind-e2e/e2e.sh
```diff
@@ -190,6 +192,25 @@ function enable_logging() {
...
+function enable_kubefed() {
+    kubectl config use-context cluster1
+    if kubectl rollout status deploy/kubefed-controller-manager -n ${KUBEFED_NS} > /dev/null 2>&1; then
+        echo Kubefed already installed, skipping setup...
+    else
+        helm repo add kubefed-charts https://raw.githubusercontent.com/kubernetes-sigs/kubefed/master/charts
+        helm install kubefed-charts/kubefed --version=0.1.0-rc2 --name kubefed --namespace ${KUBEFED_NS} --set controllermanager.replicaCount=1 > /dev/null 2>&1
+        for i in 1 2 3; do
+            kubefedctl join cluster${i} --cluster-context cluster${i} --host-cluster-context cluster1 --v=2
+            #master_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster${i}-control-plane | head -n 1)
```
mangelajo: Will we need those bits of commented code?, it'd be nice to clean up :)

scripts/kind-e2e/nginx-federated.sh
```diff
@@ -0,0 +1,92 @@
...
+set -e
+
+TEST_NS="test-ns"
+export KUBECONFIG=$(echo $(git rev-parse --show-toplevel)/output/kind-config/local-dev/kind-config-cluster{1..3} | sed 's/ /:/g')
+kubectl config use-context cluster1
+
+echo "Creating a test namespace ${TEST_NS}."
+kubectl create ns ${TEST_NS}
+
+echo "Creating a ngnix application resources in namespace ${TEST_NS}."
```
mangelajo: ngnix -> nginx

scripts/kind-e2e/nginx-federated.sh
```diff
@@ -0,0 +1,92 @@
...
+        volumeMounts:
+        - name: federated-web-file
+          mountPath: /usr/share/nginx/html/
+      volumes:
+        - name: federated-web-file
+          configMap:
+            name: federated-web-file
+EOF
+
+cat <<EOF | kubectl apply -n ${TEST_NS} -f -
```
mangelajo: I believe you can also do:

kubectl apply -n ${TEST_NS} -f - <<EOF

(I haven't tried though..)

scripts/kind-e2e/nginx-federated.sh
```diff
@@ -0,0 +1,92 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: federated-web-file
+data:
+  index.html: "Hello from Kubernetes Cluster Federation!"
+EOF
+
+cat <<EOF | kubectl apply -n ${TEST_NS} -f -
```
mangelajo: kubectl apply -n ${TEST_NS} -f - <<EOF

scripts/kind-e2e/nginx-federated.sh
```diff
@@ -0,0 +1,92 @@
...
+
+TEST_NS="test-ns"
+export KUBECONFIG=$(echo $(git rev-parse --show-toplevel)/output/kind-config/local-dev/kind-config-cluster{1..3} | sed 's/ /:/g')
+kubectl config use-context cluster1
+
+echo "Creating a test namespace ${TEST_NS}."
+kubectl create ns ${TEST_NS}
+
+echo "Creating a ngnix application resources in namespace ${TEST_NS}."
+cat <<EOF | kubectl apply -n ${TEST_NS} -f -
```
mangelajo: kubectl apply -n ${TEST_NS} -f - <<EOF

scripts/kind-e2e/e2e.sh
dimaunx: Want to keep this as a placeholder for future @skitt work. He will require it to make e2e to work without dapper.

Dockerfile.dapper
dimaunx: We can create more layers with separate RUN, but each later is 1.4 GB. We wanted to minimize the number of layers. The alternative is bunch of 1.4 GB layers laying around :).

dimaunx: IN any case we should host the build image on some public docker registry and pull it only once, this will improve the times considerably without the need to wait for dependencies on every change.

Dockerfile.dapper
```diff
@@ -4,7 +4,8 @@ FROM ubuntu:16.04
 ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPER_ENV=TAG \
     DAPPER_SOURCE=/go/src/github.com/rancher/submariner DAPPER_DOCKER_SOCKET=true \
-    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} LINT_VERSION=v1.16.0 DAPPER_OUTPUT=output
+    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} DAPPER_OUTPUT=output \
+    LINT_VERSION=v1.17.1 HELM_VERSION=v2.14.1 KIND_VERSION=v0.3.0 KUBEFED_VERSION=0.1.0-rc2
```
skitt: Could you split these up into separate lines, like #82? It means version bumps produce nice diffs in future...

pkg/client/clientset/versioned/clientset.go
```diff
@@ -28,8 +28,6 @@ import (
 type Interface interface {
 	Discovery() discovery.DiscoveryInterface
 	SubmarinerV1() submarinerv1.SubmarinerV1Interface
-	// Deprecated: please explicitly pick a version if possible.
```
skitt: This shouldn’t be part of this patch.

pkg/client/clientset/versioned/fake/clientset_generated.go
```diff
@@ -75,8 +75,3 @@ var _ clientset.Interface = &Clientset{}
 func (c *Clientset) SubmarinerV1() submarinerv1.SubmarinerV1Interface {
 	return &fakesubmarinerv1.FakeSubmarinerV1{Fake: &c.Fake}
 }
-
```
skitt: This shouldn’t be part of this patch either.

pkg/controllers/datastoresyncer/datastoresyncer.go
```diff
@@ -83,7 +83,7 @@ func (d *DatastoreSyncer) ensureExclusiveEndpoint() {
 
 	for _, endpoint := range endpoints {
 		if !util.CompareEndpointSpec(endpoint.Spec, d.localEndpoint.Spec) {
-			endpointCrdName, err := util.GetEndpointCRDName(endpoint)
+			endpointCrdName, err := util.GetEndpointCRDName(&endpoint)
```
skitt: This also looks like it was pulled in from another patch, is that right?

pkg/datastore/datastore.go
```diff
@@ -14,22 +14,33 @@ import (
 type Datastore interface {
 	// This gets all clusters that match the color code
 	GetClusters(colorCodes []string) ([]types.SubmarinerCluster, error)
+
 	// This gets a single cluster based on cluster ID and returns a v1.Cluster or error based on what it retrieves
-	GetCluster(clusterID string) (types.SubmarinerCluster, error)
```
skitt: Likewise, this is from another patch.

scripts/kind-e2e/e2e.sh
```diff
@@ -39,7 +39,7 @@ function install_helm() {
             echo Installing helm on cluster${i}.
             kubectl -n kube-system create serviceaccount tiller
             kubectl create clusterrolebinding tiller --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
-            helm init --service-account tiller
+            helm init --service-account tiller > /dev/null 2>&1
```
skitt: I know the output here is boring, but IME it’s better to keep the output because stuff ends up going wrong :-(.

scripts/kind-e2e/e2e.sh
```diff
@@ -51,7 +51,7 @@ function setup_broker() {
         echo Submariner CRDs already exist, skipping broker creation...
     else
         echo Installing broker on cluster1.
-        helm install submariner-latest/submariner-k8s-broker --name ${SUBMARINER_BROKER_NS} --namespace ${SUBMARINER_BROKER_NS}
+        helm install submariner-latest/submariner-k8s-broker --name ${SUBMARINER_BROKER_NS} --namespace ${SUBMARINER_BROKER_NS} > /dev/null 2>&1
```
skitt: Same here (and in other places further down), please keep the output. (Even if we do end up wanting to get rid of it, it should be a separate patch.)

dimaunx: > Same here (and in other places further down), please keep the output. (Even if we do end up wanting to get rid of it, it should be a separate patch.)

Yes i agree

scripts/kind-e2e/e2e.sh
dimaunx: > I know the output here is boring, but IME it’s better to keep the output because stuff ends up going wrong :-(.

Yes i know what you mean, the outputs are way too long :), i have added echo to each function but you are correct we probably should keep the output.

scripts/kind-e2e/nginx-federated.sh
dimaunx: Fixed, thank you.

pkg/client/clientset/versioned/clientset.go
dimaunx: Sorry for the noise, this should not be here.

pkg/client/clientset/versioned/fake/clientset_generated.go
dimaunx: Sorry for the noise, this should not be here.

pkg/controllers/datastoresyncer/datastoresyncer.go
dimaunx: Sorry for the noise, this should not be here, bad rebase.

pkg/datastore/datastore.go
dimaunx: Sorry for the noise, this should not be here, bad rebase.

Dockerfile.dapper
dimaunx: Done with the latest commit.

Dockerfile.dapper
mangelajo: 1.4GB! :-o why that much? :D

scripts/kind-e2e/e2e.sh
mangelajo: ack then!

Dockerfile.dapper
dimaunx: apt-get update

Dockerfile.dapper
```diff
@@ -4,7 +4,11 @@ FROM ubuntu:16.04
 ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPER_ENV=TAG \
     DAPPER_SOURCE=/go/src/github.com/rancher/submariner DAPPER_DOCKER_SOCKET=true \
-    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} LINT_VERSION=v1.16.0 DAPPER_OUTPUT=output
+    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} DAPPER_OUTPUT=output \
+    LINT_VERSION=v1.17.1 \
+    HELM_VERSION=v2.14.1 \
```
mangelajo: You're still bumping the helm version (I believe skitt was asking to separate on a different PR).

scripts/kind-e2e/README.md
```diff
@@ -103,6 +103,23 @@ the following lucene query example can be used to query the logs.
 kubernetes.namespace:"submariner" AND kubernetes.node.name: (cluster2* OR cluster3*) AND kubernetes.labels.app: submariner*
 ```
 
+#### Federation V2
+Providing **kubefed=true** parameter to **make e2e** command will setup federation.
+Federation control plain will be created on cluster1 and clusters 2/3 will be added as members.
```
mangelajo: plain=plane

Dockerfile.dapper
dimaunx: > You're still bumping the helm version (I believe skitt was asking to separate on a different PR).

I think he requested the environment variables for versions to be on separate lines.

skitt: > > You're still bumping the helm version (I believe skitt was asking to separate on a different PR).
> 
> I think he requested the environment variables for versions to be on separate lines.

Yes, but I also like changes to be isolated ;-). This PR should only be about kubefed integration, and version bumps unrelated to that should be separate IMO. If the kubefed integration part requires version bumps, that should be documented in the commit message, ideally with a link to the changelog for the new version (see #82 for Helm).

dimaunx: Yes you are right i have pushed too many changes. Will revert the versions and open a separate PR.

scripts/kind-e2e/nginx-federated.sh
dimaunx: Done

scripts/kind-e2e/nginx-federated.sh
dimaunx: Done

scripts/kind-e2e/nginx-federated.sh
dimaunx: Done

scripts/kind-e2e/e2e.sh
dimaunx: Done with latest commit.

Dockerfile.dapper
dimaunx: Done!

Dockerfile.dapper
dimaunx: @mangelajo After some testing you are right, better solution would be separate each install step in to its own RUN command. Storage is less expensive then time :).

mangelajo: Thanks!! :)

PR #83, opened 2019-06-19
Fix for ginkgo e2e recurring runs

dimaunx:
Fix issue with ginkgo between consecutive runs:
In order to **make e2e status=keep** to be executed successfully multiple times in a row, the kubeconfig for e2e tests must be read from shared persistent output directory and not from ephemeral build container, as the build container is discarded every run.

Remove hard coded directory locations:
Auto calculate directory structure based on git command and remove all the hard coded values.

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -9,20 +9,20 @@ function kind_clusters() {
             echo Cluster cluster${i} already exists, skipping cluster creation...
         else
             if [[ -n ${version} ]]; then
-                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --wait=5m --config=./kind-e2e/cluster${i}-config.yaml
+                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --wait=5m --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
             else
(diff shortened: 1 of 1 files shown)


scripts/kind-e2e/e2e.sh
```diff
@@ -9,20 +9,20 @@ function kind_clusters() {
             echo Cluster cluster${i} already exists, skipping cluster creation...
         else
             if [[ -n ${version} ]]; then
-                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --wait=5m --config=./kind-e2e/cluster${i}-config.yaml
+                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --wait=5m --config=$(git rev-parse --show-toplevel)/scripts/kind-e2e/cluster${i}-config.yaml
```
mangelajo: Just some nitpicking (about making lots of calls...)

Can we capture the directory once at start in a variable and then use the variable?

PRJ_TOPLEVEL=....

also you already have DAPPER_SOURCE that may also work inside dapper... but not outside (if we wanted that..).

dimaunx: Wanted to avoid dapper source DAPPER_SOURCE not to depend on it much.

PRJ_TOPLEVEL=....
Great idea!

dimaunx: @mangelajo Added PRJ_ROOT and tested the changes.

PR #85, opened 2019-06-21
Refactor ipsec to hide implementation

tpantelis:
The Engine and Specification structs do not need to be
exported and NewEngine can return the implemented interface
type.

Also renamed the CableEngine interface to Engine to reduce stutter -
so now it reads cableengine.Engine.


tpantelis:
@skitt

PR #86, opened 2019-06-24
Fix validate and test scripts

mangelajo:
The dirname $0 output is relative, so if we cd first,
and then try to source the lib/find_utils script, it won't
work.

This patch fixes the issue.


mangelajo:
Sending a new version of the patch :-)

PR #87, opened 2019-06-24
Fix the validate script: missing uuid module

mangelajo:
In this patch we grab all dependencies from ./test/e2e/...
golangci-lint fails otherwise because it's unable to find
the k8s uuid module which we don't vendor.

An alternative would be to start vendoring such module.

### DIFF
--- a/scripts/validate
+++ b/scripts/validate
@@ -14,4 +14,8 @@ if [[ $(goimports -l ${PACKAGES} | wc -l) -gt 0 ]]; then
     exit 1
 fi
 
+# grab e2e-test only related packages to avoid linting
+# errors for packages we didn't vendor.
+go get -t ./test/e2e/...
+
 golangci-lint run -D megacheck -E unused,gosimple,staticcheck


mangelajo:
@skitt  ^ what do you think?

An alternative could be just vendoring the missing module (used in the e2e tests)

mangelajo:
> I think we should vendor the module instead. Since `validate` is run before the build, any dependencies installed here could be used in the build too, and even if they’re only used in the tests, we won’t know what version we’re getting...

Yeah, I checked we don't do that now (validate before build) but it doesn't mean we would not do it in the future because somebody is not aware of that dept. So let's vendor that one then, I'll fix the patch to vendor it.

skitt:
> Yeah, I checked we don't do that now (validate before build) but it doesn't mean we would not do it in the future because somebody is not aware of that dept. So let's vendor that one then, I'll fix the patch to vendor it.

#88 :-D (not vendoring, running `validate` before the build).

mangelajo:
> > Yeah, I checked we don't do that now (validate before build) but it doesn't mean we would not do it in the future because somebody is not aware of that dept. So let's vendor that one then, I'll fix the patch to vendor it.
> 
> #88 :-D (not vendoring, running `validate` before the build).

Hmmm, and how does that work?, /me tries :-)

mangelajo:
Fixed in the go mod switch

PR #89, opened 2019-06-25
Deadcode

mkolesnik:
Cleaned up some dead code in cableengine & ipsec modules.

### DIFF
--- a/main.go
+++ b/main.go
@@ -134,7 +134,7 @@ func main() {
 		wg.Add(3)
 		go func() {
 			defer wg.Done()
-			if err = cableEngine.StartEngine(true); err != nil {
+			if err = cableEngine.StartEngine(); err != nil {
 				klog.Fatalf("Error starting the cable engine: %v", err)
 			}
 		}()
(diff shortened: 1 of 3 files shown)


mangelajo:
Nice @mkolesnik can you register your submariner fork to https://travis-ci.com/mkolesnik ? 

That way when you send pull requests, your fork will be tested by travis E2E.

tpantelis:
The CableEngine interface has unused/unimplemented methods. I'm not sure what the intent was - I'm fine with removing them - let's see if @Oats87 has any objections...

Oats87:
@mkolesnik @tpantelis the original reason why that dead code was left around is due to remnants from some original inspiration taken from https://github.com/rancher/ipsec . Originally there was planned functionality around being able to reload charon/stop/start charon (i.e. for graceful leader election handling instead of just dying when leader election is lost) but for now I think we should be OK to remove the handlers. It will likely make it easier to make the CableEngine interface a more uniform interface for other providers as well.

mangelajo:
Thanks for the feedback @Oats87 ! :)

PR #90, opened 2019-06-25
Fix typos in e2e test README, rm selinux=off docs

dfarrell07:
Fix various minor typos in the Kind-based end-to-end testing docs.

Remove docs about requirement for SELinux to be set to permissive, as
that is no longer a requirement.


dimaunx:
Unfortunately the SELinux issue is still there. We have experienced it.
For people who use docker version 1.13.x and not the latest docker-ce package, the solution would be 
1. edit /etc/sysconfig/docker
2. remove --selinux-enabled
3. restart docker service

PR #91, opened 2019-06-26
GCC fix for E2E ginkgo tests

dimaunx:
Because gcc is not present in the build container, the E2E ginkgo tests are failing with error:
exec: "gcc": executable file not found in $PATH

Steps to reproduce:
1. make e2e status=clean
2. docker system prune --all
3. make e2e


mangelajo:
@skitt ^ :-) (just to make you aware why gcc is back on the block :-)

PR #94, opened 2019-07-01
Update to go mod

mangelajo:
We use the -mod=vendor strategy for building, and we add a script to
let you update the vendored files.

Apparently go-mod expects semver (vX.Y.Z) naming for tags/branches, but
kubernetes-1.13.2 is not in that format so for k8s.io/xxx packages
it gets replaced to v0.0.0-commit id every time we run go mod,
see: https://github.com/golang/go/issues/27271

Go automatically detects ginkgo/gomega usage, and adds it to the
go.mod file, resulting in vendoring those pkgs, which may
be ok if we want to provide a built version of the e2e binary
as kubefed will be doing soon.


mangelajo:
@skitt

Dockerfile.dapper
```diff
@@ -13,12 +13,11 @@ ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPE
...
 
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
-    GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
+    GOPATH=/go GOFLAGS="-mod=vendor" PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
 RUN apt-get -q update && \
     apt-get install -y gcc git curl docker.io && \
-    curl https://storage.googleapis.com/golang/go1.11.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
-    go get -v github.com/rancher/trash && \
+    curl https://storage.googleapis.com/golang/go1.11.11.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
```
skitt: Could you explain the bump in the commit message?

mangelajo: Oh right, I forgot about that detail, thank you @skitt

mangelajo:
Based on today's meeting conversation I bumped to go 1.12.6, plus go modules without vendoring

tpantelis:
@skitt Are you good with the latest changes?

mangelajo:
Something I have found with this change, is that every time you run with dapper it will download all the packages etc... which makes it slower, specially if your connection is slow.

An alternative to this could be enabling the "vendor" mode of go mod, but not checking in the packages to git, so the go cache will be there between runs (we can avoid re-downloading if the vendor directory exists).

I'm going to modify the patch in such way because it's already painful to me.

tpantelis:
> Something I have found with this change, is that every time you run with dapper it will download all the packages etc... which makes it slower, specially if your connection is slow.
> 
> An alternative to this could be enabling the "vendor" mode of go mod, but not checking in the packages to git, so the go cache will be there between runs (we can avoid re-downloading if the vendor directory exists).
> 
> I'm going to modify the patch in such way because it's already painful to me.

Sounds good - so it turns out we do have a use case for keeping the vendoring 😄

skitt:
> An alternative to this could be enabling the "vendor" mode of go mod, but not checking in the packages to git, so the go cache will be there between runs (we can avoid re-downloading if the vendor directory exists).

Does the Go cache help when we’re building in containers with Dapper?

mangelajo:
It didn't seem to work, it keet re-downloading all on every run even if I
added go mod download as a RUN step in dapper.

---
irc: ajo / mangelajo
Miguel Angel Ajo Pelayo
+34 636 52 25 69
skype: ajoajoajo

skitt:
> It didn't seem to work, it keet re-downloading all on every run even if I added go mod download as a RUN step in dapper.

Right, that’s what I suspected ;-). If we want to avoid having to download everything for every run of the Dapper build, we need to either copy the Go cache into and back out of the build container, or we need to vendor the dependencies and commit them.

skitt:
This breaks e2e for me, it fails because of a missing `scripts/vendor`.

.gitignore
```diff
@@ -6,6 +6,7 @@ bin
 dist
 output
 strongswan
+vendor
```
skitt: This causes `scripts/vendor` to be ignored too; you’ll need to specify `/vendor` instead. `/vendor/` would be even better, that will only match a directory.

tpantelis: What is scripts/vendor for and why does "gitignoring" it break e2e? There is no scripts/vendor checked it.

mangelajo: ohhhh! thanks, fixing.

skitt: `make test` is broken too. I suspect `scripts/vendor` is what updates `vendor` locally; because it’s gitignored, it wasn’t included in Miguel’s commit, and therefore didn’t make it into this PR.

tpantelis: oh I see - `scripts/vendor` is a new file.

skitt:
This works for me now, thanks!

sridhargaddam:
LGTM. Works fine and even the cache is getting reused.

PR #96, opened 2019-07-03
Remove golangci-lint command flags in validate script

tpantelis:
The script was running:
    
`golangci-lint run -D megacheck -E unused,gosimple,staticcheck`
    
With these flags, the `unused` linter does not actually run (I assume `gosimple` and `staticcheck` don't run either) as there is unused code submariner (`datastoresyncer`). Running just `golangci-lint run` yields errors emitted by `unused`.
    
I'm not clear as to the purpose of these flags. It isn't necessary to explicitly enable `unused`, `gosimple` and `staticcheck` as they are enabled by default. FWICT, `megacheck` runs `staticcheck`, `gosimple` and `unused` at once for efficiency - I don't know the reason for disabling it but doing so seems to also disable the other 3 even though they are explicitly enabled.

### DIFF
--- a/pkg/controllers/datastoresyncer/datastoresyncer.go
+++ b/pkg/controllers/datastoresyncer/datastoresyncer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tpantelis:
@skitt @dimaunx

skitt:
Please merge #88 along with this one, so future linting errors don’t get merged ;-).

PR #99, opened 2019-07-04
Extract strongswan to a Driver in ipsec

mkolesnik:
In order to support other IPsec drivers, the Driver interface has been
introduced to encapsulate the requirements from the IPsec engine.
The IPsec engine has some business logic and iptables related code which
isn't related to the way it actually connects the tunnels, so that code
remains in the engine, while the tunnel specific things have been
extracted.

The implementation has been extracted as-is, with as little changes as
possible (other than the obligatory ones to actually allow the new
interface) so that the refactor is easy to understand.

Currently the StrongSwan implementation is created hard coded, but the
aim is to add a LibreSwan driver type in the next patches.

### DIFF
--- a/pkg/cableengine/ipsec/driver.go
+++ b/pkg/cableengine/ipsec/driver.go
@@ -1,0 +1,22 @@
+package ipsec
+
+import (
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+// Driver is used by the ipsec engine to actually connect the tunnels.
+type Driver interface {
+
+	// Init initializes the driver with any state it needs.
+	Init() error
+
+	// GetActiveConnections returns an array of all the active connections for the given cluster.
+	GetActiveConnections(clusterID string) ([]string, error)
+
+	// ConnectToEndpoint establishes a connection to the given endpoint and returns a string
+	// representation of the IP address of the target endpoint.
+	ConnectToEndpoint(endpoint types.SubmarinerEndpoint) (string, error)
+
+	// DisconnectFromEndpoint disconnects from the endpoint identified by the cable ID.
+	DisconnectFromEndpoint(cableID string) error
+}
--- a/pkg/cableengine/ipsec/strongswan.go
+++ b/pkg/cableengine/ipsec/strongswan.go
@@ -1,0 +1,382 @@
+package ipsec
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/bronze1man/goStrongswanVici"
+	"github.com/kelseyhightower/envconfig"
+	"github.com/submariner-io/submariner/pkg/types"
+	"k8s.io/klog"
+)
+
+const (
+	pidFile = "/var/run/charon.pid"
+
+	// DefaultReplayWindowSize specifies the replay window size for charon
+	DefaultReplayWindowSize = "1024"
+
+	// DefaultIkeSaRekeyInterval specifies the default rekey interval for IKE_SA
+	DefaultIkeSaRekeyInterval = "4h"
+
+	// DefaultChildSaRekeyInterval specifies the default rekey interval for CHILD_SA
+	DefaultChildSaRekeyInterval = "1h"
+)
+
+type strongSwan struct {
+	localSubnets              []string
+	localEndpoint             types.SubmarinerEndpoint
+	secretKey                 string
+	replayWindowSize          string
+	ipSecIkeSaRekeyInterval   string
+	ipSecChildSaRekeyInterval string
+
+	debug   bool
+	logFile string
+}
+
+type specification struct {
+	PSK     string
+	Debug   bool
+	LogFile string
+}
+
+func NewStrongSwan(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (Driver, error) {
+	ipSecSpec := specification{}
+
+	err := envconfig.Process("ce_ipsec", &ipSecSpec)
+	if err != nil {
+		return nil, fmt.Errorf("error processing environment config for ce_ipsec: %v", err)
+	}
+
+	return &strongSwan{
+		replayWindowSize:          DefaultReplayWindowSize,
+		ipSecIkeSaRekeyInterval:   DefaultIkeSaRekeyInterval,
+		ipSecChildSaRekeyInterval: DefaultChildSaRekeyInterval,
+		localEndpoint:             localEndpoint,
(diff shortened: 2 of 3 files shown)


pkg/cableengine/ipsec/driver.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+// Driver is used by the ipsec engine to actually connect the tunnels
+type Driver interface {
+
+	// Initialize the driver with any state it needs
+	Init() error
```
tpantelis: As per https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences, the comment  should begin with the name of the function, ie

    // Init performs any initialization needed by the driver.

pkg/cableengine/ipsec/driver.go
```diff
@@ -0,0 +1,22 @@
...
+
+	// Returns an array of all the active connections for the given cluster
+	GetActiveConnections(clusterID string) ([]string, error)
+
+	// Connects to the given endpoint and reutrns a string representation of
+	// the IP address of the target endpoint
```
tpantelis: How about:

    // ConnectToEndpoint establishes a connection to the given endpoint and returns a string representation of
    // the IP address of the target endpoint.

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -0,0 +1,397 @@
...
+	client, err := getClient()
+	if err != nil {
+		return "", err
+	}
+	defer client.Close()
+
+	return i.installCableInternal(endpoint, client)
+}
+
+func (i *strongSwan) installCableInternal(endpoint types.SubmarinerEndpoint, client *goStrongswanVici.ClientConn) (string, error) {
```
tpantelis: This was named to previously correspond to `InstallCable` so really should be named `connectToEndpointInternal` however this function can just be inlined within `ConnectToEndpoint`.  Originally `installCableInternal` was extracted as it was called from another function that has since been removed.

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -0,0 +1,397 @@
...
+	if err != nil {
+		return err
+
+	}
+	defer client.Close()
+
+	return i.removeCableInternal(cableID, client)
+}
+
+func (i *strongSwan) removeCableInternal(cableID string, client *goStrongswanVici.ClientConn) error {
```
tpantelis: Same as above - inline within `DisconnectFromEndpoint`.

pkg/cableengine/ipsec/driver.go
```diff
@@ -0,0 +1,22 @@
...
+	ConnectToEndpoint(endpoint types.SubmarinerEndpoint) (string, error)
+
+	// Disconnects from the endpoint identified by the cable ID
```
tpantelis: `// DisconnectFromEndpoint disconnects from the endpoint identified by the cable ID.`

On a side note, the one thing that's a bit esoteric is the meaning and relationship of _cableID_ with respect to the `SubmarinerEndpoint` passed to `ConnectToEndpoint`. _cableID_  is actually passed upstream from `SubmarinerEndpoint.Spec.CableName` so it is assumed that the `CableEngine`  implementation uses `SubmarinerEndpoint.Spec.CableName` as an internal identifier in `ConnectToEndpoint`.  To alleviate this undocumented "tribal knowledge", I think we should pass `endpoint types.SubmarinerEndpoint` to this method as well and let the implementation decide how to identify a connection internally.

pkg/cableengine/ipsec/driver.go
mangelajo: Oh, that's for IDEs to catch which function are you talking about?

pkg/cableengine/ipsec/driver.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+// Driver is used by the ipsec engine to actually connect the tunnels
+type Driver interface {
+
+	// Initialize the driver with any state it needs
+	Init() error
+
+	// Returns an array of all the active connections for the given cluster
```
mangelajo: I guess same here "GetActiveConnections ....."

pkg/cableengine/ipsec/driver.go
mkolesnik: I agree with you, but since this is a "change as little as possible" refactor I suggest we can do that in another patch since we also need to change the caller of RemoveCable itself (or change it to fetch the cable).

As for the comment I fixed that.

pkg/cableengine/ipsec/strongswan.go
mkolesnik: I think I meant to do that initially and it skipped my mind, inlined it.

mkolesnik:
Thanks for the review, I fixed it up and will upload the updated version

pkg/cableengine/ipsec/driver.go
mangelajo: @tpantelis make a very good point, but I agree with @mkolesnik about doing it on a separate patch.

tpantelis: I agree as well - I didn't mean to do it in this patch - I should've made that clear.

mangelajo:
@tpantelis I think all requested changes, I have approved, feel free to merge, not sure if I'm missing something (that's why I didn't)

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -0,0 +1,388 @@
...
+
+	// DefaultIkeSaRekeyInterval specifies the default rekey interval for IKE_SA
+	DefaultIkeSaRekeyInterval = "4h"
+
+	// DefaultChildSaRekeyInterval specifies the default rekey interval for CHILD_SA
+	DefaultChildSaRekeyInterval = "1h"
+)
+
+type strongSwan struct {
+	sync.Mutex
```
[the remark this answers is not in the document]
mkolesnik: This was actually used in loadConns which is called in Init which is called at engine start, but i reckon it should be safe to remove.

Your thoughts?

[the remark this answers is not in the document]
tpantelis: yes loadConns was only called from one site where concurrency wouldn’t occur   I suspect there were other call sites originally although I'm not sure of the reason for mutual exclusion when listing the tunnels. Either way, mutual exclusion is no longer needed in loadConns.

PR #100, opened 2019-07-05
Upgrade to kind 0.4.0

skitt:
Release notes:
https://github.com/kubernetes-sigs/kind/releases/tag/v0.4.0

Fixes:
* KUBECONFIG uses the API listener IP address, not localhost
* Images work without IPv6
* Better handling of proxy settings
* Proper API server healthchecks in the external loadbalancer for
  multi-control-plane clusters

The default K8s version with kind 0.4.0 is 1.15.0, but our
configuration fails to deploy there so this stays on 1.14.2 for now.


dimaunx:
There issue with 0.4.0 is that none on the images has support for weave, only kindnet. Unfortunately if we upgrade it will prevent from the teams to work on network policies. 0.3.0 has 1.14.1 image that supports weave without the need to build custom image.

skitt:
> Please let's not merge this now, it will introduce an issue with the network policy testing because kind-net does not support that.
> 
> I will close this for now if that's ok.

Yup, let’s close it.

PR #101, opened 2019-07-05
Libreswan

skitt:
This sits on top of #99 and should be rebased once that’s merged.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/101)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ submariner-globalnet
 /bin
 dist
 output
-strongswan
 /vendor
 package/strongswan-*
 .project
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -1,0 +1,57 @@
+package libreswan
+
+import (
+	"fmt"
+
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cable"
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+const (
+	cableDriverName = "libreswan"
+)
+
+func init() {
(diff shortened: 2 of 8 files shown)


main.go
```diff
@@ -37,11 +37,13 @@ import (
 var (
 	localMasterURL  string
 	localKubeconfig string
+	localEngine     string
 )
 
 func init() {
 	flag.StringVar(&localKubeconfig, "kubeconfig", "", "Path to kubeconfig of local cluster. Only required if out-of-cluster.")
 	flag.StringVar(&localMasterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	flag.StringVar(&localEngine, "engine", "strongswan", "The IKE implementation to use: strongswan or libreswan.")
```
tpantelis: How about we put this in ipsec.go instead of passing to ipsec.NewEngine? Also name `ipsec` instead of `engine` for clarity.

Longer term, I think we should consider using [Go plugins ](https://code.tutsplus.com/tutorials/writing-plugins-in-go--cms-29101) instead of passing CLI args to make it cleaner and easier to support other cable engine implementations.

mangelajo:
I recommend you folks talk to Paul Wouters about getting support to managing libreswan from go.

main.go
mkolesnik: Or even ipsecDriver as the name?

mangelajo: ipsecdriver ? :-)

I think that's more user friendly/clear.

skitt:
> Looks very empty.. :)

Yes, the original version of this took care of adding the framework for multiple cable drivers, but that was done differently as part of the Wireguard merge. I’ve updated this PR to build on top of that, without adding any of the Libreswan code yet (which is in #107).

PR #103, opened 2019-07-10
Remove cable endpoint

mkolesnik:
This is based on a request from PR #99 to make the remove side use endpoint instead of the cable name, and let the engine/driver use whatever it desires from the endpoint to remove the connection.

It should be rebased once #99 is merged.

### DIFF
--- a/pkg/cableengine/cableengine.go
+++ b/pkg/cableengine/cableengine.go
@@ -7,5 +7,5 @@ import (
 type Engine interface {
 	StartEngine() error
 	InstallCable(types.SubmarinerEndpoint) error
-	RemoveCable(string) error
+	RemoveCable(types.SubmarinerEndpoint) error
 }
--- a/pkg/controllers/tunnel/tunnel.go
+++ b/pkg/controllers/tunnel/tunnel.go
@@ -145,11 +145,14 @@ func (t *Controller) handleRemovedEndpoint(obj interface{}) {
 	}
 
 	klog.V(4).Infof("Informed of removed endpoint for tunnel controller object: %#v", object)
-	if err := t.ce.RemoveCable(object.Spec.CableName); err != nil {
-		utilruntime.HandleError(fmt.Errorf("error removing endpoint cable %s from engine: %v",
-			object.Spec.CableName, err))
+	myEndpoint := types.SubmarinerEndpoint{
+		Spec: object.Spec,
+	}
+	if err := t.ce.RemoveCable(myEndpoint); err != nil {
+		utilruntime.HandleError(fmt.Errorf("error removing endpoint cable %#v from engine: %v",
+			myEndpoint, err))
 		return
 	}
 
-	klog.V(4).Infof("Removed endpoint cable %s from engine", object.Spec.CableName)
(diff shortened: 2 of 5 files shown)


pkg/cableengine/cableengine.go
```diff
@@ -7,5 +7,5 @@ import (
 type Engine interface {
 	StartEngine() error
 	InstallCable(types.SubmarinerEndpoint) error
-	RemoveCable(string) error
+	RemoveCable(types.SubmarinerEndpoint) error
```
tpantelis: While we're changing the interface anyway, I think we should pass a pointer for efficiency to avoid copying. 

Also `SubmarinerEndpoint` merely just wraps an `EndpointSpec` so I think it makes sense to pass that instead. This would make the call site in _tunnel.go_ simpler and cleaner...

pkg/controllers/tunnel/tunnel.go
```diff
@@ -145,11 +145,14 @@ func (t *Controller) handleRemovedEndpoint(obj interface{}) {
 	}
 
 	klog.V(4).Infof("Informed of removed endpoint for tunnel controller object: %#v", object)
-	if err := t.ce.RemoveCable(object.Spec.CableName); err != nil {
-		utilruntime.HandleError(fmt.Errorf("error removing endpoint cable %s from engine: %v",
-			object.Spec.CableName, err))
+	myEndpoint := types.SubmarinerEndpoint{
+		Spec: object.Spec,
+	}
+	if err := t.ce.RemoveCable(myEndpoint); err != nil {
```
tpantelis: ... with the proposed changes described above, this becomes simply `t.ce.RemoveCable(&object.Spec)`.

pkg/cableengine/cableengine.go
mkolesnik: Thanks for the proposal, however I wrote this patch to comply with the current API as is.

If it indeed makes sense, I'll add more commits to accommodate the requested changes.

My main concern is that the type EndpointSpec is never used directly, but always through a wrapper (i.e. SubmarinerEndpoint). The only exception I see is util.CompareEndpointSpec which references the EndpointSpec type directly.

As for pointer v. copy, again this change was in line with the style of coding in this case. If we want to overhaul the code to be more efficient and use pointers instead of passing copies (the value of which, for performance, IMHO is negligible) then I don't mind doing it in a separate patch.

tpantelis: > My main concern is that the type EndpointSpec is never used directly, but always through a wrapper (i.e. SubmarinerEndpoint). The only exception I see is util.CompareEndpointSpec which references the EndpointSpec type directly.

right but I don't see the point for the `SubmarinerEndpoint` wrapper - seems like extra noise and overhead. We can consider removing it in another patch.

> 
> As for pointer v. copy, again this change was in line with the style of coding in this case. If we want to overhaul the code to be more efficient and use pointers instead of passing copies (the value of which, for performance, IMHO is negligible) then I don't mind doing it in a separate patch.

yeah - separate patch is fine - we've changed from pass-by-value to pass-by-pointer in other places in the code base as well.

PR #107, opened 2019-07-15
 Support Libreswan "cable" creation

skitt:
This relies on the whack command-line tool for now. Because the
full-blown status is very hard to parse, the connection-tracking code
uses the simpler trafficstatus output, which doesn’t give all the
information we’d really need but is sufficient to determine whether a
connection is active or not.

We set up one tunnel per subnet pair, one tunnel between the hosts,
and one tunnel between each host and each opposing subnet.

Fixes: #592
Fixes: #385
Fixes: #564

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -16,10 +16,13 @@ jobs:
       matrix:
         globalnet: ['', '--globalnet']
         deploytool: ['operator', 'helm']
-        cable_driver: ['', 'wireguard']
+        cable_driver: ['', 'wireguard', 'libreswan']
         exclude:
+# Our Helm setup doesn’t know how to deploy other cable drivers
           - deploytool: 'helm'
             cable_driver: 'wireguard'
+          - deploytool: 'helm'
+            cable_driver: 'libreswan'
     steps:
       - name: Checkout the repository
         uses: actions/checkout@master
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -1,11 +1,20 @@
 package libreswan
 
 import (
+	"bufio"
 	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
 
-	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/kelseyhightower/envconfig"
+	"k8s.io/klog"
+
+	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cable"
 	"github.com/submariner-io/submariner/pkg/types"
+	"github.com/submariner-io/submariner/pkg/util"
 )
 
 const (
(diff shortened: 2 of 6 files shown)


.github/workflows/e2e.yaml
```diff
@@ -16,10 +16,16 @@ jobs:
...
+        cable_driver: ['', 'wireguard', 'libreswan']
         exclude:
+# Our Helm setup doesn’t know how to deploy other cable drivers
           - deploytool: 'helm'
             cable_driver: 'wireguard'
+          - deploytool: 'helm'
+            cable_driver: 'libreswan'
+# TODO: Remove exclusion when issue #552 is fixed
+          - globalnet: '--globalnet'
+            cable_driver: 'wireguard'
```
mangelajo, line 28:
```suggestion
```
The issue was fixed by:
    https://github.com/submariner-io/submariner/pull/562

Now we have an all-together different issue because the azure-ubuntu image has broken headers for wireguard... :/ https://github.com/submariner-io/submariner/issues/607

But I think it's unrelated to your patch, let's disable globally, we can ignore those runs for your patch.

pkg/cable/libreswan/libreswan.go
```diff
@@ -1,57 +1,377 @@
...
+
+				/*
+					if err := Route(endpoint.Spec.CableName); err != nil {
+						return "", fmt.Errorf("Error routing connection %v: %v", endpoint.Spec.CableName, err)
+					}
+
+					if err := Initiate(endpoint.Spec.CableName); err != nil {
+						return "", fmt.Errorf("Error initiating connection %v: %v", endpoint.Spec.CableName, err)
+					}
+				*/
```
mangelajo, line 257:
```suggestion
```

pkg/cable/libreswan/libreswan.go
```diff
@@ -1,57 +1,377 @@
...
+					//return "", fmt
+					klog.Errorf("Error routing connection %s: %v", connectionName, err)
+				}
+
+				cmd = exec.Command("/usr/libexec/ipsec/whack", "--initiate", "--name", connectionName)
+				cmd.Stdout = os.Stdout
+				cmd.Stderr = os.Stderr
+
+				if err := cmd.Run(); err != nil {
+					//return "", fmt
```
mangelajo, line 273:
```suggestion
```

pkg/cable/libreswan/libreswan.go
```diff
@@ -1,57 +1,377 @@
...
+						return fmt.Errorf("Error deleting a connection with args %v: %v", args, err)
+					}
+				}
+			}
+		}
+	}
+
+	for j := range i.connections {
+		if i.connections[j].Endpoint.CableName == endpoint.Spec.CableName {
+			if j == 0 && j == len(i.connections)-1 {
```
mangelajo, line 322: how can j be two different things at once? I guess for len(i.connections)==1 

nit: can we use j==0 && len(i.connections)==1 ?

pkg/cable/libreswan/libreswan.go
```diff
@@ -1,57 +1,377 @@
...
+						klog.Errorf("Error deleting a connection with args %v; got exit code %d: %v", args, err.ExitCode(), err)
+					default:
+						return fmt.Errorf("Error deleting a connection with args %v: %v", args, err)
+					}
+				}
+			}
+		}
+	}
+
+	for j := range i.connections {
```
mangelajo, line 320: can we refactor it out to a separate method? (removeConnectionForEndpoint?)

pkg/cable/libreswan/libreswan.go
skitt: I prefer the current form because it matches the other checks...

.github/workflows/e2e.yaml
skitt: Done, I didn’t notice that that part had gone.

pkg/cable/libreswan/libreswan.go
skitt: Done.

pkg/cable/libreswan/libreswan.go
skitt: Done.

pkg/cable/libreswan/libreswan.go
mangelajo: ack :+1:

pkg/cable/libreswan/libreswan.go
```diff
@@ -1,57 +1,368 @@
...
 
 // Init initializes the driver with any state it needs.
-func (i *libreSwan) Init() error {
+func (i *libreswan) Init() error {
+	// Write the secrets file:
+	// %any %any : PSK "secret"
+	// TODO Check whether the file already exists
+	file, err := os.Create("/etc/ipsec.d/submariner.secrets")
+	if err != nil {
+		return fmt.Errorf("Error creating the secrets file: %v", err)
```
sridhargaddam, line 87: nit: When returning an error message in an API, it's recommended to use small letter (i.e., instead of Error, please use error)

pkg/cable/libreswan/libreswan.go
```diff
@@ -1,57 +1,368 @@
...
+	}
+	defer file.Close()
+
+	fmt.Fprintf(file, "%%any %%any : PSK \"%s\"\n", i.secretKey)
+
+	// Ensure Pluto is started
+	if err := i.runPluto(); err != nil {
+		return fmt.Errorf("Error starting Pluto: %v", err)
```
sridhargaddam, line 95: nit: as above in all occurrences.

pkg/cable/libreswan/libreswan.go
skitt: Done, thanks for the reminder!

PR #109, opened 2019-07-15
Always specify the context

skitt:
This changes the e2e script to always specify the kube context, to
avoid interactions with other uses of use-context, and allow future
parallelisation of context-sensitive steps.

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -49,43 +49,40 @@ function install_helm() {
     helm init --client-only
     helm repo add submariner-latest https://releases.rancher.com/submariner-charts/latest
     for i in 1 2 3; do
-        kubectl config use-context cluster${i}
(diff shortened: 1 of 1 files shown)


scripts/kind-e2e/e2e.sh
```diff
@@ -183,39 +177,36 @@ function test_connection() {
 
 function update_subm_pods() {
     echo Removing submariner engine pods...
-    kubectl delete pods -n submariner -l app=submariner-engine
-    kubectl wait --for=condition=Ready pods -l app=submariner-engine -n submariner --timeout=60s
+    kubectl --cluster=$1 delete pods -n submariner -l app=submariner-engine
```
mangelajo: --context= ? 

Isn't this overlapping with the other PR ?

skitt: Both `--cluster` and `--context` work, but I’ve updated the PR to only use `--context`. #110 includes this PR (see the comment in the description of #110).

PR #111, opened 2019-07-16
Run validation and tests on Travis

skitt:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ git:
   depth: 1
 
 script:
-  - make e2e status=keep
+  - make ci e2e status=keep
--- a/pkg/cableengine/ipsec/strongswan.go
+++ b/pkg/cableengine/ipsec/strongswan.go
@@ -15,8 +15,6 @@ import (
 )
 
 const (
-	pidFile = "/var/run/charon.pid"
-
 	// DefaultReplayWindowSize specifies the replay window size for charon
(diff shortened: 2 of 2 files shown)


skitt:
This is expected to fail — I want to check whether validation fails the build. I’ll push the fixed commit later so we can merge this.

mangelajo:
ha, what happens if we run make ci, then make e2e ?  @dimaunx thoughts?

mangelajo:
I see the unit tests: https://travis-ci.com/submariner-io/submariner/builds/119282452#L1219 but not the other parts of validation executing :? hmmm

dimaunx:
> ha, what happens if we run make ci, then make e2e ? @dimaunx thoughts?

Should be fine.

dimaunx:
We can run in parallel both of them similar to https://hackernoon.com/parallelizing-builds-in-travis-ci-9d913e5722eb

skitt:
> I see the unit tests: https://travis-ci.com/submariner-io/submariner/builds/119282452#L1219 but not the other parts of validation executing :? hmmm

The validation is silent when it passes. Make runs all its targets one after the other, so saying `make ci e2e status=keep` is equivalent to `make ci status=keep && make e2e status=keep`.

I had deliberately left a validation failure in the first iteration of this patch, to verify that it did fail; but I can’t find that build any more. I’ll re-push the failing version and link the failed build for posterity.

PR #120, opened 2019-07-25
Fix clusterCidr used in configuring Submariner

sridhargaddam:
Currently the clusterCidr information used to configure
Submariner has wrong values. This patch fixes it.

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -120,7 +120,7 @@ function setup_cluster2_gateway() {
             --set broker.namespace="${SUBMARINER_BROKER_NS}" \
             --set broker.ca="${SUBMARINER_BROKER_CA}" \
             --set submariner.clusterId="cluster2" \
-            --set submariner.clusterCidr="$worker_ip/32" \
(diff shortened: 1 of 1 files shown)


dimaunx:
> Currently the clusterCidr information used to configure
> Submariner has wrong values. This patch fixes it.

Yes you are correct, i was under wrong impression that clusterCidr should be the worker instances cidr.
When providing podCidr as clusterCidr, the pod to pod connectivity works correctly in addition to pod to service connectivity. I have tested this and it works perfect.
The ci now passes after recurring run test fix https://github.com/submariner-io/submariner/pull/116.

sridhargaddam:
Thanks for verifying it and confirming @dimaunx.

sridhargaddam:
@mangelajo, @tpantelis can you please take a look at this patch. Thanks.

sridhargaddam:
@mangelajo Not sure about the issue that you are referring to, may be @dimaunx can comment on it. 

While deploying Submariner using helm, we have to provide the local cluster POD and Service CIDRs. Based on the information that we are providing, submariner cluster and endpoint CRDs are populated. In the existing code, we were using wrong values in podCIDRs and this patch fixes it. So, yes we need this patch :)

PR #122, opened 2019-07-28
Push master branch docker images to quay.io with tag:latest

dimaunx:
This PR adds a deploy step to .travis.yaml to push the artifacts from master branch to quay.io with tag:latest and tag:git_sha. The artifacts include submariner and submariner-route-agent docker images.
quay.io organization exists with submariner and submariner-route-agent public repositories.
We can additionally push images to docker hub or just use docker hub.
Quay.io is **open source** and has number of advantages over docker hub.
This is open to discussion.

1. Only after PR is merged in to master branch the deploy step will kick in and push images to quay. 
2.  The images will be tagged with latest and git_sha tags.
3. For the deploy step to work, travis requires to have two secure environment variables configured for quay bot account. QUAY_PASSWORD and QUAY_USERNAME. I can provide this information to the travis owners.
4. @Oats87 and anyone of the contributors can be made admins/owners on the quay organization. Please create quay.io account and provide the username in this thread.


.travis.yml
```diff
@@ -4,14 +4,19 @@ language: go
...
-      echo "scripts/kind-e2e/e2e.sh was modified, testing again on already deployed environment.";
+  - if [[ "${CHANGED_FILES_PR[@]}" =~ "scripts/kind-e2e/e2e.sh" ]]; then
+      echo "scripts/kind-e2e/e2e.sh was modified, testing recurring run on already deployed infrastructure.";
       make e2e status=keep;
-    fi
\ No newline at end of file
+    fi
+deploy:
+  provider: script
+  script: bash scripts/kind-e2e/deploy.sh
```
mangelajo: more than "deploy.sh" wouldn't be this "publish.sh" ? 

It's the deploy action of travis, but we are actually publishing?

And may be it should not go inside scripts/kind-e2e/  may be tools/ release-tools/ something like that, I don't know, but it's not e2e :)

tpantelis: There already is a _scripts/release_ script - all it does currently is execute the _ci_ script  - not sure what it's used for (if at all) - perhaps this would be a logical place to put the deploy code.

dimaunx: Yes makes sense.

dimaunx: Added the push flow to the scripts/release script and tested on my repo the images are pushed after push to master.

skitt:
There are multiple ways of getting images published on Quay; pushing (as done here) is one of them, but another is to get Quay to pull container specifications from our git repository... Wouldn’t it be better to do the latter? It would avoid having to pass credentials around.

dimaunx:
Yes we can do the latter, after master push the image will be build, the
only issues with that approach is that the second ci run on master after PR
is merged will be useless, the build will be triggered regardless of the
master ci result(not a big issue as PR ci passed), and quay and docker hub
builds do not support any custom build args, if we would like to use them
later (different arch images from single docker file for example), we will
have to create dockerfile per arch per  quay build job per repo.
Environment variables do not solve the issue as they are not used on build
time only on runtime. I believe its less flexible approach.

mangelajo:
Is this WIP or is it ok? :)

dimaunx:
> Is this WIP or is it ok? :)

We can remove the WIP.

PR #123, opened 2019-07-29
[E2E] Use weave as default cni for all k8s and kind versions. 

dimaunx:
This PR uses [Cluster.networking.disableDefaultCNI](https://github.com/kubernetes-sigs/kind/issues/278) option for clusters 2 and 3. After the container nodes are up we are installing weave manually. In previous setup with k8s version 1.14.1, weave was ignoring podCidr provided in kind config file and used its default 10.32.0.0/12 cidr as podCidr. This PR fixes this issue.The result is a cluster with weave and podCidr that is taken from kind config file. This removes the need to be constrained to specific k8s version or kind version. After this is merged kind can be safely upgraded. Other types of cni plugins can be tested more easily.

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -19,13 +19,14 @@ function kind_clusters() {
             echo Creating cluster${i}, logging to ${logs[$i]}...
             (
             if [[ -n ${version} ]]; then
-                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --wait=5m --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
+                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
             else
-                kind create cluster --name=cluster${i} --wait=5m --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
+                kind create cluster --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
             fi
             master_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster${i}-control-plane | head -n 1)
             sed -i -- "s/user: kubernetes-admin/user: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
             sed -i -- "s/name: kubernetes-admin.*/name: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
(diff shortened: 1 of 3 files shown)


scripts/kind-e2e/e2e.sh
```diff
@@ -19,13 +19,14 @@ function kind_clusters() {
             echo Creating cluster${i}, logging to ${logs[$i]}...
             (
             if [[ -n ${version} ]]; then
-                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --wait=5m --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
+                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
```
mangelajo: Why do we remove the wait=5m? :)

scripts/kind-e2e/e2e.sh
```diff
@@ -88,6 +89,23 @@ function install_helm() {
...
 
+function setup_custom_cni(){
+    trap_commands
+    declare -A POD_CIDR=( ["cluster2"]="10.245.0.0/16" ["cluster3"]="10.246.0.0/16" )
+    for i in 2 3; do
+        if kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=60s > /dev/null 2>&1; then
+            echo "Weave already deployed cluster${i}."
+        else
+            echo "Applying weave network in to cluster${i}..."
+            kubectl --context=cluster${i} apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')&env.IPALLOC_RANGE=${POD_CIDR[cluster${i}]}"
```
mangelajo: oh wow! :D that's neat of them!

scripts/kind-e2e/e2e.sh
dimaunx: The control plain comes up without cni, --wait waits for control plain to be up and its never up until we apply weave cni.

mangelajo: ohh, makes sense, thanks :-D

mangelajo:
@sridhargaddam will love this ;)

dimaunx:
> @sridhargaddam will love this ;)

Yes :)

dimaunx:
> I have a few questions inline but looks good to me.
> 
> It'd be awesome to be able to control on an env var if we want kindnet or weave, or eventually something else. But let's get into that complication the day we really need it.
> 

Yes that's nice idea. I have also tested calico, but had some issues related to kernel modules. Seems to be fixable. Didn't want to waste too much time and used weave instead. Seems as long as cni supports podCidr customization we can utilize most of them.

PR #124, opened 2019-07-30
Adding verbosity to all bash scripts

manosnoam:


manosnoam:
> Looks good to me, working great, but its seems it conflicts #121, we should merge #121 first and then @manosnoam can rebase.

It's fairly easy to merge #121 and #124, doen't matter which one comes first.
Anyway, I rebased and Travis passed, so you can merge now.

PR #130, opened 2019-08-19
Add base Dapper image

skitt:
This will allow us to publish an image upon which we can build our
Dapper images, which avoids a lot of downloading and rebuilding all
the time. For now the image is amd64-only.


skitt:
@dimaunx I suspect you will like this!

mangelajo:
I'll merge this one, as we all agreed on this, and the change does not modify any existing behavior (the dockerfile is completely independent)

skitt:
> Time for fedora dancer

Done!

PR #135, opened 2019-08-27
Exec iptables from the host filesystem

sridhargaddam:
Depends-On: https://github.com/submariner-io/submariner-charts/pull/3

Both iptables and nftables use netfilter framework in the kernel for
packet filtering. Many distributions are moving in the direction of
using nftables over iptables. Although, nftables uses a new command
line utility (named nft), starting from iptables >=1.8, it uses
nftables under the hood while continuing to support the same iptables
syntax from the user.

Quoting from Dan's comment [#]

"In iptables 1.8, the maintainers have "deprecated" the classic ip_tables:
the iptables tool now does userspace translation from the legacy UI/UX,
and uses nf_tables under the hood. So, the commands look and feel the
same, but they're now programming a different kernel subsystem.

The problem arises when you mix and match invocations of iptables 1.6
(the previous stable) and 1.8 on the same machine, because although they
look identical, they're programming different kernel subsystems.

Empirically, this causes weird and wonderful things to happen - things
like if you trace a packet coming from a pod, you see it flowing through
both ip_tables and nf_tables, but even if both accept the packet, it then
vanishes entirely and never gets forwarded"

So, as long as we are programming either nf_tables or iptables, we would
not have any issues. Currently, there is no easy way to identify what type
of rules are programmed on the host. This patch follows the same approach
(as described here [*]) that is taken in OpenShift where the host file
system is mounted inside the docker container and iptables utility on the
host is exec'ed for programming any firewall rules.

[#] https://github.com/kubernetes/kubernetes/issues/71305#issuecomment-448052889
[*] https://github.com/kubernetes/kubernetes/issues/71305#issuecomment-521978797

### DIFF
--- a/package/Dockerfile.routeagent
+++ b/package/Dockerfile.routeagent
@@ -3,7 +3,7 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal
 WORKDIR /var/submariner
 
 # These are all available in the UBI8 base OS repository
-RUN microdnf -y install --nodocs iproute iptables && \
+RUN microdnf -y install --nodocs iproute && \
     microdnf clean all
 
 COPY submariner-route-agent.sh /usr/local/bin
@@ -11,6 +11,9 @@ COPY submariner-route-agent.sh /usr/local/bin
 RUN chmod +x /usr/local/bin/submariner-route-agent.sh
 
 COPY submariner-route-agent /usr/local/bin
+# We use iptables from the host
+COPY ./iptables /usr/sbin/
+COPY ./iptables-save /usr/sbin/
 
 # temporary sleep infinity so that we can do our debugging
 ENTRYPOINT submariner-route-agent.sh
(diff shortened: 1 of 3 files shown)


sridhargaddam:
> I think this should also drop `iptables` from the `microdnf` invocation, with a comment saying we use the host’s `iptables`.

Good point @skitt. I removed iptables from microdnf and verified that it works fine. Thanks for reviewing.

sridhargaddam:
Please note that this patch depends on the patch in Submariner-charts [https://github.com/submariner-io/submariner-charts/pull/3]

Both the patches should go together and we also have to host the updated charts in a publicly accessible repo. Currently the charts are hosted at rancher url [https://releases.rancher.com/submariner-charts/latest]

In KIND based setup, its referenced here [*]
[*] https://github.com/submariner-io/submariner/blob/master/scripts/kind-e2e/e2e.sh#L59

CC: @mangelajo , @Oats87 , @skitt, @tpantelis

mangelajo:
@tpantelis I have checked this works ok in combination with the VxLAN patch and the already merged helm chart changes.

It also works alone.

PR #140, opened 2019-09-02
Use VxLAN overlay tunnels for inter-cluster traffic

sridhargaddam:
As part of supporting Network policies and for ease of debugging, this
patch implements the following.

1. Creates VxLAN tunnels in the local Cluster between the worker nodes and
   the Cluster Gateway Node.
2. Programms the necessary iptable rules on the Cluster nodes to allow
   inter-cluster traffic.
3. This patch also avoids SNAT/MASQ for inter-cluster traffic, thereby
   preserving the original source ip of the POD all the way until the
   destination POD.
4. Programs the routing rules on the workerNodes to forward the remoteCluster
   traffic over the VxLAN interface that is created between the worker node
   and Cluster GatewayNode.

This patch depends on the following other patches

Depends-On: https://github.com/submariner-io/submariner/pull/135
Depends-On: https://github.com/submariner-io/submariner-charts/pull/3
Depends-On: https://github.com/submariner-io/submariner-charts/pull/4

### DIFF
--- a/pkg/routeagent/controllers/route/iptables.go
+++ b/pkg/routeagent/controllers/route/iptables.go
@@ -80,18 +80,17 @@ func (r *Controller) prependUnique(ipt *iptables.IPTables, table string, chain s
 		return fmt.Errorf("error listing the rules in %s chain: %v", chain, err)
 	}
 
-	/* Submariner requires certain iptable rules to be programmed at the beginning of an iptables Chain so that we can
-	preserve the sourceIP for inter-cluster traffic and avoid K8s SDN making changes to the traffic.
+	// Submariner requires certain iptable rules to be programmed at the beginning of an iptables Chain
+	// so that we can preserve the sourceIP for inter-cluster traffic and avoid K8s SDN making changes
+	// to the traffic.
+	// In this API, we check if the required iptable rule is present at the beginning of the chain.
+	// If the rule is already present and there are no stale[1] flows, we simply return. If not, we create one.
+	// [1] Sometimes after we program the rule at the beginning of the chain, K8s SDN might insert some
+	// new rules ahead of the rule that we programmed. In such cases, the rule that we programmed will
+	// not be the first rule to hit and Submariner behavior might get affected. So, we query the rules
+	// in the chain to see if the rule slipped its position, and if so, delete all such occurrences.
+	// We then re-program a new rule at the beginning of the chain as required.
 
-	In this API, we check if the required iptable rule is present at the beginning of the chain. If the rule is
-	already present and there are no stale[1] flows, we simply return. If not, we create one.
-
-	[1] Sometimes after we program the rule at the beginning of the chain, K8s SDN might insert some new rules
-	ahead of the rule that we programmed. In such cases, the rule that we programmed will not be the first rule
-	to hit and Submariner behavior might get affected. So, we query the rules in the chain to see if the rule
-	slipped its position, and if so, delete all such occurrences.  We then re-program a new rule at the beginning
-	of the chain as required.
-	*/
 	isPresentAtRequiredPosition := false
 	numOccurrences := 0
 	for index, rule := range rules {
@@ -105,8 +104,8 @@ func (r *Controller) prependUnique(ipt *iptables.IPTables, table string, chain s
 		}
 	}
 
-	/* The required rule is present in the Chain, but either there are multiple occurrences or its not at the desired
-	location */
+	// The required rule is present in the Chain, but either there are multiple occurrences or its
+	// not at the desired location
 	if numOccurrences > 1 || isPresentAtRequiredPosition == false {
 		for i := 0; i < numOccurrences; i++ {
 			if err = ipt.Delete(table, chain, ruleSpec...); err != nil {
@@ -115,7 +114,7 @@ func (r *Controller) prependUnique(ipt *iptables.IPTables, table string, chain s
 		}
 	}
 
-	/* The required rule is present only once and is at the desired location */
+	// The required rule is present only once and is at the desired location
 	if numOccurrences == 1 && isPresentAtRequiredPosition == true {
 		klog.V(4).Infof("In %s table, iptables rule \"%s\", already exists.", table, strings.Join(ruleSpec, " "))
 		return nil
--- a/pkg/routeagent/controllers/route/iptables.go
+++ b/pkg/routeagent/controllers/route/iptables.go
@@ -65,7 +65,7 @@ func (r *Controller) programIptableRulesForInterClusterTraffic(remoteCidrBlock s
 			klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
 		}
 
-		// Todo: revisit, we only have to program rules to allow traffic from the podCidr
+		// TODO: revisit, we only have to program rules to allow traffic from the podCidr
 		ruleSpec = []string{"-s", remoteCidrBlock, "-d", localClusterCidr, "-j", "ACCEPT"}
 		klog.V(4).Infof("Installing iptables rule for incoming traffic: %s", strings.Join(ruleSpec, " "))
 		if err = ipt.AppendUnique("nat", SmPostRoutingChain, ruleSpec...); err != nil {
--- a/pkg/routeagent/controllers/route/iptables.go
+++ b/pkg/routeagent/controllers/route/iptables.go
@@ -1,0 +1,91 @@
+package route
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-iptables/iptables"
+	"k8s.io/klog"
+)
+
+func (r *Controller) createIPTableChains() error {
+	ipt, err := iptables.New()
+	if err != nil {
+		klog.Errorf("Error while initializing iptables: %v", err)
+		return err
+	}
+
+	klog.V(4).Infof("Install/ensure %s chain exists", SM_POSTROUTING_CHAIN)
+	if err = ipt.NewChain("nat", SM_POSTROUTING_CHAIN); err != nil {
+		klog.Errorf("Unable to create %s chain in iptables: %v", SM_POSTROUTING_CHAIN, err)
+	}
+
+	klog.V(4).Infof("Insert %s rule that has rules for inter-cluster traffic", SM_POSTROUTING_CHAIN)
+	forwardToSubPostroutingRuleSpec := []string{"-j", SM_POSTROUTING_CHAIN}
+	if err = r.insertUnique(ipt, "nat", "POSTROUTING", 1, forwardToSubPostroutingRuleSpec); err != nil {
+		klog.Errorf("Unable to insert iptable rule in NAT table, POSTROUTING chain: %v", err)
+	}
+
+	klog.V(4).Infof("Install/ensure SUBMARINER-INPUT chain exists")
+	if err = ipt.NewChain("filter", "SUBMARINER-INPUT"); err != nil {
+		klog.Errorf("Unable to create SUBMARINER-INPUT chain in iptables: %v", err)
+	}
+
+	forwardToSubInputRuleSpec := []string{"-p", "udp", "-m", "udp", "-j", "SUBMARINER-INPUT"}
+	if err = ipt.AppendUnique("filter", "INPUT", forwardToSubInputRuleSpec...); err != nil {
+		klog.Errorf("Unable to append iptables rule \"%s\": %v\n", strings.Join(forwardToSubInputRuleSpec, " "), err)
+	}
+
+	klog.V(4).Infof("Allow VxLAN incoming traffic in SUBMARINER-INPUT Chain")
+	ruleSpec := []string{"-p", "udp", "-m", "udp", "--dport", VXLAN_PORT, "-j", "ACCEPT"}
+	if err = ipt.AppendUnique("filter", "SUBMARINER-INPUT", ruleSpec...); err != nil {
+		klog.Errorf("Unable to append iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+	}
+
+	klog.V(4).Infof("Insert rule to allow traffic over %s interface in FORWARDing Chain", VXLAN_IFACE)
+	ruleSpec = []string{"-o", VXLAN_IFACE, "-j", "ACCEPT"}
+	if err = r.insertUnique(ipt, "filter", "FORWARD", 1, ruleSpec); err != nil {
+		klog.Errorf("Unable to insert iptable rule in filter table to allow vxlan traffic: %v", err)
+	}
+
+	return nil
+}
+
+func (r *Controller) programIptableRulesForInterClusterTraffic(remoteCidrBlock string) {
+	ipt, err := iptables.New()
+	if err != nil {
+		klog.Errorf("error while initializing iptables: %v", err)
+	}
+
+	for _, localClusterCidr := range r.localClusterCidr {
+		ruleSpec := []string{"-s", localClusterCidr, "-d", remoteCidrBlock, "-j", "ACCEPT"}
+		klog.V(4).Infof("Installing iptables rule for outgoing traffic: %s", strings.Join(ruleSpec, " "))
+		if err = ipt.AppendUnique("nat", SM_POSTROUTING_CHAIN, ruleSpec...); err != nil {
+			klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+
+		// Todo: revisit we only have to program to the PODCidr
+		ruleSpec = []string{"-s", remoteCidrBlock, "-d", localClusterCidr, "-j", "ACCEPT"}
+		klog.V(4).Infof("Installing iptables rule for incoming traffic: %s", strings.Join(ruleSpec, " "))
+		if err = ipt.AppendUnique("nat", SM_POSTROUTING_CHAIN, ruleSpec...); err != nil {
+			klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+	}
+}
+
+func (r *Controller) insertUnique(ipt *iptables.IPTables, table string, chain string, position int, ruleSpec []string) error {
+	rules, err := ipt.List(table, chain)
+	if err != nil {
+		return fmt.Errorf("error listing the rules in %s chain: %v", chain, err)
+	}
+
+	if strings.Contains(rules[position], strings.Join(ruleSpec, " ")) {
+		klog.V(4).Infof("In %s table, iptables rule \"%s\", already exists.", table, strings.Join(ruleSpec, " "))
+		return nil
+	} else {
+		if err = ipt.Insert(table, chain, position, ruleSpec...); err != nil {
+			klog.Errorf("In %s table, unable to insert iptables rule \"%s\": %v\n", table, strings.Join(ruleSpec, " "), err)
+		}
+	}
+	return nil
+}
--- a/pkg/routeagent/controllers/route/iptables.go
+++ b/pkg/routeagent/controllers/route/iptables.go
@@ -2,6 +2,7 @@ package route
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
@@ -15,13 +16,13 @@ func (r *Controller) createIPTableChains() error {
 		return err
 	}
 
-	klog.V(4).Infof("Install/ensure %s chain exists", SM_POSTROUTING_CHAIN)
-	if err = ipt.NewChain("nat", SM_POSTROUTING_CHAIN); err != nil {
-		klog.Errorf("Unable to create %s chain in iptables: %v", SM_POSTROUTING_CHAIN, err)
+	klog.V(4).Infof("Install/ensure %s chain exists", SmPostRoutingChain)
+	if err = ipt.NewChain("nat", SmPostRoutingChain); err != nil {
+		klog.Errorf("Unable to create %s chain in iptables: %v", SmPostRoutingChain, err)
 	}
 
-	klog.V(4).Infof("Insert %s rule that has rules for inter-cluster traffic", SM_POSTROUTING_CHAIN)
-	forwardToSubPostroutingRuleSpec := []string{"-j", SM_POSTROUTING_CHAIN}
+	klog.V(4).Infof("Insert %s rule that has rules for inter-cluster traffic", SmPostRoutingChain)
+	forwardToSubPostroutingRuleSpec := []string{"-j", SmPostRoutingChain}
 	if err = r.insertUnique(ipt, "nat", "POSTROUTING", 1, forwardToSubPostroutingRuleSpec); err != nil {
 		klog.Errorf("Unable to insert iptable rule in NAT table, POSTROUTING chain: %v", err)
 	}
@@ -37,13 +38,13 @@ func (r *Controller) createIPTableChains() error {
 	}
 
 	klog.V(4).Infof("Allow VxLAN incoming traffic in SUBMARINER-INPUT Chain")
-	ruleSpec := []string{"-p", "udp", "-m", "udp", "--dport", VXLAN_PORT, "-j", "ACCEPT"}
+	ruleSpec := []string{"-p", "udp", "-m", "udp", "--dport", strconv.Itoa(VxLANPort), "-j", "ACCEPT"}
 	if err = ipt.AppendUnique("filter", "SUBMARINER-INPUT", ruleSpec...); err != nil {
 		klog.Errorf("Unable to append iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
 	}
 
-	klog.V(4).Infof("Insert rule to allow traffic over %s interface in FORWARDing Chain", VXLAN_IFACE)
-	ruleSpec = []string{"-o", VXLAN_IFACE, "-j", "ACCEPT"}
+	klog.V(4).Infof("Insert rule to allow traffic over %s interface in FORWARDing Chain", VxLANIface)
+	ruleSpec = []string{"-o", VxLANIface, "-j", "ACCEPT"}
 	if err = r.insertUnique(ipt, "filter", "FORWARD", 1, ruleSpec); err != nil {
 		klog.Errorf("Unable to insert iptable rule in filter table to allow vxlan traffic: %v", err)
 	}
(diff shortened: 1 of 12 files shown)


sridhargaddam:
@mangelajo , @skitt , @tpantelis , @mpeterson Can you please review the patch, thank you.

pkg/routeagent/controllers/route/route.go
```diff
@@ -24,30 +31,51 @@ type Controller struct {
...
+	remoteVTEPs []string
+
+	isGatewayNode bool
+	link          *net.Interface
 }
 
-func NewController(clusterID string, objectNamespace string, link *net.Interface, submarinerClientSet clientset.Interface,
-	clusterInformer informers.ClusterInformer, endpointInformer informers.EndpointInformer) *Controller {
+const VXLAN_IFACE = "vxlan100"
+const VXLAN_PORT = "4800"
```
skitt: Couldn’t these be integers? That would mean `strconv.Itoa` when they’re needed as strings, but it seems more appropriate to me to have these as integers than strings...

pkg/routeagent/controllers/route/route.go
```diff
@@ -24,30 +31,51 @@ type Controller struct {
...
+	vxlanGw     net.IP
+	remoteVTEPs []string
+
+	isGatewayNode bool
+	link          *net.Interface
 }
 
-func NewController(clusterID string, objectNamespace string, link *net.Interface, submarinerClientSet clientset.Interface,
-	clusterInformer informers.ClusterInformer, endpointInformer informers.EndpointInformer) *Controller {
+const VXLAN_IFACE = "vxlan100"
```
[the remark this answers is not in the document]
sridhargaddam: Yeah, I was following the C style :-) 
As you said, it looks like go-standard library uses camelCase. Let me update it.

pkg/routeagent/controllers/route/route.go
```diff
@@ -24,30 +32,55 @@ type Controller struct {
...
+	isGatewayNode bool
+	link          *net.Interface
 }
 
-func NewController(clusterID string, objectNamespace string, link *net.Interface, submarinerClientSet clientset.Interface,
-	clusterInformer informers.ClusterInformer, endpointInformer informers.EndpointInformer) *Controller {
+const (
+	VxLANIface             = "vxlan100"
+	VxLANPort              = 4800
+	VxLANVTepNetworkPrefix = 240
```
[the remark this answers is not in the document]
sridhargaddam: Done

pkg/routeagent/controllers/route/vxlan.go
```diff
@@ -0,0 +1,178 @@
...
+
+		if isVxlanConfigTheSame(iface.link, existing) {
+			klog.V(4).Infof("VxLAN interface already exists with same configuration.")
+			iface.link = existing.(*netlink.Vxlan)
+			return nil
+		}
+
+		// Config does not match, delete the existing interface and re-create it.
+		if err = netlink.LinkDel(existing); err != nil {
+			klog.V(4).Infof("Failed to delete the existing vxlan interface: %v", err)
```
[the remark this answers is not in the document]
sridhargaddam: Done

pkg/routeagent/controllers/route/vxlan.go
```diff
@@ -0,0 +1,178 @@
...
+			return err
+		}
+
+		if err = netlink.LinkAdd(iface.link); err != nil {
+			klog.V(4).Infof("Failed to re-create the the vxlan interface: %v", err)
```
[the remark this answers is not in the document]
sridhargaddam: Done

pkg/routeagent/controllers/route/vxlan.go
```diff
@@ -0,0 +1,178 @@
...
+			return err
+		}
+	} else if err != nil {
+		klog.V(4).Infof("Failed to create the the vxlan interface: %v", err)
```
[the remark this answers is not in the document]
sridhargaddam: Done

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	for _, localClusterCidr := range r.localClusterCidr {
+		ruleSpec := []string{"-s", localClusterCidr, "-d", remoteCidrBlock, "-j", "ACCEPT"}
+		klog.V(4).Infof("Installing iptables rule for outgoing traffic: %s", strings.Join(ruleSpec, " "))
+		if err = ipt.AppendUnique("nat", SmPostRoutingChain, ruleSpec...); err != nil {
+			klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+
+		// Todo: revisit, we only have to program rules to allow traffic from the podCidr
```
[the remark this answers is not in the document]
sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
```diff
@@ -24,30 +32,74 @@ type Controller struct {
...
 	clusterWorkqueue  workqueue.RateLimitingInterface
 	endpointWorkqueue workqueue.RateLimitingInterface
+	podWorkqueue      workqueue.RateLimitingInterface
 
-	gw      net.IP
-	subnets []string
+	gatewayNodeIP    net.IP
+	localClusterCidr []string
+	localServiceCidr []string
+	remoteSubnets    *util.StringSet
```
tpantelis: `remoteSubnets` is shared between the _Cluster_ and _Endpoint_  processing.  `StringSet` is thread-safe so concurrent access isn't a problem. However `remoteSubnets` is populated via the _Cluster_  processing so will this cause an issue with ordering, eg if an _Endpoint_ is observed before a _Cluster_  and `remoteSubnets` is not yet populated?

pkg/routeagent/controllers/route/route.go
```diff
@@ -24,30 +32,74 @@ type Controller struct {
...
 
-	link *net.Interface
+	vxlanDevice *vxLanIface
```
tpantelis: `vxlanDevice` is shared between the _Pod_ and _Endpoint_ processing and thus can be accessed by multiple threads so needs to be synchronized or use the same workqueue goroutine for both _Pod_ and _Endpoint_.  I think the latter would be safest to avoid any potential concurrency issues (also wrt `remoteVTEPs` and `isGatewayNode`).

pkg/routeagent/controllers/route/route.go
```diff
@@ -243,10 +476,19 @@ func (r *Controller) enqueueEndpoint(obj interface{}) {
...
+func (r *Controller) enqueuePod(obj interface{}) {
+	klog.V(6).Infof("Enqueueing pod for route controller %v", obj)
+	pod := obj.(*k8sv1.Pod)
+
+	// Add the POD event to the workqueue only if the sm-route-agent podIP does not exist in the local cache.
+	if !r.remoteVTEPs.Contains(pod.Status.HostIP) {
+		r.podWorkqueue.AddRateLimited(obj)
+	}
+}
+
```
tpantelis: The `vxlanDevice` is created via an _Endpoint_ add - should `handleRemovedEndpoint` cleanup the `vxlanDevice`?

pkg/routeagent/controllers/route/route.go
```diff
@@ -165,6 +255,136 @@ func (r *Controller) processNextCluster() bool {
...
+		defer r.podWorkqueue.Done(pod)
+		pod := pod.(*k8sv1.Pod)
+
+		klog.V(4).Infof("In processNextPod, POD HostIP is %s", pod.Status.HostIP)
+		r.populateRemoteVtepIps(pod.Status.PodIP)
+
+		// A new Node (identified via a Submariner-route-agent daemonset pod event) is added to the cluster.
+		// On the GatewayDevice, update the vxlan fdb entry (i.e., remote Vtep) for the newly added node.
+		if r.isGatewayNode {
+			ret := r.vxlanDevice.AddFDB(net.ParseIP(pod.Status.PodIP), "00:00:00:00:00:00")
```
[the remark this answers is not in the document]
tpantelis: `vxlanDevice` could be nil here if an _Endpoint_ has not yet been observed.

pkg/routeagent/controllers/route/route.go
```diff
@@ -280,23 +522,38 @@ func (r *Controller) handleRemovedCluster(obj interface{}) {
...
 }
 
+func (r *Controller) handleRemovedPod(obj interface{}) {
+	klog.V(6).Infof("Removing podIP in route controller %v", obj)
+	pod := obj.(*k8sv1.Pod)
+
+	if r.remoteVTEPs.Contains(pod.Status.HostIP) {
+		r.remoteVTEPs.Delete(pod.Status.HostIP)
+		if r.isGatewayNode {
+			ret := r.vxlanDevice.DelFDB(net.ParseIP(pod.Status.PodIP), "00:00:00:00:00:00")
```
tpantelis: `vxlanDevice` could be nil here if an _Endpoint_ has not yet been observed.

pkg/routeagent/controllers/route/route.go
```diff
@@ -243,10 +476,19 @@ func (r *Controller) enqueueEndpoint(obj interface{}) {
...
-	klog.V(4).Infof("Enqueueing endpoint for route controller %v", obj)
 	r.endpointWorkqueue.AddRateLimited(key)
 }
 
+func (r *Controller) enqueuePod(obj interface{}) {
+	klog.V(6).Infof("Enqueueing pod for route controller %v", obj)
+	pod := obj.(*k8sv1.Pod)
+
+	// Add the POD event to the workqueue only if the sm-route-agent podIP does not exist in the local cache.
+	if !r.remoteVTEPs.Contains(pod.Status.HostIP) {
```
tpantelis: I'd suggest doing this check in `processNextPod` for safety.

mangelajo:
@sridhargaddam this fails before e2e being ran (now we have the right helm charts) because of this: https://travis-ci.com/submariner-io/submariner/builds/126219311#L1244

pkg/routeagent/controllers/route/route.go:313:19: S1005: should omit value from range; this loop is equivalent to `for fdbAddress := range ...` (gosimple)

		for fdbAddress, _ := range r.remoteVTEPs.Set {

		                ^

pkg/routeagent/controllers/route/route.go:621:17: S1005: should omit value from range; this loop is equivalent to `for cidrBlock := range ...` (gosimple)

	for cidrBlock, _ := range r.remoteSubnets.Set {

	               ^

[submariner]$ ./scripts/ci keep 1.14.2 false false

pkg/routeagent/controllers/route/route.go
sridhargaddam: Good point. I'll update the code. Thanks.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
```diff
@@ -24,30 +31,51 @@ type Controller struct {
...
-		endpointsSynced:     endpointInformer.Informer().HasSynced,
-		clusterWorkqueue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Clusters"),
-		endpointWorkqueue:   workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Endpoints"),
+		clusterID:              clusterID,
+		objectNamespace:        objectNamespace,
+		localClusterCidr:       ClusterCidr,
+		localServiceCidr:       ServiceCidr,
+		submarinerClientSet:    submarinerClientSet,
+		clientSet:              clientSet,
+		link:                   link,
```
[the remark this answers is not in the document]
sridhargaddam: I renamed it to defaultHostIface to avoid confusion with vxlan iface.

pkg/routeagent/controllers/route/route.go
```diff
@@ -243,10 +476,19 @@ func (r *Controller) enqueueEndpoint(obj interface{}) {
...
 	r.endpointWorkqueue.AddRateLimited(key)
 }
 
+func (r *Controller) enqueuePod(obj interface{}) {
+	klog.V(6).Infof("Enqueueing pod for route controller %v", obj)
+	pod := obj.(*k8sv1.Pod)
+
+	// Add the POD event to the workqueue only if the sm-route-agent podIP does not exist in the local cache.
+	if !r.remoteVTEPs.Contains(pod.Status.HostIP) {
+		r.podWorkqueue.AddRateLimited(obj)
```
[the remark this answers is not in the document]
sridhargaddam: I see that for both Cluster/Endpoint the key (cache.MetaNamespaceKeyFunc) is enqueued, but while handling the event, they are again making a REST call to get that object.

i.e., 
		cluster, err := r.submarinerClientSet.SubmarinerV1().Clusters(ns).Get(name, metav1.GetOptions{})
		if err != nil {
			return fmt.Errorf("Error retrieving submariner cluster object %s: %v", name, err)
		}

Is this the expected usage? Somehow I feel this is not very optimal.

pkg/routeagent/controllers/route/route.go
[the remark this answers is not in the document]
sridhargaddam: Done. I've not seen people using camelCase for interface names. So, lets use vx-submariner which also fits within Linux max interface name length.

pkg/routeagent/controllers/route/route.go
```diff
@@ -66,6 +118,14 @@ func NewController(clusterID string, objectNamespace string, link *net.Interface
 		DeleteFunc: controller.handleRemovedEndpoint,
 	})
 
+	podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: controller.enqueuePod,
+		UpdateFunc: func(old, new interface{}) {
+			controller.enqueuePod(new)
+		},
+		DeleteFunc: controller.handleRemovedPod,
```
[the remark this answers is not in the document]
tpantelis: I know this pattern is used elsewhere in this code but it is not technically correct. If a delete occurred right after an add or update, the delete could be processed first. That's probably not likely to occur but we should be safe and ensure we process events in the order they are received.

pkg/routeagent/controllers/route/route.go
tpantelis: One potential issue is that the `vxlanDevice` variable is not synchronized so it's possible, albeit unlikely,  that the thread processing pods never sees the value for `vxlanDevice` written by the thread processing endpoints due to CPU caching. To be completely safe, all accesses of `vxlanDevice` need to be protected by a mutex or `vxlanDevice` needs to be an atomic reference.  Similarly with `isGatewayNode`. 

Another possible issue is that you set `isGatewayNode` before  `vxlanDevice` however w/o synchronization, it's possible that if the pod thread ran concurrently, it may observe the update to `vxlanDevice` but not `isGatewayNode` which would lead to incorrect state (ie it may not call `AddFDB` when it should).  There may be other concurrency issues lurking. It may be that things would eventually converge properly due to the 30 sec resync period but it's very difficult/impossible to think of and test all such scenarios and ensure correctness. So I think it's safest to take concurrency out of the equation altogether by using the same workqueue for both pods and endpoints rather than using shared memory.

pkg/routeagent/controllers/route/route.go
```diff
@@ -165,6 +258,153 @@ func (r *Controller) processNextCluster() bool {
...
+				return fmt.Errorf("Failed to add FDB entry on the Gateway Node vxlan iface %v", err)
+			}
+		}
+
+		// Enable loose mode (rp_filter=2) reverse path filtering on the vxlan interface.
+		err = ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+VxLANIface+"/rp_filter", []byte("2"), 0644)
+		if err != nil {
+			return fmt.Errorf("Unable to update vxlan rp_filter proc entry, err: %s", err)
+		} else {
+			klog.Info("Successfully configured rp_filter to loose mode(2) ")
```
tpantelis: Should this be debug level?

pkg/routeagent/controllers/route/route.go
```diff
@@ -165,6 +258,153 @@ func (r *Controller) processNextCluster() bool {
...
+	obj, shutdown := r.podWorkqueue.Get()
+	if shutdown {
+		return false
+	}
+	err := func() error {
+		defer r.podWorkqueue.Done(obj)
+
+		key := obj.(string)
+		ns, name, err := cache.SplitMetaNamespaceKey(key)
+		if err != nil {
```
tpantelis: You need to call `r.podWorkqueue.Forget(obj)` here and all return paths below that don't re-queue.

pkg/routeagent/controllers/route/route.go
```diff
@@ -195,18 +415,32 @@ func (r *Controller) processNextEndpoint() bool {
...
+		ipSlice[0] = strconv.Itoa(VxLANVTepNetworkPrefix)
+		// remoteVtepIP is used while programming the routing rules
+		remoteVtepIP := net.ParseIP(strings.Join(ipSlice, "."))
+		r.vxlanGw = remoteVtepIP
+
 		if endpoint.Spec.Hostname == hostname {
 			r.cleanRoutes()
+			r.isGatewayNode = true
+			if r.createVxLANInterface(true) != nil {
+				klog.Fatalf("Unable to create VxLAN interface on GatewayNode (%s): %v", hostname, err)
```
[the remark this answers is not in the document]
tpantelis: > So far in my testing, I did not see such errors at the same time I agree that we need the code to be tested extensively. Shall we keep it in the current state and based on further testing revisit this part (if required)?

Is it severe enough to abruptly exit the process? I'm not sure but If you think it is then I'm fine with it.

pkg/routeagent/controllers/route/route.go
tpantelis: yeah there can be hundreds of pods so shortcutting should be fine. However I'm wondering if we should only do this for an update...

pkg/routeagent/controllers/route/route.go
```diff
@@ -165,6 +258,153 @@ func (r *Controller) processNextCluster() bool {
...
+		}
+
+		klog.V(4).Infof("In processNextPod, POD HostIP is %s", pod.Status.HostIP)
+		r.populateRemoteVtepIps(pod.Status.PodIP)
+
+		// A new Node (identified via a Submariner-route-agent daemonset pod event) is added to the cluster.
+		// On the GatewayDevice, update the vxlan fdb entry (i.e., remote Vtep) for the newly added node.
+		if r.isGatewayNode {
+			if r.vxlanDevice != nil {
+				err := r.vxlanDevice.AddFDB(net.ParseIP(pod.Status.PodIP), "00:00:00:00:00:00")
```
tpantelis: It's possible `r.vxlanDevice` could be nil right after the check on line 383. To avoid this, you should  first store `r.vxlanDevice` in a local var. Or eliminate concurrency as I suggested in another comment.

pkg/routeagent/controllers/route/route.go
```diff
@@ -273,30 +543,56 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
...
 }
 
+func (r *Controller) handleRemovedPod(obj interface{}) {
+	klog.V(6).Infof("Removing podIP in route controller %v", obj)
+	pod := obj.(*k8sv1.Pod)
+
+	if r.remoteVTEPs.Contains(pod.Status.HostIP) {
+		r.remoteVTEPs.Delete(pod.Status.HostIP)
+		if r.isGatewayNode && r.vxlanDevice != nil {
+			ret := r.vxlanDevice.DelFDB(net.ParseIP(pod.Status.PodIP), "00:00:00:00:00:00")
```
tpantelis: Similar as in `processNextPod` - `r.vxlanDevice` could be nil here.

pkg/routeagent/controllers/route/vxlan.go
```diff
@@ -0,0 +1,188 @@
...
+		if isVxlanConfigTheSame(iface.link, existing) {
+			klog.V(4).Infof("VxLAN interface already exists with same configuration.")
+			iface.link = existing.(*netlink.Vxlan)
+			return nil
+		}
+
+		// Config does not match, delete the existing interface and re-create it.
+		if err = netlink.LinkDel(existing); err != nil {
+			klog.Errorf("Failed to delete the existing vxlan interface: %v", err)
+			return err
```
tpantelis: As commented earlier and for consistency, can we return a formatted error instead of logging? Same elsewhere in this file.

pkg/routeagent/main.go
```diff
@@ -53,10 +63,19 @@ func main() {
 
 	submarinerInformerFactory := submarinerInformers.NewSharedInformerFactoryWithOptions(submarinerClient, time.Second*30, submarinerInformers.WithNamespace(srcs.Namespace))
 
+	clientSet, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		klog.Errorf("Error building clientset: %s", err.Error())
+		return
```
tpantelis: This could call `Fatalf`.

pkg/routeagent/controllers/route/route.go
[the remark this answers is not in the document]
tpantelis: Actually looking at this further, we can't enqueue the _removes_ b/c we need access to the pod's state which we wouldn't be available in `processNextPod` since it only has the key.  Same with endpoints (sorry I missed that detail originally). Also it shouldn't matter if a delete is processed before an add b/c the `Get` in `processNextPod` would fail anyway.

pkg/routeagent/controllers/route/route.go
tpantelis: Given that we can't enqueue _removes_ as mentioned in a prior comment, it wouldn't buy us anything to share the same workqueue as we need shared memory anyway. However I still think we should synchronize access to `vxlanDevice` and `isGatewayNode` atomically to be safe.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Why will vxlanDevice become nil? Once we create a VxLAN interface on the host, we will not delete it. Am I missing something?

pkg/routeagent/controllers/route/route.go
sridhargaddam: IMHO I don't see any major advantage with that. I feel its currently good and we can revisit this code if required.

pkg/routeagent/controllers/route/route.go
```diff
@@ -273,30 +543,56 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
...
+	if object.Spec.ClusterID == r.clusterID {
+		klog.V(6).Infof("Endpoint matches the cluster ID of this cluster")
+		err := r.vxlanDevice.deleteVxLanIface()
+		if err != nil {
+			klog.Errorf("Failed to delete the the vxlan interface on endpoint removal: %v", err)
+			return
+		}
+		r.vxlanDevice = nil
+	}
+
```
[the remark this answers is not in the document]
sridhargaddam: As discussed in earlier comment, this can be marked as resolved.

pkg/routeagent/controllers/route/route.go
sridhargaddam: As above, please clarify why vxlanDevice would become nil after its created.

pkg/routeagent/controllers/route/vxlan.go
sridhargaddam: Done

pkg/routeagent/main.go
sridhargaddam: Done

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -45,48 +44,65 @@ func (r *Controller) createIPTableChains() error {
...
 
-	if strings.Contains(rules[position], strings.Join(ruleSpec, " ")) {
+	if strings.Contains(rules[1], strings.Join(ruleSpec, " ")) {
 		klog.V(4).Infof("In %s table, iptables rule \"%s\", already exists.", table, strings.Join(ruleSpec, " "))
 		return nil
 	} else {
-		if err = ipt.Insert(table, chain, position, ruleSpec...); err != nil {
-			klog.Errorf("In %s table, unable to insert iptables rule \"%s\": %v\n", table, strings.Join(ruleSpec, " "), err)
+		if err = ipt.Insert(table, chain, 1, ruleSpec...); err != nil {
+			return err
```
mangelajo: This can still have the issue I commented about:

Initial state

<RULE A>

You insert:

<MY RULE>
<RULE A> 

Somebody comes and inserts:

<ANOTHER RULE>
<MY RULE>
<RULE A>

You run the function next time (service reboot):
<MY RULE>
<ANOTHER RULE>
<MY RULE>
<RULE A>

unless I'm missing something

sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
tpantelis: In `handleRemovedEndpoint`, you set `r.vxlanDevice` to nil. Let's say thread 1 executes line 394 here and observes  `r.vxlanDevice` non-nil. Thread 2 then interleaves and executes line 566 in `handleRemovedEndpoint` to set it to nil. Thread 1 resumes and executes line 395 but now `r.vxlanDevice` is nil. Doing the following would alleviate that potential issue:

```
localVxlanDevice := r.vxlanDevice
if localVxlanDevice != nil {
    localVxlanDevice. AddFDB(...)
{
```

pkg/routeagent/controllers/route/route.go
tpantelis: I'm not sure why this conversation was marked as resolved. We still need to address synchronization of `vxlanDevice` and `isGatewayNode` as discussed above.

sridhargaddam: Oops, it was accidentally marked as resolved.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Thanks for the explanation. The chances of endpoint being removed AFAIU is very less (in general). Anyways, lets protect it with a Mutex, to be safe. Please take a look at the updated code.

sridhargaddam:
Thanks very much @tpantelis for reviewing the patch and approving it. I had to update the patch to fix golangci-lint errors. 

@mangelajo, can you please take a look when you find time. Thanks.

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func (r *Controller) insertUnique(ipt *iptables.IPTables, table string, chain string, position int, ruleSpec []string) error {
+	rules, err := ipt.List(table, chain)
+	if err != nil {
+		return fmt.Errorf("error listing the rules in %s chain: %v", chain, err)
+	}
+
+	if strings.Contains(rules[position], strings.Join(ruleSpec, " ")) {
```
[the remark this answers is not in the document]
mangelajo: I would do it preventively as won't ever have guarantees on how do CNIs will behave. But not necessarily in this patch, may be we can open an issue with medium priority so we don't forget ? You could keep a link to this conversation/code.

pkg/routeagent/controllers/route/route.go
mangelajo: Yeah @sridhargaddam remember the thread about memory synchronization between CPU cores we had. If we don't mutex, we have no guarantee that the vxlanDevice will be read by other threads than the one which wrote it, although, caches are small and it's very likely, what when it hits a processor with a bigger cache?

pkg/routeagent/controllers/route/route.go
mangelajo: Current code looks ok for me too in this sense.

PR #147, opened 2019-09-10
Update checksums for Go 1.13

skitt:
Go 1.13 requires longer git SHA-1 prefixes, and checks the timestamp
exactly. This patch fixes the strongSwanVICI fork prefix and
timestamp, and updates the 1.13 dependencies in go.sum.


skitt:
@mangelajo could you take a look at this? Thanks!

PR #148, opened 2019-09-11
E2E scheduling of pods to GW/NonGW nodes

mangelajo:
Before this patch, E2E didn't have any control of where pods
landed, this made the testing of different scenarios random.

With this patch we ensure the testing of the datapath when
traffic needs to transit from a worker node to a gateway node.

Cluster type has been converted to an enumeration as part of this
commit.

### DIFF
--- a/test/e2e/dataplane/tcp_pod_to_service.go
+++ b/test/e2e/dataplane/tcp_pod_to_service.go
@@ -4,7 +4,6 @@ import (
 	"strings"
 
 	"github.com/submariner-io/submariner/test/e2e/framework"
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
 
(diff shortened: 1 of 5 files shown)


test/e2e/framework/framework.go
```diff
@@ -26,10 +26,12 @@ const (
 	PollInterval = 100 * time.Millisecond
 )
 
+type TestCluster int
```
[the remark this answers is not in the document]
mangelajo: Could work too, TestCluster may be save for a more complex structure where necessary.

test/e2e/framework/network_pods.go
```diff
@@ -86,3 +88,52 @@ func (f *Framework) CreateTCPCheckConnectorPod(cluster int, remoteCheckPod *v1.P
...
+)
+
+func nodeAffinity(scheduling TestPodScheduling) *v1.Affinity {
+
+	const gatewayLabel = "submariner.io/gateway"
+
+	var nodeSelReqs []v1.NodeSelectorRequirement
+
+	switch scheduling {
+	case GatewayNode:
```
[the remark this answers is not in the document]
mangelajo: true, thanks. I'll change that.

mangelajo:
@sridhargaddam @mpeterson I posted a second commit with your requested changes, please re-review.

test/e2e/dataplane/tcp_pod_to_service.go
```diff
@@ -15,8 +15,8 @@ var _ = Describe("[dataplane] Basic Pod to Service tests across clusters without
 		testPod2ServiceTCP(f, framework.NonGatewayNode, framework.NonGatewayNode)
 	})
 
-	It("Should be able to perform a Pod to Service TCP connection and exchange data between different clusters NonGW to GW", func() {
-		testPod2ServiceTCP(f, framework.NonGatewayNode, framework.GatewayNode)
+	It("Should be able to perform a Pod to Service TCP connection and exchange data between different clusters GW to GW", func() {
```
sridhargaddam: Thanks for addressing the other comments.

Lets validate GW to GW use-case before nonGW to nonGW (i.e., move this to line before 14). 
If GW to GW fails, then the second use-case is bound to fail.

mangelajo:
The order does not matter , all tests execute in random order always.

Enviado desde mi iPhone

sridhargaddam:
> The order does not matter , all tests execute in random order always. 

I see, thanks for clarifying. LGTM.

Enviado desde mi iPhone
> […](#)
> El 16 sept 2019, a las 20:18, Sridhar Gaddam ***@***.***> escribió: @sridhargaddam requested changes on this pull request. In test/e2e/dataplane/tcp_pod_to_service.go: > @@ -15,8 +15,8 @@ var _ = Describe("[dataplane] Basic Pod to Service tests across clusters without testPod2ServiceTCP(f, framework.NonGatewayNode, framework.NonGatewayNode) }) -	It("Should be able to perform a Pod to Service TCP connection and exchange data between different clusters NonGW to GW", func() { -	testPod2ServiceTCP(f, framework.NonGatewayNode, framework.GatewayNode) +	It("Should be able to perform a Pod to Service TCP connection and exchange data between different clusters GW to GW", func() { Thanks for addressing the other comments. Lets validate GW to GW use-case before nonGW to nonGW (i.e., move this to line before 14). If GW to GW fails, then the second use-case is bound to fail. — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or mute the thread.

test/e2e/framework/network_pods.go
```diff
@@ -86,3 +88,48 @@ func (f *Framework) CreateTCPCheckConnectorPod(cluster int, remoteCheckPod *v1.P
...
+	}
+	return &affinity
+}
+
+func addNodeSelectorRequirement(nodeSelReqs *[]v1.NodeSelectorRequirement, label string, op v1.NodeSelectorOperator, values []string) {
+	*nodeSelReqs = append(*nodeSelReqs, v1.NodeSelectorRequirement{
+		Key:      label,
+		Operator: v1.NodeSelectorOpNotIn,
+		Values:   []string{"true"},
+	})
```
tpantelis: Minor thing but I think it's more intuitive and idiomatic in Go to return the new `[]v1.NodeSelectorRequirement` from the function rather than via a formal parameter.

mangelajo: I'll change it, I was trying to avoid code repetition as much as possible. I guess that pattern is good for parallelization.

mangelajo:
I had to fix conflicts with the SRC IP preservation changes.

Excuse the repetition which I'm handling in PR#149 in a different way (by making the framework a bit smarter and cleaner).. automating some of the repetitive tasks.

test/e2e/dataplane/tcp_pod_to_service.go
mangelajo: Just making a note on the Github interface that wasn't captured here (Sorry for the repetition)..

ginkgo runs the tests randomly so the order does not really matter.

PR #151, opened 2019-09-16
Add SubM operator, tests, e2e integration

dfarrell07:


dfarrell07:
Build is failing with:

```
Makefile:17: recipe for target 'ci' failed
make: *** [ci] Error 1
The command "make ci e2e status=keep" exited with 2.
```

Maybe it's related to by new `deploytool=operator` flag being thrown off by the invocation used in CI (with the extra `ci` field). I'll take a look.

mangelajo:
Something happened with CI:

[submariner]$ PACKAGES=*.go operators pkg
Incorrect formatting, please run goimports
[submariner]$ exit 1
[submariner]$ ./scripts/ci keep 1.14.2 false false helm
FATA[0216] exit status 1                                
Makefile:17: recipe for target 'ci' failed
make: *** [ci] Error 1
The command "make ci e2e status=keep" exited with 2.
Done. Your build exited with 1.

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,221 @@
...
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+  // TODO: Add deployment-creation logic here? See memc example
+
+  // Create submariner-engine SA
```
mangelajo: Makes sense, why commented?

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,221 @@
...
+  //subm_engine_sa := corev1.ServiceAccount{}
+  //subm_engine_sa.Name = "submariner-engine"
+  //reqLogger.Info("Created a new SA", "SA.Name", subm_engine_sa.Name)
+
+  // FIXME: CRDs need to be created before the Operator starts. Use OLM.
```
mangelajo: If need to be created by the OLM, may be just add the comment here that CRDs must be created by OLM itself and remove 104-138 ?

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,221 @@
...
+  // Define clusters.submariner.io CRD
+  //clusters_crd := apiextensions.CustomResourceDefinition{Spec: clusters_crd_spec, Status: clusters_crd_status}
+  //err = r.client.Create(context.TODO(), &clusters_crd)
+  //if err != nil {
+  //  return reconcile.Result{}, err
+  //}
+
+	// Define a new Pod object
+  // TODO: Make this responsive to size
+	pod := newPodForCR(instance)
```
mangelajo: you could do the same thing for the route agent from here ... and then set the controller reference too I suspect

mangelajo: Why do you manage a single pod here instead of a deployment like in the helm charts? 

https://github.com/submariner-io/submariner-charts/blob/master/submariner/templates/engine-deploy.yaml#L2

The helm charts define a deployment, with affinity to the submariner.io/gateway=true label.

Then it's kubernetes who will take care of creating one pod of the engine per node which has such flag. Here you're only deploying a single pod, where?

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// Set Submariner instance as the owner and controller
+	if err := controllerutil.SetControllerReference(instance, pod, r.scheme); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Check if this Pod already exists
```
mangelajo: You could do here the same for both types of deployments (if handling deployments instead, and handling both types of deployments from a single CRD).

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:    "submariner",
+          // TODO: Use var here
+					Image:   "submariner:local",
+          // TODO: Use var here
+					Command: []string{"submariner.sh"},
+          SecurityContext: &security_context_add_net_admin,
+          Env:     []corev1.EnvVar{subm_namespace_env_var, subm_clustercidr_env_var, subm_servicecidr_env_var, subm_token_env_var, subm_clusterid_env_var, subm_colorcodes_env_var, subm_debug_env_var, subm_natenabled_env_var, subm_broker_env_var, broker_k8s_apiserver_env_var, broker_k8s_apiservertoken_env_var, broker_k8s_remotenamespace_env_var, broker_k8s_ca_env_var, ce_ipsec_psk_env_var, ce_ipsec_debug_env_var},
```
mangelajo: you can add multiple labels like this:

https://github.com/submariner-io/submariner/blob/master/test/e2e/framework/network_pods.go#L37

Easier to maintain and to read :-)

operators/go/submariner_controller.go
dfarrell07: Yeah, agreed, I will remove this commented-out code. Was just keeping it around as a nice example of what we needed, but it's more confusing than helpful at this point.

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,185 @@
...
+  operator-sdk add controller --api-version=$api_version --kind=$kind
+
+  controller_file_src=$op_gen_dir/submariner_controller.go
+  controller_file_dst=pkg/controller/submariner/submariner_controller.go
+  cp $controller_file_src $controller_file_dst
+
+  popd
+}
+
+function add_subm_routeagent_to_operator() {
```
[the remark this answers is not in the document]
dfarrell07: Okay, as I said, this makes sense. I raised a Jira to track: https://jira.coreos.com/browse/HYCLD-247

operators/go/submariner_controller.go
dfarrell07: Agreed again, this makes sense. Jira to track: https://jira.coreos.com/browse/HYCLD-248

operators/go/submariner_controller.go
dfarrell07: As above, this change to leveraging deployments and their usefulness vs pods directly makes sense. Jira to track here: https://jira.coreos.com/browse/HYCLD-248

operators/go/submariner_controller.go
dfarrell07: Awesome, thanks - done :+1:

operators/go/submariner_controller.go
dfarrell07: This is done in the latest version.

operators/go/submariner_controller.go
dfarrell07: Okay, we can do that. Jira to use SAs for Engine/Routeagent separately from the Operator SA: https://jira.coreos.com/browse/HYCLD-249

dfarrell07:
> Something happened with CI:
> 
> [submariner]$ PACKAGES=*.go operators pkg
> Incorrect formatting, please run goimports
> [submariner]$ exit 1
> [submariner]$ ./scripts/ci keep 1.14.2 false false helm
> FATA[0216] exit status 1
> Makefile:17: recipe for target 'ci' failed
> make: *** [ci] Error 1
> The command "make ci e2e status=keep" exited with 2.
> Done. Your build exited with 1.

It was an issue with formatting. I fixed those issues in the latest changes.

dfarrell07:
I am seeing other linting errors still, via the extra checks enabled in CI (I was just running e2e locally):

```
WARN [runner/megacheck] Can't run megacheck because of compilation errors in packages [github.com/submariner-io/submariner/operators/go]: operators/go/routeagent_controller.go:1: : found packages routeagent (routeagent_controller.go) and submariner (submariner_controller.go) in /go/src/github.com/submariner-io/submariner/operators/go and 16 more errors: run `golangci-lint run --no-config --disable-all -E typecheck` to see all errors 
panic: interface conversion: types.Type is nil, not *types.Struct
```

dfarrell07:
@manosnoam @skitt - Can you guys try the latest here and report what you get? I'm currently hitting some this unexpected error, trying to track down the problem:

> $ kubectl logs submariner-pod -n submariner

```
<snip>
+ sysctl -w net.ipv4.conf.all.send_redirects=0
sysctl: setting key "net.ipv4.conf.all.send_redirects": Read-only file system
```

skitt:
> @manosnoam @skitt - Can you guys try the latest here and report what you get? I'm currently hitting some this unexpected error, trying to track down the problem:
> 
> > $ kubectl logs submariner-pod -n submariner
> 
> ```
> <snip>
> + sysctl -w net.ipv4.conf.all.send_redirects=0
> sysctl: setting key "net.ipv4.conf.all.send_redirects": Read-only file system
> ```

Same for me.

skitt:
Did you change the container’s security context recently?

dfarrell07:
> Did you change the container’s security context recently?

No, I was trying to keep things mostly stable as we finalize everything this week.

I think it's something that changed under us when I rebased. If I make a new branch from 5cfe40debfe7866e71a0e5842e2ae8c35e8ae213 and then cherry-pick this commit on top (which a little `which` vs `command -v` change to account for container base image change), I don't see this error (my Engine pod seems to deploy fine).

skitt:
> I think it's something that changed under us when I rebased. If I make a new branch from [5cfe40d](https://github.com/submariner-io/submariner/commit/5cfe40debfe7866e71a0e5842e2ae8c35e8ae213) and then cherry-pick this commit on top (with a little `which` vs `command -v` change to account for container base image change), I don't see this error (my Engine pod deploys fine and all tests pass).

Right, that’s very plausible. I’ll try bisecting.

skitt:
> > I think it's something that changed under us when I rebased. If I make a new branch from [5cfe40d](https://github.com/submariner-io/submariner/commit/5cfe40debfe7866e71a0e5842e2ae8c35e8ae213) and then cherry-pick this commit on top (with a little `which` vs `command -v` change to account for container base image change), I don't see this error (my Engine pod deploys fine and all tests pass).
> 
> Right, that’s very plausible. I’ll try bisecting.

And for some unfathomable reason the Submariner pods now deploy correctly again...

manosnoam:
> @manosnoam @skitt - Can you guys try the latest here and report what you get? I'm currently hitting some this unexpected error, trying to track down the problem:

For me it fails earlier on ./gen_subm_operator.sh:
go get k8s.io/kube-state-metrics/pkg/collector: no matching versions for query "latest"

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,181 @@
+package submariner
+
+import (
+	"context"
+
+	submarinerv1alpha1 "github.com/submariner-operator/submariner-operator/pkg/apis/submariner/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+
+	//apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
mkolesnik: nit: delete this?

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,181 @@
...
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Create submariner-engine SA
```
mkolesnik: nit: delete this commented out block?

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	//subm_engine_sa := corev1.ServiceAccount{}
+	//subm_engine_sa.Name = "submariner-engine"
+	//reqLogger.Info("Created a new SA", "SA.Name", subm_engine_sa.Name)
+
+	// Define a new Pod object
+	// TODO: Make this responsive to size
```
mkolesnik: Not sure what this TODO means, what is size?

operators/go/submariner_controller.go
```diff
@@ -0,0 +1,181 @@
...
+
+		// Pod created successfully - don't requeue
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Pod already exists - don't requeue
+	reqLogger.Info("Skip reconcile: Pod already exists", "Pod.Namespace", found.Namespace, "Pod.Name", found.Name)
+	return reconcile.Result{}, nil
```
mkolesnik: nit: reconcile.Result{} can be defined as a constant (I know this is from the scaffold. but its pretty silly repetition of it everywhere IMHO)

operators/go/routeagent_controller.go
```diff
@@ -0,0 +1,165 @@
...
+
+// newPodForCR returns a busybox pod with the same name/namespace as the cr
+func newPodForCR(cr *submarinerv1alpha1.Routeagent) *corev1.Pod {
+	labels := map[string]string{
+		"app": "submariner-routeagent",
+	}
+
+	// Create SecurityContext for Pod
+	// FIXME: Does this really need to be ALL, vs just NET_ADMIN? The current Helm-based deployment gives ALL.
+	//security_context_net_admin_cap := corev1.SecurityContext{Capabilities: &corev1.Capabilities{Add: []corev1.Capability{"NET_ADMIN"}}}
```
mkolesnik: nit: delete this?

scripts/kind-e2e/e2e.sh
```diff
@@ -111,9 +115,13 @@ function setup_broker() {
         helm --kube-context cluster1 install submariner-latest/submariner-k8s-broker --name ${SUBMARINER_BROKER_NS} --namespace ${SUBMARINER_BROKER_NS}
     fi
 
+    # FIXME: Shouldn't this be a dynamic namespace, not hard-coded "default"?
```
mkolesnik: This actually isn't related to this change..

scripts/kind-e2e/e2e.sh
```diff
@@ -322,16 +331,108 @@ setup_custom_cni
 if [[ $3 = true ]]; then
     enable_logging
 fi
-install_helm
-if [[ $4 = true ]]; then
```
mkolesnik: So no kubefed when using the operator?

scripts/kind-e2e/e2e.sh
```diff
@@ -322,16 +331,108 @@ setup_custom_cni
...
-    enable_kubefed
+
+if [[ $5 = operator ]]; then
+    operator=true
+    # TODO: Convert these Broker-deploy steps to use Operator instead of Helm
```
mkolesnik: Seems like they should be outside the if, since you do them in both cases.

scripts/kind-e2e/e2e.sh
```diff
@@ -322,16 +331,108 @@ setup_custom_cni
...
+    verify_subm_broker_secrets cluster1
+
+    for i in 2 3; do
+      # Create CRDs required as prerequisite submariner-engine
+      # TODO: Eventually OLM should handle this
+      create_subm_endpoints_crd cluster$i
+      verify_endpoints_crd cluster$i
+      create_subm_clusters_crd cluster$i
+      verify_clusters_crd cluster$i
+      if [[ $also_routeagent = true ]]; then
```
mkolesnik: Do you expect it to ever not be true?

scripts/kind-e2e/e2e.sh
```diff
@@ -322,16 +331,108 @@ setup_custom_cni
...
+        # Verify SubM CR
+        verify_subm_cr cluster$i
+        # Verify SubM Engine Pod
+        verify_subm_engine_pod cluster$i
+        # Verify SubM Engine container
+        verify_subm_engine_container cluster$i
+        # Verify Engine secrets
+        verify_subm_engine_secrets cluster$i
+      fi
+      if [[ $also_routeagent = true ]]; then
```
mkolesnik: Do you expect it to ever not be true?

scripts/kind-e2e/e2e.sh
```diff
@@ -322,16 +331,108 @@ setup_custom_cni
...
+      # Verify SubM Operator
+      verify_subm_operator cluster$i
+      # Verify SubM Operator pod
+      verify_subm_op_pod cluster$i
+      # Verify SubM Operator container
+      verify_subm_operator_container cluster$i
+
+      # Collect SubM vars for use in SubM CRs
+      collect_subm_vars cluster$i
+      if [[ $also_engine = true ]]; then
```
mkolesnik: Do you expect it to ever not be true?

scripts/kind-e2e/e2e.sh
```diff
@@ -322,16 +331,108 @@ setup_custom_cni
...
+        verify_subm_routeagent_container cluster$i
+        # Verify Routeagent secrets
+        verify_subm_routeagent_secrets cluster$i
+      fi
+    done
+
+    deploy_netshoot_cluster2
+    deploy_nginx_cluster3
+
+    test_connection
```
mkolesnik: These two should be after the condition block since you execute them in both cases

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,192 @@
...
+
+function add_subm_engine_to_operator() {
+  pushd $op_dir
+  api_version=submariner.io/v1alpha1
+  kind=Submariner
+  operator-sdk add api --api-version=$api_version --kind=$kind
+
+  # Define spec fields
+  types_file=pkg/apis/submariner/v1alpha1/submariner_types.go
+  sed -i '/SubmarinerSpec struct/a \ \ Count int32 `json:"count"`' $types_file
```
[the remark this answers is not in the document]
dfarrell07: Yeah, that's a good idea. I was trying to stick with a design that depended on as few external files as possible - just the gen-operator bash script and Go controllers. That was the form of simplicity I was shooting for, but I'm open to do something different.

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,192 @@
...
+  operator-sdk new submariner-operator --verbose
+  popd
+
+  pushd $op_dir
+  sed -i 's|REPLACE_NAMESPACE|submariner|g' deploy/role_binding.yaml
+
+  sed -i "s|REPLACE_IMAGE|quay.io/submariner/submariner-operator:$version|g" deploy/operator.yaml
+
+  # Create a definition namespace for SubM
+  ns_file=deploy/namespace.yaml
```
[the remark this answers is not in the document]
dfarrell07: Yeah, in some was that would be simpler and I'm open to that change. This is a part of the "keep it in the script when possible for simplicity" design I detailed in a similar comment below.

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,192 @@
...
+GOPATH=$HOME/go
+
+version=0.0.1
+add_engine=true
+add_routeagent=true
+openapi_checks_enabled=false
+push_image=false
+op_dir=$GOPATH/src/github.com/submariner-operator/submariner-operator
+op_gen_dir=$GOPATH/src/github.com/submariner-io/submariner/operators/go
+op_out_dir=$GOPATH/src/github.com/submariner-io/submariner/operators/go/submariner-operator
```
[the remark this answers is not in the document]
dfarrell07: Yeah, that might be clearer. I had it like that at one point but as I was refactoring I thought it was simpler to make everything relative to GOPATH vs a string of related local vars.

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,192 @@
...
+  fi
+
+  operator-sdk add controller --api-version=$api_version --kind=$kind
+
+  controller_file_src=$op_gen_dir/routeagent_controller.go
+  controller_file_dst=pkg/controller/routeagent/routeagent_controller.go
+  cp $controller_file_src $controller_file_dst
+
+  popd
+}
```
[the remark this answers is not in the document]
dfarrell07: Yeah, good idea. I raised a Jira to track this: https://jira.coreos.com/browse/HYCLD-251

The general theme of a lot of answers to other review comments applies here - "yes, good idea, I'm just trying to keep it as simple as possible for now". There are definitely opportunities to make things more DRY and easy to extend for future controllers.

operators/go/routeagent_controller.go
dfarrell07: Yeah, fair, I just kept it as an example of what I think the permissions should be (vs copying what they currently are with Helm).

scripts/kind-e2e/e2e.sh
dfarrell07: Good suggestion, thanks, done in the latest version. :+1:

scripts/kind-e2e/e2e.sh
dfarrell07: Good suggestion, thanks, done in the latest version. :+1:

scripts/kind-e2e/e2e.sh
dfarrell07: I did when I was developing them. Now, yeah you're right it doesn't seem useful to have this flag. I've removed it in the latest version.

scripts/kind-e2e/e2e.sh
dfarrell07: I did when I was developing them. Now, yeah you're right it doesn't seem useful to have this flag. I've removed it in the latest version.

scripts/kind-e2e/e2e.sh
dfarrell07: I did when I was developing them. Now, yeah you're right it doesn't seem useful to have this flag. I've removed it in the latest version.

scripts/kind-e2e/e2e.sh
dfarrell07: Okay, I removed it in the latest version.

operators/go/submariner_controller.go
dfarrell07: I think the plan is to eventually uncomment this logic and use different SAs for Engine/Routeagent/Operator. I raised this Jira to track, after discussion with @mangelajo on a different comment here: https://jira.coreos.com/browse/HYCLD-249

operators/go/submariner_controller.go
dfarrell07: Good catch, fixed in the latest, thanks!

dfarrell07:
> > > I think it's something that changed under us when I rebased. If I make a new branch from [5cfe40d](https://github.com/submariner-io/submariner/commit/5cfe40debfe7866e71a0e5842e2ae8c35e8ae213) and then cherry-pick this commit on top (with a little `which` vs `command -v` change to account for container base image change), I don't see this error (my Engine pod deploys fine and all tests pass).
> > 
> > 
> > Right, that’s very plausible. I’ll try bisecting.
> 
> And for some unfathomable reason the Submariner pods now deploy correctly again...

Yeah, it's a strange one. I'm trying to bisect as well. So far I'm pretty sure I can consistently reproduce it when I base this commit on the latest from master, and that the same Operator code fully passes for me when it's based on 3b46ebdd5ad9cf5339bd5fc011dfbf1b7e3be9aa or b0fccd7472bad4e6e0e244723543696dccaf2042.

dfarrell07:
> > @manosnoam @skitt - Can you guys try the latest here and report what you get? I'm currently hitting some this unexpected error, trying to track down the problem:
> 
> For me it fails earlier on ./gen_subm_operator.sh:
> go get k8s.io/kube-state-metrics/pkg/collector: no matching versions for query "latest"

I think this was resolved on https://jira.coreos.com/browse/HYCLD-250.

dfarrell07:
> > > > I think it's something that changed under us when I rebased. If I make a new branch from [5cfe40d](https://github.com/submariner-io/submariner/commit/5cfe40debfe7866e71a0e5842e2ae8c35e8ae213) and then cherry-pick this commit on top (with a little `which` vs `command -v` change to account for container base image change), I don't see this error (my Engine pod deploys fine and all tests pass).
> > > 
> > > 
> > > Right, that’s very plausible. I’ll try bisecting.
> > 
> > 
> > And for some unfathomable reason the Submariner pods now deploy correctly again...
> 
> Yeah, it's a strange one. I'm trying to bisect as well. So far I'm pretty sure I can consistently reproduce it when I base this commit on the latest from master, and that the same Operator code fully passes for me when it's based on [3b46ebd](https://github.com/submariner-io/submariner/commit/3b46ebdd5ad9cf5339bd5fc011dfbf1b7e3be9aa) or [b0fccd7](https://github.com/submariner-io/submariner/commit/b0fccd7472bad4e6e0e244723543696dccaf2042).

@skitt I'm pretty confident I've bisected it to be between 5306204183e7106c6117da53c6587536817a1ade (working) and 4b7def5dadd2d50e30ea0253a55f2b4718a731ba (failing). I've verified both of those "bookends" of the biset twice and seen the same results. It is possible that 4b7def5dadd2d50e30ea0253a55f2b4718a731ba is failing for a different reason however (and so that end of the search could be wrong) - it's not the same error we reported above on latest:

> F0918 21:48:18.559787       1 main.go:99] Error running route controller: createIPTableChains returned error. Unable to create SUBMARINER-POSTROUTING chain in iptables: running [/usr/sbin/iptables -t nat -S]: exit status 125: chroot: canno \_t change root directory to '/host': No such file or directory

I've also just completed a run with 28f1b1d95bc1cd0651cbf01259f1ecd7705f6ed0 that passed.

mangelajo:
Still not fully tested because I'm having issues with git.apache.org (dep from operator sdk..)

but consider this for the scripts directory:

```
$ cat scripts/gen-operator 
#!/bin/sh

mkdir -p /root/go/bin
export PATH=$PATH:/root/go/bin
export GOFLAGS=

kind delete cluster || true

RELEASE_VERSION=v0.10.0
curl -LO https://github.com/operator-framework/operator-sdk/releases/download/${RELEASE_VERSION}/operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu

chmod +x operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu 
mv operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu /root/go/bin/operator-sdk

cd operators/go
./gen_subm_operator.sh
```


That way we can just generate with
``` make gen-operator```

File: scripts/kind-e2e/lib_operator_deploy_subm.sh [hunk 23]
```diff
@@ -0,0 +1,368 @@
+set -ex
+
+#operator_helm=true
+operator_go=true
```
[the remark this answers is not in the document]
dfarrell07: Done, thanks

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
+set -ex
+
+#operator_helm=true
```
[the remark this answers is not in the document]
dfarrell07: Done, thanks

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+
+GOPATH=$HOME/go
+subm_op_dir=$GOPATH/src/github.com/submariner-operator/submariner-operator
+subm_op_scr_dir=../operators/go/submariner-operator
+mkdir -p $subm_op_dir
+
+cp -a $subm_op_scr_dir/. $subm_op_dir/
+
+# TODO Add tests that verify default ns works
+#subm_ns=default
```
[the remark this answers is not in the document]
dfarrell07: I think this is good to leave as an example of exactly what the TODO is saying we should test, but as you wish I've removed it.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+  # Accept cluster context as param
+  context=$1
+  kubectl config use-context $context
+
+  pushd $subm_op_dir
+
+  clusters_crd_file=deploy/crds/submariner_clusters_crd.yaml
+
+  # TODO: Can/should we create this with Op-SDK?
+cat <<EOF > $clusters_crd_file
```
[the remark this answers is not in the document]
dfarrell07: Agree to disagree on these types of comments I guess :)

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+  # Create clusters CRD
+  # NB: This must be done before submariner-engine pod is deployed
+  if ! kubectl get crds | grep clusters.submariner.io; then
+    kubectl create -f $clusters_crd_file
+  fi
+
+  popd
+}
+
+function create_subm_endpoints_crd() {
```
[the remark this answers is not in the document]
dfarrell07: As I've said a number of times, there are definitely nifty bash things we can do to make things more DRY. Those things do add a type of complexity however, and I'm not sure it's something we should work on until we're sure the base operator is stable.

I've documented this aspect of the "make more DRY" feedback on the Jira I raised earlier for similar.

https://jira.coreos.com/browse/HYCLD-251

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+  # Create endpoints CRD
+  # NB: This must be done before submariner-engine pod is deployed
+  if ! kubectl get crds | grep endpoints.submariner.io; then
+    kubectl create -f $endpoints_crd_file
+  fi
+
+  popd
+}
+
+function create_routeagents_crd() {
```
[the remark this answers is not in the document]
dfarrell07: Same reply - "yes DRY is good, maybe not on the critical path currently". Also doc'd this on the relevant Jira: https://jira.coreos.com/browse/HYCLD-251

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+
+  pushd $subm_op_dir
+
+  # If SubM namespace doesn't exist (ignore SubM Broker ns), create it
+  if ! kubectl get ns | grep -v $subm_broker_ns | grep $subm_ns; then
+    # TODO: Make this dynamically use any $subm_ns
+    kubectl create -f deploy/namespace.yaml
+  fi
+
+  if ! kubectl get crds | grep submariners.submariner.io; then
```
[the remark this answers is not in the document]
dfarrell07: Yes, but this would be an example of things actually becoming more complex from creating "master" DRY functions - can't be much more clear/easy to understand than this simple one-liner on a file created by the SDK.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+    kubectl create -f deploy/crds/submariner_v1alpha1_submariner_crd.yaml
+  fi
+
+  # Create SubM Operator service account if it doesn't exist
+  if ! kubectl get sa --namespace=$subm_ns | grep submariner-operator; then
+    kubectl create --namespace=$subm_ns -f deploy/service_account.yaml
```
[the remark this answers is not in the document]
dfarrell07: The Operator SDK

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+
+  # Create SubM Operator deployment if it doesn't exist
+  if ! kubectl get deployments --namespace=$subm_ns | grep submariner-operator; then
+    kubectl create --namespace=$subm_ns -f deploy/operator.yaml
+  fi
+
+  popd
+
+  # Wait for SubM Operator pod to be ready
+  kubectl get pods --namespace=$subm_ns
```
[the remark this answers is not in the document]
dfarrell07: Good catch, doesn't make sense now, just an artifact of development. Removed in latest.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+  kubectl wait --for=condition=Ready pods -l name=submariner-operator --timeout=120s --namespace=$subm_ns
+  kubectl get pods --namespace=$subm_ns
```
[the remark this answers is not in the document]
dfarrell07: Good catch, doesn't make sense now, just an artifact of development. Removed in latest.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+}
+
+function collect_subm_vars() {
```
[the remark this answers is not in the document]
dfarrell07: Okay, I changed it to `create_subm_vars` in the latest.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+    # TODO: Make this dynamically use any $subm_ns
+    kubectl create -f deploy/namespace.yaml
+  fi
+
+  if ! kubectl get crds | grep submariners.submariner.io; then
+    kubectl create -f deploy/crds/submariner_v1alpha1_submariner_crd.yaml
+  fi
+
+  # Create SubM Operator service account if it doesn't exist
+  if ! kubectl get sa --namespace=$subm_ns | grep submariner-operator; then
```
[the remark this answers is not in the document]
dfarrell07: Okay, done in the latest version.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+    kubectl create --namespace=$subm_ns -f deploy/service_account.yaml
+  fi
+
+  # Create SubM Operator role if it doesn't exist
+  if ! kubectl get roles --namespace=$subm_ns | grep submariner-operator; then
```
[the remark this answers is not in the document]
dfarrell07: Okay, done in the latest version.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+    kubectl create --namespace=$subm_ns -f deploy/role.yaml
+  fi
+
+  # Create SubM Operator role binding if it doesn't exist
+  if ! kubectl get rolebindings --namespace=$subm_ns | grep submariner-operator; then
```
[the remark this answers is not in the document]
dfarrell07: Okay, done in the latest version.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+    kubectl create --namespace=$subm_ns -f deploy/role_binding.yaml
+  fi
+
+  # Create SubM Operator deployment if it doesn't exist
+  if ! kubectl get deployments --namespace=$subm_ns | grep submariner-operator; then
```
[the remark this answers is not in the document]
dfarrell07: Okay, done in the latest version.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,368 @@
...
+
+function deploy_subm_operator() {
+  # Accept cluster context to deploy SubM Operator into as param
+  context=$1
+  kubectl config use-context $context
+
+  pushd $subm_op_dir
+
+  # If SubM namespace doesn't exist (ignore SubM Broker ns), create it
+  if ! kubectl get ns | grep -v $subm_broker_ns | grep $subm_ns; then
```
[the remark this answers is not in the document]
dfarrell07: Okay, done in the latest version.

scripts/kind-e2e/e2e.sh
```diff
@@ -324,14 +332,94 @@ setup_custom_cni
...
+collect_subm_vars cluster1
+verify_subm_broker_secrets cluster1
+
+if [[ $5 = operator ]]; then
+    operator=true
+
+    for i in 2 3; do
+      # Create CRDs required as prerequisite submariner-engine
+      # TODO: Eventually OLM should handle this
+      create_subm_endpoints_crd cluster$i
```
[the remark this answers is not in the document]
dfarrell07: Okay, done in the latest version.

scripts/kind-e2e/e2e.sh
dfarrell07: This is fixed in the latest version, thanks.

scripts/kind-e2e/e2e.sh
```diff
@@ -114,6 +118,9 @@ function setup_broker() {
     SUBMARINER_BROKER_URL=$(kubectl --context=cluster1 -n default get endpoints kubernetes -o jsonpath="{.subsets[0].addresses[0].ip}:{.subsets[0].ports[?(@.name=='https')].port}")
     SUBMARINER_BROKER_CA=$(kubectl --context=cluster1 -n ${SUBMARINER_BROKER_NS} get secrets -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='${SUBMARINER_BROKER_NS}-client')].data['ca\.crt']}")
     SUBMARINER_BROKER_TOKEN=$(kubectl --context=cluster1 -n ${SUBMARINER_BROKER_NS} get secrets -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='${SUBMARINER_BROKER_NS}-client')].data.token}"|base64 --decode)
+
+    # Verify SUBMARINER_BROKER_TOKEN is non-null/expected size
+    echo $SUBMARINER_BROKER_TOKEN | wc -c | grep 964
```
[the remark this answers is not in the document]
dfarrell07: Yeah, this was just for debugging a strange property of this value, I removed it in the latest version.

scripts/kind-e2e/e2e.sh
```diff
@@ -3,6 +3,10 @@ set -em
 
 source $(git rev-parse --show-toplevel)/scripts/lib/debug_functions
 
+# Import functions for deploying/testing with Operator
```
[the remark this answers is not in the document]
dfarrell07: Good idea, done in the latest version.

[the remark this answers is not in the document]
dfarrell07: Actually, these functions can't be imported in just the operator case. I run the verification (not deploy) functions in the current Helm-based deploy as well, as they are designed to assert the two deployments are mostly equal. I also run the broker verify function before either the Helm or Operator deploy blocks, so it needs to already be imported before that conditional.

dfarrell07:
@skitt I haven't found a commit more recent than the one I'm currently basing the PR on (5306204183e7106c6117da53c6587536817a1ade) that totally passses. I've seen everything pass with this base three times. I see the Read-only file system failure on submariner-pod on commits from the HEAD of master back to e40d95cd3f591ed0434b857415b3869523c7a4f3 (0919214bf5595bd735583cdd60234c9613663e35 and 9aad0381c7eaeded5b510500f085d424f40ae5c1 for example). There are 11 commits between that working 5306204183e7106c6117da53c6587536817a1ade and failing e40d95cd3f591ed0434b857415b3869523c7a4f3, all part of a larger change by @sridhargaddam. I see different results testing them, but I imagine that's because I'm consuming part of a larger change. On 79a29bfbc5bd214f9c660e98dfa6cd6b7062784a (the commit right after the known-working base now) I see everything deploy but the connection tests fail. On 4b7def5dadd2d50e30ea0253a55f2b4718a731ba and abe5b0123ac1c8c73edea4cb5e0d35867ee58141 (2 more of the 11 between the known working/current-failure commits) I see `cannot change root directory to '/host': No such file or directory` on submariner-routeagent-pod.

Does it consistently pass for you with this (5306204183e7106c6117da53c6587536817a1ade) as the base? Have you seen this same operator code pass on any more recent commit as the base?

sridhargaddam:
Yes @dfarrell07 , the VxLAN support patch is a big change and it depends on couple of patches in submariner-charts (as mentioned in the commit message 79a29bfbc)
The error that you pointed out "cannot change root directory to '/host': No such file or directory", will be seen if the route-agent-ds.yaml file is not updated.
Basically, you will need this fix https://github.com/submariner-io/submariner-charts/pull/3/files

Can you please check if all the three dependent patches mentioned in the commit message (79a29bfb) are in your build.

operators/go/routeagent_controller.go
```diff
@@ -0,0 +1,165 @@
...
+			Containers: []corev1.Container{
+				{
+					Name: "submariner-routeagent",
+					// TODO: Use var here
+					Image: "submariner-route-agent:local",
+					// FIXME: Should be entrypoint script, find/use correct file for routeagent
+					Command:         []string{"submariner-route-agent.sh"},
+					SecurityContext: &security_context_all_cap_allow_escal,
+					Env: []corev1.EnvVar{
+						{Name: "SUBMARINER_NAMESPACE", Value: cr.Spec.SubmarinerNamespace},
```
sridhargaddam: You will also need 
SUBMARINER_CLUSTERCIDR,
SUBMARINER_SERVICECIDR

operators/go/routeagent_controller.go
```diff
@@ -0,0 +1,165 @@
...
+						{Name: "SUBMARINER_CLUSTERID", Value: cr.Spec.SubmarinerClusterid},
+						{Name: "SUBMARINER_DEBUG", Value: cr.Spec.SubmarinerDebug},
+					},
+				},
+			},
+			// TODO: Use SA submariner-routeagent or submariner?
+			ServiceAccountName: "submariner-operator",
+			HostNetwork:        true,
+		},
```
sridhargaddam: this fix as well.
https://github.com/submariner-io/submariner-charts/pull/3/files

operators/go/routeagent_controller.go
sridhargaddam: Please take a look at the latest file for new additions - https://github.com/submariner-io/submariner-charts/blob/master/submariner/templates/route-agent-ds.yaml

operators/go/routeagent_controller.go
sridhargaddam: and this one - https://github.com/submariner-io/submariner-charts/pull/4/files

operators/go/submariner_controller.go
skitt: SAs are supposed to be created in CSVs, not in operators, AFAICT. That does mean it can be somewhat harder to deal with namespaces, but I don’t think that’s a problem here.

scripts/kind-e2e/e2e.sh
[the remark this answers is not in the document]
skitt: I agree about the `context` variable. As far as the context goes, I’d rather have it re-specified on every invocation (as in the `e2e.sh` script) — otherwise it makes it more complicated to debug clusters: the scripts rely on the current `kubectl` state, which can be changed externally.

skitt:
I get the impression the operator currently doesn’t handle running in a namespace other than the target namespace (`submariner`), does it? It would seem that this is a requirement for catalog-deployable operators. What currently happens is that the operator is deployed along with its service account, in a namespace managed by the catalog (`marketplace`, but it should be `operators`), and the service account and role bindings get created alongside it. Then when it tries to deploy pods in the target namespace, it fails...

skitt:
Why doesn’t the operator include all the CRDs used in the end-to-end tests? Are they supposed to be deployed in some other way?

operators/go/routeagent_controller.go
dfarrell07: Okay, great, thanks for the tip. I raised this Jira to track: https://jira.coreos.com/browse/HYCLD-259

dfarrell07:
> I get the impression the operator currently doesn’t handle running in a namespace other than the target namespace (`submariner`), does it? It would seem that this is a requirement for catalog-deployable operators. What currently happens is that the operator is deployed along with its service account, in a namespace managed by the catalog (`marketplace`, but it should be `operators`), and the service account and role bindings get created alongside it. Then when it tries to deploy pods in the target namespace, it fails...

@skitt - I made some tweaks and I'm now able to deploy/pass all tests in any namespace (configured by the `$subm_op` var in `lib_operator_deploy_subm.sh`). Details on the Jira I raised to track this: https://jira.coreos.com/browse/HYCLD-260

```
$ kubectl get pods --all-namespaces=true
NAMESPACE     NAME                                             READY     STATUS    RESTARTS   AGE
operators     submariner-operator-7bfbd96b5-9nmjt              1/1       Running   0          19m
operators     submariner-pod                                   1/1       Running   0          19m
operators     submariner-routeagent-pod                        1/1       Running   0          19m
```

dfarrell07:
> Why doesn’t the operator include all the CRDs used in the end-to-end tests? Are they supposed to be deployed in some other way?

@skitt - Sorry, I might not understand what you mean - what CRDs are we talking about?

With latest on PR/151, I see these CRDs.

```
$ kubectl get crds
NAME                        CREATED AT
clusters.submariner.io      2019-09-20T19:06:29Z
endpoints.submariner.io     2019-09-20T19:06:25Z
routeagents.submariner.io   2019-09-20T19:06:31Z
submariners.submariner.io   2019-09-20T19:06:34Z
```

With the latest from master (0919214bf5595bd735583cdd60234c9613663e35), I see these CRDs:

```
$ kubectl config use-context cluster2
Switched to context "cluster2".
$ kubectl get crds
NAME                      CREATED AT
clusters.submariner.io    2019-09-20T19:44:24Z
endpoints.submariner.io   2019-09-20T19:44:25Z
$ kubectl config use-context cluster1
Switched to context "cluster1".
$ kubectl get crds
NAME                      CREATED AT
clusters.submariner.io    2019-09-20T19:44:22Z
endpoints.submariner.io   2019-09-20T19:44:22Z
```

operators/go/routeagent_controller.go
dfarrell07: These are incredibly helpful @sridhargaddam, thank you! I'm hopeful this is the root explanation for the failures on recent versions of submariner/master the Operator is encountering. I raised this Jira to tack: https://jira.coreos.com/browse/HYCLD-261

dfarrell07: Actually, with fresh eyes, these are likely not the root problem of the read-only filesystem errors we're seeing with the Operator and recent versions of SubM, since these are changes to the Routeagent pods, not Engine pods where we see the error. :(

skitt:
> > Why doesn’t the operator include all the CRDs used in the end-to-end tests? Are they supposed to be deployed in some other way?
> 
> @skitt - Sorry, I might not understand what you mean - what CRDs are we talking about?

Sorry I wasn’t very clear @dfarrell07! When I run the operator generator, I get two CRDs in the `deploy` directory, RouteAgent and Submariner; but the e2e deployment scripts handle Cluster and Endpoint too. If the operator needs those, then we need to declare them in the operator CSV and make them available when we build the operator package. There’s nothing difficult, I was only wondering if there’s was a reason the generator doesn’t produce them.

operators/go/README.md
```diff
@@ -0,0 +1,42 @@
+## Submariner Operator
+
+Experiential Submariner Operator.
```
mangelajo: Experimental? :)

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,192 @@
...
+  fi
+
+  # NB: There must be a running K8s cluster pointed at by the exported KUBECONFIG
+  # for operator-sdk to work (although this dependency doesn't make sense)
+  kind create cluster || true
+  export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
+  kubectl config use-context kubernetes-admin@kind
+}
+
+function initilize_subm_operator() {
```
tpantelis: nit: _initilize_ (sic) -> _initialize_

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,326 @@
...
+  if ! kubectl get crds | grep submariners.submariner.io; then
+    kubectl create -f deploy/crds/submariner_v1alpha1_submariner_crd.yaml
+  fi
+
+  # Create SubM Operator service account if it doesn't exist
+  if ! kubectl get sa --namespace=$subm_ns submariner-operator; then
+    kubectl create --namespace=$subm_ns -f deploy/service_account.yaml
+  fi
+
+  # Create SubM Operator role if it doesn't exist
```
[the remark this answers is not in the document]
tpantelis: I think these comments are a useful synopsis for someone who's just scanning the code sections so they don't have to analyze every line to understand its purpose. Also helps folks who may not be as savvy with scripting or kubectl commands.

mkolesnik:
FYI I'm working on getting e2e to run

operators/go/README.md
dfarrell07: Done in the latest, thanks :)

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,326 @@
...
+  # Make sure go/bin is in path
+  command -v dep
+fi
+
+GOPATH=$HOME/go
+subm_op_dir=$GOPATH/src/github.com/submariner-operator/submariner-operator
+subm_op_scr_dir=../operators/go/submariner-operator
+mkdir -p $subm_op_dir
+
+cp -a $subm_op_scr_dir/. $subm_op_dir/
```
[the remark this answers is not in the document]
dfarrell07: Done, thanks

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,326 @@
...
+  # Install dep
+  curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+
+  # Make sure go/bin is in path
+  command -v dep
+fi
+
+GOPATH=$HOME/go
+subm_op_dir=$GOPATH/src/github.com/submariner-operator/submariner-operator
+subm_op_scr_dir=../operators/go/submariner-operator
```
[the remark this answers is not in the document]
dfarrell07: Done, thanks

operators/go/gen_subm_operator.sh
dfarrell07: Done, thanks :)

operators/go/routeagent_controller.go
dfarrell07: I added these new RA CIDR vars to the Operator's deployment here: https://github.com/dfarrell07/submariner/pull/13

This allows PR/151 to be rebased on master and the Operator's Routeagent pod deployments to remain successful.

Thanks again! :pray:

operators/go/routeagent_controller.go
dfarrell07: I added the /host mount change to the Operator's logic here: https://github.com/dfarrell07/submariner/pull/12

It was tracked by this Jira:
https://jira.coreos.com/browse/HYCLD-265

Thanks!

dfarrell07:
@mangelajo I rebased this on master as we discussed. I verified (a number of times) that we consistently see only the new expected-failing unit tests I mentioned.

mangelajo:
Thanks Daniel, I also checked, and it makes sense, the "src-ip"
preservation fails because
the route-agent is not deployed everywhere. Working on that now.

Helm uses Daemonset for that, so I'll do the same :-)
---
irc: ajo / mangelajo
Miguel Angel Ajo Pelayo
+34 636 52 25 69
skype: ajoajoajo

operators/go/submariner_controller.go
dfarrell07: https://github.com/dfarrell07/submariner/pull/22

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,192 @@
...
+}
+EOF
+  popd
+}
+
+function add_subm_engine_to_operator() {
+  pushd $op_dir
+  api_version=submariner.io/v1alpha1
+  kind=Submariner
+  operator-sdk add api --api-version=$api_version --kind=$kind
```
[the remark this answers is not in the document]
dfarrell07: I don't see a `--skip-generation` flag.

```
$ operator-sdk add api --skip-generation --api-version=0.0.1 --kind=Submariner
Error: unknown flag: --skip-generation
```

```
$ operator-sdk version                                                                                                                                                                                                                     
operator-sdk version: v0.10.0, commit: ff80b17737a6a0aade663e4827e8af3ab5a21170
$ operator-sdk add api -h
operator-sdk add api --kind=<kind> --api-version=<group/version> creates the
api definition for a new custom resource under pkg/apis. This command must be
run from the project root directory. If the api already exists at
pkg/apis/<group>/<version> then the command will not overwrite and return an
error.

This command runs Kubernetes deepcopy and OpenAPI V3 generators on tagged
types in all paths under pkg/apis. Go code is generated under
pkg/apis/<group>/<version>/zz_generated.{deepcopy,openapi}.go. CRD's are
generated, or updated if they exist for a particular group + version + kind,
under deploy/crds/<group>_<version>_<kind>_crd.yaml; OpenAPI V3 validation YAML
is generated as a 'validation' object.

Example:
	$ operator-sdk add api --api-version=app.example.com/v1alpha1 --kind=AppService
	$ tree pkg/apis
	pkg/apis/
	├── addtoscheme_app_appservice.go
	├── apis.go
	└── app
		└── v1alpha1
			├── doc.go
			├── register.go
			├── appservice_types.go
			├── zz_generated.deepcopy.go
			├── zz_generated.openapi.go
	$ tree deploy/crds
	├── deploy/crds/app_v1alpha1_appservice_cr.yaml
	├── deploy/crds/app_v1alpha1_appservice_crd.yaml

Usage:
  operator-sdk add api [flags]

Flags:
      --api-version string   Kubernetes APIVersion that has a format of $GROUP_NAME/$VERSION (e.g app.example.com/v1alpha1)
  -h, --help                 help for api
      --kind string          Kubernetes resource Kind name. (e.g AppService)

Global Flags:
      --verbose   Enable verbose logging
```

operators/go/gen_subm_operator.sh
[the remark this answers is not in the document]
dfarrell07: This is done in the latest version.

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,192 @@
...
+openapi_checks_enabled=false
+push_image=false
+op_dir=$GOPATH/src/github.com/submariner-operator/submariner-operator
+op_gen_dir=$GOPATH/src/github.com/submariner-io/submariner/operators/go
+op_out_dir=$GOPATH/src/github.com/submariner-io/submariner/operators/go/submariner-operator
+
+function setup_prereqs(){
+  if ! command -v dep; then
+    # Install dep
+    curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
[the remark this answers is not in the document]
dfarrell07: This is gone in the latest.

scripts/kind-e2e/e2e.sh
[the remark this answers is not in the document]
dfarrell07: This is all sorted on latest.

dfarrell07:
CI using the Operator is timing out while still correctly-running:

> The job exceeded the maximum time limit for jobs, and has been terminated.

https://travis-ci.com/submariner-io/submariner/builds/130283937

operators/go/gen_subm_operator.sh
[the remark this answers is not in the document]
dfarrell07: https://github.com/dfarrell07/submariner/pull/33

operators/go/build_subm_operator.sh
```diff
@@ -0,0 +1,20 @@
...
+
+operator-sdk build quay.io/submariner/submariner-operator:$version --verbose
+
+if [[ $push_image = true ]]; then
+  docker push quay.io/submariner/submariner-operator:$version
+else
+  echo "Skipping pushing SubM Operator image to Quay"
+fi
+
+
```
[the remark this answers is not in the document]
dfarrell07: https://github.com/dfarrell07/submariner/pull/33

operators/go/gen_subm_operator.sh
```diff
@@ -0,0 +1,158 @@
...
+set -ex
+
+# Work around https://github.com/operator-framework/operator-sdk/issues/1675
+GOROOT="$(go env GOROOT)"
+export GOROOT
+export GO111MODULE=on
+GOPATH=$HOME/go
+
+version=0.0.1
+push_image=false
```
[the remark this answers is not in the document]
dfarrell07: https://github.com/dfarrell07/submariner/pull/33

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,279 @@
+# This should only be sourced
+if [ "${0##*/}" = "lib_operator_deploy_subm.sh" ]; then
+    echo "Don't run me, source me" >&2
+    exit 1
+fi
+
+openapi_checks_enabled=false
+subm_op_src_dir=$(realpath ../operators/go/submariner-operator)
```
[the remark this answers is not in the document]
dfarrell07: https://github.com/dfarrell07/submariner/pull/33

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -0,0 +1,527 @@
+# This should only be sourced
+if [ "${0##*/}" = "lib_operator_verify_subm.sh" ]; then
```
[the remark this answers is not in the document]
dfarrell07: https://github.com/dfarrell07/submariner/pull/33

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,279 @@
+# This should only be sourced
+if [ "${0##*/}" = "lib_operator_deploy_subm.sh" ]; then
```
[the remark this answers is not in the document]
dfarrell07: https://github.com/dfarrell07/submariner/pull/33

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -0,0 +1,527 @@
...
+    exit 1
+fi
+
+function verify_subm_gateway_label() {
+  kubectl get node $context-worker -o jsonpath='{.metadata.labels}' | grep submariner.io/gateway:true
+}
+
+function verify_subm_operator() {
+  # Verify SubM namespace (ignore SubM Broker ns)
+  kubectl get ns | grep -v $subm_broker_ns | grep $subm_ns
```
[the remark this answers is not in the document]
dfarrell07: Done everywhere I could make it work in https://github.com/dfarrell07/submariner/pull/33

scripts/e2e
```diff
@@ -5,6 +5,11 @@ source $(dirname $0)/lib/debug_functions
 
 cd $(dirname $0)
 
+if [[ $5 = operator ]]; then
```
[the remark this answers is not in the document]
dfarrell07: https://github.com/dfarrell07/submariner/pull/33

scripts/kind-e2e/lib_operator_verify_subm.sh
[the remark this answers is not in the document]
skitt: This isn’t a bashism, it’s POSIX shell, supported by `dash` etc.

dfarrell07:
> CI using the Operator is timing out while still correctly-running:
> 
> > The job exceeded the maximum time limit for jobs, and has been terminated.
> 
> https://travis-ci.com/submariner-io/submariner/builds/130283937

In the latest CI run the unit tests at the end were running/passing when the job timed out. That likely means things would pass, since all the Operator verifications have already passed.

https://travis-ci.com/submariner-io/submariner/builds/130431372

(have to scrollllll)

operators/go/gen_subm_operator.sh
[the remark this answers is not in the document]
dfarrell07: This is done, thanks @mangelajo!

dfarrell07:
> > CI using the Operator is timing out while still correctly-running:
> > > The job exceeded the maximum time limit for jobs, and has been terminated.
> > 
> > 
> > https://travis-ci.com/submariner-io/submariner/builds/130283937
> 
> In the latest CI run the unit tests at the end were running/passing when the job timed out. That likely means things would pass, since all the Operator verifications have already passed.
> 
> https://travis-ci.com/submariner-io/submariner/builds/130431372
> 
> (have to scrollllll)

It looks like the maximum Travis job time for public repos is 50m and that it can't be changed.

https://docs.travis-ci.com/user/customizing-the-build#build-timeouts

**CI bat signal to signal @dimaunx**

scripts/kind-e2e/lib_operator_deploy_subm.sh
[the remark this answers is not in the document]
dfarrell07: Done

operators/go/submariner_controller.go
dfarrell07: This is long since done

operators/go/submariner_controller.go
dfarrell07: This is long-ago done/removed

scripts/kind-e2e/e2e.sh
dfarrell07: Long-ago done

operators/go/routeagent_controller.go
dfarrell07: Long-ago done

scripts/kind-e2e/lib_operator_deploy_subm.sh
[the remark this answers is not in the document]
dfarrell07: Thanks @tpantelis

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -0,0 +1,326 @@
...
+    kubectl create --namespace=$subm_ns -f deploy/role.yaml
+  fi
+
+  # Create SubM Operator role binding if it doesn't exist
+  if ! kubectl get rolebindings --namespace=$subm_ns submariner-operator; then
+    kubectl create --namespace=$subm_ns -f deploy/role_binding.yaml
+  fi
+
+  # Create SubM Operator deployment if it doesn't exist
+  if ! kubectl get deployments --namespace=$subm_ns submariner-operator; then
```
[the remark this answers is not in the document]
dfarrell07: Long-ago done

operators/go/submariner_controller.go
dfarrell07: My understanding is that these SA changes are no longer needed/desired. We no loner have as distinct Routeagent logic for example, and @skitt gave feedback elsewhere that SAs should be created by OLM logic I think, not Operators themselves. Closing for now, we can re-open the Jira if desired.

operators/go/gen_subm_operator.sh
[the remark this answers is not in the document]
dfarrell07: Marking as done

operators/go/submariner_controller.go
dfarrell07: Okay, it seems this is a non-feature, closing.

operators/go/submariner_controller.go
dfarrell07: I'm just going to stick with the scaffold, thanks.

operators/go/gen_subm_operator.sh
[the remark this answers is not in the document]
mangelajo: The plan is to remove the codegen logic and keep the code.

PR #153, opened 2019-09-17
Switch to Fedora for the Dapper image

skitt:
This gives us more up-to-date software and allows us to use packaged
versions of Go and kubectl.


Oats87:
LGTM as long as our build still works

PR #154, opened 2019-09-18
Fix network unreachable issue on AWS with multiple AZs

sridhargaddam:
When deploying a K8s/OpenShift Cluster on AWS, multiple Availability Zones (AZs) are
used for maximized service availability. Each AZ uses a CIDR (f.e., 10.165.32.0/20)
thats pulled out of the VPC Global CIDR (f.e., 10.165.0.0/16).

Currently in Submariner, we are deriving the VxLAN VTEP IP address from the host-ip as
shown below.

For example: Say, the host ipaddress is "192.168.1.100/16", we were prepending 240 to
the host-ip address and configuring the vxlan vtepIP as "240.168.1.100/16"

This works fine when all the nodes are on the same Subnet CIDR, but will have issues
when programming inter-cluster routing rules on the host via the vx-submariner interface.

Error message:
route.go:719] Error adding route {Ifindex: 58 Dst: 10.252.0.0/14 Src: <nil>
 Gw: 240.165.41.54 Flags: [] Table: 0}: network is unreachable

This patch modifies the implementation to use vtepIP Mask of 8 (i.e., 240.168.1.100/8)
so that when adding inter-cluster routing rules, the nextHop (i.e., remoteVTEPIP) is on
the same network and routing rules can be programmed successfully.

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -348,7 +348,7 @@ func (r *Controller) createVxLANInterface(ifaceType int, gatewayNodeIP net.IP) e
 			name:     VxLANIface,
 			vxlanId:  100,
 			group:    gatewayNodeIP,
-			srcAddr:  vtepIP,
+			srcAddr:  nil,
 			vtepPort: VxLANPort,
 			mtu:      1450,
 		}
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -78,7 +78,7 @@ const (
 	// So, the approach we are taking is to derive the VxLAN ip from the hostIPAddress
 	// as shown below.
 	// For example: Say, the host ipaddress is "192.168.1.100/16", we prepend 240 to the
-	// host-ip address, derive the vxlan vtepIP (i.e., 240.168.1.100/16) and configure it
+	// host-ip address, derive the vxlan vtepIP (i.e., 240.168.1.100/8) and configure it
 	// on the VxLAN interface.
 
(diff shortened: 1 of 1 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -348,7 +348,7 @@ func (r *Controller) createVxLANInterface(ifaceType int, gatewayNodeIP net.IP) e
 			name:     VxLANIface,
 			vxlanId:  100,
 			group:    gatewayNodeIP,
-			srcAddr:  vtepIP,
+			srcAddr:  nil,
```
mangelajo: ahh, good one, to bad I missed this on the initial review.

Thanks for fixing it.

pkg/routeagent/controllers/route/route.go
```diff
@@ -359,7 +359,7 @@ func (r *Controller) createVxLANInterface(ifaceType int, gatewayNodeIP net.IP) e
 		}
 	}
 
-	err = r.vxlanDevice.configureIPAddress(vtepIP, vtepMask.Mask)
+	err = r.vxlanDevice.configureIPAddress(vtepIP, net.CIDRMask(8, 32))
```
mangelajo: Technically we could even go down to /4 if we wanted.

mangelajo:
This looks good to me, if anybody can review I'll merge.

pkg/routeagent/controllers/route/route.go
sridhargaddam: True. If required, we can do that.

dimaunx:
The solution was tested and working as expected.

PR #157, opened 2019-09-20
Allow IPSec on non-standard ports via environment variables

mangelajo:
Some corporate networks block outgoing IPSEC traffic ports,
in such cases the option to specify no standard IPSEC ports
becomes necessary.

This patch allows control over the IPSEC UDP ports. All clusters
must use the same ports. In the future we should implement
the ability to declare which ports a cluster uses, and announce
those through the exchanged endpoint information.

The environment variables are:

  - CE_IPSEC_IKEPORT   defaults to 500
  - CE_IPSEC_NATTPORT  defaults to 4500

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,6 @@ require (
 	sigs.k8s.io/yaml v0.0.0-20181102190223-fd68e9863619 // indirect
 )
 
-replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190701121157-9a5ae4453bda
+replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190701121157-73b0716138ff
 
 replace github.com/onsi/ginkgo => github.com/onsi/ginkgo v0.0.0-20190716150225-054541502288
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/submariner-io/submariner
 go 1.12
 
 require (
-	github.com/bronze1man/goStrongswanVici v0.0.0-20181105005556-92d3927c899e
+	github.com/bronze1man/goStrongswanVici v0.0.0-20190921045355-4c81bd8d0bd5
 	github.com/coreos/go-iptables v0.4.0
 	github.com/evanphx/json-patch v0.0.0-20180908160633-36442dbdb585 // indirect
 	github.com/gogo/protobuf v0.0.0-20171007142547-342cbe0a0415 // indirect
@@ -41,7 +41,5 @@ require (
 	k8s.io/kube-openapi v0.0.0-20181109181836-c59034cc13d5 // indirect
 	sigs.k8s.io/yaml v0.0.0-20181102190223-fd68e9863619 // indirect
 )
-
-replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190701121157-73b0716138ff
 
 replace github.com/onsi/ginkgo => github.com/onsi/ginkgo v0.0.0-20190716150225-054541502288
--- a/pkg/cableengine/ipsec/strongswan.go
+++ b/pkg/cableengine/ipsec/strongswan.go
@@ -2,10 +2,12 @@ package ipsec
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
 	"syscall"
+	"text/template"
 	"time"
 
 	"github.com/bronze1man/goStrongswanVici"
@@ -32,21 +34,29 @@ type strongSwan struct {
 	replayWindowSize          string
 	ipSecIkeSaRekeyInterval   string
 	ipSecChildSaRekeyInterval string
+	ipSecIKEPort              string
+	ipSecNATTPort             string
 
 	debug   bool
 	logFile string
 }
 
 type specification struct {
-	PSK     string
-	Debug   bool
-	LogFile string
+	PSK      string
+	Debug    bool
+	LogFile  string
+	IKEPort  string `default:"500"`
+	NATTPort string `default:"4500"`
 }
+
+const defaultIKEPort = "500"
+const defaultNATTPort = "4500"
+const ipsecSpecEnvVarPrefix = "ce_ipsec"
 
 func NewStrongSwan(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (Driver, error) {
 	ipSecSpec := specification{}
 
-	err := envconfig.Process("ce_ipsec", &ipSecSpec)
+	err := envconfig.Process(ipsecSpecEnvVarPrefix, &ipSecSpec)
 	if err != nil {
 		return nil, fmt.Errorf("error processing environment config for ce_ipsec: %v", err)
 	}
@@ -55,6 +65,8 @@ func NewStrongSwan(localSubnets []string, localEndpoint types.SubmarinerEndpoint
 		replayWindowSize:          DefaultReplayWindowSize,
 		ipSecIkeSaRekeyInterval:   DefaultIkeSaRekeyInterval,
 		ipSecChildSaRekeyInterval: DefaultChildSaRekeyInterval,
+		ipSecIKEPort:              ipSecSpec.IKEPort,
+		ipSecNATTPort:             ipSecSpec.NATTPort,
 		localEndpoint:             localEndpoint,
 		localSubnets:              localSubnets,
 		secretKey:                 ipSecSpec.PSK,
@@ -64,7 +76,7 @@ func NewStrongSwan(localSubnets []string, localEndpoint types.SubmarinerEndpoint
 }
 
 func (i *strongSwan) Init() error {
-	if err := runCharon(i.debug, i.logFile); err != nil {
+	if err := i.runCharon(); err != nil {
 		return err
 	}
 
@@ -138,6 +150,15 @@ func (i *strongSwan) ConnectToEndpoint(endpoint types.SubmarinerEndpoint) (strin
 		Encap:       "yes",
 		Mobike:      "no",
 	}
+
+	// We point to the remote port that has proven to work by trial and error
+	// with strongswan over non-standard UDP ports
+	if i.ipSecNATTPort != defaultNATTPort {
+		ikeConf.RemotePort = i.ipSecNATTPort
+	} else if i.ipSecIKEPort != defaultIKEPort {
+		ikeConf.RemotePort = i.ipSecIKEPort
+	}
+
 	ikeConf.Children = map[string]goStrongswanVici.ChildSAConf{
 		"submariner-child-" + endpoint.Spec.CableName: childSAConf,
 	}
@@ -291,15 +312,52 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
 	return nil
 }
 
-func runCharon(debug bool, logFile string) error {
+const charonConfTemplate = `
+	charon {
+		port = {{.ipSecIKEPort}}
+		port_nat_t = {{.ipSecNATTPort}}
+		make_before_break = yes
+		ignore_acquire_ts = yes
+	}
+`
+
+func (i *strongSwan) writeCharonConfig(path string) {
+	err := os.Remove(path)
+	if err != nil {
+		klog.Infof("Error deleting %s", path)
+	}
+
+	f, err := os.Create(path)
+	if err != nil {
+		klog.Fatalf("Error creating %s: %s", path, err)
+	}
+	i.renderCharonConfigTemplate(f)
+}
+
+func (i *strongSwan) renderCharonConfigTemplate(f io.Writer) {
+	t, err := template.New("charon.conf").Parse(charonConfTemplate)
+	if err != nil {
+		klog.Fatalf("Error creating template for charon.conf")
+	}
+	if err = t.Execute(f, map[string]string{
+		"ipSecIKEPort":  i.ipSecIKEPort,
+		"ipSecNATTPort": i.ipSecNATTPort}); err != nil {
+		klog.Fatalf("Error rendering charon config file: %s", err)
+	}
+}
+
+func (i *strongSwan) runCharon() error {
+
+	i.writeCharonConfig("/etc/strongswan/strongswan.d/charon.conf")
+
 	klog.Infof("Starting Charon")
 	// Ignore error
 	os.Remove("/var/run/charon.vici")
 
 	args := []string{}
-	for _, i := range strings.Split("dmn|mgr|ike|chd|cfg|knl|net|asn|tnc|imc|imv|pts|tls|esp|lib", "|") {
-		args = append(args, "--debug-"+i)
-		if debug {
+	for _, idx := range strings.Split("dmn|mgr|ike|chd|cfg|knl|net|asn|tnc|imc|imv|pts|tls|esp|lib", "|") {
+		args = append(args, "--debug-"+idx)
+		if i.debug {
 			args = append(args, "3")
 		} else {
 			args = append(args, "1")
@@ -311,10 +369,10 @@ func runCharon(debug bool, logFile string) error {
 	cmd.Stderr = os.Stderr
 
 	var outputFile *os.File
-	if logFile != "" {
-		out, err := os.OpenFile(logFile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+	if i.logFile != "" {
+		out, err := os.OpenFile(i.logFile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
 		if err != nil {
-			return fmt.Errorf("Failed to open log file %s: %v", logFile, err)
+			return fmt.Errorf("Failed to open log file %s: %v", i.logFile, err)
 		}
 
 		cmd.Stdout = out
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -42,6 +42,6 @@ require (
 	sigs.k8s.io/yaml v0.0.0-20181102190223-fd68e9863619 // indirect
 )
 
-replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190701121157-9a5ae4453bda
+replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190701121157-73b0716138ff
```
mpeterson: (nv) for future changes it might be wise to update the version date as well

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +312,52 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+		klog.Fatalf("Error creating template for charon.conf")
+	}
+	if err = t.Execute(f, map[string]string{
+		"ipSecIKEPort":  i.ipSecIKEPort,
+		"ipSecNATTPort": i.ipSecNATTPort}); err != nil {
+		klog.Fatalf("Error rendering charon config file: %s", err)
+	}
+}
+
+func (i *strongSwan) runCharon() error {
```
mpeterson: (nit) I would have created the conversion of function to method in a different commi

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +312,52 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+}
+
+func (i *strongSwan) renderCharonConfigTemplate(f io.Writer) {
+	t, err := template.New("charon.conf").Parse(charonConfTemplate)
+	if err != nil {
+		klog.Fatalf("Error creating template for charon.conf")
+	}
+	if err = t.Execute(f, map[string]string{
+		"ipSecIKEPort":  i.ipSecIKEPort,
+		"ipSecNATTPort": i.ipSecNATTPort}); err != nil {
```
mpeterson: (nit) it's a long if. I'd handle the err in a different line

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +312,52 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
 	return nil
 }
 
-func runCharon(debug bool, logFile string) error {
+const charonConfTemplate = `
+	charon {
+		port = {{.ipSecIKEPort}}
+		port_nat_t = {{.ipSecNATTPort}}
+		make_before_break = yes
```
mpeterson: (nv) do you want to perhaps comment on why we enable these two options?

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +312,52 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+	err := os.Remove(path)
+	if err != nil {
+		klog.Infof("Error deleting %s", path)
+	}
+
+	f, err := os.Create(path)
+	if err != nil {
+		klog.Fatalf("Error creating %s: %s", path, err)
+	}
+	i.renderCharonConfigTemplate(f)
```
mpeterson: You need to close the FD after usage and check for errors on closure.

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +312,52 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+	if err = t.Execute(f, map[string]string{
+		"ipSecIKEPort":  i.ipSecIKEPort,
+		"ipSecNATTPort": i.ipSecNATTPort}); err != nil {
+		klog.Fatalf("Error rendering charon config file: %s", err)
+	}
+}
+
+func (i *strongSwan) runCharon() error {
+
+	i.writeCharonConfig("/etc/strongswan/strongswan.d/charon.conf")
```
mpeterson: make this path a constant

pkg/cableengine/ipsec/strongswan_test.go
```diff
@@ -0,0 +1,66 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Util", func() {
+	Describe("Charon port configuration", testCharonConfigPortsGen)
+})
+
+func testCharonConfigPortsGen() {
+	Context("config rendering", func() {
+		It("should render config correctly out of strongSwan parameters", func() {
```
mpeterson: add the check that the  FD is closed after

pkg/cableengine/ipsec/strongswan_test.go
```diff
@@ -0,0 +1,66 @@
...
+			ss := strongSwan{ipSecIKEPort: "500", ipSecNATTPort: "4500"}
+			buf := new(bytes.Buffer)
+
+			ss.renderCharonConfigTemplate(buf)
+
+			Expect(buf.String()).To(ContainSubstring("port = 500"))
+			Expect(buf.String()).To(ContainSubstring("port_nat_t = 4500"))
```
mpeterson: and what about the two other options?

mangelajo: I'm just testing that the dynamic parts were working correctly.

What's the value of checking that a string in code side A is equal to code side B, and then you need to maintain both?.

In the past I remember using the whole file output as a check to validate config output from code = config output from testing. And that lead to lots of unecesary testing code maintenance.

For those parts of the config file it'd be better to have some sort of functional validation IMHO, although, I haven't even checked what are they meant to be, and we have been running it without them in the right directory (dockerfile was installing it on a wrong place) :P :)

pkg/cableengine/ipsec/strongswan.go
mangelajo: What's the benefit of that? :), it's good to have a constant when you'll use that string in several places. :-1:

pkg/cableengine/ipsec/strongswan.go
mangelajo: Oh my, that's a good catch, Thanks,

pkg/cableengine/ipsec/strongswan_test.go
mangelajo: Yup, I'll check how to do that...

pkg/cableengine/ipsec/strongswan.go
mangelajo: I have absolutely no idea... may be it's worth investigating them, they were here before I touched this :), although, they were ineffective as we saw because they were installed on the wrong place.

pkg/cableengine/ipsec/strongswan.go
mangelajo: May be it will look better, I'll try, thanks.

pkg/cableengine/ipsec/strongswan.go
mangelajo: I had to work with that function and the change was simple enough... :) it's a nit, on that I agree :-) but thanks a lot for lucking at the code.

go.mod
mangelajo: I will re-point this to the upstream goStrongswanVici now again, since they merged all our patches :-)

mangelajo: Then we can point to a specific version when they tag it.

pkg/cableengine/ipsec/strongswan.go
mpeterson: yeah, I know. But now that we are "introducing" them in a new form it might be worth checking why it's there in the first place

pkg/cableengine/ipsec/strongswan.go
mpeterson: it's also good to have a constant when its value might need changing and you want to simplify that change and have it in a very visible part instead of buried down in the code. This particular path is tied to this particular package of strongswan. This might change if we change the base docker image, for example (which is what happened here btw)

pkg/cableengine/ipsec/strongswan_test.go
mpeterson: The value is to have something check that things break should that change. That is, as long as those options are actually needed.

dimaunx:
We should also create PR for helm charts to make those values configurable for helm based installations.

pkg/cableengine/ipsec/strongswan.go
mangelajo: It seems like make-before-break is related to maintaining dataplane connectivity while re-authenticating an endpoint with IKEv2.

For the other one... I dont comprehend... https://github.com/strongswan/strongswan/blob/master/conf/options/charon.opt#L152

@Oats87  ^ any idea on what ignore_acquire_ts means?

pkg/cableengine/ipsec/strongswan.go
mangelajo: I will change it although I don't agree a lot with the usefulness , selecting a decent constant name won't do any harm either :)

I normally look for things like this with grep/editor search when I need, instead of looking at specific places in code.

pkg/cableengine/ipsec/strongswan_test.go
mangelajo: I see no decent way of checking that properly, I accept ideas around how to test that on writeCharonConfig, although I believe that belongs more to funcitonal testing, that we don't have currently, and that's why I broke the render part, to be able to test it.

pkg/cableengine/ipsec/strongswan_test.go
mangelajo: Done

go.mod
mangelajo: Ok, I pointed to strongswanVici master from the u/s repo.

mangelajo:
> We should also create PR for helm charts to make those values configurable for helm based installations.

True, I will open a bug on helm charts to allow that as soon as we can.

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +315,64 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+		port_nat_t = {{.ipSecNATTPort}}
+		make_before_break = yes
+		ignore_acquire_ts = yes
+	}
+`
+
+func (i *strongSwan) writeCharonConfig(path string) {
+	err := os.Remove(path)
+	if err != nil {
+		klog.Infof("Error deleting %s", path)
```
tpantelis: Should this be Warning or Error? 

Also the `err` should be logged as well.

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +315,64 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+	}
+
+	f, err := os.Create(path)
+
+	if err != nil {
+		klog.Fatalf("Error creating %s: %s", path, err)
```
tpantelis: I think we should propagate an error for consistency. Same with the other `Fatalf` calls below.

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +315,64 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+
+	if err = f.Close(); err != nil {
+		klog.Fatalf("Unable to close %s: %s", path, err)
+	}
+}
+
+func (i *strongSwan) renderCharonConfigTemplate(f io.Writer) {
+	t, err := template.New("charon.conf").Parse(charonConfTemplate)
+	if err != nil {
+		klog.Fatalf("Error creating template for charon.conf")
```
tpantelis: We should log `err` as well.

pkg/cableengine/ipsec/strongswan_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"os"
+	"testing"
+
+	"github.com/kelseyhightower/envconfig"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Util", func() {
```
tpantelis: _"Util"_ -> _"Strongswan"_ ?

pkg/cableengine/ipsec/strongswan_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		})
+	})
+
+	Context("environment variable processing", func() {
+		It("should get the defaults from the specification definition", func() {
+			checkEnvVarParsingPorts("500", "4500")
+		})
+
+		It("should get the rigth environment variable names", func() {
```
tpantelis: _rigth_ -> _right_

pkg/cableengine/ipsec/strongswan_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+func checkEnvVarParsingPorts(ikePort string, nattPort string) {
+	ipSecSpec := specification{}
+	err := envconfig.Process(ipsecSpecEnvVarPrefix, &ipSecSpec)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(ipSecSpec.IKEPort).To(Equal(ikePort))
+	Expect(ipSecSpec.NATTPort).To(Equal(nattPort))
+}
+
+func TestUtil(t *testing.T) {
```
tpantelis: `TestUtil` -> `TestStrongswan` ?

pkg/cableengine/ipsec/strongswan_test.go
```diff
@@ -0,0 +1,68 @@
...
+		})
+	})
+}
+
+func checkEnvVarParsingPorts(ikePort string, nattPort string) {
+	ipSecSpec := specification{}
+	err := envconfig.Process(ipsecSpecEnvVarPrefix, &ipSecSpec)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(ipSecSpec.IKEPort).To(Equal(ikePort))
+	Expect(ipSecSpec.NATTPort).To(Equal(nattPort))
```
tpantelis: This is really just testing the external `envconfig.Process` lib function.  It seems it would be more useful to test `NewStrongSwan` and verify the returned `ipSecIKEPort`, `ipSecNATTPort ` et al are set properly?

mangelajo: Yeah, I understand what you mean, and I thought about it, but let me share my line of thinking:

I wasn't sure about how envconfig.Process generated the env var names, and I wanted to test End to End, and produce self-documenting test about what goes in (which env-vars) and what do you get in the end (Assuming that envconfig.Process will always work)

Does that sound ok?

pkg/cableengine/ipsec/strongswan_test.go
mangelajo: Done. Thanks

pkg/cableengine/ipsec/strongswan_test.go
mangelajo: Ooops thanks!

pkg/cableengine/ipsec/strongswan_test.go
mangelajo: oops thanks

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -138,6 +153,15 @@ func (i *strongSwan) ConnectToEndpoint(endpoint types.SubmarinerEndpoint) (strin
 		Encap:       "yes",
 		Mobike:      "no",
 	}
+
+	// We point to the remote port that has proven to work by trial and error
```
sridhargaddam: Just checking: Is this not a documented mechanism?

mangelajo: Documentation is unclear about this.. nobody tells you to point to the NATT IKE port, and if you don't do that sometimes does not work... 

Also, surprisingly, if you declare the remote port, and it's the standard port (ike or not...) dataplane will stop working, it's full of bugs.

pkg/cableengine/ipsec/strongswan_test.go
tpantelis: ok so you wrote the test originally b/c you weren't sure exactly how envconfig.Process worked - that's makes sense. But since this is mainly what `NewStrongSwan` does internally, do you think it makes sense now to test `NewStrongSwan` to get a bit more coverage and e2e (and make it more blackbox)?

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -291,15 +315,71 @@ func (i *strongSwan) loadSharedKey(endpoint types.SubmarinerEndpoint, client *go
...
+	if err != nil {
+		return fmt.Errorf("Error rendering charon config file: %s", err)
+	}
+	return nil
+}
+
+func (i *strongSwan) runCharon() error {
+
+	if err := i.writeCharonConfig(strongswanCharonConfigFilePath); err != nil {
+		klog.Fatalf("Error writing strongswan charon config: %s", err)
```
tpantelis: Can we propagate the error here as well?

pkg/cableengine/ipsec/strongswan_test.go
tpantelis: Let's merge this as is. I can push a PR later to propose the change.

pkg/cableengine/ipsec/strongswan.go
mangelajo: Done. I missed that one.

PR #158, opened 2019-09-21
Provide a small set of installation files

skitt:
This adds a script to build a small set of files (three) which can be
used to install the operator. The result matches the installation
document.


operators/go/build_publish_yaml
```diff
@@ -0,0 +1,22 @@
...
+done > publish/submariner-crds.yaml
+
+for f in submariner-operator/deploy/service_account.yaml \
+         submariner-operator/deploy/role.yaml \
+         submariner-operator/deploy/role_binding.yaml \
+         submariner-operator/deploy/operator.yaml
+do
+    echo ---
+    cat $f
+done > publish/submariner-operator.yaml
```
mangelajo: Is it possible to do it in a single yaml or are there issues with the ordering?

skitt: We need to split out the namespace creation, since the operator descriptor itself goes in the `submariner` namespace. The CRDs are installed globally, I haven’t tried installing them with a namespace specifier in combination with other files...

PR #160, opened 2019-09-22
[DNM] Operator fixes

mkolesnik:
These are fixes for PR #151 to make linter stop complaining when running "make ci".

Don't merge this patch, it should be part of #151


operators/go/gen_subm_operator.sh
```diff
@@ -13,8 +13,8 @@ add_routeagent=true
 openapi_checks_enabled=false
 push_image=false
 op_dir=$GOPATH/src/github.com/submariner-operator/submariner-operator
-op_gen_dir=$GOPATH/src/github.com/submariner-io/submariner/operators/go
-op_out_dir=$GOPATH/src/github.com/submariner-io/submariner/operators/go/submariner-operator
+op_gen_dir=$(pwd)
```
dfarrell07: Can you help me understand the motivation for this one? It definitely makes sense for script portability, but I think there are advantages to keeping everything relative to GOPATH.

Also, I worry that this would encourage folks to do development work outside of their go/src/ directory, which I thinks breaks assumptions of other tooling.

dfarrell07: Update: I went ahead and made these changes, not worth nit-picking over. :+1:

operators/go/gen_subm_operator.sh
```diff
@@ -106,7 +106,7 @@ function add_subm_engine_to_operator() {
 
   operator-sdk add controller --api-version=$api_version --kind=$kind
 
-  controller_file_src=$op_gen_dir/submariner_controller.go
+  controller_file_src=$op_gen_dir/submariner_controller.go.spec
```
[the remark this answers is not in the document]
mkolesnik: Sure, whatever suffix will work. Maybe even ".source" or some other suffix would be even clearer that this is not a part of the code but is used for the generated code.

I used .spec since that's what you use in packaging nomenclature, but it doesn't really have any significance.

mkolesnik:
Daniel incorporated fixes into his PR so closing this one

Issue #146, opened 2019-09-10
iptables is not present in GKE nodes


cyriltovena:
Hello,

Even after mounting host / to the pod, /usr/sbin/iptables is not available on GKE nodes.

```
main.go:87] Error running route controller: createIPTableChains returned error. Unable to create SUBMARINER-POSTROUTING chain in iptables: running [/usr/sbin/iptables -t nat -S]: exit status 127: chroot: failed to run command '/usr/sbin/iptables': No such file or directory
```

For now I've moved the agent Dockerfile to alpine.

Not sure if you want to address this.

Thank you

Oats87:
Are you using the published version of Submariner or are you using what is based off of master? We are in the process of making some changes in the way we handle `iptables` on the nodes

/cc @sridhargaddam

sridhargaddam:
Is iptables itself missing (i.e., not installed) on the host or its installed but is present on some other path (i.e., something other than /usr/sbin/iptables)?

cyriltovena:
No I could no find it anywhere else. I'm based off master I wanted to try the VXLAN PR. Which doesn't seems to work on GKE.

mangelajo:
Are you using the updated helm charts that mount /host to the host to access the worker node iptables binaries?

 helm repo add submariner-latest https://submariner-io.github.io/submariner-charts/chart

I guess that we need to document that ^

If after that you're still finding the issue, I guess one option would be... installing iptables on the image, and use the "host" iptables only if it exists, otherwise use the image iptables.

mangelajo:
@cyriltovena thanks for giving it a try!!! :)

mangelajo:
@cyriltovena what's your timezone? could we e-meet in some slack/IRC and try to arrange a debug session if you're able to deploy it live on GKE?

We're working to cut a release ASAP, but it'd be great if we get compatibility with GKE, please feel free to ping me on majopela@redhat.com or kubernetes/rancher slack (mangelajo)

cyriltovena:
Exactly no iptable on GKE, nodes are on a one IP subnet.

I don’t think we can sort this out quickly but yes I’ll reach out to you on slack.

cyriltovena:
@Oats87 I did update your submariner docker image to include an iptables. The result is the same you can't add a route when the destination is not part of you local network.

In GKE nodes are in a single IP space network. using their own IP as subnet mask.

sridhargaddam:
[...]

PR #161, opened 2019-09-25
Only use host's iptables when that's available

mangelajo:
This patch modifies the iptables wrapper behaviour that will only
be activated when proper host's iptables wrappers are detected.

A warning message is written on the Pod logs when the host iptables
are not available.

Related-To: https://github.com/submariner-io/submariner/issues/146


package/submariner-route-agent.sh
```diff
@@ -9,4 +9,18 @@ else
     DEBUG="-v=4"
 fi
 
-exec submariner-route-agent ${DEBUG} -alsologtostderr
\ No newline at end of file
+# if host is mounted on /host and host has it's own iptables version
+# use that one instead via the shipped iptables wrapper, we do this
+# to avoid issues with hosts
```
sridhargaddam: Instead of "to avoid issues with hosts", shall we use "to avoid configuring iptables and nftables on the host which could lead to functional failures"

package/submariner-route-agent.sh
```diff
@@ -9,4 +9,18 @@ else
...
+
+for f in iptables-save iptables; do
+	if [[ -x /host/usr/sbin/$f ]]; then
+		cp /usr/sbin/${f}.wrapper /usr/sbin/$f
+	else
+		echo "WARNING: not using iptables wrapper because /host/usr/sbin/$f was not detected."
+		echo "either the host fs isn't mounted or the host does not have iptables installed"
```
sridhargaddam: fs -> file system (for readability)

package/submariner-route-agent.sh
mangelajo: Thanks, good comment. I extended the explanation a little bit.

package/submariner-route-agent.sh
mangelajo: Done, thanks.

package/submariner-route-agent.sh
```diff
@@ -9,4 +9,20 @@ else
     DEBUG="-v=4"
 fi
 
-exec submariner-route-agent ${DEBUG} -alsologtostderr
\ No newline at end of file
+# if host is mounted on /host and host has it's own iptables version
+# use that one instead via the shipped iptables wrapper, we do this
+# to avoid configuring iptables and nftables on the host which
+# could lead to functional failures because some hosts use an iptables
+# and iptables-save which program converted nftables instead.
```
sridhargaddam: Don't have a strong opinion on this, but if you plan to update the patch, you can consider this.
"which program converted nftables instead" -> "which program nftables under the hood"

Thanks for the fix @mangelajo :+1:

mangelajo: Easy to change, done.

package/submariner-route-agent.sh
```diff
@@ -9,4 +9,20 @@ else
...
+# to avoid configuring iptables and nftables on the host which
+# could lead to functional failures because some hosts use an iptables
+# and iptables-save which program nftables under the hood. 
+
+for f in iptables-save iptables; do
+	if [[ -x /host/usr/sbin/$f ]]; then
+		cp /usr/sbin/${f}.wrapper /usr/sbin/$f
+	else
+		echo "WARNING: not using iptables wrapper because /host/usr/sbin/$f was not detected."
+		echo "either the host file system isn't mounted or the host does not have iptables"
```
tpantelis: _either_  -> _Either_

package/submariner-route-agent.sh
```diff
@@ -9,4 +9,20 @@ else
...
+		echo "installed the pod will use the image installed iptables version"
```
tpantelis: It seems there should be a separate sentence here, how about "_... installed. The pod will use the image installed iptables version instead._"?

sridhargaddam:
> I agree with Tom’s requested changes. (BTW why do we need `iptables-save`?)

Probably we may not need the iptables-save, but I'm not very sure. We can remove it and test it once to confirm.

mangelajo:
Updated with reviewer comments, please squash/merge

tpantelis:
> Updated with reviewer comments, please squash/merge

LGTM. I'll give @Oats87 a chance to review as well.

PR #163, opened 2019-09-26
[WIP] E2E flow with custom ipsec ports

dimaunx:
After #157 was merged we should start testing with non-standard ipsec ports.
This PR must be merged after submariner-io/submariner-charts#7.
The CI should pass only after the helm charts changes are merged.

Original Issue: https://github.com/submariner-io/submariner-charts/issues/6

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -128,6 +128,8 @@ function setup_cluster2_gateway() {
             --name submariner \
             --namespace submariner \
             --set ipsec.psk="${SUBMARINER_PSK}" \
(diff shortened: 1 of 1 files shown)


mangelajo:
The testing matrix will keep complicating. We will need to figure out how to handle those things.

For example, in the future we wanted to setup udp port numbers via the labels of nodes (it would be in addition to this which changes the system default). I think that will be a great opportunity to manage those labels from the E2E tests and ensure that the port changes but connectivity does not break.

mangelajo:
Ok related changes are merged, this was ok for checking on travis.

PR #164, opened 2019-09-26
Fix race condition causing NO_PROPOSAL_CHOSEN

mpeterson:
When one side of the cluster installs a cable a connection request is
automatically triggered. If the other side's `charon` has not been
configured yet with the correct IKE then it answers with a NO_PROPOSAL_CHOSEN
causing the initiator side to remove its SA and not retry again. This
leads to a situation where the cable is installed, and thus won't be
replaced/reinstalled, but without an SA to establish the tunnel. A
"stale" cable of sorts.

This problem becomes more relevant and apparent when only one side has a
public IP, as in this case the system cannot rely on the reverse SA to
establish the tunnel.

This patch handles the scenario where a cable becomes stale by removing
that cable and allowing submariner to install it again, reestablishing
a valid SA.

### DIFF
--- a/pkg/cableengine/ipsec/strongswan.go
+++ b/pkg/cableengine/ipsec/strongswan.go
@@ -14,6 +14,8 @@ import (
 	"github.com/kelseyhightower/envconfig"
 	"github.com/submariner-io/submariner/pkg/types"
 	"k8s.io/klog"
+
+	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


mpeterson:
Still have to write tests for it. The patch has been validated together with @dimaunx and it solves the issue we were experiencing.

mpeterson:
I've researched if there was an option to not delete the SA upon receiving the NO_PROPOSAL_CHOSEN event, but there is none as stated on https://wiki.strongswan.org/issues/1501#note-17

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -286,6 +296,43 @@ func (i *strongSwan) GetActiveConnections(clusterID string) ([]string, error) {
...
 }
 
+// removeStaleCable removes cables that have no active SAS and returns a
+// boolean indicating removal of the cable. The status of the error is nil
+// in case of no errors, otherwise the returns values are false and the
+// error.
+func (i *strongSwan) removeStaleCable(client *goStrongswanVici.ClientConn, cableID string) (bool, error) {
+	sas, err := client.ListSas("", "")
+	if err != nil {
+		klog.Errorf("error while retrieving active sas")
```
tpantelis: We don't need to log here are as we're propagating the error ...

pkg/cableengine/ipsec/strongswan.go
```diff
@@ -286,6 +296,43 @@ func (i *strongSwan) GetActiveConnections(clusterID string) ([]string, error) {
...
+	klog.V(4).Infof("Remove stale cable: %s", cableID)
+	// Creating this stub Endpoint is not ideal but to avoid a refactor
+	// TODO(mpeterson): refactor code to make this not a requirement
+	err = i.DisconnectFromEndpoint(types.SubmarinerEndpoint{
+		Spec: subv1.EndpointSpec{
+			CableName: cableID,
+		},
+	})
+	if err != nil {
+		klog.Errorf("error while removing stale cable")
```
tpantelis: ... same here

pkg/cableengine/ipsec/strongswan.go
mpeterson: Yup, I agree with you on this. I left it like this because even though the error is propagated, there is no actual logging of it higher in the stack (IIRC). And since I didn't want to refactor those parts now I erred on logging and propagating.

pkg/cableengine/ipsec/strongswan.go
tpantelis: Whether we log here or propagate, can you also log `cableID` with the message for additional context? - not sure if that would be useful but can't hurt.

PR #165, opened 2019-09-26
Use set -T instead of shopt -s extdebug

skitt:
This avoids issues caused by the extdebug bug in Bash < 4.4.

Closes: https://jira.coreos.com/browse/HYCLD-263


skitt:
Screenshot from 2019-09-26 16-34-31

The double trace of `find_go_pkg_dirs`, and the logs from inside that function, show that this works.

PR #168, opened 2019-10-04
Two node e2e

skitt:
This allows the Helm e2e test to run with a shared broker/engine cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/168)

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -38,15 +38,15 @@ function kind_clusters() {
         fi
     done
     if [[ ${#logs[@]} -gt 0 ]]; then
-        echo "(Watch the installation processes with \"tail -f ${logs[@]}\".)"
+        echo "(Watch the installation processes with \"tail -f ${logs[*]}\".)"
         for i in 1 2 3; do
             if [[ pids[$i] -gt -1 ]]; then
                 wait ${pids[$i]}
                 if [[ $? -ne 0 && $? -ne 127 ]]; then
                     echo Cluster $i creation failed:
-                    cat $logs[$i]
+                    cat ${logs[$i]}
                 fi
-                rm -f $logs[$i]
+                rm -f ${logs[$i]}
             fi
         done
     fi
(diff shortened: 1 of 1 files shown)


dimaunx:
if we move to 2 node e2e , shouldn't we just remove cluster1 from e2e? If we choose to install a broker on any other cluster then cluster1 it makes cluster1 redundant, as cluster1 doesn't have weave installed and can't be used as a gateway (no gw tagging).

skitt:
> if we move to 2 node e2e , shouldn't we just remove cluster1 from e2e?

This patch still allows both, 2-node and 3-node e2e; I thought it would be useful to keep the 3-node variant for now.

mangelajo:
Due to the operator changes there are conflicts now.

Jaanki:
Missing description on usage. Rest looks good.

scripts/kind-e2e/e2e.sh
```diff
@@ -311,25 +323,41 @@ if [[ $1 != keep ]]; then
...
 export KUBECONFIG=$(echo ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster{1..3} | sed 's/ /:/g')
 
-kind_clusters "$@"
-setup_custom_cni
+broker_cluster=cluster1
+if [[ $5 = cluster1 || $5 = cluster2 || $5 = cluster3 ]]; then
+    echo Broker on $5
+    broker_cluster=$5
+elif [[ -n $5 ]]; then
+    echo Invalid broker setting $5, ignoring
```
Jaanki: Or may "echo broker cluster not specified. Creating 3 clusters setup with seperate broker cluster"

scripts/kind-e2e/e2e.sh
```diff
@@ -311,25 +323,41 @@ if [[ $1 != keep ]]; then
     trap cleanup EXIT
 fi
 
-echo Starting with status: $1, k8s_version: $2, logging: $3, kubefed: $4.
+echo Starting with status: $1, k8s_version: $2, logging: $3, kubefed: $4, broker :$5.
```
Jaanki: This will be like "make ci e2e .... broker=cluster2" for example right. This should be mentioned somewhere. Maybe in testing.md from my doc PR #176

scripts/kind-e2e/e2e.sh
skitt: This message is printed if the user asks to install the broker on a non-existing cluster; it’s not the default three-node cluster case.

scripts/kind-e2e/e2e.sh
skitt: Yes, it should indeed! Once #176 is merged we’ll have to make it a requirement that user-visible behaviour-changing patches also update the documentation. (And this patch mustn’t go in before #176; in any case it needs substantial rework following the merge of the operator patch.)

mangelajo:
I don't feel brave enough to handle the conflicts here with confidence @skitt

skitt:
> I don't feel brave enough to handle the conflicts here with confidence @skitt

I know, the changes are rather complex, it’s on my queue...

tpantelis:
As an FYI, we're planning to enhance and add more tests that will use the third broker.

mkolesnik:
This seems rather outdated due to all the recent shipyard work and all, but if the idea still is valid please implement in shipyard

PR #169, opened 2019-10-07
Fix entrypoints to support debug correctly

mpeterson:
The current entrypoints were using a flag `--debug` that is not
processed by the executables they run. Therefore, when setting
`SUBMARINER_DEBUG` environmental variable the container crashed
and entered a crash loop. Moreover, the debug env var was not taking
care of putting `charon` in debug mode.

This patch fixes this issues by removing the flag `--debug`.

### DIFF
--- a/package/submariner.sh
+++ b/package/submariner.sh
@@ -7,7 +7,7 @@ export CHARON_PID_FILE=/var/run/charon.pid
 rm -f ${CHARON_PID_FILE}
 
 if [ "${SUBMARINER_DEBUG}" == "true" ]; then
-    DEBUG="--debug -v=9"
+    DEBUG="-v=9"
 else
     DEBUG="-v=4"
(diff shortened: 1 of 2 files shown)


package/submariner.sh
```diff
@@ -7,7 +7,8 @@ export CHARON_PID_FILE=/var/run/charon.pid
 rm -f ${CHARON_PID_FILE}
 
 if [ "${SUBMARINER_DEBUG}" == "true" ]; then
-    DEBUG="--debug -v=9"
+    DEBUG="-v=9"
+    export CE_IPSEC_DEBUG=true
```
mangelajo: We control that separately from operator/helm scripts.

You could want to debug submariner code itself, without debugging the ipsec part (which can be very verbose), I'd keep them separate.

PR #171, opened 2019-10-08
Edit tests for mismatched DaemonSet name Op/Helm

dfarrell07:
The Operator uses preferred DaemonSet name routeagent, Helm uses old
submariner-routeagent. Modify tests to account for difference, pass.

The CI overrides deploytool to be "operator" and passes. But with the
default value of deploytool ("helm"), verifications fail without this.


mangelajo:
I changed the operator to deploy in the same format as helm. https://github.com/submariner-io/submariner/pull/174

I didn't see this one sorry.

PR #173, opened 2019-10-09
Uncouple ci building from e2e

mangelajo:
This will avoid the double-build of engine, routeagent and operator,
because they are built once for ci, and once again for e2e which
was not expected.

We start doing what documentation says, make compiles/builds, and e2e
just runs e2e with an existing build, README.md says:

```
To build `submariner-engine` and `submariner-route-agent` you can
trigger `make`, which will perform a Dapperized build of the components.

To run basic e2e tests you can trigger `make e2e` command.
```

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -4,7 +4,12 @@ set -e
 source $(dirname $0)/lib/debug_functions
 source $(dirname $0)/lib/version
 
-cd $(dirname $0)/..
-mkdir -p bin
-echo Building submariner-engine version $VERSION
-CGO_ENABLED=0 go build -ldflags "-X main.VERSION=$VERSION" -o bin/submariner-engine main.go
+cd $(dirname $0)/
+
+./build-engine
+./build-routeagent
+
+if [[ $5 = operator ]]; then
+   test -d ../operators/go/submariner-operator || ./codegen-operator
+   ./build-operator
+fi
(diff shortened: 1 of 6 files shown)


mangelajo:
Added build-engine script which was left out of the commit.

scripts/build
```diff
@@ -4,7 +4,12 @@ set -e
...
-cd $(dirname $0)/..
-mkdir -p bin
-echo Building submariner-engine version $VERSION
-CGO_ENABLED=0 go build -ldflags "-X main.VERSION=$VERSION" -o bin/submariner-engine main.go
+cd $(dirname $0)/
+
+./build-engine
+./build-routeagent
+
+if [[ $5 = operator ]]; then
```
skitt: This needs to be handled differently now, we don’t have the arguments at this point.

mangelajo: right, there's my:

Error response from daemon: No such image: quay.io/submariner/submariner-operator:dev

thanks, I guess I can pass down the arguments to this script?

mangelajo: ok did that for now: https://github.com/submariner-io/submariner/pull/173/files#diff-799c33a5bd89478f978c187c2d3f9b79R10

mangelajo:
CI worked. @tpantelis  ^

PR #175, opened 2019-10-09
Make operator use released version 0.0.2 of submariner

mangelajo:
By the default the operator will use version 0.0.2 of the submariner
components from quay.io/submariner.


mangelajo:
CI passed, merging

PR #186, opened 2019-10-16
Fix Operator E2E polluting git tracked files

dfarrell07:
Currently, the Operator deployments in the E2E tests create/modify some
CR/CRD/namespace files such that the git repo isn't clean after running.
This cleans up better behind the logic to leave not-ignored files
unchanged.


scripts/kind-e2e/e2e.sh
```diff
@@ -238,7 +238,7 @@ function create_subm_vars() {
   ce_ipsec_ikeport=500
   ce_ipsec_nattport=4500
 
-  subm_ns=operators
+  subm_ns=submariner
```
dfarrell07: We could alternatively change the default Operator namespace to "operators", but we'd also need to update docs. In general, I just think the default for the Operator should be what we use for E2E.

mangelajo:
An alternative would also be copying the yml to a $(mktemp) file, doing the change there.. and using that file to deploy

.gitignore
```diff
@@ -6,7 +6,8 @@ submariner-route-agent
 dist
 output
 strongswan
-/vendor
+vendor
```
mangelajo: that change bit is not necessary anymore, the operator is out of the tree now.

dfarrell07:
https://github.com/submariner-io/submariner-operator/pull/37

PR #188, opened 2019-10-18
Ensure we set the default values

skitt:
We allow the Submariner CR to omit certain values (currently, the
repository and version used for the container images), but we never
set the default values. This calls setSubmarinerDefaults() to ensure
the values are populated.


mangelajo:
Merged on the submariner-operators repo

PR #190, opened 2019-10-22
Remove the operator sourcecode from this repository.

mangelajo:
We could as well remove the operator deployment scripts at some
point, we'd need to refactor CI moving the operator deploy logic
somewhere else. Same thing for helm.

But that needs a bigger change.

For now, e2e run in this repo will be pinpointed to 0.0.1 of the
operator and we can change that later.


mangelajo:
@skitt @dimaunx  ^ :) at least for now we'd have something

PR #192, opened 2019-10-25
Shrink the binaries

skitt:
This removes debugging information (but not stack trace information)
from the binaries we build, and compresses them. For the engine, this
shrinks the binary from 36MiB to 5MiB.


mangelajo:
Cool but I have some questions/concerns:

How big is the container image before/after?
Are we sure it's worth it, for example in terms of compatibility?
start-up time? I heard it's sometimes quicker for packed?
What if somebody needs to debug? (at this stage very likely..) 

I'd at least provide an env var or something to be able to build the debug version for development purposes.

scripts/build-engine
```diff
@@ -7,4 +7,5 @@ source $(dirname $0)/lib/version
 cd $(dirname $0)/..
 mkdir -p bin
 echo Building submariner-engine version $VERSION
-CGO_ENABLED=0 go build -ldflags "-X main.VERSION=$VERSION" -o bin/submariner-engine main.go
+CGO_ENABLED=0 go build -ldflags "-s -w -X main.VERSION=$VERSION" -o bin/submariner-engine main.go
```
mangelajo: What do you think of an env bar we could pass down via DAPPER_ENV_VARS to make, so we are able to build with full debug and without packing?

I think we could need that for development at least.

skitt:
> Cool but I have some questions/concerns:
> 
> How big is the container image before/after?

Engine: 422MiB before, 391MiB after; route-agent: 140MiB before, 109MiB after.

> Are we sure it's worth it, for example in terms of compatibility?

There shouldn’t be any compatibility issue.

> start-up time? I heard it's sometimes quicker for packed?

Slightly longer for packed binaries nowadays, but on the order of milliseconds...

> What if somebody needs to debug? (at this stage very likely..)

I wondered about that. When I’m debugging, I use the binary built from my IDE, not the binary built by the build scripts; but if someone wants to attach to a container to debug, then yes, it would be nice to have an option to keep the debug information.

> I'd at least provide an env var or something to be able to build the debug version for development purposes.

I agree, I’ll do that.

PR #193, opened 2019-10-25
Add trailing dots everywhere

skitt:
In its current form, find_go_pkg_dirs handles all the directories
containing Go as a single unit, because the values are fed into sed in
a single line. Thus, with operators and pkg, sed processes "operators
pkg" as a single line, producing "./operators pkg/...".

Changing the source pattern to "[^ ]*" fixes that, and results in
"./operators/... ./pkg/...".


mangelajo:
ahaaa! but why does it fail now :?

skitt:
> ahaaa! but why does it fail now :?

```
Fedora 30 - x86_64 - Updates                     13 kB/s | 3.4 kB     00:00    
Failed to download metadata for repo 'updates'
Error: Failed to download metadata for repo 'updates'
```

It should be a temporary failure. #130 would avoid this entirely ;-).

manosnoam:
> > ahaaa! but why does it fail now :?
> 
> ```
> Fedora 30 - x86_64 - Updates                     13 kB/s | 3.4 kB     00:00    
> Failed to download metadata for repo 'updates'
> Error: Failed to download metadata for repo 'updates'
> ```
> 
> It should be a temporary failure. #130 would avoid this entirely ;-).

I've rerun the Travis build, and it has passed.

PR #195, opened 2019-10-28
Stop pushing the operator

skitt:
Now that the separate operator project pushes images, we should stop
pushing them from Submariner (even before we delete the operator
source code).


mangelajo:
Let's merge this one instead? https://github.com/submariner-io/submariner/pull/190

skitt:
> Let's merge this one instead? #190

Yup, I’d prefer if we could get #190 in; the point of this was just to stop pushing the images if we can’t get reviews on #190 soon.

mangelajo:
#190 merged, thanks a lot @skitt

PR #197, opened 2019-10-30
Address comments in #149

tpantelis:
https://github.com/submariner-io/submariner/pull/149

- renamed TestPod to NetworkPod
- renamed createListenerConnectorPair t0 runAndVerifyNetworkPod2ServicePair
- renamed other function names where appropriate


test/e2e/dataplane/tcp_pod_to_service.go
```diff
@@ -30,51 +30,51 @@ var _ = Describe("[dataplane] Basic Pod to Service tests across clusters without
 	})
 })
 
-func testPod2ServiceTCP(f *framework.Framework, leftScheduling framework.TestPodScheduling, rightScheduling framework.TestPodScheduling) {
+func testPod2ServiceTCP(f *framework.Framework, leftScheduling framework.NetworkPodScheduling, rightScheduling framework.NetworkPodScheduling) {
```
Jaanki: I like "TestPod..." more as it is generic. NetworkPod... becomes more specific which at this point is not needed IMO

test/e2e/framework/network_pods.go
```diff
@@ -1,18 +1,150 @@
...
+			return false, nil // pod is still pending
+		}
+		finalPod = pod
+		return true, nil // pods is running
+	})
+	Expect(err).NotTo(HaveOccurred())
+	np.Pod = finalPod
+}
+
+func (np *NetworkPod) AwaitSuccessfulFinish() {
```
mangelajo: My version was more generic though. For example being able to wait for non-successful makes it possible to do negative tests in network policies (pods don't connect), but we can add another function on a later PR very easily. The resulting test code is cleaner your way.

test/e2e/dataplane/tcp_pod_to_service.go
mangelajo: Yeah,  there was a discussion on the other patch around this.

TestPod is shorter, I agree with you, and we can assume that pods are going to be "network" always for us... but I'm ok with the current patch, we need to make progress. And thanks tom for proactively proposing the amendments.

test/e2e/framework/network_pods.go
tpantelis: yeah that's what I was thinking, ie add `AwaitFinish` later if needed.

test/e2e/dataplane/tcp_pod_to_service.go
tpantelis: I'm OK either way - just trying to address Mike's concerns. While `TestPod` is generic, these pods really aren't generic, ie they run specific commands to test networking so `NetworkPod` seems more meaningful in this case.

PR #198, opened 2019-11-01
Add helper functions for e2e tests

tpantelis:
Added some helper functions for upcoming e2e enhancements.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -1,6 +1,7 @@
 package framework
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -8,7 +9,9 @@ import (
 
 	"github.com/onsi/ginkgo"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 
@@ -33,6 +36,22 @@ const (
 	ClusterB
 	ClusterC
 )
+
+const (
+	SubmarinerEngine = "submariner-engine"
+	GatewayLabel     = "submariner.io/gateway"
+)
+
+type PatchFunc func(pt types.PatchType, payload []byte) error
+
+type PatchStringValue struct {
+	Op    string `json:"op"`
+	Path  string `json:"path"`
+	Value string `json:"value"`
+}
+
+type DoOperationFunc func() (interface{}, error)
+type CheckResultFunc func(result interface{}) (bool, error)
 
 // Framework supports common operations used by e2e tests; it will keep a client & a namespace for you.
 // Eventual goal is to merge this with integration test framework.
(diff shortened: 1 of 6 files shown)


test/e2e/framework/nodes.go
```diff
@@ -0,0 +1,40 @@
...
+			nonGatewayName = nonGateway.Name
+		}
+	}
+
+	Logf("Found gateway node: %s, non-gateway node: %s", gatewayName, nonGatewayName)
+	return
+}
+
+func (f *Framework) SetGatewayLabelOnNode(cluster ClusterIndex, nodeName string, isGateway bool) {
+	DoPatchOperation("/metadata/labels/submariner.io~1gateway", strconv.FormatBool(isGateway),
```
mangelajo: is "~1" a typo? should be "/"?

test/e2e/framework/nodes.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"strconv"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (f *Framework) FindGatewayAndNonGatewayNodes(cluster ClusterIndex) (gateway *v1.Node, nonGateway *v1.Node) {
```
mangelajo: nit, nothing functional:

May be just GetNodes? (current function name seems overly verbose, and the fact that we will do

gateway, nonGateway := nodes.GetNodes() (even nodes.Get() ?)

would be verbose enough during usage?

test/e2e/framework/pods.go
```diff
@@ -0,0 +1,32 @@
+package framework
+
+import (
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (f *Framework) FindPodsByAppLabel(cluster ClusterIndex, appName string, namespace string, expectedCount int) *v1.PodList {
```
mangelajo: may be AwaitPodsByAppLabel? 

(since we wait for them, and not only find?)

test/e2e/framework/pods.go
```diff
@@ -0,0 +1,32 @@
...
+		if expectedCount < 0 || len(pods.Items) == expectedCount {
+			return true, nil
+		}
+
+		Logf("Actual pod count %d does not match the expected pod count %d - retrying...", len(pods.Items), expectedCount)
+		return false, nil
+	}).(*v1.PodList)
+}
+
+func (f *Framework) FindSubmarinerEnginePod(cluster ClusterIndex) *v1.Pod {
```
mangelajo: AwaitSubmarinerEnginePod?

test/e2e/framework/framework.go
```diff
@@ -34,6 +37,23 @@ const (
 	ClusterC
 )
 
+const (
+	SubmarinerNamespace = "submariner"
```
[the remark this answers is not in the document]
tpantelis: ok I'll make it configurable from the command line just in case...

test/e2e/framework/nodes.go
tpantelis: no typo - it's the special escape sequence for "/" (defined in some RFC) so it's not treated as part of the path.

test/e2e/framework/nodes.go
tpantelis: It's more than simply GetNodes - specialized to find 2 specific nodes so I prefer the verbosity for clarity.

test/e2e/framework/pods.go
tpantelis: yeah

test/e2e/framework/nodes.go
Jaanki: +1 to GetNodes

test/e2e/framework/pods.go
Jaanki: I think Find is more appropriate. we are waiting till all pods are found which is the default behavior. We have to wait for output in every case.

test/e2e/framework/nodes.go
tpantelis: OK - I'll shorten it but I think it should be `FindNodes`, as it's a search - `GetNodes` implies a lookup.

test/e2e/framework/pods.go
tpantelis: I'm find either way - ok to leave it as Find?

test/e2e/framework/nodes.go
tpantelis: I added a comment and also used the `GatewayLabel` const with `strings.Replace` to make it clearer.

test/e2e/framework/pods.go
mangelajo: I think Await explains better what the function will do (I would expect Find to only return the existing pods "now" and just exit).

mangelajo: Oh, comment does not appear here.

In this case I think it's better Await, because otherwise it's confusing, I'd expect it to output the pods right now, even if they are not present.

test/e2e/framework/nodes.go
```diff
@@ -0,0 +1,42 @@
...
+			nonGateway = &nodes.Items[i]
+			nonGatewayName = nonGateway.Name
+		}
+	}
+
+	Logf("Found gateway node: %s, non-gateway node: %s", gatewayName, nonGatewayName)
+	return
+}
+
+func (f *Framework) SetGatewayLabelOnNode(cluster ClusterIndex, nodeName string, isGateway bool) {
```
Jaanki: How about adding a "UnsetGatewayLabelOnNode" for remove cluster test use case?

tpantelis:
Pushed https://github.com/submariner-io/submariner/pull/198/commits/1d41096504512057206c0b826de947c4790aad3f to change `FindNodes` to `FindNodesByGatewayLabel`.

test/e2e/framework/nodes.go
Jaanki: Using the same function with isGateway "true" or "false" is enough. Renaming the function to "ToggleGatewayLabelOnNode" would be more appropriate.

This is not something that needs urgent change.

Rest of the patch looks good.

mangelajo: Set is better here, toggle implies that you don’t pass true/false and It will switch the existing state.

PR #201, opened 2019-11-05
Install SubM on broker cluster

Jaanki:
One of the test scenarios requires dynamically
adding/removing third cluster to the setup.
The idea is to use broker cluster as the third
cluster.

This patch just installs SubM on broker cluster
but does not tag any gateway nodes. Node is tagged
later in the test case to mimic adding third cluster.

Signed-Off-By: Janki Chhatbar <jchhatba@redhat.com>

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -122,6 +122,38 @@ function setup_broker() {
     SUBMARINER_BROKER_TOKEN=$(kubectl --context=cluster1 -n ${SUBMARINER_BROKER_NS} get secrets -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='${SUBMARINER_BROKER_NS}-client')].data.token}"|base64 --decode)
 }
 
+# Only install Submariner on broker cluster with 2 worker nodes but do not tag any worker nodes with gateway label
+function install_subm_in_cluster1() {
+    if kubectl --context=cluster1 wait --for=condition=Ready pods -l app=submariner-engine -n submariner --timeout=60s > /dev/null 2>&1; then
+            echo Submariner already installed, skipping submariner helm installation...
+            update_subm_pods cluster1
+        else
+            echo Installing submariner on cluster1...
+            helm --kube-context cluster1 install submariner-latest/submariner \
+            --name submariner \
+            --namespace submariner \
+            --set ipsec.psk="${SUBMARINER_PSK}" \
+            --set broker.server="${SUBMARINER_BROKER_URL}" \
+            --set broker.token="${SUBMARINER_BROKER_TOKEN}" \
+            --set broker.namespace="${SUBMARINER_BROKER_NS}" \
+            --set broker.ca="${SUBMARINER_BROKER_CA}" \
+            --set submariner.clusterId="cluster1" \
+            --set submariner.clusterCidr="10.244.0.0/16" \
+            --set submariner.serviceCidr="100.94.0.0/16" \
+            --set submariner.natEnabled="false" \
+            --set routeAgent.image.repository="submariner-route-agent" \
+            --set routeAgent.image.tag="local" \
+            --set routeAgent.image.pullPolicy="IfNotPresent" \
+            --set engine.image.repository="submariner" \
+            --set engine.image.tag="local" \
+            --set engine.image.pullPolicy="IfNotPresent" \
+            --set crd.create="false"
+            echo Waiting for submariner pods to be Ready on cluster1...
+            kubectl --context=cluster1 wait --for=condition=Ready pods -l app=submariner-engine -n submariner --timeout=60s
+            kubectl --context=cluster1 wait --for=condition=Ready pods -l app=submariner-routeagent -n submariner --timeout=60s
+    fi
+}
+
 function setup_cluster2_gateway() {
     if kubectl --context=cluster2 wait --for=condition=Ready pods -l app=submariner-engine -n submariner --timeout=60s > /dev/null 2>&1; then
             echo Submariner already installed, skipping submariner helm installation...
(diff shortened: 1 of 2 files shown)


Jaanki:
@tpantelis @aswinsuryan if you guys can take a look :)

scripts/kind-e2e/e2e.sh
```diff
@@ -122,6 +122,38 @@ function setup_broker() {
...
 }
 
+function install_subm_in_cluster1() {
+    if kubectl --context=cluster1 wait --for=condition=Ready pods -l app=submariner-engine -n submariner --timeout=60s > /dev/null 2>&1; then
+            echo Submariner already installed, skipping submariner helm installation...
+            update_subm_pods cluster1
+        else
+            echo Installing submariner on cluster1...
+            worker_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster1-worker | head -n 1)
+            #kubectl --context=cluster1 label node cluster1-worker "submariner.io/gateway=true" --overwrite
```
tpantelis: Can you remove this line?

scripts/kind-e2e/e2e.sh
```diff
@@ -122,6 +122,38 @@ function setup_broker() {
     SUBMARINER_BROKER_TOKEN=$(kubectl --context=cluster1 -n ${SUBMARINER_BROKER_NS} get secrets -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='${SUBMARINER_BROKER_NS}-client')].data.token}"|base64 --decode)
 }
 
+function install_subm_in_cluster1() {
+    if kubectl --context=cluster1 wait --for=condition=Ready pods -l app=submariner-engine -n submariner --timeout=60s > /dev/null 2>&1; then
+            echo Submariner already installed, skipping submariner helm installation...
+            update_subm_pods cluster1
+        else
+            echo Installing submariner on cluster1...
+            worker_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster1-worker | head -n 1)
```
tpantelis: This isn't needed

scripts/kind-e2e/e2e.sh
Jaanki: Commented line highlights that the label is not set. I will write a comment instead before the function.

scripts/kind-e2e/e2e.sh
```diff
@@ -122,6 +122,37 @@ function setup_broker() {
...
+            --set submariner.clusterId="cluster1" \
+            --set submariner.clusterCidr="10.244.0.0/16" \
+            --set submariner.serviceCidr="100.94.0.0/16" \
+            --set submariner.natEnabled="false" \
+            --set routeAgent.image.repository="submariner-route-agent" \
+            --set routeAgent.image.tag="local" \
+            --set routeAgent.image.pullPolicy="IfNotPresent" \
+            --set engine.image.repository="submariner" \
+            --set engine.image.tag="local" \
+            --set engine.image.pullPolicy="IfNotPresent"
```
tpantelis: We need to also `--set crd.create="false"`.

scripts/kind-e2e/e2e.sh
Jaanki: Done

scripts/kind-e2e/e2e.sh
Jaanki: Done

scripts/kind-e2e/e2e.sh
Jaanki: DOne

test/e2e/framework/test_context.go
```diff
@@ -32,6 +32,7 @@ func registerFlags(t *TestContextType) {
 	flag.StringVar(&t.KubeConfig, "kubeconfig", os.Getenv("KUBECONFIG"),
 		"Path to kubeconfig containing embedded authinfo.")
 	flag.Var(&TestContext.KubeContexts, "dp-context", "kubeconfig context for dataplane clusters (use several times).")
+	flag.Var(&TestContext.KubeContexts, "broker-context", "kubeconfig context for broker cluster.")
```
mangelajo: Seems fine, 

My only worry is that we should not use the "broker" cluster for dataplane testing, it's fine for checking "Cluster" or "Endpoint" CRs, but not for submariner pods, dataplane, etc...

Keep that in mind, you can use ClusterC for that (probably passing -dp-context a third time)

mangelajo:
Seems fine, 

My only worry is that we should not use the "broker" cluster for dataplane testing, it's fine for checking "Cluster" or "Endpoint" CRs, but not for submariner pods, dataplane, etc...

Keep that in mind, you can use ClusterC for that (probably passing -dp-context a third time)

test/e2e/framework/test_context.go
Jaanki: That would mean running e2e with total of 4 clusters so we thought of using broker cluster itself.

tpantelis: What's the reason we shouldn't use the broker cluster for dataplane testing? We're not requiring the broker be it's own separate cluster so it will likely also have submariner installed in real environments. 

We could also add a 4th cluster but were concerned of the extra overhead in terms of resources.

Jaanki: Another way to look at this is thinking we have 3 dataplane clusters  which broker installed on one of them :)

tpantelis: I think we should use a separate var, BrokerContext

test/e2e/framework/framework.go
```diff
@@ -31,6 +31,7 @@ type ClusterIndex int
 const (
 	ClusterA ClusterIndex = iota
 	ClusterB
+	Broker
```
tpantelis: This should be after `ClusterC`.

test/e2e/framework/test_context.go
tpantelis: Actually appending won't work properly unless the 3rd _-dp-context_ is specified. I think we should just use a fixed size `ClusterClients` array based on the `ClusterIndex` enum. In `Framework.BeforeEach`:

```
f.ClusterClients = make([]string, Broker + 1)
for i, context := range TestContext.KubeContexts {
    f.ClusterClients[i] = f.createKubernetesClient(context)
}	
ClusterClients[Broker] = f.createKubernetesClient(TestContext.BrokerContext)
```

In `TestContext.validateFlags`, check the allowed max length of `TestContext.KubeContexts`:

```
if len(t.KubeContexts) < (int(ClusterC) + 1) {
    klog.Fatalf("Too many dp-context parameters (%d) specified", len(t.KubeContexts))
}
```

tpantelis: After thinking about this some more, I think we should just use pass _cluster1_ as the 3rd _dp-context_ so it's mapped to `ClusterC` for the add/remove  cluster tests.  The test can do a sanity check to ensure there's initially no gateway nodes configured. This way the code stays as is although I think we should still validate the max length of `KubeContexts` as outlined above.

PR #206, opened 2019-11-07
Submariner IPAM controller

vthapar:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/206)


package/Dockerfile.routeagent
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.0
```
vthapar: This is to make Travis happy till we sync the fix from master.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+	// Wait for the caches to be synced before starting workers
+	klog.Info("Waiting for informer caches to sync")
+	if ok := cache.WaitForCacheSync(stopCh, i.servicesSynced, i.podsSynced); !ok {
+		return fmt.Errorf("failed to wait for caches to sync")
+	}
+
+	klog.Info("Starting workers")
+	go wait.Until(i.runServiceWorker, time.Second, stopCh)
+	go wait.Until(i.runPodWorker, time.Second, stopCh)
+	wg.Wait()
```
tpantelis: I don't see where the `wg` is released.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+			return nil
+		}
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// Retrieve the latest version of Service before attempting update
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			service, err := i.kubeClientSet.CoreV1().Services(ns).Get(name, metav1.GetOptions{})
+
+			if err != nil {
+				i.serviceWorkqueue.Forget(obj)
```
tpantelis: This is already done on line 130.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+				annotations[submarinerIpamGlobalIp] = "169.254.1.1"
+				service.SetAnnotations(annotations)
+				klog.V(4).Infof("Updating service %s", service.Name)
+				_, updateErr := i.kubeClientSet.CoreV1().Services(ns).Update(service)
+				return updateErr
+			}
+			return nil
+		})
+		if retryErr != nil {
+			i.serviceWorkqueue.Forget(obj)
```
tpantelis: Should we re-queue/retry?

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+)
+
+func NewController(spec SubmarinerIpamControllerSpecification, config InformerConfigStruct) *Controller {
```
tpantelis: The params could be pointers to avoid copy.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+		excludeNamespaces: exclusionMap,
+	}
+	klog.Info("Setting up event handlers")
+	config.ServiceInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc: ipamController.enqueueService,
+		UpdateFunc: func(old, new interface{}) {
+			ipamController.enqueueService(new)
+		},
+		DeleteFunc: ipamController.handleRemovedService,
+	}, 60*time.Second)
```
mpeterson: nit: would be better to move this time to a constant 

same on L47

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+	for i.processNextService() {
+	}
+}
+
+func (i *Controller) runPodWorker() {
+	for i.processNextPod() {
+	}
+}
+
+const submarinerIpamGlobalIp = "submariner.io/globalIp"
```
mpeterson: better at the top of the file or a constants file since it will probably be used for other things other than this controller

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if err != nil {
+		utilruntime.HandleError(err)
+		return true
+	}
+
+	return true
+}
+
+func (i *Controller) processNextPod() bool {
```
mpeterson: this function has an awful amount of repetition. Please re organize them into a function or method.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if err != nil {
+		utilruntime.HandleError(err)
+		return true
+	}
+
+	return true
+}
+
+func (i *Controller) enqueueService(obj interface{}) {
```
mpeterson: `enqueueService` and `enqueuePod` are repeating the same logic. Please use functions/methods.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+	klog.V(6).Infof("Enqueueing pod for ipam controller %v", obj)
+	i.podWorkqueue.AddRateLimited(key)
+}
+
+func (i *Controller) handleRemovedService(obj interface{}) {
```
mpeterson: `handleRemovedService` and `handleRemovedPod` are repeating the same logic. Please use functions/methods.

pkg/ipmanager/ipam/types.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	kubeInformers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+type SubmarinerIpamControllerSpecification struct {
+	ClusterID  string
+	GlobalCidr []string
```
mpeterson:
```suggestion
	GlobalCIDR []string
```

pkg/ipmanager/main.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		klog.Fatalf("Error building kubeconfig: %s", err.Error())
+	}
+
+	clientSet, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		klog.Fatalf("Error building clientset: %s", err.Error())
+	}
+
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(clientSet, time.Second*60)
```
mpeterson: nit: third place where you are using the same value

pkg/ipmanager/ipam/controller.go
sridhargaddam: I have the same question. Also, in Line 54, should it be wg.Add(2)?

pkg/ipmanager/ipam/controller.go
sridhargaddam: coastguard follows this convention, but I'd prefer what @mpeterson suggested.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+			if err != nil {
+				i.serviceWorkqueue.Forget(obj)
+				return fmt.Errorf("error retrieving submariner-ipam-controller service object %s: %v", name, err)
+			}
+			annotations := service.GetAnnotations()
+			if annotations == nil {
+				annotations = map[string]string{}
+			}
+			if annotations[submarinerIpamGlobalIp] == "" {
+				annotations[submarinerIpamGlobalIp] = "169.254.1.1"
```
sridhargaddam: I hope this (169.254.1.1) needs to be revisited right? Can you please add a comment here.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+	klog.V(6).Infof("Enqueueing service for ipam controller %v", obj)
+	i.serviceWorkqueue.AddRateLimited(key)
+}
+
+func (i *Controller) enqueuePod(obj interface{}) {
```
sridhargaddam: This method (including enqueuService) gets invoked for every 60 seconds. I've seen this as a pattern in the current code. I have a feeling that this is not scalable. Imagine we have 10,000 PODs in a cluster. We get an event for every POD and that gets repeated for every 60 secs. @tpantelis @mpeterson your thoughts on this?

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,236 @@
...
+		return
+	}
+	klog.V(6).Infof("Enqueueing pod for ipam controller %v", obj)
+	i.podWorkqueue.AddRateLimited(key)
+}
+
+func (i *Controller) handleRemovedService(obj interface{}) {
+	klog.V(6).Infof("Handling remove service in ipamController %v", obj)
+	service := obj.(*k8sv1.Service)
+	klog.V(4).Infof("Removing service %v", service.Name)
```
sridhargaddam: I hope in this method, we will mark the allocated ip (for Service and similarly in the next api for POD) as free. Please add a TODO comment.

pkg/ipmanager/ipam/controller.go
tpantelis: I agree - we shouldn't need such a frequent resync period. If we handle retries then I think we can set the resync period to 0.

pkg/ipmanager/ipam/controller.go
vthapar: Pattern taken from https://github.com/submariner-io/submariner/blob/master/pkg/routeagent/controllers/route/route.go#L144

pkg/ipmanager/ipam/controller.go
vthapar: We already do. If we return error, it will be put back in queue to be tried later. that is why we return error in next line.

pkg/ipmanager/ipam/controller.go
vthapar: What would resync 0 mean? What will happen if it failed and we return error when processing the request? Will it requeue?

pkg/ipmanager/ipam/controller.go
vthapar: Done.

pkg/ipmanager/ipam/controller.go
vthapar: Done

pkg/ipmanager/ipam/controller.go
vthapar: done

pkg/ipmanager/ipam/controller.go
tpantelis: But something has to call `wg.Done()` to decrement the counter so `wg.Wait()` will complete.  There doesn't seem a need for a `WaitGroup` here (nor in the route.go code).

pkg/ipmanager/ipam/controller.go
tpantelis: I don't see where it gets re-queued - the returned error is checked on line 138 but it just logs it. Also if you're re-queuing you don't want to call `Forget`.

pkg/ipmanager/ipam/controller.go
tpantelis: On error when processing a resource from the queue, it's up to us to re-queue it if we deem it a transient error such that a retry might succeed.  I believe a 0 resync period disables unsolicited periodic resyncing but it will still resync on reconnection to the backend.  A periodic resync is basically a fallback to re-process resources that might have previously failed. It will send an update for every resource even if nothing changed which, as @sridhargaddam points out, may be problematic when there's lot of resources, especially with a frequent period like 60 sec. That frequency is probably OK for the subm components that process Endpoint and Cluster resources as there's not a lot of those (essentially one per cluster). In this case, maybe a less frequent resync period of say every hour would be reasonable - also maybe we could shortcut the processing on update by checking if the pod/service already has the annotation before queueing it.

pkg/ipmanager/ipam/controller.go
vthapar: Oh right. I originally wrote this without Forget and added it later, wondering if we want to requeue or not. The retry loop already retries for conflict. In what cases can we fail that are not conflict and likely to be resolved next time we retry? No point in requeue if it is going to fail again and again. From my experience with ODL, either approach works, but better to have some test data in hand to make a call for it. Thoughts?

tpantelis: Maybe the backend could be temporarily busy or unavailable such that a retry might succeed.

scripts/kind-e2e/e2e.sh
```diff
@@ -212,6 +212,7 @@ function kind_import_images() {
         echo "Loading submariner images in to cluster${i}..."
```
sridhargaddam: I'm getting an error when trying to build/deploy the latest submariner repo with this patch. 
The following seems to be missing in Line200
docker tag quay.io/submariner/submariner-ipam:dev submariner-ipam:local

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,312 @@
...
+		key := obj.(string)
+		ns, name, err := cache.SplitMetaNamespaceKey(key)
+		if err != nil {
+			i.serviceWorkqueue.Forget(obj)
+			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
+		}
+		if i.excludeNamespaces[ns] {
+			i.serviceWorkqueue.Forget(obj)
+			return nil
+		}
```
sridhargaddam: Before enqueing the event in the WorkQueue, we are already checking it against the excluded namespaces. So, this looks redundant.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,312 @@
...
+		key := obj.(string)
+		ns, name, err := cache.SplitMetaNamespaceKey(key)
+		if err != nil {
+			i.podWorkqueue.Forget(obj)
+			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
+		}
+		if i.excludeNamespaces[ns] {
+			i.podWorkqueue.Forget(obj)
+			return nil
+		}
```
sridhargaddam: Before enqueing, we are already checking it against the excluded namespaces. So, this looks redundant.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,312 @@
...
+	if i.excludeNamespaces[ns] {
+		return ""
+	}
+	return key
+}
+
+func (i *Controller) handleUpdatePod(old interface{}, new interface{}) {
+	pod := new.(*k8sv1.Pod)
+	oldGlobalIp := old.(*k8sv1.Pod).GetAnnotations()[submarinerIpamGlobalIp]
+	newGlobalIp := new.(*k8sv1.Pod).GetAnnotations()[submarinerIpamGlobalIp]
```
sridhargaddam: When would this happen?

pkg/ipmanager/ipam/ippool.go
```diff
@@ -0,0 +1,115 @@
...
+type IpPool struct {
+	cidr string
+	net  *net.IPNet
+	size int
+	available map[string]bool //IP.String() is key
+	allocated map[string]string //resource name is key, ipString is value
+	sync.RWMutex
+}
+
+func NewIpPool(cidr string) (*IpPool, error) {
```
sridhargaddam: We have to handle a restart scenario as well. In such cases, as we discussed, we will have to reconstruct/re-initialize this pool from the globalIPs allocated to the pods/services.

scripts/kind-e2e/ipam/deploy-ipam-2.yaml
```diff
@@ -0,0 +1,29 @@
...
+      labels:
+        app: submariner-ipam
+    spec:
+      containers:
+        - name: submariner-ipam
+          image: submariner-ipam:local
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: SUBMARINER_GLOBALCIDR
+              value: 169.254.1.0/24
```
sridhargaddam: "/24" is too small a value. Lets use something like "/16" which is more realistic. I agree that this is only a test-case where we are not spawning even 255 pods/services, but using 16 allows us to be close to typical deployments and helps us in finding out any issues sooner.

pkg/ipmanager/ipam/controller.go
```diff
@@ -0,0 +1,312 @@
...
+	return ipamController, nil
+}
+
+func (i *Controller) Run(stopCh <-chan struct{}) error {
+	var wg sync.WaitGroup
+	wg.Add(1)
+	defer utilruntime.HandleCrash()
+
+	// Start the informer factories to begin populating the informer caches
+	klog.Info("Starting IPAM Controller")
```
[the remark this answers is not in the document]
vthapar: Yeah, I was using it during early dev but removed before pushing. Will add it back in.

pkg/ipmanager/ipam/controller.go
vthapar: Taken care in refactoring.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,300 @@
...
+
+func (i *Controller) podGetter(namespace, name string) (runtime.Object, error) {
+	return i.kubeClientSet.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+}
+
+func (i *Controller) serviceUpdater(obj runtime.Object) error {
+	service := obj.(*k8sv1.Service)
+	annotations, err := i.annotateGlobalIp(service.Name, service.GetAnnotations())
+	if err != nil {
+		return fmt.Errorf("failed to annotated GlobalIp to service %s", service.Name)
```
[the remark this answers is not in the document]
vthapar: Done

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,300 @@
...
+			objWorkqueue.Forget(obj)
+			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
+		}
+		if i.excludeNamespaces[ns] {
+			objWorkqueue.Forget(obj)
+			return nil
+		}
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// Retrieve the latest version of Pod before attempting update
```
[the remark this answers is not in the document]
vthapar: Done

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+
+		excludeNamespaces: exclusionMap,
+		pool:              pool,
+	}
+	klog.Info("Setting up event handlers")
+	config.ServiceInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			ipamController.enqueueObject(obj, ipamController.serviceWorkqueue)
+		},
+		UpdateFunc: func(old, new interface{}) {
```
mpeterson: `new` is a language keyword. I would avoid using that name. They usually use `newObj` and `oldObj`

There are more places where this is present. Only highlighting it here.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+	if i.excludeNamespaces[ns] {
+		return ""
+	}
+	return key
+}
+
+func (i *Controller) handleUpdateService(old interface{}, new interface{}) {
+	service := new.(*k8sv1.Service)
+	oldGlobalIp := old.(*k8sv1.Service).GetAnnotations()[submarinerIpamGlobalIp]
+	newGlobalIp := new.(*k8sv1.Service).GetAnnotations()[submarinerIpamGlobalIp]
```
mpeterson: `new.(*k8sv1.Service)` you already have it in `service`

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+	if oldGlobalIp != "" && oldGlobalIp != newGlobalIp && newGlobalIp != i.pool.GetAllocatedIp(service.Name) {
+		klog.V(4).Infof("GlobalIp changed from %s to %s for %s", oldGlobalIp, newGlobalIp, old.(*k8sv1.Service).Name)
+		i.enqueueObject(new, i.serviceWorkqueue)
+	}
+}
+
+func (i *Controller) handleUpdatePod(old interface{}, new interface{}) {
+	pod := new.(*k8sv1.Pod)
+	oldGlobalIp := old.(*k8sv1.Pod).GetAnnotations()[submarinerIpamGlobalIp]
+	newGlobalIp := new.(*k8sv1.Pod).GetAnnotations()[submarinerIpamGlobalIp]
```
mpeterson: `new.(*k8sv1.Pod)` you already have it in `pod`

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	return key
+}
+
+func (i *Controller) handleUpdateService(old interface{}, new interface{}) {
+	service := new.(*k8sv1.Service)
+	oldGlobalIp := old.(*k8sv1.Service).GetAnnotations()[submarinerIpamGlobalIp]
+	newGlobalIp := new.(*k8sv1.Service).GetAnnotations()[submarinerIpamGlobalIp]
+	if oldGlobalIp != "" && oldGlobalIp != newGlobalIp && newGlobalIp != i.pool.GetAllocatedIp(service.Name) {
+		klog.V(4).Infof("GlobalIp changed from %s to %s for %s", oldGlobalIp, newGlobalIp, old.(*k8sv1.Service).Name)
```
mpeterson: why not use `service.Name` instead of `old.(*k8sv1.Service).Name`

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+		i.enqueueObject(new, i.serviceWorkqueue)
+	}
+}
+
+func (i *Controller) handleUpdatePod(old interface{}, new interface{}) {
+	pod := new.(*k8sv1.Pod)
+	oldGlobalIp := old.(*k8sv1.Pod).GetAnnotations()[submarinerIpamGlobalIp]
+	newGlobalIp := new.(*k8sv1.Pod).GetAnnotations()[submarinerIpamGlobalIp]
+	if oldGlobalIp != "" && oldGlobalIp != newGlobalIp && newGlobalIp != i.pool.GetAllocatedIp(pod.Name) {
+		klog.V(4).Infof("GlobalIp changed from %s to %s for %s", oldGlobalIp, newGlobalIp, old.(*k8sv1.Pod).Name)
```
mpeterson: why not use `pod.Name` instead of `old.(*k8sv1.Pod).Name`

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+		i.enqueueObject(new, i.podWorkqueue)
+	}
+}
```
mpeterson: Again you are mostly repeating the logic for both Pods and Services. This can be consolidated.

mpeterson: same for `handleRemoved...`

mpeterson: same for `pod/serviceUpdater`

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+)
+
+func NewController(spec *SubmarinerIpamControllerSpecification, config *InformerConfigStruct) (*Controller, error) {
+	exclusionMap := make(map[string]bool)
+	for _, v := range spec.ExcludeNS {
+		exclusionMap[v] = true
+	}
+	pool, err := NewIpPool(spec.GlobalCIDR)
```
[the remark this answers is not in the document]
vthapar: Was saving that for later as use case for multiple CIDRs is not clear yet. I was thinking it will likely  be a key:value something like "services:CIDR1,pods:CIDR2" in which case it will become a map. Will convert when have clarity on how exactly we're going to use it.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
+		}
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// Retrieve the latest version of object before attempting update
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			runtimeObj, err := objGetter(ns, name)
+
+			if err != nil {
+				// mostly this means object already deleted
```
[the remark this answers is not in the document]
vthapar: Errors returned in this are already logged in caller with error string telling us the exact error, which would be "Object not found".

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func (i *Controller) enqueueObject(obj interface{}, workqueue workqueue.RateLimitingInterface) {
+	if key := i.allowEnqueue(obj); key != "" {
+		klog.V(4).Infof("Enqueueing %v for ipam controller", key)
+		workqueue.AddRateLimited(key)
+	}
+}
+
+func (i *Controller) allowEnqueue(obj interface{}) string {
```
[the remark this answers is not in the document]
vthapar: Agreed. Will refactor it.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: will do. old naming habits.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: Argh! Looks like one of my changes got lost when rebasing. this is supposed to be key and not service name anymore.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: same as above.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: Generic runtimeObject doesn't have GetAnnotations() etc. methods, so need specific objects.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: same as above

pkg/globalnet/controllers/ipam/ipam.go
vthapar: same as above

mangelajo:
___
*[scripts/release, line 14 at r3](https://reviewable.io/reviews/submariner-io/submariner/206#-LuhQS_f-I-9ipbVDvBV:-LuhQS_g1iTaBynHeb_Z:b-ygzuhs) ([raw file](https://github.com/submariner-io/submariner/blob/0e09effdfb7e3c5e5c75f9178e67944c71f970c5/scripts/release#L14)):*
> ```
> docker tag ${REPO}/submariner-route-agent:dev ${REPO}/submariner-route-agent:${DOCKER_TAG#"v"}
> docker tag ${REPO}/submariner-route-agent:dev ${REPO}/submariner-route-agent:"${TRAVIS_COMMIT:0:7}"
> docker tag ${REPO}/submariner-route-agent:dev ${REPO}/submariner-globalnet:${DOCKER_TAG#"v"}
> ```

^ typo issue: you're tagging the route-agent as globalnet

package/submariner-globalnet.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
sridhargaddam: When I deployed with this PR, some of the e2e tests are failing.


Summarizing 3 Failures:

[Fail] [dataplane] Basic TCP connectivity tests across clusters without discovery when a pod connects via TCP to a remote pod when the pod is on a gateway and the remote pod is not on a gateway [It] should have sent the expected data from the pod to the other pod 
/go/src/github.com/submariner-io/submariner/test/e2e/framework/framework.go:344

[Fail] [dataplane] Basic TCP connectivity tests across clusters without discovery when a pod connects via TCP to a remote service when the pod is not on a gateway and the remote service is on a gateway [It] should have sent the expected data from the pod to the other pod 
/go/src/github.com/submariner-io/submariner/test/e2e/framework/framework.go:344

[Fail] [dataplane] Basic TCP connectivity tests across clusters without discovery when a pod connects via TCP to a remote pod when the pod is on a gateway and the remote pod is on a gateway [It] should have sent the expected data from the pod to the other pod 
/go/src/github.com/submariner-io/submariner/test/e2e/framework/framework.go:344

Ran 10 of 10 Specs in 318.129 seconds
FAIL! -- 7 Passed | 3 Failed | 0 Pending | 0 Skipped
--- FAIL: TestE2E (318.13s)
FAIL
exit status 1
FAIL    github.com/submariner-io/submariner/test/e2e    318.146s
[submariner]$ ./scripts/e2e keep 1.14.2 false false helm false
FATA[0863] exit status 1                                
make: *** [Makefile:18: e2e] Error 1

scripts/kind-e2e/globalnet/role-globalnet.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
sridhargaddam: I tried this patch locally and seeing the following errors.

E1128 08:09:28.444999       1 reflector.go:125] k8s.io/client-go/informers/factory.go:132: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:submariner:submariner-globalnet" cannot list resource "pods" in API group "" at the cluster scope
E1128 08:09:28.446803       1 reflector.go:125] k8s.io/client-go/informers/factory.go:132: Failed to list *v1.Service: services is forbidden: User "system:serviceaccount:submariner:submariner-globalnet" cannot list resource "services" in API group "" at the cluster scope
I1128 08:09:29.445311       1 reflector.go:160] Listing and watching *v1.Pod from k8s.io/client-go/informers/factory.go:132
E1128 08:09:29.446430       1 reflector.go:125] k8s.io/client-go/informers/factory.go:132: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:submariner:submariner-globalnet" cannot list resource "pods" in API group "" at the cluster scope
I1128 08:09:29.446923       1 reflector.go:160] Listing and watching *v1.Service from k8s.io/client-go/informers/factory.go:132
E1128 08:09:29.449149       1 reflector.go:125] k8s.io/client-go/informers/factory.go:132: Failed to list *v1.Service: services is forbidden: User "system:serviceaccount:submariner:submariner-globalnet" cannot list resource "services" in API group "" at the cluster scope

It looks like this rbac is not configured.

sridhargaddam: Aaah, should I be using deploytool=operator?

package/submariner-globalnet.sh
sridhargaddam: Okay. After switching to deploytool=operator, there are no errors and Submariner-globalnet is annotating the PODs/Services properly.

pkg/globalnet/controllers/ipam/ipam.go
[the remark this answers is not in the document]
mpeterson: The way we defined it is an array of strings with the CIDRs. No need for key value since we are not going to use it like that for now. I think this needs to be changed ASAP, it can be done in a f-up patch though.

pkg/globalnet/controllers/ipam/ipam.go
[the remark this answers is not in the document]
mpeterson: yeah, but for Object not found there is no need to log nor bubble it up. That's what I'm trying to say.

pkg/globalnet/controllers/ipam/ipam.go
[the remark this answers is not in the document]
mpeterson: mmm I don't see the refactor. Perhaps you forgot to push?

pkg/globalnet/controllers/ipam/ipam.go
mpeterson: meta from apimachinery has. But as we agreed, let's leave that for later.

mangelajo:
Sorry for the yellow dot from reviewable, I have disabled that behavior... I hope it will stop doing it

mangelajo:
I re-enabled the reviewable CI part, even if it will remain yellow, otherwise it turns red. I will finally disable it when this patch is done.

pkg/globalnet/controllers/ipam/ipam.go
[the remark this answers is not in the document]
sridhargaddam: Agree with @mpeterson. No need for key/value pairs.

scripts/kind-e2e/globalnet/role-globalnet.yaml
sridhargaddam: working fine when using operator.

mangelajo:
I see all comments handled or negotiated for follow ups, so merging. Awesome work!

PR #209, opened 2019-11-09
Fix E2E setup script failure

tpantelis:
Recent changes to install subm on the brooker cluster caused a
failure when waiting for the subm pods to be ready b/c no
gateway node is configured so the engine isn't deployed. Removed
the code to wait for the pods.


tpantelis:
@Jaanki

PR #216, opened 2019-11-15
kind-e2e: updated cluster yaml kubeadm api to v1beta2

hunchback:
addressing: https://github.com/kubernetes-sigs/kind/issues/666


dimaunx:
We are currently using kind version 0.3.0 and latest k8s version  that can be used is 1.14.2.
The only api version we can currently use is v1beta1. We are planning an upgrade across all kind to 0.5.1 and we could use k8s versoin 1.15.3 and v1beta2, for lower versions we still need to use v1beta1.

hunchback:
ok ... closing PR

PR #217, opened 2019-11-15
Fix for Flannel CNI

sridhargaddam:
Reverse path filtering is a mechanism used by networking devices to check whether a receiving packet
source address is routable. When reverse path filtering is enabled, the host will first check whether
the source of the received packet is reachable through the interface it came in. If it is routable
through the interface in which it came, then the packet is accepted otherwise its dropped.

When using flannel CNI, we have to configure loose mode for reverse path filtering and this patch
automates this change.

rp_filter - INTEGER
    0 - No source validation.
    1 - Strict mode as defined in RFC3704 Strict Reverse Path
        Each incoming packet is tested against the FIB and if the interface
        is not the best reverse path the packet check will fail.
        By default failed packets are discarded.
    2 - Loose mode as defined in RFC3704 Loose Reverse Path
        Each incoming packet's source address is also tested against the FIB
        and if the source address is not reachable via any interface
        the packet check will fail.

Current recommended practice in RFC3704 is to enable strict mode
to prevent IP spoofing from DDos attacks. If using asymmetric routing
or other complicated routing, then loose mode is recommended.

Source: https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/217)

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,13 +289,6 @@ When running in Openshift, we need to grant the appropriate security context for
    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner:submariner-engine 
    ```
 
-### Flannel Notes
-
-When using flannel CNI, configure reverse path filtering to loose mode on the non-gateway nodes (i.e., worker nodes).
-Also, to preserve the setting, you can add it to /etc/sysctl.conf file
-
-sysctl -w net.ipv4.conf.flannel.1.rp_filter=2
-
 # Building
 
 To build `submariner-engine` and `submariner-route-agent` you can trigger `make`, which will perform a Dapperized build of the components.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -288,7 +288,14 @@ When running in Openshift, we need to grant the appropriate security context for
    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner:submariner-routeagent
    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner:submariner-engine 
    ```
-   
+
+### Flannel Notes
+
+When using flannel CNI, configure reverse path filtering to loose mode on the non-gateway nodes (i.e., worker nodes).
```
mangelajo: I think those notes aren't necessary anymore right? you do it in code. I'm going to update this.

Oats87:
If we are going to do this, we should very clearly call this out in the documentation. I think it’s ok otherwise.

PR #218, opened 2019-11-16
Add second E2E gateway failover scenario

tpantelis:
Two gateway nodes configured with 1 submariner engine replica.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/218)

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -8,18 +8,18 @@ import (
 	"time"
 
 	"github.com/onsi/ginkgo"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/uuid"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/rest"
-
+	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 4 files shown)


tpantelis:
> There are arguably too many implementation details in the test here, but it depends on the implementation of the engine pod anyway, so...

I assume you specifically mean waiting for a new pod and Endpoint instance. I agree that makes it a bit whitebox but I did that to eliminate possible timing issue where, after deleting the pod, we actually run the connectivity test against the previous engine instance before k8s has a chance to react to stop the process/container etc. Perhaps I don't need to check for new  a Endpoint instance but I mainly did that as a sanity check that a new engine instance is running. It also gives it a bit more of a cushion to avoid premature timeout in the connectivity test.

skitt:
> I assume you specifically mean waiting for a new pod and Endpoint instance. I agree that makes it a bit whitebox but I did that to eliminate possible timing issue where, after deleting the pod, we actually run the connectivity test against the previous engine instance before k8s has a chance to react to stop the process/container etc. Perhaps I don't need to check for new a Endpoint instance but I mainly did that as a sanity check that a new engine instance is running. It also gives it a bit more of a cushion to avoid premature timeout in the connectivity test.

Right, that’s what I was referring to, but I don’t think it’s much of a problem really. However the fact that it’s useful for timing doesn’t appear in the code; it might be worth adding a comment to that effect (so that in future, if the details change and someone fixes the test, they don’t spend ages figuring out what the point was).

tpantelis:
> > I assume you specifically mean waiting for a new pod and Endpoint instance. I agree that makes it a bit whitebox but I did that to eliminate possible timing issue where, after deleting the pod, we actually run the connectivity test against the previous engine instance before k8s has a chance to react to stop the process/container etc. Perhaps I don't need to check for new a Endpoint instance but I mainly did that as a sanity check that a new engine instance is running. It also gives it a bit more of a cushion to avoid premature timeout in the connectivity test.
> 
> Right, that’s what I was referring to, but I don’t think it’s much of a problem really. However the fact that it’s useful for timing doesn’t appear in the code; it might be worth adding a comment to that effect (so that in future, if the details change and someone fixes the test, they don’t spend ages figuring out what the point was).

Added comments and rebased.

PR #219, opened 2019-11-17
e2e: add print_logs() and wait_for_subm_pods() helpers

hunchback:


mangelajo:
I have resolved conflict on the patch with master, I hope it's ok, let's see how CI takes it.

Thanks for the contribution @hunchback , and sorry for the delay

scripts/kind-e2e/e2e.sh
```diff
@@ -17,11 +34,7 @@ function kind_clusters() {
             logs[$i]=$(mktemp)
             echo Creating cluster${i}, logging to ${logs[$i]}...
             (
-            if [[ -n ${version} ]]; then
-                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
-            else
-                kind create cluster --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
-            fi
+            kind create cluster --image=kindest/node:v${version} --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
```
skitt: This fails when `version` is set to `latest` (as the default value). Could you change it so that `version` includes the `v` when a version is specified? (Or any other appropriate fix!)

mangelajo: @hunchback can you look at @skitt comment ^

hunchback: @skitt / @mangelajo - I brought back the old code

PR #220, opened 2019-11-18
e2e: docker pull if image not preset

hunchback:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/220)

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -201,18 +201,25 @@ function setup_cluster3_gateway() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


hunchback:
> For the purpose of the E2E tests, we need to use :dev which is what the "build / package" make targets will generate.
> 
> If we want to provide the capability to deploy from "public" images, we need to add that somehow, but this would break the e2e testing (it will work in fact, but will never test the locally built images on CI, which would mean we don't validate our code)

ok - closing

mangelajo:
Thank you Aidan

Enviado desde mi iPhone

PR #221, opened 2019-11-18
e2e: fix kind cluster apiserver port only for kubefed

hunchback:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/221)

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -8,6 +8,7 @@ source $(git rev-parse --show-toplevel)/scripts/lib/debug_functions
 function kind_clusters() {
     status=$1
     version=$2
+    kubefed=$4
     pids=(-1 -1 -1)
     logs=()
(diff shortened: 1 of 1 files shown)


hunchback:
I did this to bypass error `Reason: Get http://<IP>:6443/version: dial tcp <IP>:6443: i/o timeout` which I was getting when running `make ci e2e status=keep`

mangelajo:
This change breaks CI: 

https://travis-ci.com/submariner-io/submariner/builds/136954091#L1738

Can you report a Github issue with the details to reproduce what you were experience?, then we can get @dimaunx to help us with the fix.

dimaunx:
@hunchback how this can reproduced locally? It never happens to me when running make ci e2e status=keep multiple times.

hunchback:
> @hunchback how this can reproduced locally? It never happens to me when running make ci e2e status=keep multiple times.

Was not able to reproduce (after rebasing), thus closing

PR #225, opened 2019-11-21
Fix for make ci e2e recurring runs with helm

dimaunx:
Issue:
Running make ci e2e status=keep multiple times locally.

Currently after the first initial run, the broker cluster has submariner engine installed with helm. The pods are not healthy as we don't tag gateway nodes for broker cluster and not loading the images to cluster1. This results in a check for submariner pods always to be false and forces a recurring installation of the engine with helm. The engine on the broker cluster is needed only for testing purposes so this PR makes the installation occur only once during the broker installation phase on cluster1. 

Error:
Error: a release named submariner already exists.
Run: helm ls --all submariner; to check the status of the release
Or run: helm del --purge submariner; to delete it

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/225)

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -115,6 +115,7 @@ function setup_broker() {
     else
         echo Installing broker on cluster1.
         helm --kube-context cluster1 install submariner-latest/submariner-k8s-broker --name ${SUBMARINER_BROKER_NS} --namespace ${SUBMARINER_BROKER_NS}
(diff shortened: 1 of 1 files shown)


skitt:
Wouldn’t it be more appropriate to fix the broker installation check, instead of installing the engine? So instead of looking for pods, look for the CRDs?

skitt:
> Wouldn’t it be more appropriate to fix the broker installation check, instead of installing the engine? So instead of looking for pods, look for the CRDs?

Oh, right, this is in the context of the three-cluster tests with the engine everywhere, sorry.

PR #227, opened 2019-11-26
Use chroot to verify iptables on the host

sridhargaddam:
On some platforms, its seen that iptables on the host is a symlink to a file
(f.e, /etc/alternatives/iptables) which again is a symlink to another file
(f.e., /usr/sbin/iptables-legacy). On such platforms, with the current code,
when we validate the presence of iptables using "[[ -x /host/usr/sbin/iptables ]]",
it fails since "-x" dereferences the symbolic links and the symlink would point to
a file on the container file system (instead of checking the host filesystem).

This patch uses chroot with the host-filesystem for validating the presence of iptables.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/227)

### DIFF
--- a/package/submariner-route-agent.sh
+++ b/package/submariner-route-agent.sh
@@ -9,10 +9,10 @@ else
     DEBUG="-v=4"
 fi
 
-function verify_iptables_on_host() {
-    chroot /host test -x /usr/sbin/$1 && { echo "0"; return; }
-    chroot /host test -x /sbin/$1 && { echo "1"; return; }
-    echo "-1"
+function find_iptables_on_host() {
+    chroot /host test -x /usr/sbin/$1 && { echo "/usr/sbin"; return; }
+    chroot /host test -x /sbin/$1 && { echo "/sbin"; return; }
+    echo "unknown"
 }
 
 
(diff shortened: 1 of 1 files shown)


sridhargaddam:
Another alternative is to validate the presence of file as follows.
[[ -L /host/usr/sbin/iptables ]] 
OR
[[ -x /host/usr/sbin/iptables || -L /host/usr/sbin/iptables ]]

it works. But I feel that using chroot is a more robust and efficient mechanism.

package/submariner-route-agent.sh
```diff
@@ -9,23 +9,33 @@ else
     DEBUG="-v=4"
 fi
 
+function verify_iptables_on_host() {
+    chroot /host test -x /usr/sbin/$1 && { echo "0"; return; }
+    chroot /host test -x /sbin/$1 && { echo "1"; return; }
```
skitt: Instead of hard-coding return-value-to-paths mappings, why not “return” the path directly?

```
    for file in {/usr,}/sbin/$1; do if [ -x $file ]; then echo $file; return; fi; done
    echo ""
```

then

```
iptables=$(verify_iptables_on_host)

if [ -n "$iptables" ]; then
    ...
else
    # Oops, no iptables
fi

sridhargaddam: Depending on the location of the iptables/iptables-save utility on the host, we have to copy one of the wrapper files (i.e., L30 and L32 below). If we are just checking for [ -n "$iptables" ], how can we distinguish which file to copy?

skitt: Ah, sorry, yes, I missed that.

sridhargaddam: No problem @skitt. I missed a small change and realized that i pushed a wrong file. Fixed it now. Thanks for reviewing.

PR #228, opened 2019-11-26
Allow to create a test env. without running e2e test

roytman:
Execution of the e2e tests takes around 2 minutes, if only the local
testing environment is required, the tests can be skipped.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/228)

### DIFF
--- a/scripts/kind-e2e/README.md
+++ b/scripts/kind-e2e/README.md
@@ -80,12 +80,18 @@ as with any normal k8s cluster.
 as docker images, submariner will be redeployed on the clusters from pushed images and E2E tests will be executed.
 This mode allows the developers to test their local code fast on a very close to real world scenario setup.
 
-**NOTE**: If you only want to create the test  environment without running the e2e tests, you can do it by executing 
+**NOTE**: If you only want to create the test environment without running the e2e tests, you can do it by executing 
 the following command: 
 
 ```bash
-make ci e2e status=create
+make build package e2e status=create
 ```
+or, if you don't want to rebuild your Submariner images, by:
+
+```bash
+make e2e status=create
+```
+
 #### Logging
 
 Providing **logging=true** parameter to **make e2e** command will setup ELK stack on the kind clusters.
--- a/scripts/kind-e2e/README.md
+++ b/scripts/kind-e2e/README.md
@@ -80,6 +80,12 @@ as with any normal k8s cluster.
 as docker images, submariner will be redeployed on the clusters from pushed images and E2E tests will be executed.
 This mode allows the developers to test their local code fast on a very close to real world scenario setup.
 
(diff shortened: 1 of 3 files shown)


mangelajo:
I added a commit to remove the extra whitespace.

mangelajo:
Thanks for the patch @roytman looks very useful

dimaunx:
Very useful @roytman, thank you for the contribution.

roytman:
Hi guys, in original commit, I forgot the `date`  command in the end of environment creation. I used it for time measurement. 
Therefore, I removed it now. Please re-approve the PR.

scripts/kind-e2e/README.md
```diff
@@ -80,6 +80,12 @@ as with any normal k8s cluster.
 as docker images, submariner will be redeployed on the clusters from pushed images and E2E tests will be executed.
 This mode allows the developers to test their local code fast on a very close to real world scenario setup.
 
+**NOTE**: If you only want to create the test environment without running the e2e tests, you can do it by executing 
+the following command: 
+
+```bash
+make ci e2e status=create
```
mpeterson: `ci` is not needed.

scripts/kind-e2e/e2e.sh
```diff
@@ -477,9 +476,12 @@ elif [[ $5 = helm ]]; then
 fi
 
 test_connection
-test_with_e2e_tests
 
-if [[ $1 = keep ]]; then
+if [[ $1 != create ]]; then
```
mpeterson: I find this if clause weird. Why wouldn't you check if it's `keep` which is the case where it would be run?

roytman: Done

scripts/kind-e2e/README.md
roytman: If Submariner images do not exist, the script (w/o ci) will fail.
I added both options:

- `make build package e2e status=create` (or should it be `make ci e2e status=create` ? the overhead is small, but the command is shorter) 
- and  `make e2e status=create`
    
Thoughts ? 
BTW, the `status=create` argument skips the e2e tests, but does run `test_connection`, which takes around 100  sec.  I assume that it is better to spend this time, but be sure that the deployment is configured correctly.

mpeterson: right. I forgot that the `ci` thing was overseen. That's something that should be fixed (not in this patch). i.e. `e2e` should work without requiring `ci`.

roytman:
Hi guys, I don't see why the "reviewable" check failed, can you help me ?

mangelajo:
Ignore the reviewable CI, it's because I disabled reporting when this PR already had it, I will merge now. Thanks a lot for the PR.

PR #232, opened 2019-11-29
Use a common iptables wrapper

skitt:
This avoids having path-specific wrappers. The patch also documents
why we can’t use the Kubernetes approach.


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/232)

PR #233, opened 2019-12-01
Implement globalnet ingress and egress features

sridhargaddam:
This patch implements the following
1. Spawns Globalnet Controller on the GatewayNode
2. Initializes Globalnet Controller only when the Cluster uses kubeproxy
3. Programs necessary egress rules for PODs
4. Programs necessary ingress rules for ClusterIPServices

### DIFF
--- a/pkg/globalnet/controllers/ipam/constants.go
+++ b/pkg/globalnet/controllers/ipam/constants.go
@@ -2,5 +2,25 @@ package ipam
 
 import "time"
 
-const handlerResync = time.Hour * 24
-const submarinerIpamGlobalIp = "submariner.io/globalIp"
+type Operation string
+
+const (
+	handlerResync          = time.Hour * 24
+	submarinerIpamGlobalIp = "submariner.io/globalIp"
+	submarinerGlobalNet    = "SUBMARINER-GLOBALNET"
+
+	// Currently Submariner Globalnet implementation (for services) works with kube-proxy
+	// and uses iptable chain-names programmed by kube-proxy. If the internal implementation
+	// of kube-proxy changes, globalnet needs to be modified accordingly.
+	// Reference: https://bit.ly/2OPhlwk
+	kubeProxyServiceChainPrefix = "KUBE-SVC-"
+	kubeProxyNameSpace          = "kube-system"
+	kubeProxyLabelSelector      = "k8s-app=kube-proxy"
+
+	AddRules    = true
+	DeleteRules = false
+
+	Process = "Process"
+	Ignore  = "Ignore"
+	Requeue = "Requeue"
+)
--- a/pkg/globalnet/controllers/ipam/globalnet.go
+++ b/pkg/globalnet/controllers/ipam/globalnet.go
@@ -1,0 +1,47 @@
+package ipam
+
+import (
+	k8sv1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+func (i *Controller) syncPodRules(podIP, globalIP string, addRules bool) {
+	err := i.updateEgressRulesForPod(podIP, globalIP, addRules)
+	if err != nil {
+		klog.Errorf("Error updating egress rules for pod %s: %v", podIP, err)
+		return
+	}
+}
+
+func (i *Controller) syncServiceRules(service *k8sv1.Service, globalIP string, addRules bool) {
+	chainName := i.kubeProxyClusterIpServiceChainName(service)
+	err := i.updateIngressRulesForService(globalIP, chainName, addRules)
+	if err != nil {
+		klog.Errorf("Error updating ingress rules for service %#v: %v", service, err)
+		return
+	}
+}
+
+func (i *Controller) evaluateService(service *k8sv1.Service) Operation {
+	if service.Spec.Type != v1.ServiceTypeClusterIP {
+		// Normally ClusterIPServices can be accessed only within the local cluster.
+		// When multiple K8s clusters are connected via Submariner, it enables access
+		// to ClusterIPService even from remote clusters. So, as part of Submariner
+		// Globalnet implementation, we are only interested in ClusterIP Services and
+		// not the other types of Services like LoadBalancer Services, NodePort Services
+		// etc which are externally accessible.
+		return Ignore
+	}
+
+	if len(service.Spec.Selector) != 0 {
+		chainName := i.kubeProxyClusterIpServiceChainName(service)
+		if chainExists, _ := i.doesIPTablesChainExist("nat", chainName); !chainExists {
+			return Requeue
+		}
+	} else {
+		// Ignore services that do not have selectors
+		return Ignore
+	}
+	return Process
+}
--- a/pkg/globalnet/controllers/ipam/globalnet_egress.go
+++ b/pkg/globalnet/controllers/ipam/globalnet_egress.go
@@ -1,0 +1,26 @@
+package ipam
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/routeagent/controllers/route"
+)
+
+func (i *Controller) updateEgressRulesForPod(podIP, globalIP string, addRules bool) error {
+	ruleSpec := []string{"-p", "all", "-s", podIP, "-j", "SNAT", "--to", globalIP}
+	if addRules {
+		klog.V(4).Infof("Installing iptable egress rules for pod: %s", strings.Join(ruleSpec, " "))
+		if err := i.ipt.AppendUnique("nat", route.SmPostRoutingChain, ruleSpec...); err != nil {
+			return fmt.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+	} else {
+		klog.V(4).Infof("Deleting iptable egress rules for pod: %s", strings.Join(ruleSpec, " "))
+		if err := i.ipt.Delete("nat", route.SmPostRoutingChain, ruleSpec...); err != nil {
+			return fmt.Errorf("error deleting iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+	}
+	return nil
+}
--- a/pkg/globalnet/controllers/ipam/globalnet_ingress.go
+++ b/pkg/globalnet/controllers/ipam/globalnet_ingress.go
@@ -1,0 +1,69 @@
+package ipam
+
+import (
+	"crypto/sha256"
+	"encoding/base32"
+	"fmt"
+	"strings"
+
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/util"
+)
+
+func (i *Controller) initIPTableChains() error {
+	klog.V(4).Infof("Install/ensure %s chain exists", submarinerGlobalNet)
+	if err := util.CreateChainIfNotExists(i.ipt, "nat", submarinerGlobalNet); err != nil {
+		return fmt.Errorf("error creating iptables chain %s: %v", submarinerGlobalNet, err)
+	}
+
+	forwardToSubGlobalNetRuleSpec := []string{"-j", submarinerGlobalNet}
+	if err := util.PrependUnique(i.ipt, "nat", "PREROUTING", forwardToSubGlobalNetRuleSpec); err != nil {
+		klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(forwardToSubGlobalNetRuleSpec, " "), err)
+	}
+	return nil
+}
+
+func (i *Controller) updateIngressRulesForService(globalIP, chainName string, addRules bool) error {
+	ruleSpec := []string{"-d", globalIP, "-j", chainName}
+	if addRules {
+		klog.V(4).Infof("Installing iptables rule for Service %s", strings.Join(ruleSpec, " "))
+		if err := i.ipt.AppendUnique("nat", submarinerGlobalNet, ruleSpec...); err != nil {
+			return fmt.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+	} else {
+		klog.V(4).Infof("Deleting iptable ingress rule for Service: %s", strings.Join(ruleSpec, " "))
+		if err := i.ipt.Delete("nat", submarinerGlobalNet, ruleSpec...); err != nil {
+			return fmt.Errorf("error deleting iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+	}
+	return nil
+}
+
+func (i *Controller) kubeProxyClusterIpServiceChainName(service *k8sv1.Service) string {
+	// CNIs that use kube-proxy with iptables for loadbalancing create an iptables chain for each service
+	// and incoming traffic to the clusterIP Service is directed into the respective chain.
+	// Reference: https://bit.ly/2OPhlwk
+	serviceName := service.GetNamespace() + "/" + service.GetName() + ":"
+	protocol := strings.ToLower(string(service.Spec.Ports[0].Protocol))
+	hash := sha256.Sum256([]byte(serviceName + protocol))
+	encoded := base32.StdEncoding.EncodeToString(hash[:])
+	return kubeProxyServiceChainPrefix + encoded[:16]
+}
+
+func (i *Controller) doesIPTablesChainExist(table, chain string) (bool, error) {
+	existingChains, err := i.ipt.ListChains(table)
+	if err != nil {
+		klog.V(4).Infof("Error listing iptables chains in %s table: %s", table, err)
+		return false, err
+	}
+
+	for _, val := range existingChains {
+		if val == chain {
+			klog.V(4).Infof("%s chain exists in %s table", chain, table)
+			return true, nil
+		}
+	}
+	return false, nil
+}
--- a/pkg/globalnet/controllers/ipam/ipam.go
+++ b/pkg/globalnet/controllers/ipam/ipam.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/coreos/go-iptables/iptables"
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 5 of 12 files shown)


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/233)

mangelajo:
Ignore the merge-blocked X , reviewable-- :-/

pkg/globalnet/controllers/ipam/constants.go
```diff
@@ -4,3 +4,16 @@ import "time"
 
 const handlerResync = time.Hour * 24
 const submarinerIpamGlobalIp = "submariner.io/globalIp"
+const submarinerGlobalNet = "SUBMARINER-GLOBALNET"
+const kubeProxyServiceChainPrefix = "KUBE-SVC-"
```
vthapar: nit: would prefer a newline above this.

pkg/globalnet/controllers/ipam/globalnet_egress.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (i *Controller) updateEgressRulesForPod(podIP, globalIP string, addOrDelete Operation) error {
+	ipt, err := iptables.New()
+	if err != nil {
+		return fmt.Errorf("error initializing iptables: %v", err)
+	}
+
+	ruleSpec := []string{"-p", "all", "-s", podIP, "-j", "SNAT", "--to", globalIP}
+	if addOrDelete == AddRules {
+		klog.V(4).Infof("Installing iptable egress rules for pod: %s", strings.Join(ruleSpec, " "))
```
vthapar: should this be V(6)? 4 might fill up logs pretty quick. Applies to all klog.V(4) logs. Can revisit later though.

pkg/globalnet/controllers/ipam/globalnet_egress.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-iptables/iptables"
+	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/routeagent/controllers/route"
+)
+
```
vthapar: Would be good to have a comment on what sort of rules we add at egress.

pkg/globalnet/controllers/ipam/globalnet_ingress.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-iptables/iptables"
+	"github.com/submariner-io/submariner/pkg/util"
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+func (i *Controller) createNecessaryIPTableChains() error {
```
vthapar: nit: rename to initIPTableChains() ?

pkg/globalnet/controllers/ipam/globalnet_ingress.go
```diff
@@ -0,0 +1,82 @@
...
+		if err = ipt.Delete("nat", submarinerGlobalNet, ruleSpec...); err != nil {
+			return fmt.Errorf("error deleting iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		}
+	}
+	return nil
+}
+
+func (i *Controller) kubeProxyClusterIpServiceChainName(service *k8sv1.Service) string {
+	// CNIs that use kube-proxy with iptables for loadbalancing create an iptables chain for each service
+	// and incoming traffic to the clusterIP Service is directed into the respective chain.
```
vthapar: Would be good to add link to source from which this method is taken. Will let us track any changes to original method.

pkg/globalnet/controllers/ipam/globalnet_ingress.go
```diff
@@ -0,0 +1,82 @@
...
+	serviceName := service.GetNamespace() + "/" + service.GetName() + ":"
+	protocol := strings.ToLower(string(service.Spec.Ports[0].Protocol))
+	hash := sha256.Sum256([]byte(serviceName + protocol))
+	encoded := base32.StdEncoding.EncodeToString(hash[:])
+	return kubeProxyServiceChainPrefix + encoded[:16]
+}
+
+func (i *Controller) doesIptablesChainExist(table, chain string) error {
```
vthapar: nit: Use consistent case for Iptables in all code. Earlier it was IPTables

vthapar: Why not use bool and check that? Chain not existing is different error than initializing tables.

pkg/globalnet/controllers/ipam/constants.go
```diff
@@ -4,3 +4,16 @@ import "time"
...
+const kubeProxyLabelSelector = "k8s-app=kube-proxy"
+
+const AddRules = "Add"
+const DeleteRules = "Delete"
+
+const Process = "Process"
+const Ignore = "Ignore"
+const Requeue = "Requeue"
+
+type Operation string
```
vthapar: won't an enum type work better and avoid string comparisons?

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -109,6 +125,34 @@ func (i *Controller) processNextObject(objWorkqueue workqueue.RateLimitingInterf
 			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
 		}
 
+		runtimeObj, _ := objGetter(ns, name)
+		switch runtimeObj.(type) {
+		case *k8sv1.Service:
+			service := runtimeObj.(*k8sv1.Service)
+			switch i.processServiceStatus(service) {
+			case Ignore:
```
vthapar: We can return without annotating globalIp to this service. In other words, we're adding a new condition for services that won't get globalIP.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -109,6 +125,34 @@ func (i *Controller) processNextObject(objWorkqueue workqueue.RateLimitingInterf
 			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
 		}
 
+		runtimeObj, _ := objGetter(ns, name)
```
vthapar: Can this logic be moved inside the retry loop? Will save us a Get call.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -190,6 +234,29 @@ func (i *Controller) handleUpdatePod(old interface{}, newObj interface{}) {
...
 		return
 	}
+
+	oldPodIp := old.(*k8sv1.Pod).Status.PodIP
+	updatedPodIp := newObj.(*k8sv1.Pod).Status.PodIP
+	podHostIP := newObj.(*k8sv1.Pod).Status.HostIP
+	// Pod events that are skipped during addEvent are handled here when they are assigned an ipaddress.
+	if updatedPodIp != "" && oldPodIp != updatedPodIp {
+		klog.V(4).Infof("In handleUpdatePod, pod %s is now assigned %s address, enqueing", old.(*k8sv1.Pod).Name, updatedPodIp)
+		i.enqueueObject(newObj, i.podWorkqueue)
```
vthapar: you can just use the key which we already have and add it directly without needing to call enqueueObject again. will also avoid the log msg in enqueueObject.

pkg/globalnet/controllers/ipam/constants.go
sridhargaddam: Done

pkg/globalnet/controllers/ipam/globalnet_ingress.go
vthapar: Why not boolean and error? As of now we're not really checking the error reason right? With current code I think it doesn't matter.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: No, I meant add it to queue. Yeah, just realized that we can simplify it. After refactoring I did, we don't really need to call it. Maybe can add enqueueKey method for such scenarios.

pkg/globalnet/controllers/ipam/constants.go
```diff
@@ -4,3 +4,17 @@ import "time"
 
 const handlerResync = time.Hour * 24
 const submarinerIpamGlobalIp = "submariner.io/globalIp"
+const submarinerGlobalNet = "SUBMARINER-GLOBALNET"
+
+const kubeProxyServiceChainPrefix = "KUBE-SVC-"
```
mpeterson: explain that this comes from k8s and link to the origin (to make easier to fix if something breaks because they changed something)

pkg/globalnet/controllers/ipam/globalnet_egress.go
```diff
@@ -0,0 +1,32 @@
...
+func (i *Controller) updateEgressRulesForPod(podIP, globalIP string, addOrDelete Operation) error {
+	ipt, err := iptables.New()
+	if err != nil {
+		return fmt.Errorf("error initializing iptables: %v", err)
+	}
+
+	ruleSpec := []string{"-p", "all", "-s", podIP, "-j", "SNAT", "--to", globalIP}
+	if addOrDelete == AddRules {
```
mpeterson: wouldn't a switch be preferable? (same for other instances)

mpeterson: perhaps even separate the add and the delete into two methods

pkg/globalnet/controllers/ipam/globalnet_ingress.go
```diff
@@ -0,0 +1,83 @@
...
+	"strings"
+
+	"github.com/coreos/go-iptables/iptables"
+	"github.com/submariner-io/submariner/pkg/util"
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+func (i *Controller) initIPTableChains() error {
+	ipt, err := iptables.New()
```
mpeterson: if you will call this function anyways, you might as well save the iptables object returned and avoid the initialization on all other places.

mpeterson: after reading more code I see this is not the first creation of the iptables object. The comment still applies, just put it on the outermost function

pkg/globalnet/controllers/ipam/constants.go
```diff
@@ -4,3 +4,16 @@ import "time"
```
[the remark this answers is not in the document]
sridhargaddam: Done, fixed it.

pkg/globalnet/controllers/ipam/globalnet.go
```diff
@@ -0,0 +1,47 @@
...
+		klog.Errorf("updateEgressRulesForPod returned error. %v", err)
+		return
+	}
+}
+
+func (i *Controller) syncServiceRules(service *k8sv1.Service, globalIP string, addOrDelete Operation) {
+	chainName := i.kubeProxyClusterIpServiceChainName(service)
+	err := i.updateIngressRulesForService(globalIP, chainName, addOrDelete)
+	if err != nil {
+		klog.Errorf("updateIngressRulesForService returned error. %v", err)
```
tpantelis:
```suggestion
		klog.Errorf("Error updating ingress rules for service %#v: %v", service, err)
```

pkg/globalnet/controllers/ipam/globalnet.go
```diff
@@ -0,0 +1,47 @@
+package ipam
+
+import (
+	k8sv1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+func (i *Controller) syncPodRules(podIP, globalIP string, addOrDelete Operation) {
```
tpantelis: Since the options are only add or delete, I think it makes sense to use a `bool` here instead of `Operation`. Same with `syncServiceRules`.

pkg/globalnet/controllers/ipam/globalnet.go
```diff
@@ -0,0 +1,47 @@
...
+func (i *Controller) syncServiceRules(service *k8sv1.Service, globalIP string, addOrDelete Operation) {
+	chainName := i.kubeProxyClusterIpServiceChainName(service)
+	err := i.updateIngressRulesForService(globalIP, chainName, addOrDelete)
+	if err != nil {
+		klog.Errorf("updateIngressRulesForService returned error. %v", err)
+		return
+	}
+}
+
+func (i *Controller) processServiceStatus(service *k8sv1.Service) Operation {
```
tpantelis:
```suggestion
func (i *Controller) evaluateService(service *k8sv1.Service) Operation {
```

pkg/globalnet/controllers/ipam/globalnet_egress.go
tpantelis: I suggest to use a `bool` for `addOrDelete` since the other `Operation` enums don't apply here.

pkg/globalnet/controllers/ipam/globalnet_ingress.go
tpantelis: > Why not boolean and error? As of now we're not really checking the error reason right? With current code I think it doesn't matter.

I agree.  I think a function that starts with _does_ implies a boolean answer. Non-existence isn't an error condition in this case plus makes it difficult for callers to distinguish non-existence and an actual error.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -124,6 +147,34 @@ func (i *Controller) processNextObject(objWorkqueue workqueue.RateLimitingInterf
...
+				pod := runtimeObj.(*k8sv1.Pod)
+				// Process pod event only when it has an ipaddress. Pods skipped here will be handled subsequently
+				// during podUpdate event when an ipaddress is assigned to it.
+				if pod.Status.PodIP == "" {
+					objWorkqueue.Forget(obj)
+					return nil
+				}
+
+				// Privileged pods that use hostNetwork will be ignored.
+				if pod.Status.PodIP != "" && pod.Status.PodIP == pod.Status.HostIP {
```
tpantelis: nit: no need to check `pod.Status.PodIP != ""` here since that's done above.

pkg/globalnet/controllers/ipam/globalnet.go
sridhargaddam: Done

pkg/globalnet/controllers/ipam/globalnet.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	k8sv1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+func (i *Controller) syncPodRules(podIP, globalIP string, addOrDelete Operation) {
+	err := i.updateEgressRulesForPod(podIP, globalIP, addOrDelete)
+	if err != nil {
+		klog.Errorf("updateEgressRulesForPod returned error. %v", err)
```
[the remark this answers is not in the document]
sridhargaddam: Done

pkg/globalnet/controllers/ipam/globalnet.go
sridhargaddam: Done

pkg/globalnet/controllers/ipam/globalnet.go
sridhargaddam: Done

pkg/globalnet/controllers/ipam/globalnet_egress.go
sridhargaddam: Ack, addOrDelete modified to bool

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -190,6 +241,29 @@ func (i *Controller) handleUpdatePod(old interface{}, newObj interface{}) {
...
+	if updatedPodIp != "" && oldPodIp != updatedPodIp {
+		klog.V(4).Infof("In handleUpdatePod, pod %s is now assigned %s address, enqueing", old.(*k8sv1.Pod).Name, updatedPodIp)
+		i.enqueueObject(newObj, i.podWorkqueue)
+		return
+	}
+
+	// When the POD is getting terminated, sometimes we get pod update event with podIp removed.
+	if oldPodIp != "" && updatedPodIp == "" {
+		klog.V(4).Infof("Pod %s with ip %s is being terminated", old.(*k8sv1.Pod).Name, oldPodIp)
+		i.handleRemovedPod(old)
```
tpantelis: `handleRemovedPod` expects the object's key. I'd suggest to change `handleRemovedPod` to take a `*k8sv1.Pod` factor out `extractPodFromKey` function.

pkg/util/util.go
```diff
@@ -173,3 +175,72 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
...
+		klog.V(4).Infof("In %s table, iptables rule \"%s\", already exists.", table, strings.Join(ruleSpec, " "))
+		return nil
+	} else {
+		if err = ipt.Insert(table, chain, 1, ruleSpec...); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
tpantelis: Based on best practices outlined [here](https://github.com/golang/go/wiki/CodeReviewComments#package-names), I suggest putting these functions in a separate _iptables_ package and file.

pkg/globalnet/controllers/ipam/ipam.go
sridhargaddam: Done

pkg/util/util.go
sridhargaddam: We are already using an external iptables package (github.com/coreos/go-iptables/iptables). These APIs are more like utility APIs that internally use the actual iptables package. Creating an iptables package with these APIs IMO might be confusing and shadowing the actual iptables package. WDYS?

tpantelis: I understand. Not a big deal - it's fine.

pkg/globalnet/controllers/ipam/globalnet_ingress.go
sridhargaddam: Done

pkg/globalnet/controllers/ipam/globalnet_ingress.go
```diff
@@ -0,0 +1,69 @@
...
+		return false, err
+	}
+
+	for _, val := range existingChains {
+		if val == chain {
+			klog.V(4).Infof("%s chain exists in %s table", chain, table)
+			return true, nil
+		}
+	}
+	return false, fmt.Errorf("chain %s does not exist", chain)
```
tpantelis: I don't think it should return an error in this case. Returning an error should indicate that it couldn't definitively determine whether or not the chain exists, which isn't the case here.

sridhargaddam: Good point @tpantelis .

PR #236, opened 2019-12-02
Fix build issue when re-running the build in an existing setup

sridhargaddam:
In an existing Submariner (with globalnet) deployment, when we re-run
the command "make build package e2e status=keep", the following error
is seen and this patch fixes it.

kubectl --context=cluster2 -n submariner create serviceaccount submariner-globalnet
Error from server (AlreadyExists): serviceaccounts "submariner-globalnet" already exists
[submariner]$ ./scripts/e2e keep 1.14.2 false false helm false
FATA[0224] exit status 1
make: *** [Makefile:18: e2e] Error 1

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -327,6 +327,12 @@ function test_with_e2e_tests {
 
 function deploy_globalnet {
     for i in 2 3; do
+      if kubectl --context=cluster${i} -n $subm_ns get serviceaccount submariner-globalnet > /dev/null 2>&1; then
+        echo submariner-globalnet already exists in cluster${i}, deleting the existing deployment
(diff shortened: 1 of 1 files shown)


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/236)

scripts/kind-e2e/e2e.sh
```diff
@@ -327,6 +327,12 @@ function test_with_e2e_tests {
 
 function deploy_globalnet {
     for i in 2 3; do
+      if kubectl --context=cluster${i} -n $subm_ns get serviceaccount submariner-globalnet > /dev/null 2>&1; then
```
vthapar: shouldn't an existing deployment be caught at much earlier point? Why not only create this if it doesn't already exist? something like...

if ! kubectl --context=cluster${i} get serviceaccount submariner-globalnet; then

sridhargaddam: Sorry, not able to understand your comment. 

Even if the deployment already exists, we still want to clear the existing deployment and redeploy with the updated/new config files/Image, isnt it?

PR #237, opened 2019-12-02
Fix pod scheduling in E2E part 2

tpantelis:
The NodeAffinity for non-gateway scheduling is incorrect - it has
2 terms that need to be ORed together, however grouping 2
MatchExpressions ANDs them. We need to group 2 NodeSelectorTerms
as they are ORed.


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/237)

tpantelis:
Merged as part of https://github.com/submariner-io/submariner/pull/218.

PR #238, opened 2019-12-02
E2E tests not running by default

tpantelis:
PR https://github.com/submariner-io/submariner/pull/228/ modified e2e.sh so
it only runs the E2E tests if 'status=keep' and thus the tests no longer run
with 'status=onetime', which is the default via 'make ci e2e' as documented.
Therefore modified e2e.sh to also check for 'status=onetime'.


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/238)

PR #247, opened 2019-12-07
E2E: delete service if already exists

tpantelis:
If you try to run a second service connectivity test in a spec it will
fail due to the service already existing. Modified CreateTCPService so if
it fails with IsAlreadyExists error, delete the service and try again.


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/247)

PR #250, opened 2019-12-11
IpPool.IsAvailable should lock

tpantelis:
`available` shouldn't be accessed unprotected.


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/250)

PR #251, opened 2019-12-11
Remove IsFull check in IpPool.Allocate

tpantelis:
This function is coded to return the existing allocated IP, if any, however the
IsFull check elides that if there's no available IPs. I assume this isn't
the intent so remove the up-front check as this case is handled anyway by the
subsequent code.

### DIFF
--- a/pkg/globalnet/controllers/ipam/ippool.go
+++ b/pkg/globalnet/controllers/ipam/ippool.go
@@ -54,9 +54,6 @@ func intToIP(ip int) net.IP {
 }
 
 func (p *IpPool) Allocate(key string) (string, error) {
-	if p.IsFull() {
-		return "", errors.New("IPAM: No IP available for allocation")
-	}
 	p.Lock()
 	defer p.Unlock()
 	allocatedIp := p.allocated[key]
@@ -66,7 +63,7 @@ func (p *IpPool) Allocate(key string) (string, error) {
 			delete(p.available, k)
 			return k, nil
 		}
-		return "", errors.New("IPAM: Unable to allocate IP")
+		return "", errors.New("IPAM: No IP available for allocation")
 	}
 	return allocatedIp, nil
 }
@@ -91,14 +88,6 @@ func (p *IpPool) GetAllocatedIp(key string) string {
 	ip := p.allocated[key]
 	p.RUnlock()
 	return ip
-}
-
-func (p *IpPool) IsFull() bool {
-	var result bool
-	p.RLock()
-	result = len(p.available) == 0
-	p.RUnlock()
-	return result
 }
 
 func (p *IpPool) RequestIp(key string, ip string) (string, error) {


mangelajo:
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner/251)

pkg/globalnet/controllers/ipam/ippool.go
```diff
@@ -54,9 +54,6 @@ func intToIP(ip int) net.IP {
 }
 
 func (p *IpPool) Allocate(key string) (string, error) {
-	if p.IsFull() {
```
sridhargaddam: I think this API is called only when the globalIP is not allocated for the key.

tpantelis: I'm not sure of it's intended usage/contract as there's no docs but the code does check if an IP is already allocated so it seems to make sense to do that in all cases. Or perhaps it shouldn't check if already allocated if that's not the intended usage. Also L61-66 already handle the case where it's full.

vthapar: Logic wise, yes this is not needed coz L61-66 will handle it. IsFull() uses an RLock so is a quicker check if already is full than L61-66 which uses Lock(). This allows a quicker return as we won't be locked if someone else is reading the cache. We still need to handle it later, coz it is possible that by the time we get to actual allocation IP is gone. Not sure if the earlier return adds value over extra check.

tpantelis: > Logic wise, yes this is not needed coz L61-66 will handle it. IsFull() uses an RLock so is a quicker check if already is full than L61-66 which uses Lock(). This allows a quicker return as we won't be locked if someone else is reading the cache. We still need to handle it later, coz it is possible that by the time we get to actual allocation IP is gone. Not sure if the earlier return adds value over extra check.

I don't think it does b/c the case where it's full will likely not be the norm, thus the fast path will incur 2 lock/unlock sequences. Also the earlier check elides the case where the IP is already allocated as I mentioned earlier.

mkolesnik: It's a matter of how this is going to be used and scale testing.. Your upfront check assumes that this will be called so frequently that there's a need to locally lock the pool (which is local to the process and thus could, in theory, run in multiple processes which this lock won't be useful for).
I agree with Tom that since this is already covered by a broader "Write" lock anyway, this optimization seems premature and unnecessary at this time, unless you have faced an actual performance issue that necessitates it.

tpantelis: Right but outside of that, it elides the subsequent check for existence so if it's called when an IP is already allocated and it's not full, the call will successfully return the IP. But if it happens to be full, it will return an error - that's inconsistent and not correct IMO.

vthapar: Agree. We can revisit this once we start with perf and scale testing. As Tom pointed out, this does introduce an inconsistency in behavior which we can address when and if we need to handle the perf issue.

Issue #257, opened 2019-12-17
Initial connection fails randomly, including when we add a cluster


mangelajo:
There is some "warm up" or "caching" that needs to happen somewhere on the
stack that will make the first connection very slow or fail.

By warming up with test_connection which retries, we force that "caching" to happen
before "E2E", we should investigate the root cause and fix it.

sridhargaddam:
@mangelajo can you share the steps to reproduce this issue. You mentioned that we can simulate it using e2e tests by commenting out something.

mangelajo:
I'm not being able to reproduce with those changes, I wonder if installing with kindnet and not weave-net will trigger it.

sridhargaddam:
I tried reproducing this issue both in KIND based deployment as well as between a Rancher and AWS(OCP) clusters. Sadly, in both the setups, I was not able to reproduce the issue.

For the rancher+AWS(OCP) clusters, I've used subctl commands for connecting the clusters as shown below.
rm ./broker-info.subm
./subctl deploy-broker --kubeconfig auth/kubeconfig --no-dataplane
./subctl join --kubeconfig auth/kubeconfig ./broker-info.subm
./subctl join --kubeconfig auth/rancher-cluster-kubeconfig ./broker-info.subm

Observations1:
1. Let's say that we have two (aka Rancher and AWS) Clusters deployed and running. 
2. The clusters are not connected.
3. We have two PODs, POD1 on Rancher Cluster and POD2 on AWS Cluster.
4. If we try connecting from POD1 to POD2, connectivity fails as expected.
5. Now deploy Submariner on both the Clusters to connect them.
6. Monitor the amount of time it takes for POD1 to talk to POD2. 

I've seen that it usually takes around 8 to 12 secs which IMO is a reasonable time for establishing connectivity between the clusters (a lot of things happen behind the scenes like - a leader is chosen in the cluster, endpoint/cluster CRDs are updated on the Broker, CRDs are then sync'ed on the local clusters from Broker, IPSEC connectivity and routing rules are configured, etc)

Observations2:
1. Let's say that we have two (aka Rancher and AWS) Clusters deployed and running. 
2. We deploy Submariner on both the Clusters to connect them.
3. Deploy POD1 (on Rancher Cluster) and POD2 (on AWS Cluster)
4. Monitor the amount of time it takes for POD1 to talk to POD2. 

[...]

PR #262, opened 2019-12-19
Don't test connection before E2E tests

mangelajo:
This is related to #257, there is some "warm up" or "caching" that
needs to happen somewhere on the stack that will make the first connection
very slow.

By warming up with test_connection which retries, we force that "caching" to happen
before "E2E"


mkolesnik:
Since this all moved to shipyard closing this one, if it's still relevant lets open an issue and think how best to solve it

PR #265, opened 2019-12-19
Enhancements to Submariner Globalnet Controller

sridhargaddam:
This patch implements the following features

1. Modifies the Globalnet controller to a DaemonSet with
   nodeSelector so that it gets scheduled on all the nodes
   labelled with submariner.io/gateway: "true"
2. Initializes ipamController only on the active GatewayNode
   of the Cluster. On all the other nodes, we just keep
   monitoring for any leader changes and react accordingly.
3. When a clusterLeader (aka GatewayNode) migrates to a new
   node, ipamController running on that node will be stopped
   and all the existing globalnet ingress/egress rules are
   deleted. On the new clusterLeader, ipamController will
   be initialized and it reprograms the necessary ingress,
   egress rules.
4. Inorder to support the above use-cases and to optimize
   the flow, globalnet ingress and egress rules are now
   programmed in their respective chains.

### DIFF
--- a/pkg/globalnet/controllers/ipam/ipam.go
+++ b/pkg/globalnet/controllers/ipam/ipam.go
@@ -2,7 +2,6 @@ package ipam
 
 import (
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/coreos/go-iptables/iptables"
@@ -68,18 +67,10 @@ func NewController(spec *SubmarinerIpamControllerSpecification, config *Informer
 }
 
 func (i *Controller) Run(stopCh <-chan struct{}) error {
-	var wg sync.WaitGroup
-	wg.Add(1)
 	defer utilruntime.HandleCrash()
 
 	// Start the informer factories to begin populating the informer caches
 	klog.Info("Starting IPAM Controller")
-
-	// Wait for the caches to be synced before starting workers
-	klog.Info("Waiting for informer caches to sync")
-	if ok := cache.WaitForCacheSync(stopCh, i.servicesSynced, i.podsSynced); !ok {
-		return fmt.Errorf("failed to wait for caches to sync")
-	}
 
 	// Query kube-proxy pods in the cluster
 	kubeProxyPodList, err := i.kubeClientSet.CoreV1().Pods(kubeProxyNameSpace).List(metav1.ListOptions{LabelSelector: kubeProxyLabelSelector})
@@ -97,13 +88,18 @@ func (i *Controller) Run(stopCh <-chan struct{}) error {
 		return fmt.Errorf("initIPTableChains returned error. %v", err)
 	}
 
+	// Wait for the caches to be synced before starting workers
+	klog.Info("Waiting for informer caches to sync")
+	if ok := cache.WaitForCacheSync(stopCh, i.servicesSynced, i.podsSynced); !ok {
+		return fmt.Errorf("failed to wait for caches to sync")
+	}
+
 	klog.Info("Starting workers")
 	go wait.Until(i.runServiceWorker, time.Second, stopCh)
 	go wait.Until(i.runPodWorker, time.Second, stopCh)
-	wg.Wait()
 	<-stopCh
 	klog.Info("Shutting down workers")
-
+	i.cleanupIPTableRules()
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -335,13 +343,10 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
 	}
 }
 
-func (i *Controller) annotateGlobalIp(key string, annotations map[string]string) (map[string]string, error) {
-	if annotations == nil {
-		annotations = map[string]string{}
-	}
+func (i *Controller) annotateGlobalIp(key string, globalIp string) (map[string]string, error) {
+	annotations := map[string]string{}
```
vthapar: Earlier, we were passing existing annotations and initializing this only if they were not present. With this change, if there are any other annotations, those would be lost.

pkg/globalnet/controllers/ipam/ippool.go
```diff
@@ -118,3 +118,15 @@ func (p *IpPool) RequestIp(key string, ip string) (string, error) {
 	// It is neither allocated for this key, nor available, give another.
 	return p.Allocate(key)
 }
+
+func (p *IpPool) ClearAll() {
```
vthapar: Why do we need this? Won't this be garbage collected?

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -97,16 +88,33 @@ func (i *Controller) Run(stopCh <-chan struct{}) error {
...
+func (i *Controller) cleanUp() {
+	klog.Info("In cleanUp, freeing the allocated memory and iptables rules.")
+	// Free the allocated memory and the associated iptables rules programmed on the node.
+	i.cleanupIPTableRules()
+	for k := range i.excludeNamespaces {
+		delete(i.excludeNamespaces, k)
+	}
+	i.pool.ClearAll()
+	i.pool = nil
+	i.ipt = nil
```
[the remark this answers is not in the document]
sridhargaddam: cleanupIPTableRules actually deletes the iptables rules on the host

pkg/globalnet/controllers/ipam/ipam.go
sridhargaddam: Done

pkg/globalnet/controllers/ipam/ippool.go
sridhargaddam: I agree that garbage collector will eventually clean up the memory, but I'm explicitly doing it to be on the safe side instead of leaving it to the Garbage collector. Do you see any harm in this? 

Also, as discussed with Vishal, this cleanup happens on the non-Gateway node, so even if it takes couple of millisecs its not an issue on the non-Gateway node.

pkg/globalnet/controllers/ipam/types.go
```diff
@@ -32,3 +37,15 @@ type Controller struct {
 
 	ipt *iptables.IPTables
 }
+
+type GatewayMonitor struct {
+	clusterID           string
+	KubeClientSet       kubernetes.Interface
```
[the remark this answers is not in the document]
sridhargaddam: Done

pkg/globalnet/controllers/ipam/ipam.go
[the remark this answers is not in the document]
tpantelis: Sorry I meant L110-115 where you clear out the pool/maps. This doesn't hurt but really isn't necessary as the memory will be GCed along with the `Controller` instance once it's reference is released which will happen right after the `Run` function completes after it calls this function.

pkg/globalnet/controllers/ipam/ippool.go
tpantelis: Clearing maps and nulling out variables does not immediately delete/reclaim the underlying memory - it just makes it eligible to be reclaimed by the GC (whenever it decides to) which, of course, will also occur automatically once all incoming references to the parent objects are released.  There's no harm in manually releasing memory that will be automatically reclaimed anyway but it isn't necessary and won't really cause the memory to be reclaimed any faster.  

A `ClearAll` function would be useful if the caller wants to re-init for reuse.

PR #270, opened 2019-12-29
Use Endpoint CRD to configure routes

mpeterson:
Currently the `submariner-agent` makes use of `Cluster` CRDs to set
up the routes. This, however, is information that is more natural to
the `Endpoint` CRD as it describes the subnets that are to be found
when connecting that endpoint. Moreover, having a single source of
truth instead of two simplifies the code and allows for easier
integration with `submariner-globalnet`.

This patch makes the `Endpoint` be the source of truth for setting up
routes. By doing so, there is no need for the `submariner-routeagent` to
monitor `Cluster`s anymore and therefore the code that is related has
been removed, simplifying the codebase in consequence.

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -31,7 +31,6 @@ import (
 type InformerConfigStruct struct {
 	SubmarinerClientSet clientset.Interface
 	ClientSet           kubernetes.Interface
-	ClusterInformer     informers.ClusterInformer
 	EndpointInformer    informers.EndpointInformer
 	PodInformer         podinformer.PodInformer
 }
@@ -42,11 +41,9 @@ type Controller struct {
 
 	submarinerClientSet    clientset.Interface
 	clientSet              kubernetes.Interface
-	clustersSynced         cache.InformerSynced
 	endpointsSynced        cache.InformerSynced
 	smRouteAgentPodsSynced cache.InformerSynced
 
-	clusterWorkqueue  workqueue.RateLimitingInterface
 	endpointWorkqueue workqueue.RateLimitingInterface
 	podWorkqueue      workqueue.RateLimitingInterface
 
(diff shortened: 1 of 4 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -449,9 +390,10 @@ func (r *Controller) processNextEndpoint() bool {
 		}
 
 		if endpoint.Spec.ClusterID != r.clusterID {
-			klog.V(6).Infof("Endpoint didn't match the cluster ID of this cluster")
+			klog.V(6).Infof("setting routes of endpoint object %s", name)
+			r.updateIptableRulesForInterclusterTraffic(endpoint.Spec.Subnets)
```
sridhargaddam: Are we cleaning up the iptables rules when the endpoint is removed?

sridhargaddam:
BTW, @mpeterson did you get a chance to verify the end-to-end use-case with this PR (and https://github.com/submariner-io/submariner/pull/271) patch applied together? 

Note: Overlapping-cidrs branch has the necessary functionality (already merged) that annotates PODs/Services and programs the necessary ingress/egress rules on the gateway nodes.

pkg/routeagent/controllers/route/route.go
```diff
@@ -167,16 +154,16 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
...
+		return fmt.Errorf("error while retrieving all endpoints: %v", err)
 	}
 
 	// Program iptables rules for traffic destined to all the remote cluster CIDRs
-	for _, cluster := range clusters.Items {
-		if cluster.Spec.ClusterID != r.clusterID {
-			r.updateIptableRulesForInterclusterTraffic(append(cluster.Spec.ClusterCIDR, cluster.Spec.ServiceCIDR...))
+	for _, endpoint := range endpoints.Items {
+		if endpoint.Spec.ClusterID != r.clusterID {
+			r.updateIptableRulesForInterclusterTraffic(endpoint.Spec.Subnets)
```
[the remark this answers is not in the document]
mpeterson: What do you mean? The `Endpoint` CRD is already being sync as is because there was no changes to its definition.

pkg/routeagent/controllers/route/route.go
mpeterson: We are not and we were not before either. That's some tech debt that we should tackle outside this work. What do you think?

pkg/routeagent/controllers/route/route.go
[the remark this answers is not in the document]
sridhargaddam: Okay, I see that its already addressed as part of https://github.com/submariner-io/submariner/pull/271/files

pkg/routeagent/controllers/route/route.go
sridhargaddam: Sure

PR #271, opened 2019-12-29
Support Global CIDR in submariner-engine

mpeterson:
The Cluster CRD has now support for a Global CIDR segment. However,
`submariner-engine` is not aware of it and thus it cannot properly
react to it. The `submariner-routeagent` uses the information stored on
each Endpoint CRDs to set routes based on the Subnets each Endpoint acts
as a gateway for. Therefore, `submariner-agent` needs to populate that
information with the correct Subnets. That means current behavior if
the Global CIDR is not present or the Global CIDR otherwise.

This patch allows `submariner-agent` to properly handle this case. For
this it introduces a new environmental variable `SUBMARINER_GLOBALCIDR`
which when present tells `submariner-agent` that the Global CIDR is to
be preferred over the Service and Pod CIDRs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -79,19 +79,27 @@ func main() {
 		submarinerInformers.WithNamespace(submSpec.Namespace))
 
 	start := func(context.Context) {
+		var localSubnets []string
+
 		localCluster, err := util.GetLocalCluster(submSpec)
 		if err != nil {
 			klog.Fatalf("Fatal error occurred while retrieving local cluster from %#v: %v", submSpec, err)
 		}
 
+		if len(submSpec.GlobalCidr) > 0 {
+			localSubnets = submSpec.GlobalCidr
+		} else {
+			localSubnets = append(submSpec.ServiceCidr, submSpec.ClusterCidr...)
+		}
+
 		localEndpoint, err := util.GetLocalEndpoint(submSpec.ClusterID, "ipsec", nil, submSpec.NatEnabled,
-			append(submSpec.ServiceCidr, submSpec.ClusterCidr...), util.GetLocalIP())
+			localSubnets, util.GetLocalIP())
 
 		if err != nil {
 			klog.Fatalf("Fatal error occurred while retrieving local endpoint from %#v: %v", submSpec, err)
 		}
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -79,19 +79,27 @@ func main() {
...
 		localEndpoint, err := util.GetLocalEndpoint(submSpec.ClusterID, "ipsec", nil, submSpec.NatEnabled,
-			append(submSpec.ServiceCidr, submSpec.ClusterCidr...), util.GetLocalIP())
+			subnets, util.GetLocalIP())
 
 		if err != nil {
 			klog.Fatalf("Fatal error occurred while retrieving local endpoint from %#v: %v", submSpec, err)
 		}
 
-		cableEngine, err := ipsec.NewEngine(append(submSpec.ClusterCidr, submSpec.ServiceCidr...), localCluster, localEndpoint)
+		cableEngine, err := ipsec.NewEngine(subnets, localCluster, localEndpoint)
```
sridhargaddam: nit: To be consistent with the other variables, shall we name the variable as localSubnets?

main.go
```diff
@@ -79,19 +79,27 @@ func main() {
...
 
 	start := func(context.Context) {
+		var subnets []string
+
 		localCluster, err := util.GetLocalCluster(submSpec)
 		if err != nil {
 			klog.Fatalf("Fatal error occurred while retrieving local cluster from %#v: %v", submSpec, err)
 		}
 
+		if len(submSpec.GlobalCidr) > 0 {
```
sridhargaddam: Are you planning to propose a patch in helm/operator that supplies the environment variable (SUBMARINER_GLOBALCIDR) during initialization?

f.e., https://github.com/submariner-io/submariner-charts/blob/e1a313755ceba38c35b0da40f3aa63acb28562d4/submariner/templates/engine-deploy.yaml#L59

mpeterson: @mkolesnik is working on that.

main.go
mpeterson: Yeah, sounds okey to me.

Issue #256, opened 2019-12-17
Refactoring E2E sh to use armada and separate helm/subctl


mangelajo:
1) Use armada to deploy kind clusters
2) Separate helm and subctl logic to different "sh" libs that we can import from e2e.sh, the functions to setup stuff would have the same name in the helm and subctl versions... 
3) Remove all the conditionals: we can now call setup_broker, install_subm, setup_xxx_gateway  ....

dfarrell07:
Done:

> 1. Separate helm and subctl logic to different "sh" libs that we can import from e2e.sh, the functions to setup stuff would have the same name in the helm and subctl versions...
> 2. Remove all the conditionals: we can now call setup_broker, install_subm, setup_xxx_gateway ....

https://github.com/submariner-io/submariner/pull/273

Broke this out into a different issue:

> 1. Use armada to deploy kind clusters

https://github.com/submariner-io/submariner/issues/291

PR #273, opened 2020-01-06
Cleanup e2e script, separate subctl/helm logic

dfarrell07:
Relates to #256

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -57,35 +57,6 @@ function kind_clusters() {
     print_logs "${logs[@]}"
 }
 
-function install_helm() {
-    helm init --client-only
-    helm repo add submariner-latest https://submariner-io.github.io/submariner-charts/charts
-    pids=(-1 -1 -1)
-    logs=()
-    for i in 1 2 3; do
-        # Skip other clusters on operator deployment, we only need it on the first
-        if [ "$deploy_operator" = true ]; then
-            echo "Skipping other clusters since we're deploying with operator."
-            break
-        fi
-
-        if kubectl --context=cluster${i} -n kube-system rollout status deploy/tiller-deploy > /dev/null 2>&1; then
-            echo Helm already installed on cluster${i}, skipping helm installation...
-        else
-            logs[$i]=$(mktemp)
-            echo Installing helm on cluster${i}, logging to ${logs[$i]}...
-            (
-            kubectl --context=cluster${i} -n kube-system create serviceaccount tiller
-            kubectl --context=cluster${i} create clusterrolebinding tiller --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
-            helm --kube-context cluster${i} init --service-account tiller
-            kubectl --context=cluster${i} -n kube-system rollout status deploy/tiller-deploy
-            ) > ${logs[$i]} 2>&1 &
-            set pids[$i] = $!
-        fi
-    done
-    print_logs "${logs[@]}"
-}
-
 function setup_custom_cni(){
     declare -A POD_CIDR=( ["cluster2"]="10.245.0.0/16" ["cluster3"]="10.246.0.0/16" )
     for i in 2 3; do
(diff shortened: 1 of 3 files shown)


scripts/kind-e2e/e2e.sh
```diff
@@ -95,117 +66,13 @@ function setup_custom_cni(){
             echo "Applying weave network in to cluster${i}..."
             kubectl --context=cluster${i} apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')&env.IPALLOC_RANGE=${POD_CIDR[cluster${i}]}"
             echo "Waiting for weave-net pods to be ready cluster${i}..."
-            kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=700s
+            kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=360s
```
dfarrell07: I might should do this in a different change. @manosnoam, you were the one asking to make this timeout longer ages ago - do you still think it needs to be so long?

manosnoam:
The timeout in my previous issues, was not the cause of failure apperently.

In general, timeout configuration will only be required when running e2e on
real environment, where the connection takes time... but it's not required
in KIND clusters.

dfarrell07:
The connection tests are failing in CI (x2), but consistently passing locally for me. Could there possibly be something wrong with the CI infra? I see the other most-recent PR, an automated version bump in #282, is also failing for the same reason. /cc @dimaunx @skitt @mangelajo @tpantelis

scripts/kind-e2e/e2e.sh
```diff
@@ -363,84 +228,64 @@ logging=$3
...
 deploy=$5
 
+echo Starting with status: $status, k8s_version: $version, logging: $logging, kubefed: $kubefed, deploy: $deploy
+
 if [[ $status = clean ]]; then
     cleanup
     exit 0
 fi
 if [[ $status != keep && $status != create ]]; then
+    # FIXME: Does not cover status=onetime, which is used below
```
mangelajo: I suspect for onetime we didn't want a cleanup (I don't remember the purpose of onetime though)

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -5,40 +5,90 @@ if [ "${0##*/}" = "lib_operator_deploy_subm.sh" ]; then
...
+    get_subctl
 }
 
-function del_subm_gateway_label() {
-  kubectl label node $context-worker "submariner.io/gateway-" --overwrite
+function setup_broker() {
+    context=$1
+    if kubectl --context=$context get crd clusters.submariner.io > /dev/null 2>&1; then
+        echo Submariner CRDs already exist, skipping broker creation...
+    else
```
mangelajo: I believe the if is not necessary anymore and subctl can be called twice to setup the broker without crash.

scripts/kind-e2e/e2e.sh
tpantelis: onetime is the default, ie run the tests and clean up.

scripts/kind-e2e/lib_helm_deploy_subm.sh
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function deploytool_prereqs() {
+    install_helm
+}
+
+function preinstall_cleanup_subm() {
+    context=$1
+    delete_subm_pods $context submariner
+}
```
tpantelis: This is duplicated in the operator deploy script. Is there a particular reason the base code doesn't directly call `delete_subm_pods $context $sub_ns`?

scripts/kind-e2e/lib_operator_deploy_subm.sh
dfarrell07: Okay, changed in the latest version - thanks!

scripts/kind-e2e/e2e.sh
dfarrell07: Ah, I see what it was meant to to do now, thanks. Refactored to make more clear in the latest version.

scripts/kind-e2e/lib_helm_deploy_subm.sh
dfarrell07: Because the Helm/Operator deploys use different namespaces by default (submariner vs submariner-operator). This design hides that difference behind function calls with the same call signature (so we can do this dual Helm/Operator library import design).

scripts/kind-e2e/e2e.sh
dfarrell07: Thanks for the clarification @manosnoam! I added a FIXME inline to reduce the timeout, but didn't change it in this commit to avoid confusion and maintain the scope of the commit.

mangelajo:
Github lost my comments :?

@dfarrell07, I think that the setup_subm_vars part deserves a 2nd round of cleanups.

Would it be feasible to setup the env vars somewhere in the main e2e.sh, and then make sure that any shared/needed details from there where used on each module? (for example the CIDRs that we need  to maintain in two places)

Other vars like subm_ns are intrisic to each method, but those could be declared as on the top level of each .sh (helm/operator file).

scripts/kind-e2e/lib_helm_deploy_subm.sh
tpantelis: yeah the namespaces are different but the libraries already expose a `$sub_ns` var that is used elsewhere in the base code. So it seems like it could be used here as well. But not really a big deal either way.

dfarrell07:
@mangelajo okay, agreed. I raised https://github.com/submariner-io/submariner/issues/292 to track that refactoring and the good suggestion regarding  @tpantelis raised above related to `$subm_ns` in `delete_subm_pods`/`preinstall_cleanup_subm`.

scripts/kind-e2e/lib_helm_deploy_subm.sh
dfarrell07: That makes sense. As @mangelajo raised below, the variables in e2e and the related libs need another set of refactoring. I raised #292 to track, and quoted/linked to this feedback. Imho this chunk of refactoring has been reviewed enough that it should go in as-is, handle #292 in a follow-up PR.

tpantelis: sounds good.

PR #254, opened 2019-12-16
Only build master branch, ignoring local branches

mangelajo:
With local branches, and automatically created ones travis is doing extra unnecessary work duplication. This fixes it. 

Pull requests will be tested regardless of this, but changes to specific branches  (not through PRs) will be ignored.

PR #278, opened 2020-01-09
Fix build issues on the overlapping-cidrs branch

sridhargaddam:
This PR fixes two issues.

1. Patch [https://github.com/submariner-io/submariner/pull/251] removed an
   API from ippool, but the associated unit tests were not updated. This
   patch fixes those issues.
2. PR [https://github.com/submariner-io/submariner/pull/232] modified the
   iptable wrappers. Similar changes are required in the globalnet controller.

Signed-off-by: Sridhar Gaddam <sgaddam@redhat.com>

PR #279, opened 2020-01-09
Enable CI on overlapping-cidrs branch

sridhargaddam:
The following change [https://github.com/submariner-io/submariner/pull/254]
modifies travis configuration to validate only master branch ignorning local
branches. Since we are interested in validating overlapping-cidrs branch,
this PR enables overlapping-cidrs as part of travis configuration.

Along with this, it also fixes a build issue with submariner-operator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: go
 branches:
   only:
   - master
+  - overlapping-cidrs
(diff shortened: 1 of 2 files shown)


mpeterson:
This should be added together with #278

PR #285, opened 2020-01-14
Output useful status info during E2E

mangelajo:


dfarrell07:
> ./../scripts/kind-e2e/e2e.sh: line 445: free: command not found

scripts/kind-e2e/e2e.sh
```diff
@@ -442,6 +442,21 @@ elif [[ $deploy = helm ]]; then
     setup_cluster3_gateway
 fi
 
+free -h
```
skitt: This requires the `procps-ng` package (in `Docker.dapper`).

scripts/kind-e2e/e2e.sh
```diff
@@ -442,6 +442,21 @@ elif [[ $deploy = helm ]]; then
...
+
+for context in cluster{1..3}; do
+  kubectl config use-context $context
+
+  POD=$(kubectl get pod -n $subm_ns -l app=submariner-engine -o jsonpath="{.items[0].metadata.name}")
+
+  echo "= ${context} ============================================="
+  kubectl exec $POD -n $subm_ns strongswan status
+  echo "--"
+  kubect get pods -n $subm_ns
```
skitt: `kubectl`

scripts/kind-e2e/e2e.sh
mangelajo: thanks

scripts/kind-e2e/e2e.sh
mangelajo: thanks :)

mangelajo:
> I see procps-ng was installed successfully, but you still don't have free.
> 
> > ./../scripts/kind-e2e/e2e.sh: line 445: free: command not found

I suspect we need to push the dapper image first.

skitt:
> > I see procps-ng was installed successfully, but you still don't have free.
> > > ./../scripts/kind-e2e/e2e.sh: line 445: free: command not found
> 
> I suspect we need to push the dapper image first.

Yes, that’s why I suggested adding the package to `Docker.dapper` rather than the base image ;-). Adding it to `Docker.dapper` makes the PR self-contained; we can add it to the base image later if we decide it’s useful there.

skitt:
`procps-ng` appears installed in the log twice, but both times are during image builds, and we don’t use the updated Dapper base image: that’s 54bf5c783db4, but the build then goes and uses 88c763283be3.

mangelajo:
> > > I see procps-ng was installed successfully, but you still don't have free.
> > > > ./../scripts/kind-e2e/e2e.sh: line 445: free: command not found
> > 
> > 
> > I suspect we need to push the dapper image first.
> 
> Yes, that’s why I suggested adding the package to `Docker.dapper` (`Dockerfile.dapper`, sorry) rather than the base image ;-). Adding it to `Dockerfile.dapper` makes the PR self-contained; we can add it to the base image later if we decide it’s useful there.

I think that we should add logic to rebuild the dapper base image before anything else if the base image changed.

skitt:
> I think that we should add logic to rebuild the dapper base image before anything else if the base image changed.

That’s already there, the base Dapper image is being rebuilt, but it’s not being used. It works locally though, so perhaps there’s a conflict with Travis’ container image cache (IIRC that’s a thing)...

scripts/package
```diff
@@ -23,9 +23,18 @@ DAPPER_BASE_IMAGE=${REPO}/dapper-base:${TAG}
 # We expect the Dapper base image to change rarely, so pull it first
 docker pull ${REPO}/dapper-base:latest || :
 
+CHANGED_FILES_LOCAL=$(git diff --name-only)
```
skitt: Changing this to `git diff --name-only origin/master` would ensure that the base image is rebuilt for any local change, whether committed or otherwise, assuming there’s an `origin` remote.

mangelajo:
> Yeah but there’s not always origin, I use upstream sometimes for the upstream repo and origin for mine (in which in don’t sync master) And what if we are working on a branch? Hmmm
> […](#)

Lol that was me, not sure why github picked it up as submariner-bot ;)

scripts/package
mangelajo: As I said, that's not the case on my side.  And there are corner cases like.. PRs that go on top of a branch (easy to detect in travis, but hard to detect on  local git, etc).

Saying that, I think we should also add this detection on the deploy phase, so deployment of this image only happens when a change has been really made.

And anyway, before it wasn't better, because before we were building dapper-base:*dev*, which does not lead to running dapper with that version on later steps (since we use dapper-base:later), I could be missing something

mangelajo:
I need to revise the e2e.sh logic based on the conflict

scripts/package
```diff
@@ -23,9 +23,18 @@ DAPPER_BASE_IMAGE=${REPO}/dapper-base:${TAG}
...
+
+if [[ "${CHANGED_FILES_PR[@]}" =~ "package/Dockerfile.dapper-base" ]] ||
+   [[ "${CHANGED_FILES_LOCAL[@]}" =~ "package/Dockerfile.dapper-base" ]] ; then
+      echo "Dockerfile.dapper-base was modified, rebuilding dapper image"
+      docker build -t ${DAPPER_BASE_IMAGE} -f Dockerfile.dapper-base .
+      docker tag ${DAPPER_BASE_IMAGE} ${REPO}/daper-base:latest
```
skitt: This should be `dapper-base` with two `p`s.

scripts/package
mangelajo: Resolved on different PR

scripts/package
mangelajo: Right, fixed on the other PR ':)

mangelajo:
ha, so now travis does not build composite patches... ':-)

Issue #288, opened 2020-01-16
submariner-engine fails to start with error about /var/run/charon.vici


mangelajo:
E0115 13:35:07.406556       1 strongswan.go:502] Failed to connect to charon: dial unix /var/run/charon.vici: connect: no such file or directory

E0115 13:35:07.500297       1 strongswan.go:502] Failed to connect to charon: dial unix /var/run/charon.vici: connect: no such file or directory

E0115 13:35:08.407199       1 strongswan.go:502] Failed to connect to charon: dial unix /var/run/charon.vici: connect: no such file or directory

E0115 13:35:08.500558       1 strongswan.go:502] Failed to connect to charon: dial unix /var/run/charon.vici: connect: no such file or directory

F0115 13:35:09.407418       1 main.go:138] Error starting the cable engine: Failed to load connections from charon: dial unix /var/run/charon.vici: connect: no such file or directory

goroutine 39 [running]:

k8s.io/klog.stacks(0xc00024a200, 0xc000186000, 0xb7, 0x279)

	/go/src/github.com/submariner-io/submariner/vendor/k8s.io/klog/klog.go:828 +0xb1

k8s.io/klog.(*loggingT).output(0x1e32e80, 0xc000000003, 0xc00021d0a0, 0x1dbd204, 0x7, 0x8a, 0x0)

	/go/src/github.com/submariner-io/submariner/vendor/k8s.io/klog/klog.go:779 +0x2d9

k8s.io/klog.(*loggingT).printf(0x1e32e80, 0x3, 0x129ad50, 0x23, 0xc00039ffa0, 0x1, 0x1)

	/go/src/github.com/submariner-io/submariner/vendor/k8s.io/klog/klog.go:678 +0x14e

k8s.io/klog.Fatalf(...)

	/go/src/github.com/submariner-io/submariner/vendor/k8s.io/klog/klog.go:1207

main.main.func1.1(0xc000462160, 0x145cac0, 0xc000456240, 0xc000458020)

	/go/src/github.com/submariner-io/submariner/main.go:138 +0xe5

created by main.main.func1

	/go/src/github.com/submariner-io/submariner/main.go:135 +0xa14

PR #289, opened 2020-01-16
Fixes strongswan vici socket path

mangelajo:
A recent change in strongswan modified the default vici socket path,
making submariner-engine unable to find the vici interface socket.

This patch explicitly specifies the vici socket path in charon.conf
and fixes the issue.

Fixes: #288


mangelajo:
Ok CI is passing with this patch, please review and we can merge it:

[scripts]$ kubectl --context=cluster2 exec netshoot-785ffd8c8-pt952 -- curl --output /dev/null -m 30 --silent --head --fail 100.96.167.210
Connection test was successful!

PR #290, opened 2020-01-16
Test artifact upload

mangelajo:


mangelajo:
It seems like artifact upload does not work for external PRs for security reasons, which is... a bummer... bad job @github

PR #293, opened 2020-01-19
Use globalCIDR from ClusterCRD

sridhargaddam:
Currently, globalCIDR is configured/stored as part of clusterCRD and
Submariner uses this value to populate the endpoint subnets. At the
same time, globalnet controller is reading this value as an environment
parameter in the globalnet POD. Storing/reading globalCIDR from two
different places could be error prone if not configured/deployed properly.
This patch modifies globalnet controller to read the globalCIDR from the
clusterCRD which is the source of truth for a cluster.

### DIFF
--- a/pkg/globalnet/controllers/ipam/gatewaymonitor.go
+++ b/pkg/globalnet/controllers/ipam/gatewaymonitor.go
@@ -135,10 +135,33 @@ func (i *GatewayMonitor) processNextEndpoint() bool {
 		// If the endpoint hostname matches with our hostname, it implies we are on gateway node
 		if endpoint.Spec.Hostname == hostname {
 			klog.V(4).Infof("We are now on GatewayNode %s", endpoint.Spec.PrivateIP)
+
+			clusterList, err := i.submarinerClientSet.SubmarinerV1().Clusters(i.ipamSpec.Namespace).List(metav1.ListOptions{})
+			if err != nil {
+				i.endpointWorkqueue.AddRateLimited(key)
+				return fmt.Errorf("error while retrieving the cluster info: %v", err)
+			}
+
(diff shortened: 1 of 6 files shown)


pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -220,7 +220,30 @@ func (i *GatewayMonitor) initializeIpamController() {
 	}
 
 	klog.V(4).Infof("On Gateway Node, initializing ipamController.")
-	ipamController, err := NewController(i.ipamSpec, &informerConfig)
+	clusterList, err := i.submarinerClientSet.SubmarinerV1().Clusters(i.ipamSpec.Namespace).List(metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("Error while retrieving the cluster info: %v", err)
```
tpantelis: I think this would likely be a transient error so I think we should return an error and have the caller re-queue-retry.

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -220,7 +220,30 @@ func (i *GatewayMonitor) initializeIpamController() {
...
+		return
+	}
+
+	globalCIDR := ""
+	for _, cluster := range clusterList.Items {
+		if cluster.Spec.ClusterID == i.clusterID {
+			klog.V(4).Infof("GlobalCidr configured in %s is %s", i.clusterID, cluster.Spec.GlobalCIDR)
+			if cluster.Spec.GlobalCIDR != nil {
```
tpantelis: I think we should check `len(cluster.Spec.GlobalCIDR) > 0` to also handle an empty array.

PR #295, opened 2020-01-21
Refactor vars in e2e script/libs

### LINKED ISSUE #292: Cleanup vars in e2e script/libs
Follow-up on #273 review comments related to additional variable refactoring that should be done in the e2e script and related Helm/subctl libraries.

> I think that the setup_subm_vars part deserves a 2nd round of cleanups.
> Would it be feasible to setup the env vars somewhere in the main e2e.sh, and then make sure that any shared/needed details from there where used on each module? (for example the CIDRs that we need to maintain in two places)
> Other vars like subm_ns are intrisic to each method, but those could be declared as on the top level of each .sh (helm/operator file).

https://github.com/submariner-io/submariner/pull/273#issuecomment-575209204

> yeah the namespaces are different but the libraries already expose a $sub_ns var that is used elsewhere in the base code. So it seems like it could be used here as well. But not really a big deal either way.

https://github.com/submariner-io/submariner/pull/273#discussion_r367509912

dfarrell07:
Closes #292

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -3,6 +3,11 @@ set -em
 
 source $(git rev-parse --show-toplevel)/scripts/lib/debug_functions
 
+### Variables ###
+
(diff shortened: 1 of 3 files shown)


mangelajo:
Thanks for handling this!

Issue #298, opened 2020-01-23
connection from hostNetwork to remote Pod or ClusterIP does not work


mangelajo:
In k8s, connecting from a worker node (or pod with hostNetwork) to another pod or clusterIP works, see example:

```bash
$ kubectl get pods -o wide
NAME                       READY   STATUS    RESTARTS   AGE   IP           NODE          NOMINATED NODE   READINESS GATES
netshoot-785ffd8c8-t2tn5   1/1     Running   509        21d   10.42.4.34   subm-a-w-1    <none>           <none>
test-5db4cfb4bd-z5q7l      1/1     Running   0          16d   10.42.3.24   subm-a-gw-1   <none>           <none>

$ ./netshoot-hostnet.sh subm-a-w-1
bash-5.0# ping 10.42.4.34
PING 10.42.4.34 (10.42.4.34) 56(84) bytes of data.
64 bytes from 10.42.4.34: icmp_seq=1 ttl=64 time=0.063 ms
64 bytes from 10.42.4.34: icmp_seq=2 ttl=64 time=0.072 ms
^C
--- 10.42.4.34 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 999ms
rtt min/avg/max/mdev = 0.063/0.067/0.072/0.004 ms

bash-5.0# ping 10.42.3.24
PING 10.42.3.24 (10.42.3.24) 56(84) bytes of data.
64 bytes from 10.42.3.24: icmp_seq=1 ttl=63 time=0.691 ms
64 bytes from 10.42.3.24: icmp_seq=2 ttl=63 time=0.639 ms
^C
--- 10.42.3.24 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 999ms
rtt min/avg/max/mdev = 0.639/0.665/0.691/0.026 ms

bash-5.0# ip a | grep 10.42 -B 2
6: flannel.1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8951 qdisc noqueue state UNKNOWN group default 
    link/ether 7e:4f:17:bc:77:db brd ff:ff:ff:ff:ff:ff
    inet 10.42.4.0/32 scope global flannel.1
bash-5.0# exit
```

[...]

PR #305, opened 2020-01-29
Support Connectivity from HostNetwork to RemoteCluster

sridhargaddam:
Currently, Submariner configures necessary tunnels and routing rules to
support connectivity between the PODs in different Clusters. However, if
we try to connect from the underlying host to a remotePOD/Service it does
not work. While working on some of the Storage use-cases in K8s, we realized
that connectivity from underlying Host to remoteCluster is required. This
patch addresses this use-case.

Fixes Issue# https://github.com/submariner-io/submariner/issues/298


pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,57 @@ package route
...
+	_, err := netlink.LinkByName(iface)
+	return err == nil
+}
 
-func isFlannelCNI() error {
-	_, err := netlink.LinkByName(FlannelInterface)
-	if err != nil {
-		return fmt.Errorf("error retrieving link by name %s: %v", FlannelInterface, err)
+func discoverCNIInterface(clusterCIDR string) string {
+	cniInterfaceList := []string{
```
mangelajo: May be a we can implement a more generic way... like listing the interfaces and figuring if one of them has an IP inside the pod network.

In such way we don't need to care about which interface the CNI uses.

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,57 @@ package route
...
+			}
 		}
 	}
+	return ""
+}
+
+func toggleCNISpecificConfiguration(iface string) error {
+	err := ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+iface+"/rp_filter", []byte("2"), 0644)
+	if err != nil {
+		return fmt.Errorf("unable to update rp_filter for "+iface+", err: %s", err)
```
mangelajo: nit/idea, no action required: May be in such error case we can also explain "hostNetwork to remote podIP/clusterIPs won't work" ? ..

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,57 @@ package route
...
 	"github.com/vishvananda/netlink"
 	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/util"
 )
 
-const FlannelInterface = "flannel.1"
+func interfaceExists(iface string) bool {
+	_, err := netlink.LinkByName(iface)
+	return err == nil
```
tpantelis: It seems `LinkByName` could return an err for reasons other than not found. Perhaps we should specifically check for `netlink.LinkNotFoundError` and also propagate any other error...

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,57 @@ package route
...
-	// FlannelInterface exists, return nil
-	return nil
-}
+	for _, iface := range cniInterfaceList {
+		if interfaceExists(iface) {
+			klog.Infof("Found CNI interface [%s] on the Host", iface)
+			ipv4Addr, err := util.GetIPv4AddressOnInterface(iface)
+			if err != nil {
+				klog.Errorf("error reading IPv4 address on CNI interface [%s]: %v", iface, err)
+				return ""
```
tpantelis: Should we propagate the error here (and below)?

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,57 @@ package route
...
 
-	// FlannelInterface exists, return nil
-	return nil
-}
+	for _, iface := range cniInterfaceList {
+		if interfaceExists(iface) {
+			klog.Infof("Found CNI interface [%s] on the Host", iface)
+			ipv4Addr, err := util.GetIPv4AddressOnInterface(iface)
+			if err != nil {
+				klog.Errorf("error reading IPv4 address on CNI interface [%s]: %v", iface, err)
```
tpantelis:
```suggestion
				klog.Errorf("error reading IPv4 address on CNI interface %q: %v", iface, err)
```

nit - so use quoted string instead of []? Same in other places.

pkg/routeagent/controllers/route/driver.go
mangelajo: ack thanks

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,59 @@ package route
...
+					// Verify that interface has an address from cluster CIDR
+					if clusterNetwork.Contains(address) {
+						klog.V(4).Infof("Found CNI Interface %q that has IP %q from ClusterCIDR %q",
+							iface.Name, ipAddr.String(), clusterCIDR)
+						return &cniInterface{ipAddress: ipAddr.String(), name: iface.Name}
+					}
+				}
+			}
+		}
+	}
```
mangelajo: Thanks! :-)

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -17,35 +17,46 @@ func (r *Controller) createIPTableChains() error {
...
 	if err = r.createChainIfNotExists(ipt, "nat", SmPostRoutingChain); err != nil {
-		return fmt.Errorf("Unable to create %s chain in iptables: %v", SmPostRoutingChain, err)
+		return fmt.Errorf("unable to create %s chain in iptables: %v", SmPostRoutingChain, err)
 	}
 
 	klog.V(4).Infof("Insert %s rule that has rules for inter-cluster traffic", SmPostRoutingChain)
 	forwardToSubPostroutingRuleSpec := []string{"-j", SmPostRoutingChain}
 	if err = r.prependUnique(ipt, "nat", "POSTROUTING", forwardToSubPostroutingRuleSpec); err != nil {
-		klog.Errorf("Unable to insert iptable rule in NAT table, POSTROUTING chain: %v", err)
+		klog.Errorf("unable to insert iptable rule in NAT table, POSTROUTING chain: %v", err)
```
mangelajo: why klog.Error versus return in other cases (L20,L31,L36 ..) may be we should also bubble up the error here for consistency

sridhargaddam: Oops, I missed it, thanks for pointing out. Will update the PR.

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,59 @@ package route
...
+		if err != nil {
+			klog.Errorf("For interface %q, iface.Addrs returned error: %v", iface.Name, err)
+			return nil
+		}
+
+		if len(addrs) > 0 {
+			for i := range addrs {
+				ipAddr, _, err := net.ParseCIDR(addrs[i].String())
+				if err != nil {
+					klog.Errorf("Unable to ParseCIDR : %v", addrs[i].String())
```
tpantelis:
```suggestion
					klog.Errorf("Unable to ParseCIDR %q : %v", addrs[i].String(), err)
```

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,59 @@ package route
...
 
-	// FlannelInterface exists, return nil
+	for _, iface := range hostInterfaces {
+		addrs, err := iface.Addrs()
+		if err != nil {
+			klog.Errorf("For interface %q, iface.Addrs returned error: %v", iface.Name, err)
+			return nil
+		}
+
+		if len(addrs) > 0 {
```
tpantelis: nit - this check isn't needed as `range` covers it on the next line.

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,59 @@ package route
...
-		err := ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+FlannelInterface+"/rp_filter", []byte("2"), 0644)
-		if err != nil {
-			return fmt.Errorf("unable to update rp_filter for "+FlannelInterface+", err: %s", err)
-		} else {
-			klog.Info("Successfully configured rp_filter to loose mode(2) on " + FlannelInterface)
-		}
+func toggleCNISpecificConfiguration(iface string) error {
+	err := ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+iface+"/rp_filter", []byte("2"), 0644)
+	if err != nil {
+		return fmt.Errorf("unable to update rp_filter for "+iface+", err: %s", err)
```
tpantelis:
```suggestion
		return fmt.Errorf("unable to update rp_filter for cni interface %q", err: %s", iface, err)
```

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -17,35 +17,46 @@ func (r *Controller) createIPTableChains() error {
...
 	if err = r.prependUnique(ipt, "filter", "FORWARD", ruleSpec); err != nil {
-		klog.Errorf("Unable to insert iptable rule in filter table to allow vxlan traffic: %v", err)
+		return fmt.Errorf("unable to insert iptable rule in filter table to allow vxlan traffic: %v", err)
+	}
+
+	if r.cniIface != nil {
+		// Program rules to support communication from HostNetwork to remoteCluster
+		sourceAddress := strconv.Itoa(VxLANVTepNetworkPrefix) + ".0.0.0/8"
+		ruleSpec = []string{"-s", sourceAddress, "-o", VxLANIface, "-j", "SNAT", "--to", r.cniIface.ipAddress}
+		klog.V(4).Infof("Installing rule for hostNetwork to remoteCluster communication:"+
```
tpantelis:
```suggestion
		klog.V(4).Infof("Installing rule for host network to remote cluster communication: %s, strings.Join(ruleSpec, " "))
```

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -17,35 +17,46 @@ func (r *Controller) createIPTableChains() error {
...
 	klog.V(4).Infof("Install/ensure SUBMARINER-INPUT chain exists")
 	if err = r.createChainIfNotExists(ipt, "filter", "SUBMARINER-INPUT"); err != nil {
-		return fmt.Errorf("Unable to create SUBMARINER-INPUT chain in iptables: %v", err)
+		return fmt.Errorf("unable to create SUBMARINER-INPUT chain in iptables: %v", err)
 	}
 
 	forwardToSubInputRuleSpec := []string{"-p", "udp", "-m", "udp", "-j", "SUBMARINER-INPUT"}
 	if err = ipt.AppendUnique("filter", "INPUT", forwardToSubInputRuleSpec...); err != nil {
-		klog.Errorf("Unable to append iptables rule \"%s\": %v\n", strings.Join(forwardToSubInputRuleSpec, " "), err)
+		return fmt.Errorf("unable to append iptables rule \"%s\": %v\n", strings.Join(forwardToSubInputRuleSpec, " "), err)
```
tpantelis:
```suggestion
		return fmt.Errorf("unable to append iptables rule %q: %v\n", strings.Join(forwardToSubInputRuleSpec, " "), err)
```

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -17,35 +17,46 @@ func (r *Controller) createIPTableChains() error {
...
 	}
 
 	klog.V(4).Infof("Allow VxLAN incoming traffic in SUBMARINER-INPUT Chain")
 	ruleSpec := []string{"-p", "udp", "-m", "udp", "--dport", strconv.Itoa(VxLANPort), "-j", "ACCEPT"}
 	if err = ipt.AppendUnique("filter", "SUBMARINER-INPUT", ruleSpec...); err != nil {
-		klog.Errorf("Unable to append iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
+		return fmt.Errorf("unable to append iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
```
tpantelis:
```suggestion
		return fmt.Errorf("unable to append iptables rule %q: %v\n", strings.Join(ruleSpec, " "), err)
```

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -17,35 +17,46 @@ func (r *Controller) createIPTableChains() error {
...
+	}
+
+	if r.cniIface != nil {
+		// Program rules to support communication from HostNetwork to remoteCluster
+		sourceAddress := strconv.Itoa(VxLANVTepNetworkPrefix) + ".0.0.0/8"
+		ruleSpec = []string{"-s", sourceAddress, "-o", VxLANIface, "-j", "SNAT", "--to", r.cniIface.ipAddress}
+		klog.V(4).Infof("Installing rule for hostNetwork to remoteCluster communication:"+
+			" %s", strings.Join(ruleSpec, " "))
+		if err = ipt.AppendUnique("nat", SmPostRoutingChain, ruleSpec...); err != nil {
+			return fmt.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
```
tpantelis:
```suggestion
			return fmt.Errorf("error appending iptables rule %q: %v\n", strings.Join(ruleSpec, " "), err)
```

pkg/routeagent/controllers/route/driver.go
```diff
@@ -3,32 +3,59 @@ package route
...
 	"fmt"
 	"io/ioutil"
+	"net"
 
-	"github.com/vishvananda/netlink"
 	"k8s.io/klog"
 )
 
-const FlannelInterface = "flannel.1"
+func discoverCNIInterface(clusterCIDR string) *cniInterface {
```
tpantelis: Should this function propagate errors? And does an error here warrant aborting start up?

sridhargaddam: I'm afraid No. We are only trying to discover the CNI interface in this API. If it does not exist (like some new CNI), its fine. We should not abort the startup.

pkg/routeagent/controllers/route/driver.go
sridhargaddam: Done

pkg/routeagent/controllers/route/driver.go
sridhargaddam: Done

pkg/routeagent/controllers/route/driver.go
sridhargaddam: Done

pkg/routeagent/controllers/route/iptables.go
sridhargaddam: Done

sridhargaddam:
There seem to be some issues during merge, so proposed a new PR with the same changes - https://github.com/submariner-io/submariner/pull/316 and closing this PR.

PR #306, opened 2020-01-29
Update subctl 0.0.2 to 0.0.3

dfarrell07:


skitt:
> I wonder if we could have some way of detecting the latest version, and pulling such one by default.

https://github.com/submariner-io/submariner/pull/310

dfarrell07:
> > I wonder if we could have some way of detecting the latest version, and pulling such one by default.
> 
> #310

+1, let's do it with #310. Closing this.

PR #307, opened 2020-01-29
Decrease weave-net pod deployment timeout

dfarrell07:
In long-past debugging/fixing activities, our CI folks thought this
timeout needed to be longer for some deployments. In recent discussions,
I've confirmed that it no longer needs to be so long. Set it to the same
300s used elsewhere in similar timeouts.

See comments here:

github.com/submariner-io/submariner/pull/273#discussion_r367435417

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -70,8 +70,7 @@ function setup_custom_cni(){
             echo "Applying weave network in to cluster${i}..."
             kubectl --context=cluster${i} apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')&env.IPALLOC_RANGE=${cluster_CIDRs[cluster${i}]}"
             echo "Waiting for weave-net pods to be ready cluster${i}..."
-            # FIXME: This timeout doesn't need to be so long
-            kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=700s
+            kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=300s
             echo "Waiting for core-dns deployment to be ready cluster${i}..."
             kubectl --context=cluster${i} -n kube-system rollout status deploy/coredns --timeout=300s
         fi


scripts/kind-e2e/e2e.sh
```diff
@@ -70,8 +70,7 @@ function setup_custom_cni(){
             echo "Applying weave network in to cluster${i}..."
             kubectl --context=cluster${i} apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')&env.IPALLOC_RANGE=${cluster_CIDRs[cluster${i}]}"
             echo "Waiting for weave-net pods to be ready cluster${i}..."
-            # FIXME: This timeout doesn't need to be so long
-            kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=700s
+            kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=300s
```
tpantelis: I'm fine with lowering it but it should be uncommon for it to timeout so having it longer to give more safety cushion shouldn't hurt.

dfarrell07: The problem is when it correctly-fails and you have to wait 700s to find it out.

tpantelis: yeah but I'd think that **should never** happen? Either way I'm fine to lower it as long as we're sure 5 min gives it plenty of cushion.

manosnoam:
The 700s timeout was originally requested when I used to run kind-e2e on my laptop, as 300s was not enough. Since nowadays I test Submariner deployment on real clusters (AWS and OSP), not in local KIND k8s, this is not a problem I'm seeing anymore, so it's fine to decrease back to 300s.

tpantelis:
> The 700s timeout was originally requested when I used to run kind-e2e on my laptop, as 300s was not enough. Since nowadays I test Submariner deployment on real clusters (AWS and OSP), not in local KIND k8s, this is not a problem I'm seeing anymore, so it's fine to decrease back to 300s.

But what about for others running kind-e2e on their laptops? Based on that, It seems the 700s timeout is a safer cushion to cover all environments.

manosnoam:
The question is, do we still see this error due to timeout lenght, or perhaps the real culprit is now resolved (it was long time ago), so 300s is good to go ?

dfarrell07:
> The question is, do we still see this error due to timeout lenght, or perhaps the real culprit is now resolved (it was long time ago), so 300s is good to go ?

Right, there was an underlying issue that was causing this part of the deployment to fail (hanging old containers if I recall correctly). The long timeout wasn't because something was actually too slow for laptops (it's not, this part is fast), it's that something else was broken under the hood and having a longer timeout was a step in debugging.

scripts/kind-e2e/e2e.sh
dfarrell07: It has happened to me quite a bit as I've been developing the e2e scripts. If everything is working correctly, it shouldn't be hit, but as it's being developed or if something is wrong, it is sometimes hit.

dfarrell07:
Thanks @tpantelis!

PR #309, opened 2020-01-30
Handle local dapper-base changes when necessary

mangelajo:
When packages/Dockerfile.dapper-base is modified it will be detected
and the image will be regenerated locally.

The base dapper image is built and published from the master branch
so if any change has happened between the upstream master branch
and the local files, the local dapper image will be rebuilt
before any other build procedure is handled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ kubefed ?= false
 deploytool ?= helm
 debug ?= false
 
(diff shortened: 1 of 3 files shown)


mangelajo:
@skitt I have separated this patch into it's own :-)

mangelajo:
Oh, doing manual rebase.

Issue #301, opened 2020-01-29
MTU is smaller than necessary for vx-submariner path


mangelajo:
We use vx-submariner to achieve src-ip preservation of pods.

But vx-submariner MTU is not calculated and uses a fixed 1450. We should identify the outgoing interface, and adapt the interface to use interface.mtu - 50 (Vxlan overhead).

mangelajo:
@sridhargaddam I added you to this one since you were working on the other related fix, and you are very familiar with this part, but let me know if you are busy and wand me to look at it.

sridhargaddam:
> @sridhargaddam I added you to this one since you were working on the other related fix, and you are very familiar with this part, but let me know if you are busy and wand me to look at it.

No problem @mangelajo, I'll work on the necessary changes and push the PR.

sridhargaddam:
The necessary fix is now merged in Submariner repo.

PR #311, opened 2020-01-30
Fix MTU on vx-submariner interface

sridhargaddam:
Currently the MTU on Submariner VXLAN interface is hardcoded to 1450.
In some of the environments Jumbo frames are used. So, ideally
Submariner should identify the outgoing interface and configure
the appropriate MTU. This patch fixes it.

Fixes Issue: https://github.com/submariner-io/submariner/issues/301

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -67,6 +67,7 @@ const (
 	VxInterfaceWorker  = 0
 	VxInterfaceGateway = 1
 	VxLANPort          = 4800
+	VxLANOverhead      = 50
 
 	// Why VxLANVTepNetworkPrefix is 240?
 	// On VxLAN interfaces we need a unique IPAddress which does not collide with the
@@ -318,6 +319,9 @@ func (r *Controller) createVxLANInterface(ifaceType int, gatewayNodeIP net.IP) e
 		return fmt.Errorf("failed to derive the vxlan vtepIP for %s, %v", ipAddr, err)
 	}
 
+	// Derive the MTU based on the default outgoing interface
+	vxlanMtu := r.defaultHostIface.MTU - VxLANOverhead
+
 	if ifaceType == VxInterfaceGateway {
 		attrs := &vxLanAttributes{
 			name:     VxLANIface,
@@ -325,7 +329,7 @@ func (r *Controller) createVxLANInterface(ifaceType int, gatewayNodeIP net.IP) e
 			group:    nil,
 			srcAddr:  nil,
 			vtepPort: VxLANPort,
-			mtu:      1450,
+			mtu:      vxlanMtu,
 		}
 
 		r.vxlanDevice, err = newVxlanIface(attrs)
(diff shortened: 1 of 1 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -67,6 +67,7 @@ const (
 	VxInterfaceWorker  = 0
 	VxInterfaceGateway = 1
 	VxLANPort          = 4800
+	VxLANOverhead      = 50
```
tpantelis: So I assume this is subtracted from the system's default MTU as estimated overhead added by vxLAN stuff. But why 50? Should we make this configurable?

sridhargaddam: 50 is the VxLAN header size which is fixed.

sridhargaddam: Yes, its the VxLAN header size which is fixed. I don't see a need for making it configurable.

tpantelis: ok - makes sense.

pkg/routeagent/controllers/route/route.go
```diff
@@ -318,14 +319,17 @@ func (r *Controller) createVxLANInterface(ifaceType int, gatewayNodeIP net.IP) e
 		return fmt.Errorf("failed to derive the vxlan vtepIP for %s, %v", ipAddr, err)
 	}
 
+	// Derive the MTU based on the default outgoing interface
+	vxlanMtu := r.defaultHostIface.MTU - VxLANOverhead
```
mangelajo: I'm trying to think if in some environments this would not be perfect, for example, deployments where it some hosts will span multiple L2 domains, where some of the links are smaller.

There's no simple solution for such thing.  May be we can add an environment variable to mitigate that, and let people configure down the MTU size?

May be we can add a log entry explaining from which interface we derived the MTU, and the mtu assigned to vxlan?


Variables can be added here: https://github.com/submariner-io/submariner/blob/929d9bf1991bb67b7dcc2e351644a6585d7ad69c/pkg/types/types.go#L16

And captured here:
https://github.com/submariner-io/submariner/blob/929d9bf1991bb67b7dcc2e351644a6585d7ad69c/main.go#L55

like SUBMARINER_NAT_ENABLED, etc...

May be I'm going to far, but I'm trying to think of past issues we encoutered with neutron/etc, that then we make easier to mitigate in the field if happen.

sridhargaddam: @mangelajo I see your point, but I'm not sure if such environments are commonly used in cloud environments. My main concern is about introducing **new environment variables** which will affect the user-experience. I agree that we can/should do it if necessary, but I feel that we can hold on until we come across such a use-case - at the most the effect would be slightly degraded performance (submariner functionality would still continue as you know).

mangelajo: Fair point. Give that the worst scenario is likely to be slight performance degradation in those corner cases I agree with you that making UX worse it's better to wait and see if that's really ever a problem.

Issue #302, opened 2020-01-29
Mark diagnostic E2E tests which can be used outside CI E2E by customers without special requisites.


mangelajo:
We must organize the subset of E2E tests which can be used to validate connectivity and a submariner install without specific setup of the environment (gateway redundancy, etc.).

This should also include clear documentation on how to use such E2E tests, and what's required to do so.

This is also related to: https://github.com/submariner-io/submariner-operator/issues/103

tpantelis:
Rather than exposing the Go E2E tests to end-users, I would propose we package such validation in a CLI, ie subctl.

mangelajo:
Yes, but my intention is to use these tests directly (at least for now) from subctl, by compiling them in. And then subctl can select specifically the "diagnostic" ones.

PR #312, opened 2020-01-30
Mark tests which could be used as diagnostic from subctl

mangelajo:
Issue: #302


mangelajo:
It's kind of a silly change, but I will use this tag from subctl.

tpantelis:
> It's kind of a silly change, but I will use this tag from subctl.

I'm not clear on how this would be used from subctl. I'm assuming you would scan top-level `Describes` and look for substring "diagnostics"? 

Either way, rather than a tagging approach to identify which tests subctl would run, I was imagining  subctl command(s) would target specific diagnostic tests. Eg, a user just joined a new cluster and wants to verify connectivity to another so might run something like: _subctl verify-connectivity <cluster1> <cluster2>_.

mangelajo:
Exactly as you describe but we would use that tag to select only specific cases, and under other verifications we could use other types of tests, with warnings and special requirements, etc...

tpantelis:
> Exactly as you describe but we would use that tag to select only specific cases, and under other verifications we could use other types of tests, with warnings and special requirements, etc...

I'm not sure about the use of such a "diagnostic" tag. It seems like it could be fragile, ie what if someone tags something we don't want for a specific type of test, possible accidentally. I think it would be better to have subctl specifically know and target the spec(s) it wants to run. Or maybe even just use subm E2E as a library and invoke functions like `dataplane.RunConnectivityTest` directly, maybe even bypassing Ginkgo altogether if possible.

mangelajo:
> > Exactly as you describe but we would use that tag to select only specific cases, and under other verifications we could use other types of tests, with warnings and special requirements, etc...
> 
> I'm not sure about the use of such a "diagnostic" tag. It seems like it could be fragile, ie what if someone tags something we don't want for a specific type of test, possible accidentally. I think it would be better to have subctl specifically know and target the spec(s) it wants to run. Or maybe even just use subm E2E as a library and invoke functions like `dataplane.RunConnectivityTest` directly, maybe even bypassing Ginkgo altogether if possible.

It's our work as reviewer do do ensure thing, may be we can mark them "subctl" which would be clearer.

tpantelis:
> 
> It's our work as reviewer do do ensure thing, may be we can mark them "subctl" which would be clearer.

I'm sorry but the whole idea of specially marking tests in this manner just doesn't strike my fancy.

tpantelis:
> > It's our work as reviewer do do ensure thing, may be we can mark them "subctl" which would be clearer.
> 
> I'm sorry but the whole idea of specially marking tests in this manner just doesn't strike my fancy.

To clarify - the subctl `verify-connectivity` command already specifies the _dataplane_ focus string and imports the dataplane package which targets and runs the appropriate tests. So I don't see a need to add anything else on top of that.

PR #313, opened 2020-01-30
Support multiple individual contextless kubeconfig files

mangelajo:
The idea behind this framework change is to allow consumntion of the
E2E tests from subctl, where customers normally have multiple kubeconfig
files with same context names.

Related-to: https://github.com/submariner-io/submariner-operator/issues/103


mangelajo:
True, but it will help us solve the existing use case. Thanks for the note and review @skitt

test/e2e/framework/framework.go
```diff
@@ -123,9 +123,18 @@ func (f *Framework) BeforeEach() {
 
 	ginkgo.By("Creating kubernetes clients")
 
+	// We accept either KubeContexts + a single KubeConfig variable, or lots of individual
+	// KubeConfigs
+	Expect(len(TestContext.KubeConfigs) > 0 && len(TestContext.KubeContexts) > 0).NotTo(BeTrue())
```
tpantelis: Can you add a user-friendly message otherwise the  error output will be vague, eg "Invalid kube configuration. You must specify either a single \"kubeconfig\" parameter containing  multiple kube config paths or multiple \"kubeconfig-path\" parameters, one for kube config path"?

test/e2e/framework/framework.go
```diff
@@ -123,9 +123,18 @@ func (f *Framework) BeforeEach() {
...
+
 	for _, context := range TestContext.KubeContexts {
-		f.ClusterClients = append(f.ClusterClients, f.createKubernetesClient(context))
-		f.SubmarinerClients = append(f.SubmarinerClients, f.createSubmarinerClient(context))
+		f.ClusterClients = append(f.ClusterClients, f.createKubernetesClient(TestContext.KubeConfig, context))
+		f.SubmarinerClients = append(f.SubmarinerClients, f.createSubmarinerClient(TestContext.KubeConfig, context))
```
tpantelis: If using multiple kubeconfigs then `TestContext.KubeConfig` may/would be empty in which case it looks like `loadConfig` would fail or am I missing something?

test/e2e/framework/test_context.go
```diff
@@ -11,8 +11,10 @@ import (
 type contextArray []string
 
 type TestContextType struct {
+	KubeConfigs         []string // KubeConfigs provides an alternative to KubeConfig + KubeContexts
```
tpantelis: This needs to be added to `registerFlags` below.  Also I suggest we name the parameter "kubeconfig-path" so usage would be:

  _-kubeconfig-path path1  kubeconfig-path path2_

rather than:

  _-kubeconfigs path1  kubeconfigs path2_

test/e2e/framework/framework.go
mangelajo: Yeah, that's true. I will either add an error on the check on previous lines or also support that case.

test/e2e/framework/test_context.go
mangelajo: I didn't want to make this available from the command line. Just to expose it on the framework for subctl usage.

I could add support for that if we think it'd be beneficial. @skitt said the community and kind is moving away from that model, but I needed support for now on subctl.

I'd skip this, and add it if we think we need it.

mangelajo:
@tpantelis waiting for you to read my comments about the flags before merging, all your other comments where handled.

test/e2e/framework/framework.go
```diff
@@ -123,9 +123,31 @@ func (f *Framework) BeforeEach() {
 
 	ginkgo.By("Creating kubernetes clients")
 
+	Expect(len(TestContext.KubeConfigs) > 0 && len(TestContext.KubeContexts) > 0).NotTo(BeTrue(),
+		"Mixing multiple KubeConfigs (contextless) and one KubeConfig variable"+
+			"with mutiple contexts is not supported.")
```
tpantelis: So we've decided not to expose `KubeConfigs` as a CLI arg for now but perhaps we shouldn't make that assumption here. `ParseFlags` sets `TestContext.ClusterIDs = TestContext.KubeContexts` so it is expected that `KubeContexts` is set which will break these checks if `KubeConfigs` is set since it is expected that `len(ClusterIDs)` matches that of `KubeConfigs` below. 

Also all these checks and the interaction between the settings is a bit confusing. I think we can simplify by just making `KubeConfig` and `KubeConfigs` mutually exclusive, ie

```
if len(TestContext.KubeConfig) > 0 {
    Expect(len(TestContext.KubeConfigs)).To(BeZero(),
			"Either KubeConfig or KubeConfigs must be specified but not both")
    for _, context := range TestContext.KubeContexts {
        ...
    }
} else if len(TestContext.KubeConfigs) > 0 {
    Expect(len(TestContext.KubeConfig)).To(BeZero(),
			"Either KubeConfig or KubeConfigs must be specified but not both")
    Expect(len(TestContext.KubeConfigs)).To(Equal(len(TestContext.ClusterIDs)), "...")
    for _, kubeConfig := range TestContext.KubeConfigs {
        ....
    }
} else {
    Fail("One of KubeConfig or KubeConfigs must be specified")
}
```

Also, to simplify further,  I don't think we need both `KubeContexts` and `ClusterIDs` since the tests have been using the `KubeContexts` as cluster IDs anyway (and you've changed all other uses of `KubeContexts` to `ClusterIDs`) and you explicitly make then equal in `ParseFlags`.  So if using `KubeConfig` then `KubeContexts` and `ClusterIDs` are one and the same; if using `KubeConfigs` then `KubeContexts` isn't used - all that tells me that `KubeContexts` isn't needed.

mangelajo: On the check, I agree, I like your proposal better.

On the difference between ClusterIDs and contexts, I disagree, our E2E tests were assuming that clusterIDs == contexts, which is important specially in the e2e tests which check clusters, but also important for printing.

Reusing the E2E tests from subctl needs to dissociate those two things which are logically different, just happen to be the same in our environment.

May be I should move the ClusterID array population to the BeforeEach part, explaining the logic.

mangelajo:
Hey @tpantelis I didn't address all your concerns in https://github.com/submariner-io/submariner/pull/313/commits/c6398fb38515bf8f04764b7ddf1ea7bee8fb0b25 . But some of them.

ClusterIDs and Context names are definitely different, for example, when using this with openshift config files you would find:

kubeconfig-cluster-1, with context "admin"
kubeconfig-cluster-2, with context "admin"

And that doesn't work with our E2E framework. 

I can provide access to that type of configuration from the E2E in-tree too, which I guess noam could make some good use of. 

I'd prefer to do that as a separate PR, may be we could accept commands in the format of:

```
./e2e --kubeconfig kubeconfig-1:kubeconfig-2:kubeconfig-3  --dp-cluster <context>:<id> --dp-cluster <context>:<id> --dp-cluster <context>:<id>
```
or, alternatively:
```
./e2e --dp-cluster kubeconfig-1:<context>:<id> --dp-cluster kubeconfig-2:<context>:<id>  ...
```

mangelajo:
I can also do that as part of this PR if we find it right, as, very likely will change the interface of the TestContext too..

test/e2e/framework/framework.go
tpantelis: > May be I should move the ClusterID array population to the BeforeEach part, explaining the logic.

yeah - I was thinking that too, ie assign `ClusterIDs` to `KubeContexts` only if it's nil/empty.

tpantelis:
I like the idea of this form `<context>:<id>`. I think we can keep `-dp-context`. So if the `:<id>` part isn't specified, then assume `id=context`. 

I think this can be done in a separate PR if/when it's necessary.

PR #314, opened 2020-01-30
Merge overlapping-cidrs branch into master

tpantelis:


mangelajo:
Approved, IMHO the changes are not modifying behaviour as long as globalnet is not enabled, and the CRD changes are compatible for an update.

PR #316, opened 2020-01-31
Support Connectivity from HostNetwork to RemoteCluster

sridhargaddam:
Currently, Submariner configures necessary tunnels and routing rules to
support connectivity between the PODs in different Clusters. However, if
we try to connect from the underlying host to a remotePOD/Service it does
not work. While working on some of the Storage use-cases in K8s, we realized
that connectivity from underlying Host to remoteCluster is required. This
patch addresses this use-case.

Fixes Issue: https://github.com/submariner-io/submariner/issues/298

### DIFF
--- a/pkg/routeagent/controllers/route/driver.go
+++ b/pkg/routeagent/controllers/route/driver.go
@@ -3,32 +3,57 @@ package route
 import (
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


ardenwick:
Does the SNAT rule also apply to packets originated from pods running on the host?

PR #317, opened 2020-01-31
Add Armada-based kind k8s cluster deploys

### LINKED ISSUE #291: Prototype Armada-based deployments, related e2e script refactoring
This is to cover part of #256 that should be broken out into its own Issue.

> 1. Use armada to deploy kind clusters

Refactor the e2e scripts to deploy kind-based clusters using Armada. Note that Armada is still a PR under review, being refactored on a feature branch (https://github.com/submariner-io/armada/pull/1/files).

dfarrell07:
Support Armada-abstracting-kind k8s cluster deployments in CI.

Allows us to use Armada for multicluster management and custom
networking, vs duplicating common bash between many repositories.

Defaults to using Armada, but keeps the old bash-based logic for now. It
can be removed once we have confidence enough to fully move to Armada.
Can also be removed in a distinct commit to make it easy to revert.

Closes #291

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ version ?= 1.14.2
 logging ?= false
 kubefed ?= false
 deploytool ?= helm
+armada ?= true
 debug ?= false
 
 TARGETS := $(shell ls scripts | grep -v dapper-image)
@@ -21,7 +22,7 @@ shell:
 	./.dapper -m bind -s
 
 $(TARGETS): .dapper dapper-image vendor/modules.txt
-	DAPPER_ENV="OPERATOR_IMAGE"  ./.dapper -m bind $@ $(status) $(version) $(logging) $(kubefed) $(deploytool) $(debug)
+	DAPPER_ENV="OPERATOR_IMAGE"  ./.dapper -m bind $@ $(status) $(version) $(logging) $(kubefed) $(deploytool) $(armada) $(debug)
 
 vendor/modules.txt: .dapper go.mod
 	./.dapper -m bind vendor
--- a/package/Dockerfile.dapper-base
+++ b/package/Dockerfile.dapper-base
@@ -30,6 +30,7 @@ ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${ARC
 # findutils      | e2e cleanup (xargs)
 # upx            | binary compression
 # jq             | JSON processing (GitHub API)
+# armada         | abstracting kind to deploy multiple k8s clusters in CI
 RUN dnf -y distrosync && \
     dnf -y install --nodocs --setopt=install_weak_deps=False gcc git-core curl moby-engine make golang kubernetes-client findutils upx jq && \
     dnf -y clean all && \
@@ -39,5 +40,7 @@ RUN dnf -y distrosync && \
     curl -LO "https://github.com/kubernetes-sigs/kubefed/releases/download/v${KUBEFED_VERSION}/kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz" && \
     tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
     curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
+    # FIXME: Use a release from submariner-io/armada GitHub once one is available
+    curl -L https://github.com/dimaunx/armada/releases/download/v0.1.3/armada_0.1.3_linux_amd64.tar.gz | tar -xzf - && mv ./armada /usr/bin/armada && chmod a+x /usr/bin/armada && \
     GOFLAGS="" go get -v github.com/onsi/ginkgo/ginkgo && \
     GOFLAGS="" go get -v golang.org/x/tools/cmd/goimports
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -6,9 +6,6 @@ source $(git rev-parse --show-toplevel)/scripts/lib/version
 
 ### Variables ###
 
-declare -A cluster_CIDRs=( ["cluster1"]="10.244.0.0/16" ["cluster2"]="10.245.0.0/16" ["cluster3"]="10.246.0.0/16" )
-declare -A service_CIDRs=( ["cluster1"]="100.94.0.0/16" ["cluster2"]="100.95.0.0/16" ["cluster3"]="100.96.0.0/16" )
-
 ### Functions ###
 
 function print_logs() {
@@ -28,65 +25,11 @@ function print_logs() {
     fi
 }
 
-function kind_clusters() {
-    status=$1
-    version=$2
-    pids=(-1 -1 -1)
-    logs=()
-    for i in 1 2 3; do
-        if [[ $(kind get clusters | grep cluster${i} | wc -l) -gt 0  ]]; then
-            echo Cluster cluster${i} already exists, skipping cluster creation...
-        else
-            logs[$i]=$(mktemp)
-            echo Creating cluster${i}, logging to ${logs[$i]}...
-            (
-            if [[ -n ${version} ]]; then
-                kind create cluster --image=kindest/node:v${version} --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
-            else
-                kind create cluster --name=cluster${i} --config=${PRJ_ROOT}/scripts/kind-e2e/cluster${i}-config.yaml
-            fi
-            master_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster${i}-control-plane | head -n 1)
-            sed -i -- "s/user: kubernetes-admin/user: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
-            sed -i -- "s/name: kubernetes-admin.*/name: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
-            sed -i -- "s/current-context: kubernetes-admin.*/current-context: cluster$i/g" $(kind get kubeconfig-path --name="cluster$i")
-
-            if [[ ${status} = keep ]]; then
-                cp -r $(kind get kubeconfig-path --name="cluster$i") ${PRJ_ROOT}/output/kind-config/local-dev/kind-config-cluster${i}
-            fi
-
-            sed -i -- "s/server: .*/server: https:\/\/$master_ip:6443/g" $(kind get kubeconfig-path --name="cluster$i")
-            cp -r $(kind get kubeconfig-path --name="cluster$i") ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster${i}
-            ) > ${logs[$i]} 2>&1 &
-            set pids[$i] = $!
-        fi
-    done
-    print_logs "${logs[@]}"
-}
-
-function setup_custom_cni(){
-    for i in 2 3; do
-        if kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=60s > /dev/null 2>&1; then
-            echo "Weave already deployed cluster${i}."
-        else
-            echo "Applying weave network in to cluster${i}..."
-            kubectl --context=cluster${i} apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')&env.IPALLOC_RANGE=${cluster_CIDRs[cluster${i}]}"
-            echo "Waiting for weave-net pods to be ready cluster${i}..."
-            kubectl --context=cluster${i} wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout=300s
-            echo "Waiting for core-dns deployment to be ready cluster${i}..."
-            kubectl --context=cluster${i} -n kube-system rollout status deploy/coredns --timeout=300s
-        fi
-    done
-}
-
-function kind_import_images() {
-    docker tag quay.io/submariner/submariner:$VERSION submariner:local
-    docker tag quay.io/submariner/submariner-route-agent:$VERSION submariner-route-agent:local
-
-    for i in 1 2 3; do
-        echo "Loading submariner images into cluster${i}..."
-        kind --name cluster${i} load docker-image submariner:local
-        kind --name cluster${i} load docker-image submariner-route-agent:local
-    done
+function export_kubeconfig() {
+    # TODO: Are both these mkdirs really necessary?
+    mkdir -p ${DAPPER_SOURCE}/output/kind-config/dapper/
+    mkdir -p ${DAPPER_SOURCE}/output/kind-config/local-dev/
+    export KUBECONFIG=$(echo $kubecfgs_dir/kind-config-cluster{1..3} | sed 's/ /:/g')
 }
 
 function test_connection() {
@@ -113,12 +56,12 @@ function enable_logging() {
     else
         echo Installing Elasticsearch...
         es_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' cluster1-control-plane | head -n 1)
-        kubectl --context=cluster1 apply -f ${PRJ_ROOT}/scripts/kind-e2e/logging/elasticsearch.yaml
-        kubectl --context=cluster1 apply -f ${PRJ_ROOT}/scripts/kind-e2e/logging/filebeat.yaml
+        kubectl --context=cluster1 apply -f ${DAPPER_SOURCE}/scripts/kind-e2e/logging/elasticsearch.yaml
+        kubectl --context=cluster1 apply -f ${DAPPER_SOURCE}/scripts/kind-e2e/logging/filebeat.yaml
         echo Waiting for Elasticsearch to be ready...
         kubectl --context=cluster1 wait --for=condition=Ready pods -l app=elasticsearch --timeout=300s
         for i in 2 3; do
-            kubectl --context=cluster${i} apply -f ${PRJ_ROOT}/scripts/kind-e2e/logging/filebeat.yaml
+            kubectl --context=cluster${i} apply -f ${DAPPER_SOURCE}/scripts/kind-e2e/logging/filebeat.yaml
             kubectl --context=cluster${i} set env daemonset/filebeat -n kube-system ELASTICSEARCH_HOST=${es_ip} ELASTICSEARCH_PORT=30000
         done
     fi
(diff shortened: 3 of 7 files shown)


Makefile
```diff
@@ -3,6 +3,7 @@ version ?= 1.14.2
 logging ?= false
 kubefed ?= false
 deploytool ?= helm
+armada ?= true
```
mangelajo: I think, adopting armada means dropping all the bash/kind part. Otherwise we need to maintain 2 different versions.

Let's cleanup and reduce the complexity.

scripts/kind-e2e/lib_armada_deploy_kind.sh
```diff
@@ -0,0 +1,47 @@
...
+         -o /go/bin/armada
+    chmod a+x /go/bin/armada
+    /go/bin/armada version
+}
+
+function create_kind_clusters() {
+    install_armada
+
+    # Armada -n basically adds clusters to make sure there are at least n
+    /go/bin/armada create clusters -n 1 --tiller
```
mangelajo: I think we only need tiller when deploying with helm.

scripts/kind-e2e/lib_armada_deploy_kind.sh
```diff
@@ -0,0 +1,47 @@
...
+
+kubecfgs_rel_dir=scripts/output/kube-config/container/
+kubecfgs_dir=${PRJ_ROOT}/$kubecfgs_rel_dir
+
+### Functions ###
+
+function install_armada() {
+    test -x /go/bin/armada && return
+    # FIXME: Use a release from GitHub once one is available
+    curl -L https://drive.google.com/u/0/uc\?id\=17r6E0RwbHcvlaFl0LKFICuAc_eGb13W9\&export\=download \
```
mangelajo: I believe we should add this binary to the base dapper image instead.

scripts/kind-e2e/lib_armada_deploy_kind.sh
```diff
@@ -0,0 +1,47 @@
...
+
+function import_subm_images() {
+    docker tag quay.io/submariner/submariner:dev submariner:local
+    docker tag quay.io/submariner/submariner-route-agent:dev submariner-route-agent:local
+
+    /go/bin/armada load docker-images --clusters cluster1,cluster2,cluster3 --images submariner:local,submariner-route-agent:local
+}
+
+function destroy_kind_clusters() {
+    # FIXME: Don't depend on armada outside of container
```
mangelajo: What do you mean?

scripts/kind-e2e/lib_bash_deploy_kind.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+# This should only be sourced
+if [ "${0##*/}" = "lib_helm_deploy_subm.sh" ]; then
```
mangelajo: Let's ditch bash, no need to maintain and test all combinations of all things

scripts/kind-e2e/lib_armada_deploy_kind.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+# This should only be sourced
+if [ "${0##*/}" = "lib_helm_deploy_subm.sh" ]; then
+    echo "Don't run me, source me" >&2
+    exit 1
+fi
+
+### Variables ###
+
+# FIXME: Armada support for setting pod/service CIRDs via flags?
```
mangelajo: I would keep the CIDRs may be as part of the top level e2e.sh or any other script where we set variables with the deployment details.

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -30,13 +30,13 @@ function deploytool_prereqs() {
 function setup_broker() {
     context=$1
     echo Installing broker on $context.
-    subctl --kubeconfig ${PRJ_ROOT}/output/kind-config/dapper/kind-config-$context deploy-broker --no-dataplane
+    subctl --kubeconfig ${kubecfgs_dir}/kind-config-$context deploy-broker --no-dataplane
```
mangelajo: Don't we need to do similar change on the helm side?

scripts/kind-e2e/lib_operator_deploy_subm.sh
```diff
@@ -30,13 +30,13 @@ function deploytool_prereqs() {
 function setup_broker() {
     context=$1
     echo Installing broker on $context.
```
mangelajo: unrelated to this patch, but before I forget, we need to add "kubectl config use-context $context" here, otherwise if we run the deploy a second time it will fail here, because the config files are at some point modified for an specific cluster config, and then it doesn't work here (using cluster1 here, but context has been set to cluster3 on all config files after a first run)

scripts/kind-e2e/lib_armada_deploy_kind.sh
mkolesnik: +1

I suggest we can take the binary from dima's repo [1] for now until it's merged in submariner-io

[1] https://github.com/dimaunx/armada/releases

package/Dockerfile.dapper-base
```diff
@@ -39,5 +40,7 @@ RUN dnf -y distrosync && \
     curl -LO "https://github.com/kubernetes-sigs/kubefed/releases/download/v${KUBEFED_VERSION}/kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz" && \
     tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
     curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
+    # FIXME: Use a release from GitHub once one is available
+    curl -Lo /usr/bin/armada https://drive.google.com/u/0/uc\?id\=17r6E0RwbHcvlaFl0LKFICuAc_eGb13W9\&export\=download && chmod a+x /usr/bin/armada && \
```
mkolesnik: I suggest we can take the binary from dima's repo [1] for now until it's merged in submariner-io

[1] https://github.com/dimaunx/armada/releases

scripts/kind-e2e/lib_armada_deploy_kind.sh
```diff
@@ -0,0 +1,46 @@
...
+function create_kind_clusters() {
+    # FIXME: Remove this function after CI validates the rest of the patch
+    install_armada
+
+    # FIXME: Install tiller when deploying with helm
+    /usr/bin/armada create clusters -n 3 --weave
+}
+
+function import_subm_images() {
+    docker tag quay.io/submariner/submariner:dev submariner:local
```
mkolesnik: Do you really need to re-tag them? Can't armada just load the original image tags?

mangelajo: I think the deployment expects to find :local images. It's also an indicative we use to know that the deployments should not "pull always" (subctl detects that, and we force it into helm I think)

scripts/kind-e2e/e2e.sh
```diff
@@ -5,8 +5,7 @@ source $(git rev-parse --show-toplevel)/scripts/lib/debug_functions
 
 ### Variables ###
 
-declare -A cluster_CIDRs=( ["cluster1"]="10.244.0.0/16" ["cluster2"]="10.245.0.0/16" ["cluster3"]="10.246.0.0/16" )
-declare -A service_CIDRs=( ["cluster1"]="100.94.0.0/16" ["cluster2"]="100.95.0.0/16" ["cluster3"]="100.96.0.0/16" )
+PRJ_ROOT=$(git rev-parse --show-toplevel)
```
mangelajo: you don't need to calculate that, there's a variable in dapper for the prj root.

scripts/kind-e2e/lib_armada_deploy_kind.sh
mangelajo: I think the deployment tools expect the :local tag when we don't want to re-pull a remote image.

mangelajo:
@dfarrell07 ping

scripts/kind-e2e/lib_armada_deploy_kind.sh
dfarrell07: Removed, thanks

scripts/kind-e2e/lib_armada_deploy_kind.sh
dfarrell07: Fixed in the latest (and #357 added to test), thanks!

scripts/kind-e2e/lib_operator_deploy_subm.sh
dfarrell07: This was done elsewhere, has since been added via rebase.

Makefile
dfarrell07: As we discussed briefly on a call a while back, keeping the option while Armada is a PoC reduces risk to Submariner-proper if Armada turns out to have problems. I could remove the option in a follow-up commit, which would allow us to easily revert if necessary.

scripts/kind-e2e/e2e.sh
dfarrell07: Nice, done in the latest, thanks!

scripts/kind-e2e/lib_armada_deploy_kind.sh
dfarrell07: I'm not sure I understand this one. The CIDRs vars are in different libs because they have different values for the two deployment methods. They can't be extracted to a common place like e2e.sh without naming them per-deployment-method, and breaking the model of using libs with parallel vars/fns names and no conditionals in common code.

dfarrell07:
That's twice now that the Helm job has failed, then passed when retriggered. :\

scripts/kind-e2e/lib_armada_deploy_kind.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# This should only be sourced
+if [ "${0##*/}" = "lib_helm_deploy_subm.sh" ]; then
```
tpantelis: Shouldn't this be "lib_armada_deploy_kind.sh"?

scripts/kind-e2e/lib_bash_deploy_kind.sh
tpantelis: Shouldn't this be "lib_bash_deploy_kind.sh"?

scripts/kind-e2e/lib_armada_deploy_kind.sh
```diff
@@ -0,0 +1,39 @@
...
+    echo "Don't run me, source me" >&2
+    exit 1
+fi
+
+### Variables ###
+
+# FIXME: Armada support for getting these values, vs hardcoding?
+declare -A cluster_CIDRs=( ["cluster1"]="10.4.0.0/14" ["cluster2"]="10.8.0.0/14" ["cluster3"]="10.12.0.0/14" )
+declare -A service_CIDRs=( ["cluster1"]="100.1.0.0/16" ["cluster2"]="100.2.0.0/16" ["cluster3"]="100.3.0.0/16" )
```
tpantelis: FWIR, you can query the cluster for the CIDRs via `kubectl cluster-info dump`.

scripts/kind-e2e/lib_armada_deploy_kind.sh
dfarrell07: Good catch, thanks. Fixed in the latest.

scripts/kind-e2e/lib_bash_deploy_kind.sh
dfarrell07: Good catch, thanks. Fixed in the latest.

scripts/kind-e2e/lib_armada_deploy_kind.sh
dfarrell07: I see pod_CIDRs there for each node ("podCIDR": "10.8.0.0/24"), which seem to be related but not the same as the cluster_CIDRs here. I also see the cluster and service CIDR vars ({ "name": "SUBMARINER_CLUSTERCIDR", "value": "10.8.0.0/14" }, { "name": "SUBMARINER_SERVICECIDR", "value": "100.2.0.0/16" }) that are a result of these being used by Helm, but that of course is a chicken-and-egg situation. Am I missing something else, or can you be more specific about what you have in mind? Thanks!

dfarrell07: Tracking with https://github.com/submariner-io/submariner/issues/372.

sridhargaddam:
> Checking runtimes, don't merge yet

@mkolesnik we are kind of blocked on this PR for overlapping cidrs use-case. I understand your concern regarding the runtime. Since we are not deleting the existing bash installation, can we get this PR in and in-parallel work on improving/fixing the deployment time?

mangelajo:
@mkolesnik can you revisit your vote if you think it's ok to go on with armada?

dfarrell07:
@mkolesnik talking with @tpantelis, @mangelajo, @skitt, @sridhargaddam, @manosnoam, @aswinsuryan and others about the de-duplication effort (including Go vs Bash Armada-esque logic), my summary of the relevant bits:

* Go is easier to test and maintain than Bash.
* The current Go-based implementation is much more flexible than the current Bash-based implementation. For example, it can deploy any number of nodes with any combination of CNIs and other extras.
* The current Go-based implementation has support for deploying with overlapping CIDRs, which is needed quickly by others. That feature would have to be added to the Bash-based logic.
* We don't expect significant performance differences between the two implementations, as the costly steps are similar. Either implementation will need refactoring in that regard.

IMHO let's merge this ASAP, making it easier for myself and others to do related follow-up work.

mkolesnik:
Really sorry guys I was out sick most of last week and didn't even touch the computer.

What I observed is that the time is more or less the same, BASH is much less complex, and armada is much more taxing on the machine (one run caused load to go to 50x, i.e. 5000% CPU utilization).

I was working on an alternative to improve the bash scripts deployment to make it more maintainable and extendible. Since I was sick I didn't have time to continue on it most of last week, but I pushed it as PR #425 so that you can take a look.

I was rooting for Armada, but replacing ~80 lines of bash code with ~3000 lines of go code (and many hundreds of YAML lines) won't make it better IMHO. The main promise of Armada was to improve the build, and I feel it didn't deliver.

BTW regarding the local registry I think regardless of bash/armada we should check out this possibility to reduce the build time and perhaps make the whole build even simpler.

Let's discuss this a bit further before we push a whole bunch of complicated code that doesn't add much benefit.

tpantelis:
> I was rooting for Armada, but replacing ~80 lines of bash code with ~3000 lines of go code (and many hundreds of YAML lines) won't make it better IMHO. The main promise of Armada was to improve the build, and I feel it didn't deliver.

From my perspective and understanding, the main purpose of armada was for de-duplication.

dfarrell07:
> I was working on an alternative to improve the bash scripts deployment to make it more maintainable and extendible. Since I was sick I didn't have time to continue on it most of last week, but I pushed it as PR #425 so that you can take a look.

Some nice ideas there, but I'm concerned that we're continuing to develop both implementations, vs making a decision and focusing on one. For example, your work to make the create nginx/netshoot functions generic is already implemented as `armada deploy netshoot nginx-demo` in Armada. The work to refactor the `create_kind_cluster` function would also become obsolete.

> I was rooting for Armada, but replacing ~80 lines of bash code with ~3000 lines of go code (and many hundreds of YAML lines) won't make it better IMHO.

We've talked about this concern quite a lot, and it seems there's more nuance. Using Go certainly does add more code, but it also seems to make the code more flexible, maintainable and testable. I tried to summarize some of those thoughts above, but here are a few more:

* Most of the Go-based Armada code is actually test code that adds substantial coverage not provided by the Bash path.
* A substantial additional chunk of the additional LoC are related to the proper command structure of Go-based Armada, which isn't a feature currently provided in Bash. It'd take a lot of scripting around the Bash control flow (already overly complex) to do more than our current static/prescribed deployments. This flexibility out-of-the-box with Go-based Armada would enable devs to easily add scenarios, ie scale deployments or mixed-feature deployments.
* From what I see of the YAML, it's either providing a feature not provided by Bash (Calico, Flannel) or already exist in the Bash path (Netshoot, Nginx, Weave).

> The main promise of Armada was to improve the build, and I feel it didn't deliver.

As @tpantelis said and we talked about quite a lot last week, the consensus seems to be that the goal of "Armada" in the architure is to de-duplicate managing multicluster deployments in CI. The question is how to best architect that. The current ideas are basically:

1) Refactor the Bash-based "Armada" to be generic/flexible enough to be used in all submariner-io/* deployment needs. Share that Bash in CI, likely via the dapper-base image.
2) Extract all the "deploying K8s multiclusters with kind" logic into Go-based Armada (as is already done). Share and consume in CI parallel to the current use of `subctl` for the Submariner abstraction layer. Very simple Bash scripting to basically pass params from the Makefile to Armada.

Finally, I'll highlight that at least having Go-based Armada as an option would unblock a substantial number of PRs that have been waiting on this one to go in.

mkolesnik:
> > I was working on an alternative to improve the bash scripts deployment to make it more maintainable and extendible. Since I was sick I didn't have time to continue on it most of last week, but I pushed it as PR #425 so that you can take a look.
> 
> Some nice ideas there @mkolesnik, but I'm concerned that we're continuing to develop both implementations, vs making a decision and focusing on one. For example, your work to make the create nginx/netshoot functions generic is already implemented as `armada deploy netshoot nginx-demo` in Armada. The work to refactor the `create_kind_cluster` function would also become obsolete.

Yes though my patch turns it into a generic 5 line function (first 4 lines are more for clarity than actual logic) which can deploy anything.
The armada, OTOH, has dedicated "command" which are specific to "netshoot" and "nginx" and are 100s of LOC. Say you want to deploy another thing, mysql, then what would you do? Add 100s more lines for such a command?

> 
> > I was rooting for Armada, but replacing ~80 lines of bash code with ~3000 lines of go code (and many hundreds of YAML lines) won't make it better IMHO.
> 
> We've talked about this concern quite a lot, and it seems there's more nuance. Using Go certainly does add more code, but it also seems to make the code more flexible, maintainable and testable. I tried to summarize some of those thoughts above, but here are a few more:

Using go also adds a lot of boierplate code, don't forget.

> 
>     * Most of the Go-based Armada code is actually test code that adds substantial coverage not provided by the Bash path.

If you test needless functions, is that test really worth something?

> 
>     * Another substantial chunk of the additional LoC is related to the proper command structure of Go-based Armada, which isn't a feature currently provided in Bash. It'd take a lot of scripting around the Bash control flow (already overly complex) to do more than our current static/prescribed deployments. This flexibility out-of-the-box with Go-based Armada would enable devs to easily add scenarios, ie scale deployments or mixed-feature deployments.

Which is overkill right now for the bash, but can be added with not so much effort should we actually decide we need it.

> 
>     * From what I see of the YAML, it's either providing a feature not provided by Bash (Calico, Flannel) or already exist in the Bash path (Netshoot, Nginx, Weave).

Yes as I said there's a lot of YAMLs there as well with logic hidden in them, it's not better of worse than the bash, except for all those "features" that you mentioned that we aren't actually even using.

> 
> 
> > The main promise of Armada was to improve the build, and I feel it didn't deliver.
> 
> As @tpantelis said and we talked about quite a lot last week, the consensus seems to be that the goal of "Armada" in the architure is to de-duplicate managing multicluster deployments in CI. The question is how to best architect that. The current ideas are basically:

From my many discussions with Dima on this project, he emphasized the "time saving" as one of the key features in Aramada.

> 
>     1. Refactor the Bash-based "Armada" to be generic/flexible enough to be used in all submariner-io/* deployment needs. Share that Bash in CI, likely via the dapper-base image.
> 
>     2. Extract all the "deploying K8s multiclusters with kind" logic into Go-based Armada (as is already done). Share and consume in CI parallel to the current use of `subctl` for the Submariner abstraction layer. Very simple Bash scripting to basically pass params from the Makefile to Armada.

On this one I agree, deduplication can be either done solely in bash (and probably on the dapper image level as you suggest) or in Armada. Though in the end you'll still have a bash wrapper that you'll need to (possibly) use everywhere as well, so not sure how much it really saves us.

> 
> 
> Finally, I'll highlight that at least having Go-based Armada as an option would unblock a substantial number of PRs that have been waiting on this one to go in.

Do you have a list of them?

dfarrell07:
After extensive discussions (and this prototyping work, to show the details) with @tpantelis, @mkolesnik and many others, we decided not to go with Go-based Armada.

PR #335, opened 2020-02-10
Fix pipefail travis

mangelajo:


mangelajo:
Thanks for the quick reviews.

PR #342, opened 2020-02-11
Log level constants

mangelajo:


mangelajo:
ooops
mixed up commit, sorry

pkg/log/loglevels.go
```diff
@@ -0,0 +1,8 @@
+package log
+
+const (
+	TRACE_LEVEL   = 100
+	DEBUG_LEVEL   = 10
+	VERBOSE_LEVEL = 4
+	INFO_LEVEL    = 0
```
mangelajo: Sorry I think I got it wrong, and default should be 1?

tpantelis: I don't think we need INFO_LEVEL - one would just use `klog.Infof(...)`.

mangelajo: Probably we should remove "INFO_LEVEL" altogether and then, use
klog.Info( ... ) for "always-on" logs.

pkg/log/loglevels.go
```diff
@@ -0,0 +1,8 @@
+package log
+
+const (
+	TRACE_LEVEL   = 100
+	DEBUG_LEVEL   = 10
+	VERBOSE_LEVEL = 4
```
tpantelis: What does VERBOSE_LEVEL mean? 

I was thinking:

TRACE = 3
DEBUG = 2

mangelajo: sounds good

pkg/log/loglevels.go
```diff
@@ -0,0 +1,22 @@
+package log
+
+// Log levels defined for use with:
+//   klog.V(__).Info
+
+const (
+	// INFO_LEVEL : This level is for anything which does not happen very often, and while not being
+	//              an error, needs to be on the logs, because it's important information:
```
tpantelis:
```suggestion
	//              an error, is important information and helpful to be in the logs, eg:
```

pkg/log/loglevels.go
```diff
@@ -0,0 +1,22 @@
...
+	//                * startup information
+	//                * HA failovers
+	//                * re-connections/disconnections
+	//                * ...
+	//              This level is not defined because you must use klog.Info helpers
```
tpantelis:
```suggestion
	//              This level is not specifically defined as you would use the klog.Info helpers.
```

pkg/log/loglevels.go
```diff
@@ -0,0 +1,22 @@
...
+	//
+	// DEBUG_LEVEL : should be used to provide logs on often occurring events but, which could be helpful
```
tpantelis:
```suggestion
	// DEBUG_LEVEL : used to provide logs for often occurring events that could be helpful
```

pkg/log/loglevels.go
```diff
@@ -0,0 +1,22 @@
...
+	//               for debugging errors.
+	DEBUG_LEVEL = 2
+	// TRACE_LEVEL : must be used for debug level errors dumping lots of information which generally would
```
tpantelis:
```suggestion
	// TRACE_LEVEL : used for logging that occurs often or may dump a lot of information which generally 
        //               would be less useful for debugging but can be useful in some cases, for example tracing
	//               function entry/exit, parameters, structures, etc..
```

pkg/log/loglevels.go
```diff
@@ -0,0 +1,22 @@
...
+	//              an error, needs to be on the logs, because it's important information:
+	//                * startup information
+	//                * HA failovers
+	//                * re-connections/disconnections
+	//                * ...
+	//              This level is not defined because you must use klog.Info helpers
+	//
+	// DEBUG_LEVEL : should be used to provide logs on often occurring events but, which could be helpful
+	//               for debugging errors.
+	DEBUG_LEVEL = 2
```
tpantelis: Do you think we can omit the "_LEVEL" part to reduce real estate? I think `klog.V(log.DEBUG).Infof(...)` reads clearly.

sridhargaddam: Big +1 to this.

mangelajo: I like that, will do it.

mangelajo:
Talking to tpantelis on whatsapp, this should be fine to merge, he's on the road and PTO.

PR #344, opened 2020-02-12
Hide the Set field and add Elements method to StringSet

tpantelis:
StringSet exposed the Set field publicly which is unsafe.
The Set was accessed and iterated unprotected by code in route.go.

Made the field package-scoped and added an Elements method to return
a new []string. Also updated the unit tests.

### DIFF
--- a/pkg/util/stringset.go
+++ b/pkg/util/stringset.go
@@ -6,21 +6,21 @@ import (
 
 type StringSet struct {
 	syncMutex *sync.Mutex
-	Set       map[string]bool
+	set       map[string]bool
 }
 
 func NewStringSet() *StringSet {
 	return &StringSet{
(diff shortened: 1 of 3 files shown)


pkg/util/stringset.go
```diff
@@ -6,42 +6,56 @@ import (
...
+	delete(set.set, s)
+}
+
+func (set *StringSet) Elements() []string {
+	set.syncMutex.Lock()
+	defer set.syncMutex.Unlock()
+
+	elements := make([]string, len(set.set))
+	i := 0
+	for v, _ := range set.set {
```
skitt: This trips linting, it should be

    for v := range set.set {

tpantelis: Done

Issue #341, opened 2020-02-11
During E2E gateway failovers the route-agent will crash sometimes


mangelajo:
```
I0211 13:49:45.613051       1 reflector.go:202] github.com/submariner-io/submariner/pkg/client/informers/externalversions/factory.go:117: forcing resync
I0211 13:49:45.613181       1 route.go:476] Enqueueing endpoint for route controller &{{ } {cluster2-submariner-cable-cluster2-172-17-0-5  submariner-operator /apis/submariner.io/v1/namespaces/submariner-operator/endpoints/cluster2-submariner-cable-cluster2-172-17-0-5 31886450-4cd5-11ea-98c1-0242ac110009 176653 1 2020-02-11 13:48:33 +0000 UTC <nil> <nil> map[] map[] [] nil [] } {cluster2 submariner-cable-cluster2-172-17-0-5 cluster2-worker [100.95.0.0/16 10.245.0.0/16] 172.17.0.5  false ipsec map[]}}
I0211 13:49:45.613261       1 route.go:476] Enqueueing endpoint for route controller &{{Endpoint submariner.io/v1} {cluster3-submariner-cable-cluster3-172-17-0-6  submariner-operator /apis/submariner.io/v1/namespaces/submariner-operator/endpoints/cluster3-submariner-cable-cluster3-172-17-0-6 d7a2545e-4bf1-11ea-98c1-0242ac110009 1114 1 2020-02-10 10:41:06 +0000 UTC <nil> <nil> map[] map[] [] nil [] } {cluster3 submariner-cable-cluster3-172-17-0-6 cluster3-worker [100.96.0.0/16 10.246.0.0/16] 172.17.0.6  false ipsec map[]}}
I0211 13:49:45.618420       1 route.go:394] Handling object in handleEndpoint
I0211 13:49:45.618442       1 route.go:395] Processing endpoint object: submariner-operator/cluster3-submariner-cable-cluster3-172-17-0-6
I0211 13:50:01.601481       1 route.go:500] Handling object in handleEndpoint
[...]

PR #345, opened 2020-02-12
Route agent logs refactor

mangelajo:
This cleans up logging on the route-agent and fixes some of the issues seen, logging them and failing more gracefully. 

Related-Issue: #341

### DIFF
--- a/pkg/apis/submariner.io/v1/string.go
+++ b/pkg/apis/submariner.io/v1/string.go
@@ -1,0 +1,16 @@
+package v1
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// String method returns a string representation, for human readability.
+func (e *Endpoint) String() string {
+
+	bytes, err := json.Marshal(e)
+	if err != nil {
+		return fmt.Sprintf("%#v", e)
+	}
+	return string(bytes)
+}
--- a/pkg/apis/submariner.io/v1/string_test.go
+++ b/pkg/apis/submariner.io/v1/string_test.go
@@ -1,0 +1,36 @@
+package v1
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const expectedString = `{"metadata":{"creationTimestamp":null},"spec":{"cluster_id":"cluster-id","cable_name":` +
+	`"cable-1","hostname":"","subnets":["10.0.0.0/24","172.0.0.0/24"],"private_ip":"1.1.1.1",` +
+	`"public_ip":"","nat_enabled":false,"backend":""}}`
+
+var _ = Describe("API v1", func() {
+	When("Endpoint String representation called", func() {
+		It("Should return a human readable string", func() {
+
+			endpoint := Endpoint{
+				Spec: EndpointSpec{
+					ClusterID: "cluster-id",
+					Subnets:   []string{"10.0.0.0/24", "172.0.0.0/24"},
+					CableName: "cable-1",
+					PublicIP:  "",
+					PrivateIP: "1.1.1.1",
+				},
+			}
+
+			Expect(endpoint.String()).To(Equal(expectedString))
+		})
+	})
+})
+
+func TestApiMethods(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "V1 Api Method suite")
+}
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -15,6 +15,7 @@ import (
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	clientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	informers "github.com/submariner-io/submariner/pkg/client/informers/externalversions/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/log"
 	"github.com/submariner-io/submariner/pkg/util"
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
@@ -147,7 +148,7 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
 	klog.Infof("Starting Route Controller. ClusterID: %s, localClusterCIDR: %v, localServiceCIDR: %v", r.clusterID, r.localClusterCidr, r.localServiceCidr)
 
 	// Wait for the caches to be synced before starting workers
-	klog.Info("Waiting for endpoint informer caches to sync.")
+	klog.Info("Waiting for Endpoint informer caches to sync.")
 	if ok := cache.WaitForCacheSync(stopCh, r.endpointsSynced, r.smRouteAgentPodsSynced); !ok {
 		return fmt.Errorf("failed to wait for caches to sync")
 	}
@@ -187,16 +188,16 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
 	}
 
 	for index, pod := range podList.Items {
-		klog.V(4).Infof("In %s, podIP of submariner-route-agent[%d] is %s", r.clusterID, index, pod.Status.PodIP)
+		klog.V(log.DEBUG).Infof("In %s, podIP of submariner-route-agent[%d] is %s", r.clusterID, index, pod.Status.PodIP)
 		r.populateRemoteVtepIps(pod.Status.PodIP)
 	}
 
-	klog.Info("Starting workers")
 	go wait.Until(r.runEndpointWorker, time.Second, stopCh)
 	go wait.Until(r.runPodWorker, time.Second, stopCh)
 	wg.Wait()
+	klog.Info("Route agent workers started")
 	<-stopCh
-	klog.Info("Shutting down workers")
+	klog.Info("Route agent stopping")
 	return nil
 }
 
@@ -302,7 +303,7 @@ func (r *Controller) createVxLANInterface(ifaceType int, gatewayNodeIP net.IP) e
 		if err != nil {
 			return fmt.Errorf("unable to update vxlan rp_filter proc entry, err: %s", err)
 		} else {
-			klog.Info("Successfully configured rp_filter to loose mode(2) ")
+			klog.V(log.DEBUG).Infof("Successfully configured rp_filter to loose mode(2) on %s", VxLANIface)
 		}
 
 	} else if ifaceType == VxInterfaceWorker {
@@ -357,7 +358,9 @@ func (r *Controller) processNextPod() bool {
 			return fmt.Errorf("error retrieving submariner-route-agent pod object %s: %v", name, err)
 		}
 
-		klog.V(4).Infof("In processNextPod, POD HostIP is %s", pod.Status.HostIP)
+		klog.V(log.DEBUG).Infof("Processing submariner-route-agent pod %q with host IP %q, pod IP %q",
+			name, pod.Status.HostIP, pod.Status.PodIP)
+
 		r.populateRemoteVtepIps(pod.Status.PodIP)
 
 		r.gwVxLanMutex.Lock()
@@ -372,6 +375,8 @@ func (r *Controller) processNextPod() bool {
 					r.podWorkqueue.Forget(obj)
 					return fmt.Errorf("failed to add FDB entry on the Gateway Node vxlan iface %v", err)
 				}
+				klog.Infof("FDB entry added on the Gateway node's vxlan iface for "+
+					"route-agent pod %q, IP %q", pod.Name, pod.Status.PodIP)
 			} else {
 				r.podWorkqueue.AddRateLimited(obj)
 				klog.Errorf("vxlanDevice is not yet created on the Gateway node")
@@ -380,7 +385,7 @@ func (r *Controller) processNextPod() bool {
 		}
 
 		r.podWorkqueue.Forget(obj)
-		klog.V(4).Infof("Pod event processed by route controller")
+		klog.V(log.DEBUG).Infof("Successfully processed submariner-route-agent pod %q", name)
 		return nil
 	}()
 
@@ -399,9 +404,8 @@ func (r *Controller) processNextEndpoint() bool {
 	}
 	err := func() error {
 		defer r.endpointWorkqueue.Done(obj)
-		klog.V(4).Infof("Handling object in handleEndpoint")
-		klog.V(4).Infof("Processing endpoint object: %v", obj)
 		key := obj.(string)
+
 		ns, name, err := cache.SplitMetaNamespaceKey(key)
 		if err != nil {
 			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
@@ -419,35 +423,44 @@ func (r *Controller) processNextEndpoint() bool {
 			return fmt.Errorf("error retrieving submariner endpoint object %s: %v", name, err)
 		}
 
+		klog.V(log.TRACE).Infof("Processing endpoint %q : %s", name, endpoint)
+
 		if endpoint.Spec.ClusterID != r.clusterID {
-			klog.V(6).Infof("setting routes of endpoint object %s", name)
+			klog.Infof("Updating iptable rules for remote cluster's Endpoint %q with subnets %v",
+				endpoint.Name, endpoint.Spec.Subnets)
 			r.updateIptableRulesForInterclusterTraffic(endpoint.Spec.Subnets)
 			r.endpointWorkqueue.Forget(obj)
 			return nil
 		}
+
+		klog.Infof("Processing local Endpoint %s with IP %s, Host %s for this cluster",
+			endpoint.Name, endpoint.Spec.PrivateIP, endpoint.Spec.Hostname)
 
 		hostname, err := os.Hostname()
 		if err != nil {
 			klog.Fatalf("unable to determine hostname: %v", err)
 		}
 
-		klog.V(6).Infof("Local Cluster Gateway Node IP is %s", endpoint.Spec.PrivateIP)
-
 		// If the endpoint hostname matches with our hostname, it implies we are on gateway node
 		if endpoint.Spec.Hostname == hostname {
-			r.cleanRoutes()
+			r.cleanVxSubmarinerRoutes()
+			klog.Infof("This route agent is running on the active gateway node")
+
 			r.gwVxLanMutex.Lock()
 			defer r.gwVxLanMutex.Unlock()
 
 			r.isGatewayNode = true
+			klog.Infof("Creating the vxlan interface: %s on the gateway node", VxLANIface)
 			err = r.createVxLANInterface(VxInterfaceGateway, nil)
 			if err != nil {
-				klog.Fatalf("Unable to create VxLAN interface on GatewayNode (%s): %v", hostname, err)
+				klog.Fatalf("Unable to create VxLAN interface on gateway node (%s): %v", hostname, err)
 			}
-			klog.V(6).Infof("not reconciling routes because we appear to be the gateway host")
+
 			r.endpointWorkqueue.Forget(obj)
 			return nil
 		}
+
+		klog.Infof("This route agent is running on a non-gateway/non-active node")
 
 		localClusterGwNodeIP := net.ParseIP(endpoint.Spec.PrivateIP)
 		remoteVtepIP, err := r.getVxlanVtepIPAddress(localClusterGwNodeIP.String())
@@ -458,6 +471,7 @@ func (r *Controller) processNextEndpoint() bool {
 
 		r.gwVxLanMutex.Lock()
 		r.isGatewayNode = false
+		klog.Infof("Creating the vxlan interface %s with gateway node IP %s", VxLANIface, localClusterGwNodeIP)
 		err = r.createVxLANInterface(VxInterfaceWorker, localClusterGwNodeIP)
 		if err != nil {
 			klog.Fatalf("Unable to create VxLAN interface on non-GatewayNode (%s): %v", endpoint.Spec.Hostname, err)
@@ -475,7 +489,7 @@ func (r *Controller) processNextEndpoint() bool {
 		}
 
 		r.endpointWorkqueue.Forget(obj)
-		klog.V(4).Infof("endpoint processed by route controller")
+		klog.V(log.DEBUG).Infof("Successfully processed local Endpoint %q", endpoint.Name)
 		return nil
 	}()
 
@@ -494,7 +508,7 @@ func (r *Controller) enqueueEndpoint(obj interface{}) {
 		utilruntime.HandleError(err)
 		return
 	}
-	klog.V(4).Infof("Enqueueing endpoint for route controller %v", obj)
+	klog.V(log.DEBUG).Infof("Enqueueing endpoint for route controller %v", obj)
 	r.endpointWorkqueue.AddRateLimited(key)
 }
 
@@ -509,7 +523,7 @@ func (r *Controller) enqueuePod(obj interface{}) {
 	pod := obj.(*k8sv1.Pod)
 	// Add the POD event to the workqueue only if the sm-route-agent podIP does not exist in the local cache.
 	if pod.Status.HostIP != "" && !r.remoteVTEPs.Contains(pod.Status.HostIP) {
-		klog.V(4).Infof("Enqueueing sm-route-agent-pod event, ip: %s", pod.Status.HostIP)
+		klog.V(log.DEBUG).Infof("Enqueueing sm-route-agent-pod event, ip: %s", pod.Status.HostIP)
 		r.podWorkqueue.AddRateLimited(key)
 	}
 }
@@ -518,7 +532,7 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
 	// ideally we should attempt to remove all routes if the endpoint matches our cluster ID
 	var object *v1.Endpoint
 	var ok bool
-	klog.V(4).Infof("Handling object in handleEndpoint")
+	klog.V(log.TRACE).Infof("Handling object in handleRemoveEndpoint")
 	if object, ok = obj.(*v1.Endpoint); !ok {
 		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
 		if !ok {
@@ -530,35 +544,47 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
 			klog.Errorf("Could not convert object tombstone %v to an Endpoint", tombstone.Obj)
 			return
 		}
-		klog.V(4).Infof("Recovered deleted object '%s' from tombstone", object.GetName())
 	}
-	klog.V(4).Infof("Informed of removed endpoint for route controller object: %v", object)
+	klog.V(log.DEBUG).Infof("Informed of removed endpoint: %v", object.String())
 	hostname, err := os.Hostname()
 	if err != nil {
 		klog.Fatalf("Could not retrieve hostname: %v", err)
 	}
 	if object.Spec.Hostname == hostname {
-		r.cleanRoutes()
+		r.cleanVxSubmarinerRoutes()
+		klog.V(log.DEBUG).Infof("Removed routes from host")
 	}
 
 	if object.Spec.ClusterID == r.clusterID {
 		r.gwVxLanMutex.Lock()
 		defer r.gwVxLanMutex.Unlock()
 
-		klog.V(6).Infof("Endpoint matches the cluster ID of this cluster")
+		if object.Spec.Hostname == hostname {
+			klog.Infof("Local Gateway Endpoint (IP %s) removed: deleting vxlan interface",
+				object.Spec.PrivateIP)
+		} else {
+			klog.Infof("Gateway Endpoint (IP %s) for our cluster removed: deleting vxlan interface",
+				object.Spec.PrivateIP)
+		}
+
 		err := r.vxlanDevice.deleteVxLanIface()
+
 		if err != nil {
-			klog.Errorf("Failed to delete the the vxlan interface on endpoint removal: %v", err)
+			klog.Errorf("Failed to delete the the vxlan interface on endpoint removal: %s", err.Error())
 			return
 		}
 		r.vxlanDevice = nil
-	}
 
-	klog.V(4).Infof("Removed routes from host")
+	} else {
+		// TODO: Handle a remote endpoint removal use-case
+		//         - remove routes to remote cluster
+		//         - remove related iptable rules
+		return
+	}
 }
 
 func (r *Controller) handleRemovedPod(obj interface{}) {
-	klog.V(6).Infof("Removing podIP in route controller %v", obj)
+	klog.V(log.DEBUG).Infof("Removing podIP in route controller %v", obj)
 	pod := obj.(*k8sv1.Pod)
 
 	if r.remoteVTEPs.Contains(pod.Status.HostIP) {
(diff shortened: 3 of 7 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -419,31 +420,36 @@ func (r *Controller) processNextEndpoint() bool {
...
 
 			r.isGatewayNode = true
+			klog.Infof("Creating vxlan interface: %s", VxLANIface)
 			err = r.createVxLANInterface(VxInterfaceGateway, nil)
 			if err != nil {
 				klog.Fatalf("Unable to create VxLAN interface on GatewayNode (%s): %v", hostname, err)
 			}
-			klog.V(6).Infof("not reconciling routes because we appear to be the gateway host")
+
+			r.cleanRoutes()
```
sridhargaddam: Any specific reason @mangelajo for moving this API below createVxLANInterface?

pkg/routeagent/controllers/route/route.go
```diff
@@ -526,9 +534,8 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
 			klog.Errorf("Could not convert object tombstone %v to an Endpoint", tombstone.Obj)
 			return
 		}
-		klog.V(4).Infof("Recovered deleted object '%s' from tombstone", object.GetName())
 	}
-	klog.V(4).Infof("Informed of removed endpoint for route controller object: %v", object)
+	klog.V(log.DEBUG_LEVEL).Infof("Informed of removed endpoint for route controller object: %v", object)
```
sridhargaddam: Can you include the following check with a TODO (we can handle it properly in a separate PR)
if object.Spec.ClusterID != r.clusterID {
   // TODO: Handle a remote endpoint removal use-case 
   return
}

CC: @tpantelis

pkg/routeagent/controllers/route/route.go
```diff
@@ -526,9 +534,8 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
...
 	hostname, err := os.Hostname()
```
sridhargaddam: Thanks to @tpantelis for pointing this out. 

If we are on the gateway node, it's a good idea to cleanup the xfrm policies as well.
i.e., r.cleanXfrmPolicies

sridhargaddam: Small note: Actually the r.cleanRoutes() is not required in this API because as soon as we delete the VxLAN interface the routes associated on that interface will be deleted. I'm fine if you want to remove that call from this API.

pkg/routeagent/controllers/route/route.go
```diff
@@ -541,20 +548,31 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
...
+			klog.Errorf("r.vxlanDevice == nil , this should never happen, endpoint creation event missed?.")
+		} else {
+			err := r.vxlanDevice.deleteVxLanIface()
+			if err != nil {
+				klog.Errorf("Failed to delete the the vxlan interface on endpoint removal: %s", err.Error())
+				return
+			}
+			r.vxlanDevice = nil
 		}
-		r.vxlanDevice = nil
```
tpantelis: I suggest this structure:

```
if object.Spec.ClusterID != r.clusterID {
    return
}
	
r.gwVxLanMutex.Lock()
defer r.gwVxLanMutex.Unlock()
	
if r.vxlanDevice == nil {
    klog.Errorf("vxlanDevice is not set - ignoring removed endpoint %q", object.Name)
    return
}

if r.isGatewayNode {
    klog.Infof("Gateway Endpoint %q with IP %q was removed: deleting vxlan interface", object.Name, object.Spec.PrivateIP)
    r.cleanRoutes()
} else {
    klog.Infof("Non-gateway Endpoint %q with IP %q was removed: deleting vxlan interface", object.Name, object.Spec.PrivateIP)
}

err = r.vxlanDevice.deleteVxLanIface()
r.vxlanDevice = nil
if err != nil {
    klog.Errorf("Failed to delete the the vxlan interface on endpoint removal: %v", err)
} 
```

pkg/routeagent/main.go
```diff
@@ -37,6 +37,8 @@ func filterRouteAgentPods(options *v1.ListOptions) {
 	options.LabelSelector = route.SmRouteAgentFilter
 }
 
+const resyncInterval = time.Hour * 24
```
tpantelis: 24 hours is essentially useless. I think we should just set it to 0 to disable it. As long as we're re-queueing appropriately in `processNextEndpoint` and `processNextPod` on failures then we shouldn't need this.

pkg/routeagent/controllers/route/route.go
tpantelis: I assume to move it inside the critical section. Same in `handleRemovedEndpoint`. I was thinking the same thing to be safe.

mangelajo: Als, because you can't clean routes if vxlan interface does not exist, that resulted always in a logging error "cannot find vx-submariner" which seems like we are doing something wrong, although we aren't

pkg/routeagent/controllers/route/route.go
mangelajo: I don't feel comfortable making "structural" changes before a release, I've been bitten by unexpected things in the past. I just want to cleanup the logging as much as I can. Do you think it's fine if we do this after release?

pkg/routeagent/controllers/route/route.go
mangelajo: What would we need to handle in that case, so I can include it as part of the comments.

pkg/routeagent/main.go
mangelajo: Yeah, may be it's even something I should add to be configurable, for just in case it has unintended consequences... as I told @sridhargaddam on the leader-election patch. Sridhar, what do you think?.

sridhargaddam: IMHO, a resync interval of 24 hours is reaallyyy a long time and I don't see any value in having it. If we want to increase the current value of 30 secs, we could use something close (like in minutes) or disable it all-together as @tpantelis suggested (assuming that we are properly handling the re-queing event). 

Regarding making it configurable params, TBH I'm not a big fan of introducing too many configurable params which would affect the user-experience. I'd rather prefer to use some reasonable defaults where applicable and only create configurable params where it really matters. My 2 cents.

pkg/routeagent/controllers/route/route.go
```diff
@@ -419,31 +420,36 @@ func (r *Controller) processNextEndpoint() bool {
 		}
 
 		if endpoint.Spec.ClusterID != r.clusterID {
-			klog.V(6).Infof("setting routes of endpoint object %s", name)
+			klog.V(log.DEBUG_LEVEL).Infof("setting routes for endpoint  %s", name)
```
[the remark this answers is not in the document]
mangelajo: :+1:

pkg/routeagent/controllers/route/route.go
mangelajo: It was really missing in that case, but after the work-queue related patches by tom I'm not seeing that.

pkg/routeagent/controllers/route/route.go
mangelajo: Oh right, that's a good one.

pkg/routeagent/main.go
mangelajo: Yeah, but the parameters, with sensible defaults are useful when problems arise on production, you can mitigate instead of keeping a customer waiting for a patch which could take some time to produce and ship.

I'm not a big fan of parameters or asking the user for input. I think it's fine if all this is hidden to the normal user via subctl, etc...

tpantelis: The resync seems problematic in `processNextEndpoint`. Every period we'll create a new vxlan interface, w/o deleting the old one as we do in `handleRemovedEndpoint`.  We would need to make `processNextEndpoint` idempotent.  It is intended to re-process in case something went wrong previously and may not have been re-queued but we haven't seen a case where that would come into play. So I don't think we need auto resync at all and, in fact, it seems it may be detrimental. Unless I'm missing something?

mangelajo:
I hope it will look better now.

pkg/routeagent/controllers/route/route.go
```diff
@@ -419,31 +420,36 @@ func (r *Controller) processNextEndpoint() bool {
...
 		if err != nil {
 			klog.Fatalf("unable to determine hostname: %v", err)
 		}
 
-		klog.V(6).Infof("Local Cluster Gateway Node IP is %s", endpoint.Spec.PrivateIP)
-
 		// If the endpoint hostname matches with our hostname, it implies we are on gateway node
 		if endpoint.Spec.Hostname == hostname {
-			r.cleanRoutes()
+			klog.Infof("This has become an active gateway node (IP %s)", endpoint.Spec.PrivateIP)
```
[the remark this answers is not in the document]
mangelajo: That would be fine, let me update again.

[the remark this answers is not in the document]
mangelajo: I'm using %s btw, %q is only necessary if you need escape sequences, that IMO just make logs harder to read,

pkg/routeagent/controllers/route/route.go
mangelajo: no, if it's active GW we should not clean xfrm policies (those are set by strongswan). The other routes yes, because those are the routes which would be used on the non-gw node.
I suggest, that after release we could perform a small refactor of this agent to make it clearer. Probably smaller functions and isolating better the GW / non-gw behaviors/details.

pkg/routeagent/controllers/route/route.go
mangelajo: Yeah, I think it is... I'm just pre-release-scared ;)

But a separate PR may be better, as this is more about the logs, I guess.

pkg/routeagent/controllers/route/route.go
tpantelis: yeah I've already have patches lined up and other changes planned.

pkg/routeagent/controllers/route/route.go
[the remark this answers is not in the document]
tpantelis: %q surrounds with double quotes which I think that makes it easier to read with names. For IPs I'd agree sans double quotes is better.

pkg/apis/submariner.io/v1/string.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"k8s.io/klog"
+)
+
+// String method returns a string representation, for human readability.
+func (e *Endpoint) String() string {
+
+	bytes, err := json.Marshal(e.Spec)
+	if err != nil {
+		klog.Fatalf("Error serializing Endpoint spec to string: %s", err.Error())
```
tpantelis: I think it's rather drastic to exit the process for a debug string error. If it fails then format it with "%#v".

pkg/apis/submariner.io/v1/string.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"encoding/json"
+
+	"k8s.io/klog"
+)
+
+// String method returns a string representation, for human readability.
+func (e *Endpoint) String() string {
+
+	bytes, err := json.Marshal(e.Spec)
```
tpantelis: I think we should marshal the whole object - it's useful to see at least some of the Object metadata too.

pkg/apis/submariner.io/v1/string_test.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+const expectedString = "{\"cluster_id\":\"cluster-id\",\"cable_name\":\"cable-1\",\"hostname\":" +
+	"\"\",\"subnets\":[\"10.0.0.0/24\",\"172.0.0.0/24\"],\"private_ip\":\"1.1.1.1\"," +
+	"\"public_ip\":\"\",\"nat_enabled\":false,\"backend\":\"\"}"
+
+var _ = Describe("API v1", func() {
+	Describe("Endpoint String representation", func() {
+		When("Called", func() {
+			It("Should return a human readable string", func() {
```
tpantelis: I think you can combine the `Describe` and `When`, ie

When("Endpoint String function is called", func() {
    It("should return a human readable string", ...
})

pkg/routeagent/controllers/route/route.go
```diff
@@ -372,6 +373,8 @@ func (r *Controller) processNextPod() bool {
 					r.podWorkqueue.Forget(obj)
 					return fmt.Errorf("failed to add FDB entry on the Gateway Node vxlan iface %v", err)
 				}
+				klog.Infof("FDB entry added on the Gateway node's vxlan iface for "+
+					"route agent pod %s, IP %s", pod.Name, pod.Status.PodIP)
```
tpantelis: I do prefer double-quoting names and general strings...

pkg/routeagent/controllers/route/route.go
```diff
@@ -357,7 +358,7 @@ func (r *Controller) processNextPod() bool {
 			return fmt.Errorf("error retrieving submariner-route-agent pod object %s: %v", name, err)
 		}
 
-		klog.V(4).Infof("In processNextPod, POD HostIP is %s", pod.Status.HostIP)
+		klog.V(log.DEBUG).Infof("In processNextPod, POD HostIP is %s", pod.Status.HostIP)
```
tpantelis: I suggest making this Info so we can see each pod that was processed:

`klog.Infof("Processing submariner-route-agent pod %q with host IP %q, pod IP %q", name, pod.Status.HostIP, pod.Status.PodIP)`

pkg/routeagent/controllers/route/route.go
```diff
@@ -380,7 +383,7 @@ func (r *Controller) processNextPod() bool {
 		}
 
 		r.podWorkqueue.Forget(obj)
-		klog.V(4).Infof("Pod event processed by route controller")
+		klog.V(log.DEBUG).Infof("Pod event processed by route controller")
```
tpantelis: I suggest making this Info as well so we can see when it finished processing and the elapsed time:

`klog.Infof("Successfully processed submariner-route-agent pod %q", name)`

pkg/routeagent/controllers/route/route.go
```diff
@@ -419,36 +422,47 @@ func (r *Controller) processNextEndpoint() bool {
...
 			err = r.createVxLANInterface(VxInterfaceGateway, nil)
 			if err != nil {
-				klog.Fatalf("Unable to create VxLAN interface on GatewayNode (%s): %v", hostname, err)
+				klog.Fatalf("Unable to create VxLAN interface on gateway node (%s): %v", hostname, err)
 			}
-			klog.V(6).Infof("not reconciling routes because we appear to be the gateway host")
+
+			r.cleanRoutes()
+
+			klog.V(log.DEBUG).Infof("not reconciling gateway routes because we appear to be the gateway host")
```
tpantelis: Not sure this really adds any value since we've already reported we're on the gateway host above.

pkg/routeagent/controllers/route/route.go
```diff
@@ -475,7 +490,7 @@ func (r *Controller) processNextEndpoint() bool {
 		}
 
 		r.endpointWorkqueue.Forget(obj)
-		klog.V(4).Infof("endpoint processed by route controller")
+		klog.V(log.DEBUG).Infof("endpoint processed by route controller")
```
tpantelis:
```suggestion
		klog.Infof("Successfully processed local Endpoint %q", endpoint.Name)
```

pkg/routeagent/controllers/route/route.go
```diff
@@ -419,36 +422,47 @@ func (r *Controller) processNextEndpoint() bool {
 			return fmt.Errorf("error retrieving submariner endpoint object %s: %v", name, err)
 		}
 
+		klog.V(log.TRACE).Infof("Processing endpoint %s : %s", name, endpoint.String())
```
tpantelis: Don't specifically call `String()` so we don't incur the string conversion overhead on the fast path, ie when TRACE level is not enabled.

pkg/routeagent/controllers/route/route.go
```diff
@@ -399,13 +402,13 @@ func (r *Controller) processNextEndpoint() bool {
...
 		defer r.endpointWorkqueue.Done(obj)
-		klog.V(4).Infof("Handling object in handleEndpoint")
-		klog.V(4).Infof("Processing endpoint object: %v", obj)
 		key := obj.(string)
+
 		ns, name, err := cache.SplitMetaNamespaceKey(key)
 		if err != nil {
 			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
 		}
+		klog.V(log.DEBUG).Infof("Processing endpoint %s", name)
```
tpantelis: I don't think this is necessary.

pkg/routeagent/controllers/route/route.go
```diff
@@ -530,35 +545,49 @@ func (r *Controller) handleRemovedEndpoint(obj interface{}) {
...
-	klog.V(4).Infof("Removed routes from host")
+		if r.vxlanDevice == nil {
+			klog.Errorf("r.vxlanDevice == nil , this should never happen, endpoint creation event missed?.")
+		} else {
+			err := r.vxlanDevice.deleteVxLanIface()
+			if err != nil {
+				klog.Errorf("Failed to delete the the vxlan interface on endpoint removal: %s", err.Error())
+				return
+			}
+			r.vxlanDevice = nil
```
tpantelis: This should be right after L574.

pkg/routeagent/main.go
```diff
@@ -27,10 +27,11 @@ var (
...
 type SubmarinerRouteControllerSpecification struct {
-	ClusterID   string
-	Namespace   string
-	ClusterCidr []string
-	ServiceCidr []string
+	ClusterID             string
+	Namespace             string
+	ClusterCidr           []string
+	ServiceCidr           []string
+	ResyncIntervalMinutes int `default:"0"`
```
tpantelis: I prefer we do not expose this, at least until we make the code handle resync properly and see a need for it.

pkg/routeagent/controllers/route/route.go
```diff
@@ -458,6 +472,7 @@ func (r *Controller) processNextEndpoint() bool {
 
 		r.gwVxLanMutex.Lock()
 		r.isGatewayNode = false
+		klog.Infof("Creating the vxlan interface: %s (non-gw)", VxLANIface)
```
tpantelis:
```suggestion
		klog.Infof("Creating the vxlan interface %s with gateway node IP %s", VxLANIface, localClusterGwNodeIP)
```

More context can't hurt here.

pkg/routeagent/controllers/route/route.go
mangelajo: I think it's better to leave this one as debug (with the new formatting) since it will replicate in some cases the one in L376 which is the real action out of this processing.

pkg/routeagent/controllers/route/route.go
mangelajo: a, skitt clarifies \"%s\" for the pod name. Makes sense

pkg/routeagent/controllers/route/route.go
mangelajo: I think it's better as debug, if we want to check elapsed times, etc, which should be very low. Although I will change the formatting.

pkg/routeagent/controllers/route/route.go
mangelajo: True, :+1:

pkg/routeagent/controllers/route/route.go
mangelajo: Ack, let's remove it.

pkg/routeagent/controllers/route/route.go
mangelajo: But better as debug. Again, the initial purpose of this patch was to unclutter the logs, we are certainly making them better, but IMO that event belongs more into debug from the point of view of un-cluttering

pkg/routeagent/controllers/route/route.go
mangelajo: What if the failure was for other reason? "could not be deleted?" at least it would be retried on next similar event.

Although I'm not convinced on the delete strategy.

May be we should just keep those interfaces and just delete/create (=update) when necessary. But no deletion if it's not going to be re-created in a different form. That would feel safer in general. Probably something we want to try after release.

pkg/routeagent/main.go
mangelajo: I feel safer keeping it there, what if our change has unintended consequences, or if we introduce a bug which could be reverted by tweaking the setting. What if we discover this on a production deployment? (apparently there are production deployments now out there..) 

I think that's may be far now. But I've been able to provide workarounds for customers by settings of this type.

pkg/routeagent/controllers/route/route.go
mangelajo: > Small note: Actually the r.cleanRoutes() is not required in this API because as soon as we delete the VxLAN interface the routes associated on that interface will be deleted. I'm fine if you want to remove that call from this API.

Let's leave it there for now, I think it will be better not to delete those interfaces where we don absolutely need- I want to propose a patch after the release.

pkg/apis/submariner.io/v1/string.go
mangelajo:
```
 {
            "apiVersion": "submariner.io/v1",
            "kind": "Endpoint",
            "metadata": {
                "creationTimestamp": "2020-02-14T10:39:28Z",
                "generation": 1,
                "name": "cluster2-submariner-cable-cluster2-172-17-0-3",
                "namespace": "submariner-operator",
                "resourceVersion": "373557",
                "selfLink": "/apis/submariner.io/v1/namespaces/submariner-operator/endpoints/cluster2-submariner-cable-cluster2-172-17-0-3",
                "uid": "46945da8-4f16-11ea-af6e-0242ac110008"
            },
            "spec": {
                "backend": "ipsec",
                "cable_name": "submariner-cable-cluster2-172-17-0-3",
                "cluster_id": "cluster2",
                "hostname": "cluster2-worker",
                "nat_enabled": false,
                "private_ip": "172.17.0.3",
                "public_ip": "",
                "subnets": [
                    "100.95.0.0/16",
                    "10.245.0.0/16"
                ]
            }
        },
```

vs

```
 {
                "backend": "ipsec",
                "cable_name": "submariner-cable-cluster2-172-17-0-3",
                "cluster_id": "cluster2",
                "hostname": "cluster2-worker",
                "nat_enabled": false,
                "private_ip": "172.17.0.3",
                "public_ip": "",
                "subnets": [
                    "100.95.0.0/16",
                    "10.245.0.0/16"
                ]
        }
```

pkg/routeagent/controllers/route/route.go
mangelajo: Done, but as DEBUG

pkg/routeagent/controllers/route/route.go
mangelajo: Removed.

pkg/routeagent/controllers/route/route.go
mangelajo: Done

pkg/routeagent/controllers/route/route.go
mangelajo: Done.

pkg/routeagent/controllers/route/route.go
mangelajo: Done.

pkg/routeagent/controllers/route/route.go
mangelajo: I'll move it inside, since the most probable cause of error would be if it didn't exist...

mangelajo: Done.

pkg/routeagent/main.go
mangelajo: For ease of mind I debug-logged what the nodes do during resync (I set 1min manually):

The non-gateway:

```
I0214 12:20:04.031327       1 route.go:512] Enqueueing endpoint for route controller {"cluster_id":"cluster3","cable_name":"submariner-cable-cluster3-172-17-0-4","hostname":"cluster3-worker","subnets":["100.96.0.0/16","10.246.0.0/16"],"private_ip":"172.17.0.4","public_ip
":"","nat_enabled":false,"backend":"ipsec"}
I0214 12:20:04.031524       1 route.go:512] Enqueueing endpoint for route controller {"cluster_id":"cluster2","cable_name":"submariner-cable-cluster2-172-17-0-3","hostname":"cluster2-worker","subnets":["100.95.0.0/16","10.245.0.0/16"],"private_ip":"172.17.0.3","public_ip
":"","nat_enabled":false,"backend":"ipsec"}
I0214 12:20:04.042278       1 route.go:429] Updating iptable rules for remote cluster's Endpoint "cluster3-submariner-cable-cluster3-172-17-0-4" with subnets [100.96.0.0/16 10.246.0.0/16]
I0214 12:20:04.050802       1 route.go:436] Processing local Endpoint cluster2-submariner-cable-cluster2-172-17-0-3 with IP 172.17.0.3, Host cluster2-worker for this cluster
I0214 12:20:04.050898       1 route.go:464] This route agent is running on a non-gateway/non-active node
I0214 12:20:04.050948       1 route.go:475] Creating the vxlan interface vx-submariner with gateway node IP 172.17.0.3
I0214 12:20:04.051597       1 vxlan.go:62] VxLAN interface already exists with same configuration.
I0214 12:20:04.059206       1 route.go:671] Reconciling routes to gw: 240.17.0.3
I0214 12:20:04.059558       1 route.go:687] Processing route {Ifindex: 7680 Dst: 10.246.0.0/16 Src: <nil> Gw: 240.17.0.3 Flags: [] Table: 254}
I0214 12:20:04.059702       1 route.go:692] Found route {Ifindex: 7680 Dst: 10.246.0.0/16 Src: <nil> Gw: 240.17.0.3 Flags: [] Table: 254} with gw 240.17.0.3 already installed
I0214 12:20:04.059726       1 route.go:687] Processing route {Ifindex: 7680 Dst: 100.96.0.0/16 Src: <nil> Gw: 240.17.0.3 Flags: [] Table: 254}
I0214 12:20:04.059772       1 route.go:692] Found route {Ifindex: 7680 Dst: 100.96.0.0/16 Src: <nil> Gw: 240.17.0.3 Flags: [] Table: 254} with gw 240.17.0.3 already installed
I0214 12:20:04.059787       1 route.go:687] Processing route {Ifindex: 7680 Dst: 240.0.0.0/8 Src: 240.17.0.5 Gw: <nil> Flags: [] Table: 254}
I0214 12:20:04.059814       1 route.go:689] Found nil gw or dst
I0214 12:20:04.060107       1 route.go:728] Found equivalent route, not adding
I0214 12:20:04.060299       1 route.go:728] Found equivalent route, not adding
I0214 12:20:04.060355       1 route.go:493] Successfully processed local Endpoint "cluster2-submariner-cable-cluster2-172-17-0-3"
```
On the gw:

```
I0214 12:20:04.032232       1 route.go:512] Enqueueing endpoint for route controller {"cluster_id":"cluster2","cable_name":"submariner-cable-cluster2-172-17-0-3","hostname":"cluster2-worker","subnets":["100.95.0.0/16","10.245.0.0/16"],"private_ip":"172.17.0.3","public_ip
":"","nat_enabled":false,"backend":"ipsec"}
I0214 12:20:04.032376       1 route.go:512] Enqueueing endpoint for route controller {"cluster_id":"cluster3","cable_name":"submariner-cable-cluster3-172-17-0-4","hostname":"cluster3-worker","subnets":["100.96.0.0/16","10.246.0.0/16"],"private_ip":"172.17.0.4","public_ip
":"","nat_enabled":false,"backend":"ipsec"}
I0214 12:20:04.042951       1 route.go:436] Processing local Endpoint cluster2-submariner-cable-cluster2-172-17-0-3 with IP 172.17.0.3, Host cluster2-worker for this cluster
I0214 12:20:04.043036       1 route.go:446] This route agent is running on the active gateway node
I0214 12:20:04.043079       1 route.go:452] Creating the vxlan interface: vx-submariner on the gateway node
I0214 12:20:04.043814       1 vxlan.go:62] VxLAN interface already exists with same configuration.
I0214 12:20:04.044028       1 vxlan.go:158] Successfully added the bridge fdb entry 172.17.0.5 00:00:00:00:00:00
I0214 12:20:04.044191       1 vxlan.go:158] Successfully added the bridge fdb entry 172.17.0.3 00:00:00:00:00:00
I0214 12:20:04.044379       1 route.go:306] Successfully configured rp_filter to loose mode(2) on vx-submariner
I0214 12:20:04.045028       1 route.go:619] Processing route {Ifindex: 7680 Dst: 240.0.0.0/8 Src: 240.17.0.3 Gw: <nil> Flags: [] Table: 254}
I0214 12:20:04.045099       1 route.go:621] Found nil gw or dst
I0214 12:20:04.053169       1 route.go:429] Updating iptable rules for remote cluster's Endpoint "cluster3-submariner-cable-cluster3-172-17-0-4" with subnets [100.96.0.0/16 10.246.0.0/16]
```

There seems to be no destructive event, so I think it could potentially be of help if anything arises (stale connections?, does the watcher ping the http connection being watched?, can be remediated with TCP keepalived settings though..)

tpantelis: We're also cleaning routes and XRFM policies. `processNextEndpoint` is really coded assuming `handleRemovedEndpoint` occurred prior. I understand what your saying about making it configurable but I'd be wary of turning this on in a production environment w/o fully understanding its effects and testing it even though it's already been on so it would **seem** to be OK. This also would it to the public API so we'd be stuck with it. I'd still prefer not to add the config at this point but if others agree we should then I'm OK.

tpantelis: Actually `SubmarinerRouteControllerSpecification ` is not really public so this would effectively be a hidden config. So it should be fine should we want to remove it at some point.

pkg/apis/submariner.io/v1/string.go
mangelajo: True about the trace level, I guess the generation/resourceversion/timestamps could be useful.

pkg/routeagent/main.go
mangelajo: Right, I don't propose advertising or documenting it, and it we do, under the proper warning. I'm testing it though in a loop with 1 min resync, and the gw failover/connectivity does not seem to fail.

I think the "220 table" and your "workqueue" patches actually fixed it for good/now ;)

mangelajo:
@tpantelis sent me a message on whatsapp, he's PTO today and on the road, but this should be fine to merge after the last changes.

mangelajo:
Ok, leaving this out of the release until we figure out what's wrong with this patch, although I'm suspicious because I've also seen master failing sometimes, but it could have been my laptop.

pkg/routeagent/controllers/route/route.go
```diff
@@ -575,24 +601,24 @@ func (r *Controller) handleRemovedPod(obj interface{}) {
 	}
 }
 
-func (r *Controller) cleanRoutes() {
+func (r *Controller) cleanVxSubmarinerRoutes() {
 	link, err := netlink.LinkByName(VxLANIface)
 	if err != nil {
-		klog.Errorf("Error retrieving link by name %s: %v", VxLANIface, err)
+		// if the vx-submariner interface does not exist then there are no routes to be cleaned up, we're fine.
```
tpantelis: Looking at the `netlink.LinkByName` code, it will return `netlink.LinkNotFoundError` in that case.  I think we should check for that to be complete.

mangelajo:
@sridhargaddam could you re-eye-ball :-) I hope it will pass CI now.

pkg/routeagent/controllers/route/route.go
mangelajo: ack, I will do that.

tpantelis:
It's passing and looks good - merging. My comment about using `netlink.LinkNotFoundError` is minor and @mangelajo is on PTO for a bit so I'll address that in a follow-up.

PR #347, opened 2020-02-12
Refactoring for WireGuard support

roytman:
Code refactoring preparation for support WireGuard driver


roytman:
will create a separate pull request with squashed commits

PR #348, opened 2020-02-12
Reduce gateway failover detection interval

sridhargaddam:
In an HA setup where multiple nodes are labelled as Gateway
nodes, when the active SubmarinerEngine/Node goes down, it takes
about 15+ seconds for one of the backup Submariner Gateway PODs
to become active. This patch makes the leader election values
configurable along with some defaults that reduce this interval.

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,6 +43,19 @@ func init() {
 	flag.StringVar(&localMasterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 }
 
+type leaderConfig struct {
+	LeaseDuration int64
+	RenewDeadline int64
+	RetryPeriod   int64
+}
+
+const (
+	leadershipConfigEnvPrefix = "leadership"
+	defaultLeaseDuration      = 5 // In Seconds
+	defaultRenewDeadline      = 3 // In Seconds
+	defaultRetryPeriod        = 2 // In Seconds
+)
+
 func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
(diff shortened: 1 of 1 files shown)


sridhargaddam:
> I think it's safer to make this configurable, probably with the low defaults you're suggesting, but being safer because I'm not sure of the impact in busy k8s.

Done

main.go
```diff
@@ -43,6 +43,19 @@ func init() {
 	flag.StringVar(&localMasterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 }
 
+type leaderConfig struct {
+	LeaseDuration int64
```
mangelajo: you can also annotate with `default:"5"` 

and the module will handle it, but I'm fin with it now, we need to release.

sridhargaddam: Actually, I tried it and it did not seem to work when I tested it. So, I modified it this way.

main.go
```diff
@@ -174,6 +187,28 @@ func main() {
 }
 
 func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder record.EventRecorder, run func(ctx context.Context)) {
+	gwLeadershipConfig := leaderConfig{}
+
+	err := envconfig.Process(leadershipConfigEnvPrefix, &gwLeadershipConfig)
+	if err != nil {
+		klog.Fatalf("error processing environment config for %s: %v", leadershipConfigEnvPrefix, err)
+	}
```
tpantelis: I wonder if we should start using a ConfigMap for this type of config ...

sridhargaddam: There are many params that can be moved to a ConfigMap. Is it ok if we take it up in a separate patch?

mangelajo: I think it’s something we should do, and also the secrets. There’s an open issue for that. May be something we can do after this release

main.go
mangelajo: oh, ok, makes sense then. I guess the lib is buggy

main.go
tpantelis: +1

PR #349, opened 2020-02-13
Refactoring for support pluggable drivers. 

roytman:
Code refactoring for support pluggable drivers. 
We will start from WireGuard


pkg/cable/driver.go
```diff
@@ -1,9 +1,17 @@
...
 
 import (
+	"fmt"
+
+	"github.com/submariner-io/submariner/pkg/cable/ipsec"
 	"github.com/submariner-io/submariner/pkg/types"
 )
 
+const (
+	IPSec     = "ipsec"
```
skitt: Could you rename this to StrongSwan? We want to add a LibreSwan cable too (see #101 and #107).

pkg/cableengine/cableengine.go
```diff
@@ -1,11 +1,100 @@
...
+// An Engine cooperates with, and delegates work to, a cable.Driver for implementing
+// a secure connection to remote clusters.
 type Engine interface {
+	// StartEngine performs any general set up work needed independent of any remote connections.
 	StartEngine() error
-	InstallCable(types.SubmarinerEndpoint) error
-	RemoveCable(types.SubmarinerEndpoint) error
+	// InstallCable performs any set up work needed for connecting to given remote endpoint.
+	// Once InstallCable completes, it should be possible to connect to remote
+	// Pods or Services behind the given endpoint.
```
skitt: Yay, documentation! :+1:

pkg/cable/driver.go
roytman: Done

roytman:
@mangelajo and I had a slack conversation about cable and driver abstractions. 
A future requirement is supporting other IPsec implementations (drivers), e.g. `LibreSwan`. 
In general it can open a lot of different interoperability use cases (protocol, implementation, configurations, ...) that go beyond this PR.
In order to go forward, I'd suggest the following:
- Submariner gateway will get a configuration parameter with format `cable-engine[/driver]` (driver is optional).
- The cable-engine info (such as `IPsec`, `WireGuard`) will be shared via `Endpoints` CRD so they can be known to peers
- The optional driver can help to distinguish between different cable implementations. (`StrongSwan`, `LiberSwan`) and is relevant only to the local endpoint (at least for now)
- If the driver is not specified, the default one will be used. 

I'd emphasis that the PR doesn't change current Submariner behavior, but creates options to extend it. 
What do you think ?

skitt:
[...]
> What do you think ?

Thinking “aloud” here...

I’m concerned that distinguishing between protocols and drivers feels somewhat premature; I’m not sure that there’s much point in doing so. We know that we want different _drivers_, but whether any two (or more) drivers share a protocol doesn’t necessarily matter — I agree it would be nice, so for example one cluster using LibreSwan could join another using StrongSwan (which is theoretically possible of course), but enabling this adds constraints and increases the test matrix.

I know that the two are different concepts, and it only makes sense to share _protocol_ requirements, not driver requirements. But I think the design needs refinement, and a PR is probably not the best place to do so ;-). A PR using only drivers, not protocols, would meet the overall requirements, and be simpler to understand and less constraining on the design. Put another way, it bothers me that the protocol/driver handling is closely intertwined in the current implementation...

(I need to look into general plugin handling a bit more. For example, could we handle protocols as shared requirements, and have plugins which provide protocols, without hard-coding protocol/driver mappings?)

pkg/cable/driver.go
```diff
@@ -0,0 +1,57 @@
...
+	case IPsec:
+		driver := StrongSwan
+		if localEndpoint.Spec.BackendConfig != nil {
+			if d, ok := localEndpoint.Spec.BackendConfig[DriverImpl]; ok {
+				driver = d
+			}
+		}
+		switch driver {
+		case StrongSwan:
+			return ipsec.NewStrongSwan(localSubnets, localEndpoint)
```
mangelajo: This also works for me..

mangelajo:
> [...]
> 
> > What do you think ?
> 
> Thinking “aloud” here...
> 
> I’m concerned that distinguishing between protocols and drivers feels somewhat premature; I’m not sure that there’s much point in doing so. We know that we want different _drivers_, but whether any two (or more) drivers share a protocol doesn’t necessarily matter — I agree it would be nice, so for example one cluster using LibreSwan could join another using StrongSwan (which is theoretically possible of course), but enabling this adds constraints and increases the test matrix.
> 
> I know that the two are different concepts, and it only makes sense to share _protocol_ requirements, not driver requirements. But I think the design needs refinement, and a PR is probably not the best place to do so ;-). A PR using only drivers, not protocols, would meet the overall requirements, and be simpler to understand and less constraining on the design. Put another way, it bothers me that the protocol/driver handling is closely intertwined in the current implementation...
> 
> (I need to look into general plugin handling a bit more. For example, could we handle protocols as shared requirements, and have plugins which provide protocols, without hard-coding protocol/driver mappings?)

May be an alternative as I proposed on the U/S meeting, could be just having cable engines, and then each cable engine declares what protocol it supports (just a string). That would keep it flexible enough, and the code would be a bit simpler.

But I really think we need to expose the protocol we use on the endpoints (more than the implementation), also at some point we may also want to expose port information, etc..

Would that be fine?.

Sorry @roytman to keep you circling.

sridhargaddam:
> > [...]
> > > What do you think ?
> > 
> > 
> > Thinking “aloud” here...
> > I’m concerned that distinguishing between protocols and drivers feels somewhat premature; I’m not sure that there’s much point in doing so. We know that we want different _drivers_, but whether any two (or more) drivers share a protocol doesn’t necessarily matter — I agree it would be nice, so for example one cluster using LibreSwan could join another using StrongSwan (which is theoretically possible of course), but enabling this adds constraints and increases the test matrix.
> > I know that the two are different concepts, and it only makes sense to share _protocol_ requirements, not driver requirements. But I think the design needs refinement, and a PR is probably not the best place to do so ;-). A PR using only drivers, not protocols, would meet the overall requirements, and be simpler to understand and less constraining on the design. Put another way, it bothers me that the protocol/driver handling is closely intertwined in the current implementation...
> > (I need to look into general plugin handling a bit more. For example, could we handle protocols as shared requirements, and have plugins which provide protocols, without hard-coding protocol/driver mappings?)
> 
> May be an alternative as I proposed on the U/S meeting, could be just having cable engines, and then each cable engine declares what protocol it supports (just a string). That would keep it flexible enough, and the code would be a bit simpler.
> 
> But I really think we need to expose the protocol we use on the endpoints (more than the implementation), also at some point we may also want to expose port information, etc..
> 
> Would that be fine?.

I'm fine with the idea of having a cable engine and each cable engine declaring the protocol it supports. Say we have two cable engines, one that supports LibreSwan(todo) and the other StrongSwan, one of the challenges would be to ensure that IPSEC configuration (which currently is hardcoded) matches on the joining clusters. 
> 
> Sorry @roytman to keep you circling.

tpantelis:
> > 
> > May be an alternative as I proposed on the U/S meeting, could be just having cable engines, and then each cable engine declares what protocol it supports (just a string). That would keep it flexible enough, and the code would be a bit simpler.
> > But I really think we need to expose the protocol we use on the endpoints (more than the implementation), also at some point we may also want to expose port information, etc..
> > Would that be fine?.
> 
> I'm fine with the idea of having a cable engine and each cable engine declaring the protocol it supports. Say we have two cable engines, one that supports LibreSwan(todo) and the other StrongSwan, one of the challenges would be to ensure that IPSEC configuration (which currently is hardcoded) matches on the joining clusters.
> 

I agree it would be simpler to have `Endpoint.Spec.Backend` contain the name of the driver implementation and expose driver info like protocol, port information, etc, via `Endpoint.Spec.BackendConfig`.

pkg/cable/driver.go
mangelajo: To match what we were talking about on the subctl side, I guess it would make sense to map:

"ipsec" -> Strongswan
"ipsec/strongswan" -> Strongswan
"ipsec/libreswan" -> Libreswan

And we must make sure all checks are not case sensitive.

mangelajo:
See also my comments here: https://github.com/submariner-io/submariner/pull/349/files#diff-58d734873ce9b60e90b658b0859d9228R47

@tpantelis I didn't know we already had BackendConfig field.

We could use Backend "ipsec", 

and on the fields provide more details:
* implemetation: "ipsec/strongswan"
(and the next ones specific to the type of backend) : -unrelated to this specific patch-, just thinking of more uses.
* ike-port: "500"
* natt-port: "4500"

mangelajo:
Ah, I think @tpantelis and I are proposing the inverse :-)

I think he's proposing:

Backend = "ipsec/strongswan"
BackendConfig = { "protocol": "ipsec"
                               "...." }

In that case it only hurts me that "protocol" is not a configuration of the backend.

May be we should simply introduce another field called protocol? 

May be that would be more natural.

roytman:
Maybe to use `cable` as a protocol, e.g. `ipsec` or `wireguard`. And to store the specific driver implementation in the BackendConfig = {"driver": "strongswan" ... }
According to the operator input `cable[\driver]`, the driver type is an optional one (wiregiuard doesn't have any ) and its name can be shared with other backend config parameters such as "port"

tpantelis:
> Ah, I think @tpantelis and I are proposing the inverse :-)
> 
> I think he's proposing:
> 
> Backend = "ipsec/strongswan"
> BackendConfig = { "protocol": "ipsec"
> "...." }
> 
> In that case it only hurts me that "protocol" is not a configuration of the backend.
> 
> May be we should simply introduce another field called protocol?
> 
> May be that would be more natural.

Protocol could be part of the `BackendConfig` but using the "protocol[/driver]" convention would be fine.  

@roytman is proposing to store the optional driver impl in the `BackendConfig` which is also viable but I'd prefer one field to identify the backend driver as it's simpler.

pkg/cable/driver.go
tpantelis: I suggest using a static registration map instead of a switch, eg

```
type DriverInitFunc func(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (Driver, error)

var Drivers = map[string]DriverInitFunc{}
```

Then in each driver file, eg strongswan.go:

```
func init() {
	cable.Drivers["ipsec"] = ipsec.NewStrongSwan
	cable.Drivers["ipsec/strongswan"] = ipsec.NewStrongSwan
}

```

roytman: Yeah, the static map is better, done

pkg/cable/driver.go
```diff
@@ -20,3 +20,9 @@ type Driver interface {
 	// DisconnectFromEndpoint disconnects from the connection to the given endpoint.
 	DisconnectFromEndpoint(endpoint types.SubmarinerEndpoint) error
 }
+
+// Function prototype to create a new driver
+type DriverCreateFunc func(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (Driver, error)
+
+// Static map of supported drivers
+var Drivers = map[string]DriverCreateFunc{}
```
tpantelis: I think we should go thru a function to add drivers to prevent dups: 

```
var drivers = map[string]DriverCreateFunc{}

func AddDriver(name string, driverCreate DriverCreateFunc) {
	if drivers[name] != nil {
		panic(fmt.Sprintf("Multiple cable engine drivers attempting to register with name %q", name))
	}
	
	drivers[name] = driverCreate
}
```

main.go
```diff
@@ -4,34 +4,37 @@ import (
...
 	"github.com/submariner-io/submariner/pkg/controllers/tunnel"
+	"github.com/submariner-io/submariner/pkg/datastore"
 	subk8s "github.com/submariner-io/submariner/pkg/datastore/kubernetes"
 	"github.com/submariner-io/submariner/pkg/datastore/phpapi"
 	"github.com/submariner-io/submariner/pkg/signals"
+	"github.com/submariner-io/submariner/pkg/types"
+	"github.com/submariner-io/submariner/pkg/util"
+
+	// Add supported drivers
+	_ "github.com/submariner-io/submariner/pkg/cable/ipsec"
```
tpantelis: I think this really belongs in _driver.go_.

pkg/types/types.go
```diff
@@ -24,6 +24,7 @@ type SubmarinerSpecification struct {
 	ColorCodes  []string
 	NatEnabled  bool
 	Broker      string
+	Cable       string
```
tpantelis:
```suggestion
	CableDriver       string
```

roytman: Done

main.go
roytman: I agree that logically it belongs to _driver.go_, however we cannot put it there because it creates imports cycle. I moved it to _cableengine.go_

pkg/cable/driver.go
roytman: done

pkg/cable/ipsec/strongswan.go
```diff
@@ -36,6 +38,12 @@ const (
 	charonViciSocket = "/var/run/charon.vici"
 )
 
+func init() {
+	cable.AddDriver("", NewStrongSwan) //default
```
tpantelis: Looking at this some more, I'm thinking maybe we shouldn't specify the default driver in this manner as the engine's `Endpoint` object would also advertise an empty `Backend` driver field. That doesn't really matter now other than one inspecting the `Endpoint` object from the API server would need knowledge of the internal code to know which driver is being used. But perhaps remote clusters may utilize the `Backend` field in the future. 

We could add a specific `cable.SetDefautDriver("ipsec/strongswan")` call and in, **main.go**:

```
if submSpec.CableDriver == "" {
    submSpec.CableDriver = cable.GetDefautDriver()
}
```

roytman: **Question**: should we replace `ipsec` to `ipsec/strongswan` ? or assume that if a user specified cable only it should be used as is.

tpantelis: yeah - good question - I was wondering that too. It seems to me the `Backend` that's advertised to other clusters should be the actual driver being used, ie if _"ipsec"_ was specified as the `CableDriver` then `Backend` should contain the full name _"ipsec/strongswan"_. This makes me wonder if we're unnecessarily complicating it too much by formally defining a _"protocol[/driver]"_ convention and allowing just _"ipsec"_.  I think we should just keep it simple and take the `CableDriver` as is - for the Strongswan driver we can still use _"ipsec/strongswan"_ as the name just for additional information clarity or just _"strongswan"_.

roytman: Done

pkg/cable/ipsec/strongswan.go
```diff
@@ -34,8 +36,18 @@ const (
...
 	charonViciSocket = "/var/run/charon.vici"
+
+	cableDriverName = "ipsec/strongswan"
 )
 
+func init() {
+	cable.SetDefautCableDriver(cableDriverName)
+	cable.AddDriver(cableDriverName, NewStrongSwan)
+	cable.AddDriver("ipsec", NewStrongSwan)
+	cable.AddDriver("strongswan", NewStrongSwan)
```
tpantelis: On the call today, it was decided to keep it simple and register the exact driver name sans aliases/defaults and sans _"ipsec"_ prefix.  So driver name would just be _"strongswan"_.

roytman: Done, please review

pkg/cableengine/cableengine.go
```diff
@@ -1,11 +1,118 @@
...
+	return &engine{
+		localCluster:  localCluster,
+		localEndpoint: localEndpoint,
+		localSubnets:  localSubnets,
+		driver:        driver,
+	}, nil
+}
+
+func (i *engine) StartEngine() error {
+	klog.Infof("Starting IPSec Engine (Charon)")
```
mangelajo: This is IPSEC/Strongswan specific, although, just a nit, something we can fix on a follow up patch.

pkg/types/types.go
```diff
@@ -24,6 +24,7 @@ type SubmarinerSpecification struct {
 	ColorCodes  []string
 	NatEnabled  bool
 	Broker      string
+	CableDriver string
```
mangelajo: I see that we will publish "strongswan" as a "cabledriver" to the broker.

It should really be either "ipsec" or "ipsec/strongswan".


But we can iterate over this on follow up patches.

mangelajo: Also, to the broker, what we should probably just share is just Protocol, it doesn't really matter which driver it is as long as it's the same protocol (theoretically, of course..)

pkg/cable/ipsec/strongswan.go
mangelajo: > On the call today, it was decided to keep it simple and register the exact driver name sans aliases/defaults and sans _"ipsec"_ prefix. So driver name would just be _"strongswan"_.

Sorry, I missed this comment.

I think it should also be fine to go with cabledriver, and that's it ;)

mangelajo:
Ignore my comments, merging :)

mangelajo:
@tpantelis I merged master, feel free to merge once CI is passed.

pkg/cableengine/cableengine.go
tpantelis: you're right - I'll push a follow-up for this.

Issue #346, opened 2020-02-12
After gateway failover sometimes an extra 220 routing table (strongswan) will be left and break connectivity


mangelajo:
After a gateway failover, a worker that previously was a master gateway would leave some entries in routing table 220 (strongswan) which take precedence to the default routing rules. 

This will make connectivity to the remote cluster from such node. This breaks the E2E tests sometimes.

NOTE: 10.246.224.3 is a remote cluster pod

root@cluster2-worker:/# ip r get 10.246.224.3                                                                                                                                                
10.246.224.3 via **172.17.0.6 dev eth0** table 220 src 172.17.0.5 uid 0 

root@cluster2-worker:/# ip r                                                                                                                                                                 
default via 172.17.0.1 dev eth0                                                                                                                                                              
10.245.0.0/16 dev weave proto kernel scope link src 10.245.0.1                                                                                                                               
10.246.0.0/16 via 240.17.0.8 dev vx-submariner proto static                                                                                                                                  
100.96.0.0/16 via 240.17.0.8 dev vx-submariner proto static                                                                                                                                  
172.17.0.0/16 dev eth0 proto kernel scope link src 172.17.0.5                                                                                                                                
240.0.0.0/8 dev vx-submariner proto kernel scope link src 240.17.0.5 

root@cluster2-worker:/# ip r show table 220
**10.246.0.0/16 via 172.17.0.6 dev eth0 proto static src 172.17.0.5 
100.96.0.0/16 via 172.17.0.6 dev eth0 proto static src 172.17.0.5 
172.17.0.6 via 172.17.0.6 dev eth0 proto static src 172.17.0.5** 

fixed: 

root@cluster2-worker:/# ip r flush table 220
root@cluster2-worker:/# ip r get 10.246.224.3
**10.246.224.3 via 240.17.0.8 dev vx-submariner src 240.17.0.5 uid 0**

PR #350, opened 2020-02-13
Cleanup strongswan table 220 when master gw becomes non-master.

mangelajo:
HA Failover tests start working flawlesly after this change, otherwise
sometimes the node that previously was a master gw, left over the 220
routing table, causing conflict with the default routes managed by
the route agent.

Fixes: #346

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -597,6 +601,15 @@ func (r *Controller) cleanRoutes() {
 	}
 }
 
+func (r *Controller) cleanStrongswanRoutingTable() {
+	cmd := exec.Command("/sbin/ip", "r", "flush", "table", "220")
```
sridhargaddam: This API is tied to Strongswan implementation and now that there are some efforts in adding other drivers, shall we add a comment to revisit this API?

mangelajo: True, let me extend L467 note to the functions.

I believe this should not be altogether here. Also, in cases where the route agent won't live on every node this won't work.

We need to find a better place, the alternative would be having a separate daemonset to run on nodes, but that sounds expensive... we need to keep thinking about it.

mangelajo: I updated the comments, let me know if it looks ok :)

sridhargaddam: Thanks for adding the comments.

Issue #304, opened 2020-01-29
Raise timeouts for pod creation in E2E tests


mangelajo:
Some environments are slower to react, and some E2E tests eventually timeout, while submariner is not at fault.

Related initial report by @manosnoam : https://github.com/submariner-io/submariner/issues/267

PR #351, opened 2020-02-13
Re-enable and fix Gateway failover tests

mangelajo:
This increases the timeouts for tests, which are not enough sometimes and
generate random failures, and also fixes the gateway failover tests which
could detect a pod in the wrong host just by the fact that the new label
did't have time to trigger / complete the transition.

Fixes: #304

### DIFF
--- a/test/e2e/redundancy/gateway_failover.go
+++ b/test/e2e/redundancy/gateway_failover.go
@@ -2,15 +2,18 @@ package redundancy
 
 import (
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
 	"github.com/submariner-io/shipyard/test/e2e/tcp"
+	v1 "k8s.io/api/core/v1"
+
 	subFramework "github.com/submariner-io/submariner/test/e2e/framework"
 )
 
-var _ = PDescribe("[redundancy] Gateway fail-over tests", func() {
+var _ = Describe("[redundancy] Gateway fail-over tests", func() {
 	f := subFramework.NewFramework("gateway-redundancy")
 
 	When("one gateway node is configured and the submariner engine pod fails", func() {
@@ -94,9 +97,17 @@ func testGatewayFailOverScenario(f *subFramework.Framework) {
 	f.SetGatewayLabelOnNode(framework.ClusterA, initialGatewayNode.Name, false)
 
 	// Ensure the new engine pod is started before we run the connectivity tests to eliminate possible timing issue where,
-	// after deleting the old pod, we actually run the connectivity test against the oold engine instance before k8s has
+	// after deleting the old pod, we actually run the connectivity test against the old engine instance before k8s has
 	// a chance to react to stop the process/container etc.
-	newEnginePod := f.AwaitSubmarinerEnginePod(framework.ClusterA)
+	var newEnginePod *v1.Pod
+	for retries := 1; retries < 10; retries++ {
+		newEnginePod = f.AwaitSubmarinerEnginePod(framework.ClusterA)
+		if newEnginePod.Spec.NodeName == initialGatewayNode.Name {
+			time.Sleep(5 * time.Second)
+		} else {
+			break
+		}
+	}
 	Expect(newEnginePod.Spec.NodeName).To(Equal(initialNonGatewayNode.Name),
 		"The new engine pod is not running on the expected node")
 	By(fmt.Sprintf("Found new submariner engine pod %q", newEnginePod.Name))
--- a/test/e2e/redundancy/gateway_failover.go
+++ b/test/e2e/redundancy/gateway_failover.go
@@ -54,6 +54,7 @@ func testEnginePodRestartScenario(f *subFramework.Framework) {
 		FromClusterScheduling: framework.GatewayNode,
 		ToCluster:             framework.ClusterA,
 		ToClusterScheduling:   framework.GatewayNode,
+		ToEndpointType:        defaultEndpointType(),
 	})
 
 	By(fmt.Sprintf("Verifying TCP connectivity from non-gateway node on %q to non-gateway node on %q", clusterBName, clusterAName))
@@ -63,7 +64,15 @@ func testEnginePodRestartScenario(f *subFramework.Framework) {
 		FromClusterScheduling: framework.NonGatewayNode,
 		ToCluster:             framework.ClusterA,
 		ToClusterScheduling:   framework.NonGatewayNode,
+		ToEndpointType:        defaultEndpointType(),
 	})
+}
+
+func defaultEndpointType() tcp.EndpointType {
+	if framework.TestContext.GlobalnetEnabled {
+		return tcp.GlobalIP
+	}
+	return tcp.PodIP
 }
 
 func testGatewayFailOverScenario(f *subFramework.Framework) {
@@ -124,6 +133,7 @@ func testGatewayFailOverScenario(f *subFramework.Framework) {
 		FromClusterScheduling: framework.GatewayNode,
 		ToCluster:             framework.ClusterA,
 		ToClusterScheduling:   framework.GatewayNode,
+		ToEndpointType:        defaultEndpointType(),
 	})
 
 	By(fmt.Sprintf("Verifying TCP connectivity from non-gateway node on %q to non-gateway node on %q", clusterBName, clusterAName))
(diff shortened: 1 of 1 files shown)


test/e2e/framework/network_pods.go
```diff
@@ -198,7 +198,16 @@ func (np *NetworkPod) buildTCPCheckConnectorPod() {
...
+					Command: []string{"sh", "-c", ` function msg() {
+														for in in $(seq 50); do
+															echo [dataplane] connector says $SEND_STRING;
+														done
+													}
+													for i in $(seq $CONN_TRIES); do
+														if msg | nc -v $REMOTE_IP $REMOTE_PORT -w $CONN_TIMEOUT; then
+															break;
+														fi;
+													done >/dev/termination-log 2>&1`},
```
tpantelis: This looks indented too far, at least in the github web UI.

test/e2e/framework/test_context.go
```diff
@@ -41,9 +41,9 @@ func registerFlags(t *TestContextType) {
 	flag.StringVar(&TestContext.ReportPrefix, "report-prefix", "", "Optional prefix for JUnit XML reports. Default is empty, which doesn't prepend anything to the default name.")
 	flag.StringVar(&TestContext.ReportDir, "report-dir", "", "Path to the directory where the JUnit XML reports should be saved. Default is empty, which doesn't generate these reports.")
 	flag.StringVar(&TestContext.SubmarinerNamespace, "submariner-namespace", "submariner", "Namespace in which the submariner components are deployed.")
-	flag.UintVar(&TestContext.ConnectionTimeout, "connection-timeout", 6, "The timeout in seconds per connection attempt when verifying communication between clusters.")
+	flag.UintVar(&TestContext.ConnectionTimeout, "connection-timeout", 90, "The timeout in seconds per connection attempt when verifying communication between clusters.")
```
tpantelis: This seems rather large - with 10 attempts the total timeout will be 900 sec (like 15 min).  Did you intend for 90 sec total? If so then this would be 9.

test/e2e/framework/network_pods.go
mangelajo: Ooops, right, let me handle this.

test/e2e/framework/test_context.go
mangelajo: Then may be we need to reduce the connection attempts. One issue I found is that you need to do this across clusters, and while you are spawning the connector pod, the listener pod could timeout and go away. 

I will reduce the connection attempts.

tpantelis: I'm not sure if it's better to have lower timeout and more attempts or vice versa.

test/e2e/framework/test_context.go
```diff
@@ -41,9 +41,9 @@ func registerFlags(t *TestContextType) {
 	flag.StringVar(&TestContext.ReportPrefix, "report-prefix", "", "Optional prefix for JUnit XML reports. Default is empty, which doesn't prepend anything to the default name.")
 	flag.StringVar(&TestContext.ReportDir, "report-dir", "", "Path to the directory where the JUnit XML reports should be saved. Default is empty, which doesn't generate these reports.")
 	flag.StringVar(&TestContext.SubmarinerNamespace, "submariner-namespace", "submariner", "Namespace in which the submariner components are deployed.")
-	flag.UintVar(&TestContext.ConnectionTimeout, "connection-timeout", 6, "The timeout in seconds per connection attempt when verifying communication between clusters.")
-	flag.UintVar(&TestContext.ConnectionAttempts, "connection-attempts", 10, "The number of connection attempts when verifying communication between clusters.")
-	flag.UintVar(&TestContext.OperationTimeout, "operation-timeout", 60, "The general operation timeout in seconds.")
+	flag.UintVar(&TestContext.ConnectionTimeout, "connection-timeout", 60, "The timeout in seconds per connection attempt when verifying communication between clusters.")
```
mangelajo: As tom said, in case of failure this will be long 60*4 = 2 minutes.

But it should be fine on ok-cases, I'm trying to avoid false negatives happening sometimes because
the listener timeouts while the connector is being spawned, 6 seconds is too short.

tpantelis: It was 6 sec but with 10 attempts for total 1 min.  I'm fine with increasing too 2 min.  I'm just not sure if it's better to have shorter timeout and more attempts or vice versa or if it really matters. Let's see how this works.

mangelajo:
https://travis-ci.com/submariner-io/submariner/jobs/290988735#L2167

Look at this ^ I think we would benefit from a slightly higher timeout for the failover tests, I will bump it a bit more.

Alternatively we can differentiate between listerner timeouts and connector timeouts?
Oh I see we do already
```
						{Name: "CONN_TIMEOUT", Value: strconv.Itoa(int(np.Config.ConnectionTimeout * np.Config.ConnectionAttempts))},
```

Let me balance them in a different way. May be we could have a timeouts multiplier for this case (gateway failovers)..

test/e2e/cluster/add_remove_cluster.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/submariner-io/submariner/test/e2e/tcp"
 )
 
+const clusterManagementTimeoutMultiplier = 3
```
mangelajo: As I was filling out the new parameter on other types of tests I figured out these disabled tests could benefit also from a multiplier.

test/e2e/redundancy/gateway_failover.go
```diff
@@ -83,7 +88,17 @@ func testGatewayFailOverScenario(f *framework.Framework) {
...
+	var newEnginePod *v1.Pod
+	for retries := 1; retries < 10; retries++ {
+		newEnginePod = f.AwaitSubmarinerEnginePod(framework.ClusterA)
+		if newEnginePod.Spec.NodeName == initialGatewayNode.Name {
+			time.Sleep(5 * time.Second)
+		} else {
+			break
+		}
+	}
+
```
tpantelis: What is this extra retry loop for? `AwaitSubmarinerEnginePod` already has a retry loop - if the timeout needs to be longer in this case then we should pass it in.

test/e2e/cluster/add_remove_cluster.go
tpantelis: I'd prefer not to add another config knob, especially a hidden internal one.  I'd prefer to either set the defaults high enough to accommodate all tests or add separate user configurable knobs for various test suites as needed, eg `dataplane-connection-timeout` and `redundancy-connection-timeout`.

test/e2e/redundancy/gateway_failover.go
mangelajo: Ahh, ok, I didn't see that, we need to increase the timeout there then. good point

test/e2e/cluster/add_remove_cluster.go
mangelajo: ok, I'll change it to be that way, as you said, probably that's cleaner.

test/e2e/redundancy/gateway_failover.go
tpantelis: This is already governed by the `OperationTimeout` - would bumping that to 120 sec suffice for this case?

mangelajo:
This has rotten, and I need to rework it for 0.3.0 along with the shipyard changes.

mangelajo:
@tpantelis I need to re-add this bit: 

https://github.com/submariner-io/submariner/commit/33a1fbfa907ede4743636ff0711cdc5e9ab623f4#diff-1d0d8aa6a74711b24485168c8c111b21R89

```
	var newEnginePod *v1.Pod
	for retries := 1; retries < 10; retries++ {
		newEnginePod = f.AwaitSubmarinerEnginePod(framework.ClusterA)
		if newEnginePod.Spec.NodeName == initialGatewayNode.Name {
			time.Sleep(5)
		} else {
			break
		}
	}

```

Because the next thing is checking the node name... and it will fail if the daemonset still didn't move the gateway around and it picked up the old host for the gateway.. AwaitSubmarinerEnginePod really waits for an Engine Pod... it really doesn't care about the host, which we check next.

test/e2e/redundancy/gateway_failover.go
```diff
@@ -94,9 +94,17 @@ func testGatewayFailOverScenario(f *subFramework.Framework) {
...
-	newEnginePod := f.AwaitSubmarinerEnginePod(framework.ClusterA)
+	var newEnginePod *v1.Pod
+	for retries := 1; retries < 10; retries++ {
+		newEnginePod = f.AwaitSubmarinerEnginePod(framework.ClusterA)
+		if newEnginePod.Spec.NodeName == initialGatewayNode.Name {
+			time.Sleep(5)
+		} else {
+			break
+		}
+	}
```
tpantelis, line 107: If the pod doesn't start for some reason then this will be a really long timeout 😄  We should be able to bake this into the underlying `AwaitUntil` call but that can be looked into later - let's get this in so we can get the tests running again.

mangelajo: Yes :+1: I thought the same :)

Let me update the patch, I was using the web interface and forgot to add some imports

mangelajo:
Oh @sridhargaddam this fails now for globalnet, I guess we need to make it globalnet aware.

PR #353, opened 2020-02-17
Vendor as needed

skitt:
vendor/modules.txt is updated whenever "go mod vendor" is run, so we
can use that to track vendor freshness compared to go.mod. This avoids
problems with an already-present but out-dated vendor directory when
we add or update dependencies.


tpantelis:
Needs a manual rebase.

Issue #283, opened 2020-01-14
Use periodic Travis triggers to expand test matrix


dfarrell07:
We currently only run CI on PR-based triggers, which requires us to keep job run times short enough to provide timely feedback to developers. However, as our CI has expanded, we have parts of our test matrix that are not being automatically tested. Ie, currently the KubeFed deploys are broken and no one knew because we had to run them manually/locally to see.

Add cron-type Travis jobs to cover additional parts of the test matrix:

* Helm-based deploys
* KubeFed deploys with Helm
* Kubefed deploys with Operator

dfarrell07:
Thinking about this more, it seems better to follow the OpenStack model of always running all checks against every change vs depending on periodic tests for some checks. We can avoid longer wait times on PRs by splitting Travis builds to use multiple parallel jobs, as I proposed in #357.

PR #357, opened 2020-02-17
Split CI into jobs for more concurrency, coverage

dfarrell07:
Split Travis CI configuration into three jobs to reduce execution time
through parallelism and achieve better test coverage of both
Operator-based and Helm-based deployments.

Instead of using the meta "ci" make target in both Operator/Helm jobs,
use the more narrow build/package targets and extract the common
test/validate to a third job. This reduces duplication and total can
shorten time-till-results.

Closes: #283

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,12 @@ branches:
 git:
   depth: false
 
+jobs:
+  include:
+  - env: CMD="make test validate"
+  - env: CMD="make build package e2e status=keep deploytool=operator"
+  - env: CMD="make build package e2e status=keep deploytool=helm"
+
 install:
   - sudo apt-get install moreutils # make ts available
 services:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,12 +23,12 @@ before_script:
   - CHANGED_FILES_PR=$(git diff --name-only HEAD $(git merge-base HEAD $TRAVIS_BRANCH))
 script:
   - set -o pipefail ;
-    make ci e2e status=keep deploytool=operator 2>&1 | ts '[%H:%M:%.S]' -s
+    $CMD 2>&1 | ts '[%H:%M:%.S]' -s
 after_success:
   - if [[ "${CHANGED_FILES_PR[@]}" =~ "scripts/kind-e2e/e2e.sh" ]]; then
```
dfarrell07: I think we should either remove this check (prefered), or additionally check when e2e libs are modified.

mangelajo: We could trigger a separate job on this case, but I think it's fine for now.

PR #359, opened 2020-02-18
Refactor engine logging

tpantelis:
This is a series of commits to enhance logging in the various engine components, one commit per Go file.  In general the changes encompass:

-  Use log.DEBUG approriately
-  Enhance log messages and formatting
-  Renamed some local vars to enhance clarity

### DIFF
--- a/main.go
+++ b/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/cableengine/ipsec"
 	"github.com/submariner-io/submariner/pkg/controllers/datastoresyncer"
 	"github.com/submariner-io/submariner/pkg/datastore"
+	"github.com/submariner-io/submariner/pkg/log"
 	"github.com/submariner-io/submariner/pkg/types"
 	"github.com/submariner-io/submariner/pkg/util"
 	v1 "k8s.io/api/core/v1"
@@ -60,7 +61,8 @@ func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
 
-	klog.V(2).Info("Starting submariner")
+	klog.Info("Starting the submariner gateway engine")
+
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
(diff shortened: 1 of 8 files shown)


pkg/controllers/datastoresyncer/datastoresyncer.go
```diff
@@ -31,8 +33,8 @@ type DatastoreSyncer struct {
 	submarinerClusterInformer  submarinerInformers.ClusterInformer
 	submarinerEndpointInformer submarinerInformers.EndpointInformer
 	datastore                  datastore.Datastore
-	localCluster               types.SubmarinerCluster
-	localEndpoint              types.SubmarinerEndpoint
+	localCluster               *types.SubmarinerCluster
+	localEndpoint              *types.SubmarinerEndpoint
```
skitt: Could you explain this in the commit message?

main.go
```diff
@@ -175,23 +181,22 @@ func main() {
 
 	leClient, err := kubernetes.NewForConfig(rest.AddUserAgent(cfg, "leader-election"))
 	if err != nil {
-		klog.Fatal(err)
+		klog.Fatalf("Error creating leader election kubernetes clientset: %s", err.Error())
 	}
 
 	eventBroadcaster := record.NewBroadcaster()
-	eventBroadcaster.StartLogging(klog.V(4).Infof)
+	eventBroadcaster.StartLogging(klog.V(log.DEBUG).Infof)
```
sridhargaddam: @tpantelis We are changing this from INFO (i.e., log level 4) to DEBUG (i.e., 2). I hope this is not a typo?

tpantelis: We're collapsing all previous levels (eg 2, 4, 6, 8) to just 2. I thought DEBUG was appropriate here - do you think it should be TRACE (3)?

pkg/controllers/datastoresyncer/datastoresyncer.go
tpantelis: It was to avoid more copying of the objects but it's negligible and really shouldn't be part of this PR. I changed it back.

main.go
sridhargaddam: I think its fine.

PR #363, opened 2020-02-24
Support for deploying Globalnet Controller with KIND/Helm

sridhargaddam:
This patch includes support in KIND (deploytool=helm) for deploying
Globalnet controller.

Usage:
make ci e2e status=keep deploytool=helm globalnet=true

Fixes issue: https://github.com/submariner-io/submariner/issues/381

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -98,8 +98,35 @@ function kind_import_images() {
     fi
 }
 
+function get_globalip() {
+    svcname=$1
+    gip=
+    attempt_counter=0
+    max_attempts=5
+    # It takes a while for globalIp annotation to show up on a service
+    until [[ $gip ]]; do
+        sleep 1
+        gip=$(kubectl get svc nginx-demo -o jsonpath='{.metadata.annotations.submariner\.io/globalIp}')
+        if [[ ${attempt_counter} -eq ${max_attempts} ]];then
+          echo "Max attempts reached, failed to get globalIp!"
+          exit 1
+        fi
+        attempt_counter=$(($attempt_counter+1))
+    done
+    echo $gip
+}
+
 function test_connection() {
-    nginx_svc_ip_cluster3=$(kubectl --context=cluster3 get svc -l app=nginx-demo | awk 'FNR == 2 {print $3}')
+    if [[ $globalnet = true ]]; then
+        nginx_svc_ip_cluster3=$(get_globalip nginx-demo)
+    else
+        nginx_svc_ip_cluster3=$(kubectl --context=cluster3 get svc -l app=nginx-demo | awk 'FNR == 2 {print $3}')
+    fi
+
+    if [[ -z "$nginx_svc_ip_cluster3" ]]; then
+        echo "Failed to get nginx-demo IP"
+        exit 1
+    fi
     netshoot_pod=$(kubectl --context=cluster2 get pods -l app=netshoot | awk 'FNR == 2 {print $1}')
 
     echo "Testing connectivity between clusters - $netshoot_pod cluster2 --> $nginx_svc_ip_cluster3 nginx service cluster3"
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -87,6 +87,15 @@ function kind_import_images() {
         kind --name cluster${i} load docker-image submariner:local
         kind --name cluster${i} load docker-image submariner-route-agent:local
     done
+
+    if [ ! -z "${global_CIDRs[cluster1]}" ]
+    then
+        docker tag quay.io/submariner/submariner-globalnet:$VERSION submariner-globalnet:local
+        for i in 1 2 3; do
+            echo "Loading globalnet image into cluster${i}..."
+            kind --name cluster${i} load docker-image submariner-globalnet:local
+        done
+    fi
 }
 
 function test_connection() {
@@ -191,14 +200,12 @@ function test_with_e2e_tests {
 function delete_subm_pods() {
     context=$1
     ns=$2
-    if kubectl --context=$context wait --for=condition=Ready pods -l app=submariner-engine -n $ns --timeout=60s > /dev/null 2>&1; then
-        echo Removing submariner engine pods...
-        kubectl --context=$context delete pods -n submariner -l app=submariner-engine
-    fi
-    if kubectl --context=$context wait --for=condition=Ready pods -l app=submariner-routeagent -n $ns --timeout=60s > /dev/null 2>&1; then
-        echo Removing submariner route agent pods...
-        kubectl --context=$context delete pods -n submariner -l app=submariner-routeagent
-    fi
+    for app in submariner-engine submariner-routeagent submariner-globalnet; do
+        if kubectl --context=$context wait --for=condition=Ready pods -l app=$app -n $ns --timeout=60s > /dev/null 2>&1; then
+            echo Removing $app pods...
+            kubectl --context=$context delete pods -n $ns -l app=$app
+        fi
+    done
 }
 
 function cleanup {
@@ -232,8 +239,14 @@ version=$2
 logging=$3
 kubefed=$4
 deploy=$5
+debug=$6
+globalnet=$7
 
-echo Starting with status: $status, k8s_version: $version, logging: $logging, kubefed: $kubefed, deploy: $deploy
+echo Starting with status: $status, k8s_version: $version, logging: $logging, kubefed: $kubefed, deploy: $deploy, debug: $debug, globalnet: $globalnet
+
+if [[ $globalnet = true ]]; then
+  declare -A global_CIDRs=( ["cluster1"]="169.254.1.0/24" ["cluster2"]="169.254.2.0/24" ["cluster3"]="169.254.3.0/24" )
+fi
 
 if [[ $status = clean ]]; then
     cleanup
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -81,43 +81,39 @@ function setup_custom_cni(){
 function kind_import_images() {
     docker tag quay.io/submariner/submariner:$VERSION submariner:local
     docker tag quay.io/submariner/submariner-route-agent:$VERSION submariner-route-agent:local
+    if [[ $globalnet = "true" ]]; then
+        docker tag quay.io/submariner/submariner-globalnet:$VERSION submariner-globalnet:local
+    fi
 
     for i in 1 2 3; do
         echo "Loading submariner images into cluster${i}..."
         kind --name cluster${i} load docker-image submariner:local
         kind --name cluster${i} load docker-image submariner-route-agent:local
-    done
-
-    if [ ! -z "${global_CIDRs[cluster1]}" ]
-    then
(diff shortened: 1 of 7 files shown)


sridhargaddam:
Request for review - @skitt @vthapar @tpantelis

scripts/kind-e2e/e2e.sh
```diff
@@ -86,6 +86,17 @@ function kind_import_images() {
         kind --name cluster${i} load docker-image submariner:local
         kind --name cluster${i} load docker-image submariner-route-agent:local
     done
+
+    if [ -z "${global_CIDRs[cluster1]}" ]
+    then
+        echo "Globalnet is not configured..."
```
vthapar: This is not needed. We already have an echo that shows all configured flags.

scripts/kind-e2e/lib_helm_deploy_subm.sh
```diff
@@ -56,9 +56,17 @@ function setup_broker() {
...
 function helm_install_subm() {
     cluster_id=$1
-    cluster_cidr=$2
-    service_cidr=$3
-    crd_create=$4
+    crd_create=$2
+    cluster_cidr=$3
+    service_cidr=$4
+    global_cidr=$5
+    if [ -z $global_cidr ]
```
vthapar: can be simplified as:

`globalnet_enable="true" && [[ -z $global_cidr ]] && globalnet_enable="false"`

sridhargaddam: Done

scripts/kind-e2e/e2e.sh
```diff
@@ -86,6 +86,15 @@ function kind_import_images() {
         kind --name cluster${i} load docker-image submariner:local
         kind --name cluster${i} load docker-image submariner-route-agent:local
     done
+
+    if [ ! -z "${global_CIDRs[cluster1]}" ]
+    then
+        docker tag quay.io/submariner/submariner-globalnet:dev submariner-globalnet:local
+        for i in 1 2 3; do
```
vthapar: Don't think we need globalnet image in cluster1. This adds un-necessary delay if image not needed.

sridhargaddam: Lets keep the code consistent with existing code. As you can see above, we are loading submariner images into cluster1 and the plan is to use cluster1 for some failover/redundancy tests.

scripts/kind-e2e/e2e.sh
```diff
@@ -86,6 +86,15 @@ function kind_import_images() {
         kind --name cluster${i} load docker-image submariner:local
         kind --name cluster${i} load docker-image submariner-route-agent:local
     done
+
+    if [ ! -z "${global_CIDRs[cluster1]}" ]
+    then
+        docker tag quay.io/submariner/submariner-globalnet:dev submariner-globalnet:local
```
mangelajo: please use $VERSION instead of dev, you may need to rebase this patch to latest master for that to work though.
```suggestion
        docker tag quay.io/submariner/submariner-globalnet:$VERSION submariner-globalnet:local
```

sridhargaddam: Done

scripts/kind-e2e/e2e.sh
```diff
@@ -87,10 +87,46 @@ function kind_import_images() {
...
+            kind --name cluster${i} load docker-image submariner-globalnet:local
+        done
+    fi
+}
+
+function get_globalip() {
+    svcname=$1
+    gip=
+    attempt_counter=0
+    max_attempts=5
```
tpantelis: Will 5 sec be enough all the time...  it wouldn't hurt to give this plenty more cushion, say 60.

scripts/kind-e2e/e2e.sh
```diff
@@ -87,10 +87,46 @@ function kind_import_images() {
...
+    # It takes a while for globalIp annotation to show up on a service
+    until [[ $gip ]]; do
+        sleep 1
```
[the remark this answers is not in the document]
vthapar: It was moved to beginning coz it is never initially present. The way e2e scripts are, we test connection right after creating service and there will always be some delay before globalIp shows up for a service.

scripts/kind-e2e/e2e.sh
```diff
@@ -87,10 +87,46 @@ function kind_import_images() {
...
+        gip=$(kubectl get svc nginx-demo -o jsonpath='{.metadata.annotations.submariner\.io/globalIp}')
+        if [[ ${attempt_counter} -eq ${max_attempts} ]];then
+          echo "Max attempts reached, failed to get globalIp!"
+          exit 1
+        fi
```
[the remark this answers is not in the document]
vthapar: Or just move globalIp get to after attempt check. Check after L109 and break would mean we're doing one more attempt than configured which would be wrong.

scripts/kind-e2e/e2e.sh
```diff
@@ -191,14 +227,12 @@ function test_with_e2e_tests {
...
     ns=$2
-    if kubectl --context=$context wait --for=condition=Ready pods -l app=submariner-engine -n $ns --timeout=60s > /dev/null 2>&1; then
-        echo Removing submariner engine pods...
-        kubectl --context=$context delete pods -n submariner -l app=submariner-engine
-    fi
-    if kubectl --context=$context wait --for=condition=Ready pods -l app=submariner-routeagent -n $ns --timeout=60s > /dev/null 2>&1; then
-        echo Removing submariner route agent pods...
-        kubectl --context=$context delete pods -n submariner -l app=submariner-routeagent
-    fi
+    for app in submariner-engine submariner-routeagent submariner-globalnet; do
```
mkolesnik: I would rather get the app name in the parameters and then do the iteration from outside the function, this makes it more reusable.

scripts/kind-e2e/e2e.sh
```diff
@@ -87,10 +87,46 @@ function kind_import_images() {
         kind --name cluster${i} load docker-image submariner:local
         kind --name cluster${i} load docker-image submariner-route-agent:local
     done
+
+    if [ ! -z "${global_CIDRs[cluster1]}" ]
```
mkolesnik: Why not put this conditional inside the loop on L85-L89? (Obviously check `"${global_CIDRs[cluster${i}]}"` in that case)

As a matter of fact you can check "$globalnet" = "true" which would be cleaner IMHO

scripts/kind-e2e/e2e.sh
```diff
@@ -87,10 +87,46 @@ function kind_import_images() {
...
+    until [[ $gip ]]; do
+        sleep 1
+        gip=$(kubectl get svc nginx-demo -o jsonpath='{.metadata.annotations.submariner\.io/globalIp}')
+        if [[ ${attempt_counter} -eq ${max_attempts} ]];then
+          echo "Max attempts reached, failed to get globalIp!"
+          exit 1
+        fi
+        attempt_counter=$(($attempt_counter+1))
+    done
+    echo $gip
```
mkolesnik: Let me suggest an alternative construct which will be clearer:

```suggestion
# It takes a while for globalIp annotation to show up on a service
function get_globalip() {
    svcname=$1
    gip=
    for i in {1..5}; do
        gip=$(kubectl get svc $svcname -o jsonpath='{.metadata.annotations.submariner\.io/globalIp}')
        if [ "$gip" != "" ]; then
            echo $gip
            return
        fi
    done
    
    echo "Max attempts reached, failed to get globalIp!"
    exit 1
```

P.S. you can skip the echo & exit in the end since you already check them externally.

scripts/kind-e2e/lib_helm_deploy_subm.sh
```diff
@@ -56,9 +56,12 @@ function setup_broker() {
...
 function helm_install_subm() {
     cluster_id=$1
-    cluster_cidr=$2
-    service_cidr=$3
-    crd_create=$4
+    crd_create=$2
+    cluster_cidr=$3
+    service_cidr=$4
+    global_cidr=$5
+    globalnet_enable="true" && [[ -z $global_cidr ]] && globalnet_enable="false"
```
mkolesnik: I think this will be more comprehensible:

```suggestion
    if [ -z $global_cidr ]; then globalnet_enable="false"; else globalnet_enable="true"; fi
```

scripts/kind-e2e/e2e.sh
sridhargaddam: Since I have to tag the image along with loading them into clusters, I used a separate if check.
I can update it, not an issue.

scripts/kind-e2e/e2e.sh
sridhargaddam: I was doing exactly the same thing (i.e., moved to for loop instead of until). I feel that let's keep the echo/exit atleast for sometime until we fully test globalnet.

scripts/kind-e2e/e2e.sh
sridhargaddam: Done

scripts/kind-e2e/lib_helm_deploy_subm.sh
sridhargaddam: my previous commit used to have this. I updated it based on @vthapar's comments. I too agree that its better to use if ... then .. else ... which is more readable.
@vthapar hope you are ok :)

test/e2e/e2e.go
```diff
@@ -48,6 +52,40 @@ var _ = ginkgo.SynchronizedAfterSuite(func() {
...
+
+	submClusterCRD, err := submarinerClient.SubmarinerV1().Clusters(testContext.SubmarinerNamespace).Get(testContext.KubeContexts[framework.ClusterB], metav1.GetOptions{})
+	if err != nil {
+		klog.Fatalf("Error retrieving clusterCRD from cluster %s, %v", testContext.KubeContexts[framework.ClusterB], err)
+	}
+
+	if submClusterCRD.Spec.GlobalCIDR != nil && len(submClusterCRD.Spec.GlobalCIDR) != 0 {
+		// Based on the status of GlobalnetEnabled, certain tests will be skipped/executed.
+		framework.TestContext.GlobalnetEnabled = true
+	}
```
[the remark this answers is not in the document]
sridhargaddam: Done, updated the api to use AwaitUtil

scripts/kind-e2e/lib_helm_deploy_subm.sh
tpantelis: I agree.  I found this line difficult to comprehend - we want it readable for non-scripting experts too 😄

test/e2e/e2e.go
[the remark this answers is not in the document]
tpantelis: I understand what you're saying but we're already recreating the clientsets etc in `BeforeEach` so I think querying the Cluster CRD is negligible. Frankly, I prefer that over duplicating code here. Also `BeforeEach` is on the normal test code path so we can use assertions etc.  That said, I think the kube config parsing, clientset creation and this new check really belong in a `BeforeSuite`.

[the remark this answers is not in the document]
tpantelis: How about we put this check in `BeforeEach` for now and I'll do a follow-up to move this and other code to `BeforeSuite`?  Also I'd be inclined to move the `GlobalnetEnabled` flag to `Framework`.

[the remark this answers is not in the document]
tpantelis: I can look into this in a follow-up.

PR #370, opened 2020-02-25
Disable route-agent resync interval

mangelajo:
This will remove unnecessary and noisy logging during debug,
and load on the k8s cluster.

The informer should receive any events, and our logic is robust to
make sure we handle all cases.

### DIFF
--- a/pkg/routeagent/main.go
+++ b/pkg/routeagent/main.go
@@ -27,10 +27,11 @@ var (
 )
 
 type SubmarinerRouteControllerSpecification struct {
-	ClusterID   string
-	Namespace   string
-	ClusterCidr []string
-	ServiceCidr []string
+	ClusterID             string
+	Namespace             string
+	ClusterCidr           []string
+	ServiceCidr           []string
+	ResyncIntervalMinutes int `default:"0"`
 }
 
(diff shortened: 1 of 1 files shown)


pkg/routeagent/main.go
```diff
@@ -61,14 +64,14 @@ func main() {
 		klog.Fatalf("Error building submariner clientset: %s", err.Error())
 	}
 
-	submarinerInformerFactory := submarinerInformers.NewSharedInformerFactoryWithOptions(submarinerClient, time.Second*30, submarinerInformers.WithNamespace(srcs.Namespace))
+	submarinerInformerFactory := submarinerInformers.NewSharedInformerFactoryWithOptions(submarinerClient, resyncInterval, submarinerInformers.WithNamespace(srcs.Namespace))
```
mangelajo: @tpantelis it seems like this was the breaking change, may be setting to 0 does not disable the resync interval?

tpantelis: It does but even if it didn't we shouldn't be seeing failures as resync has been occurring all along. It seems we're actually indirectly relying on the resync, ie something fails or doesn't work properly the first time around such that a subsequent resync fixes it. If so, I also wonder if this contributes to longer startup or fail-overs.

PR #389, opened 2020-02-26
DO NOT MERGE - Testing timing with Armada

mkolesnik:


mangelajo:
Please note that this takes also longer because the dapper image is rebuilt, and that's slow

Issue #396, opened 2020-02-26
Cleanup how CI jobs deploy artifacts after success


dfarrell07:
The current Travis jobs run a deployment section after each job. This has a number of issues:
* Verify jobs don't have an artifact to deploy, so they fail at that step.
* Both e2e jobs push artifacts, when only one needs to.

PR #400, opened 2020-02-27
Cleanup how CI jobs deploy artifacts after success

dfarrell07:
Use an environment variable to signal that exactly one job should run
CI's deployment stage on success. Fixes problems like:

* Verify jobs error on deploy because they have no artifact to push.
* Both e2e jobs push artifacts, which is an unnecessary duplication.

Closes: #396


dfarrell07:
Example of a failing verify job, per @mangelajo:

https://travis-ci.com/submariner-io/submariner/jobs/291822380

Issue #372, opened 2020-02-25
Get Cluster/Service CIDRs for Helm deploys dynamically from subctl or Armada, don't hardcode


dfarrell07:
[Part of the review](https://github.com/submariner-io/submariner/pull/317#discussion_r383649597) of #317 showed the need to remove a hardcoded link between the Cluster/Service CIDR vars passed to Helm deploys and the underlying deployment implementation (ie Armada internal defaults). That connection wasn't introduced by the proposed change, and has a fairly different scope, so it was agreed it should be handled separately.

Ideas so far:
* Use `subctl info` to scrape network details - via @mangelajo 
  * Well-developed, used by Operator path
* Scrape network details from Armada output - via @tpantelis
  * Doesn't make a dependency on subctl

dfarrell07:
Per feedback from @mkolesnik, and I agree, it seems the implementation (https://github.com/submariner-io/submariner/pull/406) isn't worth the complexity now that we aren't trying to support more generic deploys.

PR #406, opened 2020-02-28
Detect CIDRs vs hardcoding for Helm deploys

dfarrell07:
Use subctl's networking information tooling to dynamically detect
Cluster/Service CIDRs required by Helm. Avoids a fragile dependency on
the internal details of the deployment logic.

Closes: #372

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -206,6 +206,23 @@ function del_subm_gateway_label() {
     kubectl --context=$context label node $context-worker "submariner.io/gateway-" --overwrite
 }
 
+function get_latest_subctl_tag() {
+    curl https://api.github.com/repos/submariner-io/submariner-operator/releases | jq -r '.[0].tag_name'
+}
+
+function travis_retry() {
+    # We don't pretend to support commands with multiple words
+    $1 || (sleep 2 && $1) || (sleep 10 && $1)
+}
+
+function install_subctl() {
+    test -x /go/bin/subctl && return
+    version=$(travis_retry get_latest_subctl_tag || echo v0.1.1)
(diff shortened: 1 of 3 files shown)


dfarrell07:
Per recent discussions, I should probably swap the order of these commits to make this fix able to go in before the main adding-Armada commit, as that might not go in anytime soon (if at all).

mangelajo:
There's an armada patch in the middle

scripts/kind-e2e/e2e.sh
```diff
@@ -201,6 +201,23 @@ function del_subm_gateway_label() {
...
+}
+
+function travis_retry() {
+    # We don't pretend to support commands with multiple words
+    $1 || (sleep 2 && $1) || (sleep 10 && $1)
+}
+
+function install_subctl() {
+    test -x /go/bin/subctl && return
+    version=$(travis_retry get_latest_subctl_tag || echo v0.1.0)
```
skitt: Might as well bump to 0.1.1 now ;-).

dfarrell07: Done, thanks:)

scripts/kind-e2e/lib_helm_deploy_subm.sh
```diff
@@ -54,12 +55,33 @@ function setup_broker() {
...
+    echo $service_cidr
+}
+
 function helm_install_subm() {
     cluster_id=$1
     crd_create=$2
-    cluster_cidr=$3
-    service_cidr=$4
-    global_cidr=$5
+    global_cidr=$3
```
sridhargaddam: Thanks for fixing it, the CI logs look good.

dfarrell07:
> I'd like to question the necessity of this change in light of the decision to stay with bash scripts.
> 
> To me this adds a lot of complexity which doesn't give any tangible value since we're not going forward with armada.

Yeah as I think about it I guess you're right, this isn't worth the complexity anymore.

PR #411, opened 2020-03-02
Remove unused functions in Datastore interface

tpantelis:
Simplifies implementations and unit tests. We can add functions
back if ever needed.

### DIFF
--- a/pkg/datastore/phpapi/phpapi.go
+++ b/pkg/datastore/phpapi/phpapi.go
@@ -31,7 +31,7 @@ type Specification struct {
 	Server string
 }
 
-func NewPHPAPI(apitoken string) (*PHPAPI, error) {
+func NewPHPAPI(apitoken string) (datastore.Datastore, error) {
 	var pais Specification
 	err := envconfig.Process("backend_phpapi", &pais)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/datastore/phpapi/phpapi.go
```diff
@@ -31,7 +31,7 @@ type Specification struct {
 	Server string
 }
 
-func NewPHPAPI(apitoken string) (*PHPAPI, error) {
+func NewPHPAPI(apitoken string) (datastore.Datastore, error) {
```
skitt: Why change the return type?

tpantelis: No need to expose the internal type but more so the compiler will verify that `PHPAPI` implements `datastore.Datastore` here as well at call sites. But it's not that important - I can change it back if you want.

skitt: Oh, I read it the other way round, so I thought this would remove one level of type-checking ;-). So the new variant is better, thanks for the explanation!

mangelajo: @Oats87 also said we are free to kill the php api bits

tpantelis: Ok.  I'll do that later in a follow-up.

PR #413, opened 2020-03-03
Fix Travis env var syntax

dfarrell07:
The previous syntax resulted in only the second environment variable
being set. Put all env vars for each job on a single line.

Relates to: #396

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,7 @@ git:
 jobs:
   include:
   - env: CMD="make test validate"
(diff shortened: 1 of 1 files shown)


dfarrell07:
Can see the issue with the post-merge jobs here:
https://travis-ci.com/submariner-io/submariner/builds/151563213

In particular, this job:
https://travis-ci.com/submariner-io/submariner/jobs/293450454

Issue #412, opened 2020-03-03
Support in KIND (via Armada) to deploy clusters with Overlapping CIDRs


sridhargaddam:
Currently, KIND clusters are deployed with non-overlapping CIDRs (pod and svc). Inorder to validate Globanetnet Controller, we require clusters with overlapping CIDRs. This issue is to track the necessary enhancements in KIND/Armada for this use-case.

dfarrell07:
We decided not to go with Go-based Armada, so my patch became irrelevant, but @sridhargaddam ended up implementing this logic on the Bash-based path here: https://github.com/submariner-io/submariner/pull/436/files

PR #414, opened 2020-03-03
Add flag to deploy with overlapping CIDRs

dfarrell07:
Add flag to expose Armada's support for deploying clusters with
overlapping CIDRs.

Closes: #412

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ logging ?= false
 kubefed ?= false
 deploytool ?= helm
 armada ?= true
+overlap_cidrs ?= true
 debug ?= false
 
 TARGETS := $(shell ls scripts | grep -v dapper-image)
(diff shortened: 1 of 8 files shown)


tpantelis:
> If required, we can have some flag supported in armada (something like --overlap=true). The point I'm trying to make is - shall we hide this overlap_cidrs during KIND deployment and reuse globalnet (while internally mapping it to overlap_cidrs/any-other-flag-that-armada-uses)? Your thoughts?
> 

That makes sense to me.

dfarrell07:
> The point I'm trying to make is - shall we hide this overlap_cidrs during KIND deployment and reuse globalnet (while internally mapping it to overlap_cidrs/any-other-flag-that-armada-uses)? Your thoughts?

Yeah, we can do that, makes sense.

dfarrell07:
Per discussions with @tpantelis, @mkolesnik and many others - we decided not to move in the Go-based Armada direction. Overlapping CIDR support was added to the Bash path in different patches.

Issue #369, opened 2020-02-25
Share common deployment/e2e scripts in submariner-io/* repos


dfarrell07:
There is currently a duplication of kind/submariner deployment and e2e testing scripting between a number of subprojects of submariner-io. The duplication is resulting in fragmentation, as fixes and features go into some projects but not others. In order to avoid costly cross-porting, find a way to share common scripts between projects.

Ideas:
* Git submodule
  * Adds some extra complexity. Requires folks to clone them when cloning a repo. Might end up with complexity related to versions.
* Curling scripts from submariner-io/submariner
  * dfarrell prototyped this a bit and it turned out to be pretty messy, not a great design to dynamically add new code to the repo at runtime.

The three projects I know need attention are main-SubM, the Operator, and Lighthouse. These are their scripts directories.

```
[~/go/src/github.com/submariner-io/submariner/scripts]$ tree                                                                                                                                                           *[armada_kind_deploy]
.
├── broker-info.subm
├── broker-info.subm.2020-02-17T17_59_50Z
├── broker-info.subm.2020-02-18T13_06_35Z
├── build
├── build-engine
├── build-globalnet
├── build-routeagent
├── ci
├── codegen
├── dapper-image
├── e2e
├── entry
├── kind-e2e
│   ├── cluster1-config.yaml
│   ├── cluster2-config.yaml
│   ├── cluster3-config.yaml
│   ├── e2e.sh
│   ├── lib_armada_deploy_kind.sh
│   ├── lib_bash_deploy_kind.sh
│   ├── lib_helm_deploy_subm.sh
│   ├── lib_operator_deploy_subm.sh
│   ├── logging
│   │   ├── elasticsearch.yaml
│   │   └── filebeat.yaml
│   ├── netshoot.yaml
│   ├── nginx-demo.yaml
│   ├── nginx-federated.sh
│   └── README.md
├── lib
│   ├── debug_functions
│   ├── find_functions
│   └── version
├── output
│   ├── kind-clusters
│   └── kube-config
│       ├── container
│       └── local-dev
├── package
├── release
├── scripts
│   └── output
│       └── kube-config
│           └── local-dev
├── test
├── validate
└── vendor

[~/go/src/github.com/submariner-io/lighthouse/scripts]$ tree                                                                                                                                                                         [master]
.
├── build-controller
├── build-coredns
├── build-dnsserver
├── ci
├── e2e
├── entry
├── kind-e2e
│   ├── config
[...]

PR #419, opened 2020-03-06
Use portable paths for kind-e2e files

dfarrell07:
Allows for more robust use across submariner-io/* CIs, where the scripts
will be live in and be addressed from different locations.

Relates to: #369

### DIFF
--- a/scripts/kind-e2e/README.md
+++ b/scripts/kind-e2e/README.md
@@ -6,7 +6,7 @@ E2E environment consists of:
 - 3 k8s clusters deployed with [kind].
   - Cluster1: One master node for broker and elasticsearch if required.
   - Cluster{2..3}: One master node and two workers for gateways nodes.
-  The configuration can be viewed or changed at **scripts/kind-e2e/cluster{1..3}-config.yaml**.
+  The configuration can be viewed or changed at **/kind-e2e/cluster{1..3}-config.yaml**.
 - Submariner installed and configured on top of the clusters.
 
 [kind] is a tool for running local Kubernetes clusters using Docker container “nodes”.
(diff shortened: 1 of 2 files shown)


mangelajo:
Base changed on top of armada patch, don't merge until that one is merged.

dfarrell07:
> Base changed on top of armada patch, don't merge until that one is merged.

I did that for quite a number of patches, but the Go vs Bash Armada discussion is taking much longer than expected, so just basing on master so I can maybe make some progress.

mangelajo:
> Alternatively maybe we can add it to PATH in the Dockerfile?

Yes but with this change it becomes irrelevant where the script is copied and made available on the dockerfile, I guess that's the point. 

I agree we could add it to the path

mangelajo:
@dfarrell07 won't we need to do the same for other files, like yamls, yaml templates, etc?, example:

https://github.com/submariner-io/submariner/blob/master/scripts/kind-e2e/e2e.sh#L215

Anything referencing PRJ_ROOT?

dfarrell07:
@mangelajo Good catch, thanks! I think it's anything referencing the $prj_root/kind-e2e, but not $prj_root/output. That would keep the kubeconfig files in the same output dir they currently go in. Does that seem right to you too?

dfarrell07:
> Won't it be better to use DAPPER_SOURCE anyway?

It seems like you're doing that in #425, so if it's okay with you let's just keep this as-is?

mangelajo:
> @mangelajo Good catch, thanks! I think it's anything referencing the $prj_root/kind-e2e, but not $prj_root/output. That would keep the kubeconfig files in the same output dir they currently go in. Does that seem right to you too?

Correct, good catch too, let me look at the new version.

mangelajo:
@mkolesnik I think we should probably refactor to DAPPER_SOURCE, I remember we were using that at some point, I don't remember why we switched.

PR #422, opened 2020-03-06
Cleanup and merge my CI patch with dfarrel's

mangelajo:


scripts/release
```diff
@@ -7,16 +7,6 @@ DEPLOY="${DEPLOY:-false}"
...
-
-# This flag is passed from travis, so only the right jobs will deploy the
-# container images to quay
-
-if [ "$DEPLOY" != "true" ] ; then
-    echo "We don't need to deploy from this job, if you're trying to deploy manually set DEPLOY=true"
-    exit 0
-fi
-
-
```
dfarrell07: This logic was removed in shipyard/scripts/release, so it will go away here when we move to Shipyard. I think we don't need this PR anymore.

Issue #420, opened 2020-03-06
Convert e2e.sh to use getopts to accept params, allowing use in various repos


dfarrell07:
As we move to sharing e2e deployment logic better between submariner-io/* repos (https://github.com/submariner-io/submariner/issues/369), the `e2e.sh` script will become an entry point for multiple, somewhat-different CI systems. We currently expect the params to be an explicit, full, ordered list ($1 $2 $3...), which is very difficult to share. Should move to getopts to support proper flags, which will allow callers to not need knowledge about all params.

PR #423, opened 2020-03-06
Use getopt for robust params in e2e.sh

dfarrell07:
Convert e2e.sh to use getopt to accept parameters, vs positional
agreements. This makes calls more robust against breakages from changes
and doesn't require callers to be aware of all params.

Supports reusing e2e.sh and related scripts in other submariner-io/*
project CI systems.

Closes: #420

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -6,6 +6,27 @@ source $(dirname $0)/lib/version
 
 cd $(dirname $0)/
 
-./build-engine $6
-./build-routeagent $6
-./build-globalnet $6
+LONGOPTS=build_debug:
+# Only accept longopts, but must pass null shortopts or first param after "--" will be incorrectly used
+SHORTOPTS=""
+! PARSED=$(getopt --options=$SHORTOPTS --longoptions=$LONGOPTS --name "$0" -- "$@")
+eval set -- "$PARSED"
+
+while true; do
+    case "$1" in
+        --build_debug)
+            build_debug="$2"
+            ;;
+        --)
+            break
+            ;;
+        *)
+            echo "Ignoring unknown option: $1 $2"
+            ;;
+    esac
+    shift 2
+done
+
+./build-engine $build_debug
+./build-routeagent $build_debug
+./build-globalnet $build_debug
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -6,6 +6,8 @@ source $(git rev-parse --show-toplevel)/scripts/lib/version
 
 ### Variables ###
 
+KIND_REGISTRY=kind-registry
+
 ### Functions ###
 
 function print_logs() {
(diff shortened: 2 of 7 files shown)


scripts/build
```diff
@@ -6,6 +6,21 @@ source $(dirname $0)/lib/version
 
 cd $(dirname $0)/
 
-./build-engine $6
-./build-routeagent $6
-./build-globalnet $6
+# Outer loop causes getopts to not-quit on unknown arguments
```
skitt: It might be easier to understand if we move `set -e` after the arguments are parsed, instead of wrapping `getopts` in another loop.

scripts/kind-e2e/e2e.sh
```diff
@@ -274,15 +274,36 @@ function cleanup {
...
 ### Main ###
 
-status=$1
-version=$2
-logging=$3
-kubefed=$4
-deploy=$5
-debug=$6
-globalnet=$7
+# Outer loop causes getopts to not-quit on unknown arguments
```
mkolesnik: 1. Why won't we want to quit on unknown params? It's an error.
2. Perhaps it's better to use enhanced getopt which seems more robust, per [1].

[1] https://stackoverflow.com/a/29754866

scripts/build
mkolesnik: Why won't we want to quit on unknown params? It's an error.

I know the Makefile does some magic to declare the targets, but honestly I think we can live with them being defined statically and more correctly (i.e. not sending parameters where they're not needed).

scripts/kind-e2e/e2e.sh
dfarrell07: 1. Why won't we want to quit on unknown params? It's an error.

IMHO, it's actually a key feature in this case, not a bug. ;) The goal of this work is to make sharing scripts between different CI easier. By ignoring unknown params, different CIs can pass whatever params are relevant to them, and different scripts can catch only the ones they know how to handle.

So for example, the -b option only applies to building images with debug flags, not the e2e.sh script. The build scripts can respond to that flag, while the other scripts can ignore it, and vice versa for the other flags.

dfarrell07: > 2. Perhaps it's better to use enhanced getopt which seems more robust, per [1].

I thought about that, but went with getopts because it's a Bash built-in. But maybe that's not a good reason, we know the CI box will be Linux. We can try getopts, it'd be good to have long options.

mkolesnik: I'm not sure I agree, see my other comment about the unknown params from more details.
Also @skitt just told me he's working on improving the Makefile, so peerhaps it's best to coordinate with him on it?

scripts/build
dfarrell07: You're saying you want to get rid of the "make targets are automatically searched for as scripts in scripts/ dir" system? I'm not sure I'm on board with that one.

skitt: I do want to get rid of that part of the Makefiles, to avoid the issues with, *e.g.* `bin/subctl` in the operator project. However that’s a separate discussion, which shouldn’t affect this PR. I think this PR should preserve the current behaviour, so ignoring errors in arguments is acceptable IMO.
I stand by my comment on the outer loop, it shouldn’t be necessary.

mkolesnik: We could preserve it for now but I would expect it to be removed later on since it's quite error prone to do it so dynamically, and surely there's a better way to write the makefile :)

dfarrell07: Okay, understood and agreed @skitt and @mkolesnik. I think the current version gets rid of the loop y'all didn't like, moves to getopt, but doesn't scope-creep into additional changes to the makefile. Thanks for the good feedback!

dfarrell07:
CI is failing, but hard to imagine how it's related to this change.

```
[00:04:33.561099] [scripts]$ setup_custom_cni
[00:04:33.562116] 
[00:04:33.656369] Applying weave network in to cluster2...
[00:04:33.749912] [scripts]$ kubectl --context=cluster2 apply -f https://cloud.weave.works/k8s/net?k8s-version=Q2xpZW50IFZlcnNpb246IHZlcnNpb24uSW5mb3tNYWpvcjoiMSIsIE1pbm9yOiIxNSIsIEdpdFZlcnNpb246InYxLjE1LjIiLCBHaXRDb21taXQ6ImY2Mjc4MzAwYmViYmI3NTAzMjhhYzE2ZWU2ZGQzYWE3ZDM1NDk1NjgiLCBHaXRUcmVlU3RhdGU6ImFyY2hpdmUiLCBCdWlsZERhdGU6IjIwMTktMDgtMTZUMTQ6MTQ6NTFaIiwgR29WZXJzaW9uOiJnbzEuMTIuNyIsIENvbXBpbGVyOiJnYyIsIFBsYXRmb3JtOiJsaW51eC9hbWQ2NCJ9Cg==&env.IPALLOC_RANGE=10.245.0.0/16
[00:04:33.751556] [scripts]$ kubectl version
[00:04:33.752774] [scripts]$ base64
[00:04:33.756281] [scripts]$ tr -d \n
[00:04:33.841925] The connection to the server localhost:8080 was refused - did you specify the right host or port?
[00:04:33.933720] error: context "cluster2" does not exist
```

PR #427, opened 2020-03-09
Local registry

mkolesnik:
Use local registry for the images instead of loading them into the clusters


docs/testing.md
```diff
@@ -141,8 +141,8 @@ To test specific k8s version, additional **version** parameter can be passed to
 make ci e2e version=1.14.1
 ```
 
-Full list of supported k8s versions can found on [kind release page] page. We are using kind vesion 0.3.0.
-Default **version** is 1.14.2.
+Full list of supported k8s versions can found on [kind release page] page. We are using kind vesion 0.6.1.
+Default **version** is 1.14.6
```
mangelajo: We could even use something newer if we needed...

mkolesnik: I agree but it's not the focus of the patch..

It was surprisingly hard enough to track down a working version (and it's not even the latest on the 1.14 line), so I didn't want to spend too much time on this.

We can, and should, bump the versions in a different PR.

PR #431, opened 2020-03-10
Add GetName to drivers

roytman:
Current implementation of `[StartEngine](https://github.com/submariner-io/submariner/blob/b83e23dacbbbfc1c534dae7eba2d87a9fbb3c16b/pkg/cableengine/cableengine.go#L59)`  has the following log statement `Starting IPSec Engine (Charon)"`, which is not true for other driver implementations. 
This PR adds a method GetName to drivers.


pkg/cableengine/cableengine.go
```diff
@@ -56,7 +56,7 @@ func NewEngine(localSubnets []string, localCluster types.SubmarinerCluster, loca
 }
 
 func (i *engine) StartEngine() error {
-	klog.Infof("Starting IPSec Engine (Charon)")
+	klog.Infof("Starting %s", i.driver.GetName())
```
tpantelis: I pushed https://github.com/submariner-io/submariner/pull/429 to actually remove this line and let the implementation announce that it's starting or whatever it wants to in its `Init` method (which the current lone implementation already does). This is fine too but I think we should be clear about the intended usage of `GetName` in the API contract, eg will it be used as an identifier, how does it differ from the name specified on registration...  If we're intending this only for logging, perhaps change the method name to `GetDisplayName`?  

Or maybe I'm just overthinking this 😄

mangelajo: Yeah, I'd agree with "GetDisplayName"

tpantelis: But is it worth adding a method to the API contract just for this one log message instead of just letting the implementation do logging as it sees fit (and as ipsec already does)?

mangelajo: We could use this for reporting on cable status later, etc... It's not only about logging.

mangelajo: I think it's better to add it, because we can use that for other reasons: status reporting API, etc.. which is one of the items I have for this sprint.

mangelajo: Anyway I think this discussion is splitting hairs...

mangelajo: With the above in mind (using it for status report via API etc..) I think "GetName" is perfectly fine

tpantelis: Ok

PR #434, opened 2020-03-11
Extend EndpointSpec comparisons

roytman:
Original EndpointSpec comparisons were limited to `ClusterID`, `CableName` and the `Hostname`. 
In the case of wireguard it is not enough. 
This PR adds to comparison `Backend` and `BackendConfig`.


mangelajo:
Thanks!

Issue #438, opened 2020-03-12
Post merge job doesn't rebuild the dapper base image


mkolesnik:
We noticed on this build [1] that the dapper base image wasn't rebuilt, even though the PR was updating it and the CI for the PR itself passed with no problems.

[1] https://travis-ci.com/github/submariner-io/submariner/builds/152954775

PR #439, opened 2020-03-12
Builds the dapper image when a PR is merged to master

mangelajo:
Our scripts had an error and didn't detect dapper image changes on
a master merge.

Fixes: #438

### DIFF
--- a/scripts/dapper-image
+++ b/scripts/dapper-image
@@ -23,8 +23,15 @@ DAPPER_BASE_IMAGE=${REPO}/dapper-base:${TAG}
 UPSTREAM_REMOTE=$(git remote -v | grep submariner-io/submariner | awk '{ print $1 }' | head -n 1)
 CHANGED_FILES_LOCAL=$(git diff --name-only remotes/$UPSTREAM_REMOTE/master)
 
+# Detect when we are on a merged patch
+LAST_COMMIT_LOCAL=$(git rev-parse HEAD)
+LAST_COMMIT_MASTER=$(git rev-parse remotes/$UPSTREAM_REMOTE/master)
+
+
+
 if [[ "${CHANGED_FILES_PR[@]}" =~ "package/Dockerfile.dapper-base" ]] ||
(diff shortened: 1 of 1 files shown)


scripts/dapper-image
```diff
@@ -23,8 +23,15 @@ DAPPER_BASE_IMAGE=${REPO}/dapper-base:${TAG}
...
+# Detect when we are on a merged patch
+LAST_COMMIT_LOCAL=$(git rev-parse HEAD)
+LAST_COMMIT_MASTER=$(git rev-parse remotes/$UPSTREAM_REMOTE/master)
+
+
+
 if [[ "${CHANGED_FILES_PR[@]}" =~ "package/Dockerfile.dapper-base" ]] ||
-   [[ "${CHANGED_FILES_LOCAL[@]}" =~ "package/Dockerfile.dapper-base" ]] ; then
+   [[ "${CHANGED_FILES_LOCAL[@]}" =~ "package/Dockerfile.dapper-base" ]] ||
+   [[ "${LAST_COMMIT_LOCAL}" == "${LAST_COMMIT_MASTER}" ]]; then
```
mkolesnik: Wouldn't this rebuild it for each merge, no matter if the dapper was modified or not?

dfarrell07: That seems to be what he wants, according to the commit message. +1 for clarification.

mkolesnik: Yeah but this AFAICT will rebuild on each time the last commit is a merge, whether the PR changed the base image or not, which I'm not sure is desired since it'll generate these images a whole lot more but most of them will be unchanged

mangelajo: I think it's fine :) we're hitting this again.

It's an ok cost to pay for the meanwhile, we can optimize this later ... the big churn of builds is PRs, not merges.

mangelajo:
> Requesting clarification

Sorry, I didn't see your message. 

As we said over IRC, let's merge it, and then in shipyard side we can optimize, this will fix the dapper image which wasn't built.

PR #441, opened 2020-03-12
Initial WireGuard driver implementation

deanlorenz:


pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+
+	// del routes
+	var wg netlink.Link
+	if wg, err = netlink.LinkByName(DefaultDeviceName); err != nil {
+		return fmt.Errorf("failed to find wireguard device by name: %v", err)
+	}
+	var cidr *net.IPNet
+	for _, sn := range remoteEndpoint.Spec.Subnets {
+		if _, cidr, err = net.ParseCIDR(sn); err != nil {
+			return fmt.Errorf("failed to parse subnet %s: %v", sn, err)
```
tpantelis: Perhaps just log the error and continue? Either way we shouldn't get a bad CIDR...

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+		route := netlink.Route{
+			LinkIndex: wg.Attrs().Index,
+			Dst:       cidr,
+		}
+		if err = netlink.RouteDel(&route); err != nil {
+			return fmt.Errorf("failed to delete route %s: %v", route.String(), err)
+		}
+	}
+
+	klog.V(log.TRACE).Infof("Done removing endpoint for cluster %s", remoteEndpoint.Spec.ClusterID)
```
tpantelis: I notice you use TRACE level throughout. Any particular reason to not use DEBUG? We expect DEBUG to be the norm with TRACE only used for really noisy logs or ones that may dump a lot of data.

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+	peerCfg := []wgtypes.PeerConfig{{
+		PublicKey: remoteKey,
+		Remove:    true,
+	}}
+	if err = w.client.ConfigureDevice(DefaultDeviceName, wgtypes.Config{
+		ReplacePeers: false,
+		Peers:        peerCfg,
+	}); err != nil {
+		return fmt.Errorf("failed to remove wireguard peer with key %s: %v", remoteKey.String(), err)
+	}
```
tpantelis: L322-332 are largely duplicated in `ConnectToEndpoint` - please create a function.

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+			Port: DefaultListenPort,
+		},
+		PersistentKeepaliveInterval: nil,
+		ReplaceAllowedIPs:           true,
+		AllowedIPs:                  allowedIPs,
+	}}
+	if err = w.client.ConfigureDevice(DefaultDeviceName, wgtypes.Config{
+		ReplacePeers: false,
+		Peers:        peerCfg,
+	}); err != nil {
```
tpantelis: Please separate lines to improve readability.

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+			LinkIndex: wg.Attrs().Index,
+			Dst:       &peerNet,
+		}
+		if err = netlink.RouteAdd(&route); err != nil {
+			return "", fmt.Errorf("failed to add route %s: %v", route.String(), err)
+		}
+	}
+
+	klog.V(log.TRACE).Infof("Successfully connected endpoint peer %+v", peerCfg)
+
```
tpantelis: Can you split this into separate functions where appropriate?

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+		if !found {
+			klog.Errorf("Failed to verify peer configuration")
+		}
+	}
+	w.peers[remoteEndpoint.Spec.ClusterID] = remoteKey
+
+	// Add routes to peer
+	//TODO save old routes for removal
+	var wg netlink.Link
+	if wg, err = netlink.LinkByName(DefaultDeviceName); err != nil {
```
tpantelis: Can't use use `w.link` instead of retrieving it again?

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	if keyMismatch {
+		klog.Warningf("Key mismatch for peer cluster %s, keeping existing routes", remoteEndpoint.Spec.ClusterID)
+		return nil
+	}
+	delete(w.peers, remoteEndpoint.Spec.ClusterID)
+
+	// del routes
+	var wg netlink.Link
+	if wg, err = netlink.LinkByName(DefaultDeviceName); err != nil {
```
tpantelis: Can't use use `w.link` instead of retrieving it again?

pkg/cable/wireguard/WGdriver.go
tpantelis: I'm not clear on why you proceed with removing the peer if the keys didn't match. Is this case expected under certain circumstances? Can you add comments to explain this?

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
+package wireguard
```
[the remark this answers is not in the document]
mangelajo: being a nit, but I agree, we try to avoid repetition, this would be 

cable/wireguard/driver , quite meaningful :)

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+	peers         map[string]wgtypes.Key // clusterID -> publicKey
+	mutex         sync.Mutex
+	client        *wgctrl.Client
+	link          netlink.Link
+	//debug   bool
+	//logFile string
+}
+
+// NewWGDriver creates a new Wireguard driver
+func NewWGDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (cable.Driver, error) {
```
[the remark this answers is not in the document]
mangelajo: :+1: as when used in code it would read wireguard.NewDriver(...)

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+		FirewallMark: nil,
+		ReplacePeers: false,
+		Peers:        peerConfigs,
+	}
+	if err = wg.client.ConfigureDevice(DefaultDeviceName, cfg); err != nil {
+		return nil, fmt.Errorf("failed to configure wireguard device: %v", err)
+	}
+
+	klog.V(log.TRACE).Infof("Initialized wireguard %s with publicKey %s", DefaultDeviceName, pub.String())
+	return &wg, nil
```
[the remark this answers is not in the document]
mangelajo: :+1:

pkg/cable/wireguard/WGdriver.go
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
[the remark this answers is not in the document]
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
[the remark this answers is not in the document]
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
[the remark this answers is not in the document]
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+
+	// public key
+	var key string
+	if key, found = remoteEndpoint.Spec.BackendConfig[PublicKey]; !found {
+		return fmt.Errorf("missing peer public key")
+	}
+	var remoteKey wgtypes.Key
+	if remoteKey, err = wgtypes.ParseKey(key); err != nil {
+		return fmt.Errorf("failed to parse public key %s: %v", key, err)
+	}
```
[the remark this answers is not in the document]
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+	// wg remove
+	klog.V(log.TRACE).Infof("Removing wireguard peer with key %s", remoteKey.String())
+	peerCfg := []wgtypes.PeerConfig{{
+		PublicKey: remoteKey,
+		Remove:    true,
+	}}
+	if err = w.client.ConfigureDevice(DefaultDeviceName, wgtypes.Config{
+		ReplacePeers: false,
+		Peers:        peerCfg,
+	}); err != nil {
```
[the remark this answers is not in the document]
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
```diff
@@ -0,0 +1,365 @@
...
+		return fmt.Errorf("failed to remove wireguard peer with key %s: %v", remoteKey.String(), err)
+	}
+	if keyMismatch {
+		klog.Warningf("Key mismatch for peer cluster %s, keeping existing routes", remoteEndpoint.Spec.ClusterID)
+		return nil
+	}
+	delete(w.peers, remoteEndpoint.Spec.ClusterID)
```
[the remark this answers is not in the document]
deanlorenz: Added comments in code. Cannot delete peers entry until checking mismatch -- clusterID may have a valid entry with a different key.

pkg/cable/wireguard/WGdriver.go
deanlorenz: Done, ignoring malformed subnets. Might create peers with no allowed IPs.

pkg/cable/wireguard/WGdriver.go
deanlorenz: Moved all to DEBUG

pkg/cable/wireguard/WGdriver.go
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
deanlorenz: done

pkg/cable/wireguard/WGdriver.go
deanlorenz: added comment in code to explain meaning of key mismatch

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,411 @@
...
+	return nil
+}
+
+func (w *wireguard) GetActiveConnections(clusterID string) ([]string, error) {
+	// force caller to skip duplicate handling
+	return make([]string, 0), nil
+}
+
+// Create new wg link and assign addr from local subnets
+func setWGLink(w *wireguard, localSubnets []string) error {
```
tpantelis: minor thing but you could `w *wireguard` the receiver instead of explicitly passing in.

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,411 @@
...
+
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return "", fmt.Errorf("net.InterfaceAddrs() returned error : %v", err)
+	}
+
+	for _, a := range addrs {
+		ip, _, err := net.ParseCIDR(a.String())
+		if err != nil {
+			klog.V(log.DEBUG).Infof("Skipping local address %v, unable to ParseCIDR: %v", a, err)
```
tpantelis: It seems this should be logged as error or warning as we do on L365? I assume this isn't expected...

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	if err := netlink.LinkAdd(link); err == nil {
+		w.link = link
+	} else {
+		return fmt.Errorf("failed to add WireGuard device: %v", err)
+	}
+
+	// parse localSubnets and get internal address
+	w.localSubnets = parseSubnets(localSubnets)
+	ip, err := discoverInternalIP(w.localSubnets)
```
tpantelis: What if `discoverInternalIP` returned an error indicating it wasn't able to determine if an internal address exists? Should we still proceed with using the endpoint IP?

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+// parse CIDR string and skip errors
+func parseSubnets(subnets []string) []net.IPNet {
+	nets := make([]net.IPNet, 0, len(subnets))
+	for _, sn := range subnets {
+		_, cidr, err := net.ParseCIDR(sn)
+		if err != nil {
+			// this should not happen. Log and continue
+			klog.Errorf("failed to parse subnet %s: %v", sn, err)
```
tpantelis: Should continue here and not fall thru to L367.

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,411 @@
...
+			Remove:    true,
+		},
+	}
+	err := c.ConfigureDevice(DefaultDeviceName, wgtypes.Config{
+		ReplacePeers: false,
+		Peers:        peerCfg,
+	})
+	if err != nil {
+		klog.Errorf("Failed to remove WireGuard peer with key %s: %v", key, err)
+		return err
```
tpantelis: Since the call sites ignore the returned error, I'd prefer to not return it. I think call sites look cleaner w/o  having to suppress it with `_`.

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,411 @@
...
+
+type wireguard struct {
+	localSubnets  []net.IPNet
+	localEndpoint types.SubmarinerEndpoint
+	peers         map[string]wgtypes.Key // clusterID -> publicKey
+	mutex         sync.Mutex
+	client        *wgctrl.Client
+	link          netlink.Link
+	//debug   bool
+	//logFile string
```
tpantelis: Please remove commented lines.

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,411 @@
...
+		ReplacePeers: false,
+		Peers:        peerCfg,
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to configure peer: %v", err)
+	}
+
+	// verify peer was added
+	// TODO verify configuration
+	if p, err := peerByKey(w.client, remoteKey); err != nil {
```
tpantelis: Shouldn't success from `ConfigureDevice` above be enough to know that the peer was added?

pkg/cable/wireguard/driver.go
deanlorenz: I don't have a strong opinion either way, but the errors are slightly different. 
L365 is submariner error -- malformed subnets -- we can continue but something is probably not working as sub CIDRs will not be added to allowed IPs and not routed to other cluster.
L342 is netlink library error or some weird device that returns an unexpected address -- we simply don't use it and look for another internal address. Only if we cannot find any address this becomes an error at L355.

pkg/cable/wireguard/driver.go
deanlorenz: that is what is attempted at L311. 
Using endpoint IP will probably cause problems for pods on host network.

pkg/cable/wireguard/driver.go
deanlorenz: The error path at the caller is currently very basic we may want to improve it later. I find that ignoring an error explicitly indicates that. I'd rather have the utility return the error as a failure indication. 
I actually find it cleaner as is, but if you prefer I can remove it.

pkg/cable/wireguard/driver.go
deanlorenz: It should, but we had some silent failures in our tests. So added this for better debugging.
Peer configurations is not that common, so I don't think it is a problem to verify.

pkg/cable/wireguard/driver.go
deanlorenz: oops. good catch. Done.

pkg/cable/wireguard/driver.go
tpantelis: 👍 makes sense.

pkg/cable/wireguard/driver.go
deanlorenz: done

pkg/cable/wireguard/driver.go
tpantelis: I understand what you're saying. It makes sense to return an error and not assume the caller will ignore it, although we are kind of assuming that by logging the error on L386 😄 It's fine the way it is.

pkg/cable/wireguard/driver.go
tpantelis: L311 is a log statement - did you mean L314? 😄 The call site here can't distinguish between an internal address not found and an error trying to determine if an internal address exists from `discoverInternalIP`.  For the former,  we certainly want to proceed to L309 to use the endpoint IP.  I'm unclear about the latter but if we still want to proceed with attempting to use the endpoint IP in this case as well then I think we should at least log the error. I propose to return nil from `discoverInternalIP` if no address is found and log the returned error here, ie:

```
ip, err := discoverInternalIP(w.localSubnets)
if err != nil {
    klog.Errorf("Error attempting to discover an internal address: %v", err)
}

if ip == nil {
    klog.V(log.DEBUG).Infof("Using endpoint IP as internal address; %v", err)
    ip = endpointIP(&w.localEndpoint)
}

```

pkg/cable/wireguard/driver.go
deanlorenz: done

pkg/cable/wireguard/driver.go
deanlorenz: Actually L309 (removed 2 comments above...).
I see your point; however, `discoverInternalIP` returns `""` in both cases.
I can change `discoverInternalIP` to only log the 'not found' case, but I still see both cases as error, and still think the best thing we can do is try the ep IP.

tpantelis: > Actually L309 (removed 2 comments above...).
> I see your point; however, `discoverInternalIP` returns `""` in both cases.

yeah, sorry, return `""` instead of nil.

> I can change `discoverInternalIP` to only log the 'not found' case, but I still see both cases as error, and still think the best thing we can do is try the ep IP.

yeah try the endpoint IP in both cases - I'm just suggesting to log the error if something failed preventing the determination so we don't lose it in case it might be helpful for troubleshooting. So the last line in `discoverInternalIP`  where it's not found  would return  `"", nil`.

deanlorenz: done

tpantelis:
Can you use another PR for https://github.com/submariner-io/submariner/pull/441/commits/4a015470685b53b66a1c2121493c31410e103b7b? The e2e scripting is being refactored right now with most of it moving to the shipyard project so it's in a holding pattern for new features right now until that work is done.

deanlorenz:
Added option to make e2e to allow setting wireguard as cable driver:
`make e2e cable_driver=wireguard`
Removed code to set wireguard as default driver

@mangelajo the make option did not work for me without changing `scripts/e2e` to `kind-e2e/e2e.sh` (not use `/opt/`. Even `make dapper-image` did not help.

deanlorenz:
@tpantelis reverted the make change and added PR [#476](https://github.com/submariner-io/submariner/pull/476)

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,416 @@
...
+			klog.V(log.DEBUG).Infof("Skipping local address %+v: not IP4", ip)
+			continue
+		}
+		for _, c := range cidrs {
+			if c.Contains(ip) {
+				return ip.String(), nil
+			}
+		}
+	}
+	klog.Warningf("could not find internal address in local subnets")
```
tpantelis:
```suggestion
	klog.Warningf("Could not find an internal address in %v that matches a local subnet in %v", addrs, cidrs)
```

Sorry one more request - since this is a warning, it seems it would be useful to include some context for possible troubleshooting. If you think otherwise then we can leave as is.

pkg/cable/wireguard/driver.go
```diff
@@ -0,0 +1,416 @@
...
+	DefaultListenPort = 5871
+
+	// DefaultDeviceName specifies name of WireGuard network device
+	DefaultDeviceName = "subwg0"
+
+	// PublicKey is name (key) of publicKey entry in back-end map
+	PublicKey = "publicKey"
+
+	// we assume Linux
+	//deviceType = wgtypes.LinuxKernel
```
mkolesnik: This needed?

pkg/cable/wireguard/README.md
```diff
@@ -0,0 +1,53 @@
...
+- WireGuard needs to be [installed](https://www.wireguard.com/install "WireGuard installation instructions") on the gateway nodes. For example, (Ubuntu < 19.04),
+  ```ShellSession
+  $ sudo add-apt-repository ppa:wireguard/wireguard
+  $ sudo apt-get update
+  $ sudo apt-get install wireguard
+  ```
+
+- The driver needs to be enabled with
+  ```ShellSession
+  $ bin/subctl join --kubeconfig wg3-conf --cable-driver wireguard --disable-nat  --repository roytman --version  latest broker-info.subm
```
mkolesnik: I'd drop the repo & version flags (unless they have some significance specifically for WG?)

pkg/cable/wireguard/driver.go
deanlorenz: This was a placeholder for user-space wireguard support. 
I removed for now and will add comment to README

pkg/cable/wireguard/README.md
deanlorenz: They are no longer significant, as latest subctl now supports the --cable-driver option. Removed.

Issue #440, opened 2020-03-12
Bash error message when globalnet controller fails to annotate a service


sridhargaddam:
When deploying a KIND cluster with globalnet=true, and for some
reason if the globalnet POD fails to annotate nginx-demo service
with a globalIp, KIND should fail with an appropriate error message.

Currently build fails with the following error

<SNIP>
[scripts]$ kubectl get svc nginx-demo -o jsonpath={.metadata.annotations.submariner\.io/globalIp}
./../scripts/kind-e2e/e2e.sh: line 105: [: !=: unary operator expected
[scripts]$ sleep 1
[scripts]$ gip=
[scripts]$ kubectl get svc nginx-demo -o jsonpath={.metadata.annotations.submariner\.io/globalIp}
./../scripts/kind-e2e/e2e.sh: line 105: [: !=: unary operator expected
[scripts]$ sleep 1
[scripts]$ gip=
[scripts]$ kubectl get svc nginx-demo -o jsonpath={.metadata.annotations.submariner\.io/globalIp}
./../scripts/kind-e2e/e2e.sh: line 105: [: !=: unary operator expected
[scripts]$ sleep 1
[scripts]$ gip=
[scripts]$ kubectl get svc nginx-demo -o jsonpath={.metadata.annotations.submariner\.io/globalIp}
./../scripts/kind-e2e/e2e.sh: line 105: [: !=: unary operator expected
[scripts]$ sleep 1
[scripts]$ gip=
[scripts]$ kubectl get svc nginx-demo -o jsonpath={.metadata.annotations.submariner\.io/globalIp}

PR #442, opened 2020-03-12
Fix error message when globalnet fails to annotate a service

sridhargaddam:
When deploying a KIND cluster with globalnet=true, and for some
reason if the globalnet controller fails to annotate the nginx-demo
service with a globalIp, KIND should fail with an appropriate error
message.

Fixes Issue: https://github.com/submariner-io/submariner/issues/440

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -116,11 +116,12 @@ function kind_import_images() {
 
 function get_globalip() {
     svcname=$1
+    context=$2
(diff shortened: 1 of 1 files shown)


sridhargaddam:
This PR is failing in CI because of some recent changes in submariner-charts repo. 
Basically, we need a new chart with the following PR - https://github.com/submariner-io/submariner-charts/commit/116615040a3df2a6212b935e342ce346c227cae3

PR #444, opened 2020-03-16
Shrink the dapper-base image

skitt:
This makes the following changes:
* avoid pulling in weak dependencies on distrosync
* clean out a number of unnecessary packages
* compress the large Go binaries we end up installing

removing 400 MiB from the image (1.83 GiB down to 1.43 GiB).

### DIFF
--- a/package/Dockerfile.dapper-base
+++ b/package/Dockerfile.dapper-base
@@ -30,14 +30,21 @@ ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${ARC
 # findutils      | e2e cleanup (xargs)
 # upx            | binary compression
 # jq             | JSON processing (GitHub API)
-RUN dnf -y distrosync && \
+# diffutils      | required for goimports
+RUN dnf -y distrosync --nodocs --setopt=install_weak_deps=False && \
     dnf -y install --nodocs --setopt=install_weak_deps=False gcc git-core curl moby-engine make golang kubernetes-client findutils upx jq && \
+    dnf -y remove acl dnf-yum langpacks-en libevent libsemanage libsss_idmap libxcrypt-compat mkpasswd openssl python3-pip python3-setuptools \
+           sssd-client unbound-libs whois-nls && \
+    rpm -e --nodeps selinux-policy-targeted && \
     dnf -y clean all && \
     curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin ${LINT_VERSION} && \
     curl "https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-${ARCH}.tar.gz" | tar -xzf - && \
-    cp linux-${ARCH}/helm /usr/bin/ && chmod a+x /usr/bin/helm && \
-    curl -LO "https://github.com/kubernetes-sigs/kubefed/releases/download/v${KUBEFED_VERSION}/kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz" && \
-    tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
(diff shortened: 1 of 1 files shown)


package/Dockerfile.dapper-base
```diff
@@ -40,4 +42,6 @@ RUN dnf -y distrosync && \
     tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
     curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
     GOFLAGS="" go get -v github.com/onsi/ginkgo/ginkgo && \
-    GOFLAGS="" go get -v golang.org/x/tools/cmd/goimports
+    GOFLAGS="" go get -v golang.org/x/tools/cmd/goimports && \
```
mangelajo: May be we can also remove the downloaded sources for that, what do you think?, we only want the binary, as other deps come via "vendor" directory later.

skitt: I’d rather upgrade to Fedora 31 and replace this with the `golang-x-tools-goimports` package ;-).

package/Dockerfile.dapper-base
```diff
@@ -40,4 +42,6 @@ RUN dnf -y distrosync && \
     tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
```
mkolesnik:
```suggestion
    tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && mv kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl && \
```

package/Dockerfile.dapper-base
```diff
@@ -40,4 +42,6 @@ RUN dnf -y distrosync && \
...
     curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
     GOFLAGS="" go get -v github.com/onsi/ginkgo/ginkgo && \
-    GOFLAGS="" go get -v golang.org/x/tools/cmd/goimports
+    GOFLAGS="" go get -v golang.org/x/tools/cmd/goimports && \
+    upx /usr/bin/kubectl /usr/bin/dockerd /usr/bin/docker /usr/bin/containerd /usr/bin/helm /usr/bin/kubefedctl && \
```
mkolesnik: You can also shrink the following:
/go/bin/golangci-lint
/usr/bin/ctr
/linux-amd64/tiller
/usr/bin/runc

Also for some reason these seem to be duplicate:
40840	/linux-amd64/helm
40840	/usr/bin/helm

mkolesnik:
i think you can also safely run `rpm -e --nodeps golang-src`
should save some ~40MB
and if you're really ballsy, delete `selinux-policy-targeted` while at it for extra ~50MB ;)

skitt:
> i think you can also safely run `rpm -e --nodeps golang-src`
> should save some ~40MB
> and if you're really ballsy, delete `selinux-policy-targeted` while at it for extra ~50MB ;)

No, I checked every single installed package; those two are needed (along with all the others which remain). Some packages _shouldn’t_ be necessary but are...

package/Dockerfile.dapper-base
skitt: Thanks, I got rid of the dupe and compressed the other binaries.

mkolesnik:
> No, I checked every single installed package; those two are needed (along with all the others which remain). Some packages _shouldn’t_ be necessary but are...

Huh weird that the source code for go is needed, maybe for the compilation?
Also for selinux it seems to be disabled so hence I thought we can get rid of that one :)
`[root@f11c784ef407 submariner]# getenforce
Disabled`

skitt:
> > No, I checked every single installed package; those two are needed (along with all the others which remain). Some packages _shouldn’t_ be necessary but are...
> 
> Huh weird that the source code for go is needed, maybe for the compilation?

Yes, Go rebuilds everything from source all the time, dependencies included, so the library source code needs to be shipped with the compiler...

Issue #365, opened 2020-02-24
e2e tests to validate Globalnet Controller 


sridhargaddam:
Globalnet controller is a new feature that supports Clusters with Overlapping CIDRs. This is an issue reported to track the corresponding e2e tests.

PR #454, opened 2020-03-18
Globalnet Controller e2e dataplane tests

sridhargaddam:
This PR implements dataplane tests to validate Globalnet Controller use-cases.
The tests will be executed only when Clusters are deployed with Globalnet Controller and will be skipped on normal clusters.

Fixes Issue: https://github.com/submariner-io/submariner/issues/365

### DIFF
--- a/pkg/globalnet/controllers/ipam/constants.go
+++ b/pkg/globalnet/controllers/ipam/constants.go
@@ -6,10 +6,11 @@ type Operation string
 
 const (
 	handlerResync          = time.Hour * 24
-	submarinerIpamGlobalIp = "submariner.io/globalIp"
-	submarinerIngress      = "SUBMARINER-GN-INGRESS"
-	submarinerEgress       = "SUBMARINER-GN-EGRESS"
-	submarinerMark         = "SUBMARINER-GN-MARK"
+	SubmarinerIpamGlobalIp = "submariner.io/globalIp"
+
+	submarinerIngress = "SUBMARINER-GN-INGRESS"
+	submarinerEgress  = "SUBMARINER-GN-EGRESS"
+	submarinerMark    = "SUBMARINER-GN-MARK"
 
 	// Globalnet uses MARK target to mark traffic destined to remote clusters.
 	// Some of the CNIs also use iptable MARK targets in the pipeline. This should not
--- a/pkg/globalnet/controllers/ipam/constants.go
+++ b/pkg/globalnet/controllers/ipam/constants.go
@@ -6,11 +6,10 @@ type Operation string
 
 const (
 	handlerResync          = time.Hour * 24
-	SubmarinerIpamGlobalIp = "submariner.io/globalIp"
-
-	submarinerIngress = "SUBMARINER-GN-INGRESS"
-	submarinerEgress  = "SUBMARINER-GN-EGRESS"
-	submarinerMark    = "SUBMARINER-GN-MARK"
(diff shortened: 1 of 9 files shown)


pkg/globalnet/controllers/ipam/constants.go
```diff
@@ -6,10 +6,11 @@ type Operation string
 
 const (
 	handlerResync          = time.Hour * 24
-	submarinerIpamGlobalIp = "submariner.io/globalIp"
-	submarinerIngress      = "SUBMARINER-GN-INGRESS"
-	submarinerEgress       = "SUBMARINER-GN-EGRESS"
-	submarinerMark         = "SUBMARINER-GN-MARK"
+	SubmarinerIpamGlobalIp = "submariner.io/globalIp"
```
dfarrell07: Seems like keeping it lowerSnake would be more consistent with the parallel vars.

sridhargaddam: Instead of duplicating the constant in e2e tests, I wanted to re-use the same constant (any changes to this annotation in the future will be foolproof). So mainly for this purpose, I modified it to "S"

test/e2e/framework/services.go
```diff
@@ -47,3 +47,16 @@ func (f *Framework) CreateTCPService(cluster ClusterIndex, selectorName string,
 		return service, err
 	}, NoopCheckResult).(*v1.Service)
 }
+
+// AwaitServiceByAnnotation queries the service and looks for the presence of annotation.
+func (f *Framework) AwaitServiceByAnnotation(cluster ClusterIndex, annotation string, svcName string, namespace string) *v1.Service {
+	return AwaitUntil("get"+annotation+" annotation for service "+svcName, func() (interface{}, error) {
+		return f.ClusterClients[cluster].CoreV1().Services(namespace).Get(svcName, metav1.GetOptions{})
```
tpantelis: It seems we check for `IsNotFound` and not assume the Service exists yet to be complete, ie:

```
service, err := f.ClusterClients[cluster].CoreV1().Services(namespace).Get(svcName, metav1.GetOptions{})
    return nil, nil
}
return service, err

```

test/e2e/framework/services.go
```diff
@@ -47,3 +47,16 @@ func (f *Framework) CreateTCPService(cluster ClusterIndex, selectorName string,
...
+	}, func(result interface{}) (bool, string, error) {
```
tpantelis: and here ....

```
if result == nil {
    return false, "No Service found", nil
}
```

test/e2e/framework/services.go
sridhargaddam: Done, thanks for reviewing @tpantelis

test/e2e/framework/services.go
sridhargaddam: Done

pkg/globalnet/controllers/ipam/constants.go
mangelajo: It's better to duplicate, because in that case if somebody changes it and doesn't realize that it's going to break compatibility at least E2E tests will chime in and fail, and then you see that the change being done is changing the E2E tests which at least would be suspicious.

We are following the same principle of replicating the constants to tests.

sridhargaddam: Okay, make sense. Thanks.

PR #457, opened 2020-03-18
install WireGuard

roytman:
e2e submariner WireGuard cable tests require WireGuard modules on the testing host. 
This PR installs WireGuard.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,10 @@ jobs:
 
 install:
   - sudo apt-get install moreutils # make ts available
+  - sudo add-apt-repository -y ppa:wireguard/wireguard # add WireGuard support
+  - sudo apt-get update
+  - sudo apt-get install wireguard -y
+
 services:
   - docker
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,6 +18,10 @@ jobs:
 
 install:
   - sudo apt-get install moreutils # make ts available
+  - sudo add-apt-repository -y ppa:wireguard/wireguard # add WireGuard support
```
dfarrell07: If you do these installs in the base container images it can be done once at image-building time, vs at each runtime.

https://github.com/submariner-io/submariner/tree/master/package
https://github.com/submariner-io/submariner/blob/master/package/Dockerfile.dapper-base

roytman: It is a Linux kernel module, it is should be installed on the VM running the e2e tests.

dfarrell07: Oh that makes sense, okay thanks for clarifying.

Issue #460, opened 2020-03-20
Globalnet Controller is not annotating Services without selectors


sridhargaddam:
Setup: Overlapping Clusters with Submariner Globalnet deployed. 

Services without selectors i.e no backend pod, are not accessible across clusters if globalnet is enabled. It works when clusters don't have overlapping CIDRs and globalnet is not used.

mangelajo:
But what if the service without selectors is pointing to local cluster pods?

mangelajo:
Ahh sorry I see it was implemented, cool

PR #461, opened 2020-03-20
Globalnet: Annotate services without selectors

sridhargaddam:
Services without selectors i.e no backend pod, are not accessible across clusters if globalnet is enabled. It works when clusters don't have overlapping CIDRs and globalnet is not used.

This PR addresses this issue by supporting services without selectors along with fixing an issue in the iptables chain name used by kube-proxy.

Fixes issue: https://github.com/submariner-io/submariner/issues/460

### DIFF
--- a/pkg/globalnet/controllers/ipam/globalnet_ingress.go
+++ b/pkg/globalnet/controllers/ipam/globalnet_ingress.go
@@ -31,7 +31,7 @@ func (i *Controller) kubeProxyClusterIpServiceChainName(service *k8sv1.Service) 
 	// CNIs that use kube-proxy with iptables for loadbalancing create an iptables chain for each service
 	// and incoming traffic to the clusterIP Service is directed into the respective chain.
 	// Reference: https://bit.ly/2OPhlwk
-	serviceName := service.GetNamespace() + "/" + service.GetName() + ":"
+	serviceName := service.GetNamespace() + "/" + service.GetName() + ":" + service.Spec.Ports[0].Name
 	protocol := strings.ToLower(string(service.Spec.Ports[0].Protocol))
 	hash := sha256.Sum256([]byte(serviceName + protocol))
 	encoded := base32.StdEncoding.EncodeToString(hash[:])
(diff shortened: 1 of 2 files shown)


vthapar:
Can you change the commit msg to something like this:

"Services without selectors i.e no backend pod, are not accessible across clusters if globalnet is enabled. It works when clusters don't have overlapping CIDRs and globalnet is not used.

This PR addresses this issue by supporting services without selectors.

sridhargaddam:
> "Services without selectors i.e no backend pod, are not accessible across clusters if globalnet is enabled. It works when clusters don't have overlapping CIDRs and globalnet is not used.
> 
> This PR addresses this issue by supporting services without selectors.

Sure, will update it. Thanks.

pkg/globalnet/controllers/ipam/globalnet_ingress.go
```diff
@@ -31,7 +31,7 @@ func (i *Controller) kubeProxyClusterIpServiceChainName(service *k8sv1.Service)
 	// CNIs that use kube-proxy with iptables for loadbalancing create an iptables chain for each service
 	// and incoming traffic to the clusterIP Service is directed into the respective chain.
 	// Reference: https://bit.ly/2OPhlwk
-	serviceName := service.GetNamespace() + "/" + service.GetName() + ":"
+	serviceName := service.GetNamespace() + "/" + service.GetName() + ":" + service.Spec.Ports[0].Name
```
mangelajo: Ports can sometimes be "name" or sometimes be number, or both I think. 

Can we make a function so we try to get name, if that fails, just the port number converted to string?
would that make sense?

sridhargaddam: This piece of code uses internal implementation details of kube-proxy when it comes to chain names.
As you can see here [*], kube-proxy is using only the Port.Name (this was missed in the initial implementation and is now fixed). For all the normal services that we were creating, we did not specify the name, so it was empty and was working fine all these days. But while testing kubernetes services, I've seen that chain-name was not matching thats when I had a close look at kubeproxy code again and realized that we missed specifying the Port.Name.

[*] https://github.com/kubernetes/kubernetes/blob/b1766b707a0f61e4a2640359c4482dcb2c689df5/pkg/proxy/iptables/proxier.go#L124

mangelajo: Ah, thanks for the explanation, I see now.

PR #474, opened 2020-03-25
Switch to shipyard's E2E scripts

mkolesnik:
This PR includes work to switch the E2E to use the scripts and capabilities from shipyard.


Dockerfile.dapper
```diff
@@ -1,8 +1,9 @@
-FROM quay.io/submariner/dapper-base
+FROM quay.io/submariner/shipyard-dapper-base
 
 ENV DAPPER_ENV=REPO DAPPER_ENV=TAG \
     DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner DAPPER_DOCKER_SOCKET=true \
-    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} DAPPER_OUTPUT=output
+    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE}
+ENV DAPPER_OUTPUT=${DAPPER_SOURCE}/output
```
skitt:
```suggestion
    TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} DAPPER_OUTPUT=${DAPPER_SOURCE}/output
```

This avoids adding a layer.

mkolesnik: Its wrong, that's why I fixed it. Otherwise it doesn't expand $DAPPER_SOURCE since its not defined yet.

skitt: What about all the other variables that depend on `DAPPER_SOURCE` then?

mangelajo: @skitt I guess we are not really using them then? I don't know.

mangelajo: Although I remember DAPPER_OUTPUT always worked..

mangelajo: As discussed in slack, merging it, while I opened https://github.com/submariner-io/submariner/issues/477 to check this item.

PR #476, opened 2020-03-25
add option to setup cable driver on make e2e.

deanlorenz:
For example, 
```
make e2e cable_driver=wireguard
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ kubefed ?= false
 deploytool ?= operator
 globalnet ?= false
 build_debug ?= false
+cable_driver ?= default
 
 TARGETS := $(shell ls scripts | grep -v dapper-image)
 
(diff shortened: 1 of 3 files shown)


deanlorenz:
For changes to work on dev environment:

- `scripts/` gets copied into  `/opt` during dapper-image creation; however `make dapper-image` does not check all dependencies, so the `scripts/...` changes will not make into image

- As a workaround, changing `scripts/e2e` wrapper to call `kind-e2e/e2e.sh` (remove `/opt` prefix) will use the scripts in local dev directory

- subctl version must be new enough to support the `cable-driver` option. `latest` is good.

dfarrell07:
Thanks for the PRs! FYI, much of the related scripting is going to be extracted (for deduplication across projects via Shipyard) in https://github.com/submariner-io/submariner/pull/474.

mangelajo:
Yes, let's add the patch in shipyard at the right moment, sorry about that ':)

mkolesnik:
Closing as it has been added to shipyard https://github.com/submariner-io/shipyard/pull/102

PR #481, opened 2020-03-30
Use controller-runtime signals lib

tpantelis:
The signals code in submariner was apparently copied from
somewhere originally. The same code exists in k8s controller-runtime
so use that.


mkolesnik:
I wonder why it was copied in the first place

tpantelis:
> I wonder why it was copied in the first place

I suspect it wasn't in a reusable place like controller-runtime at that time. I also found a copy in the apiserver project which isn't intended for general use - maybe it was copied from there.

PR #483, opened 2020-04-01
Use the shared Makefile.inc

skitt:
This also merges e2e and vendor into Makefile.

This needs https://github.com/submariner-io/shipyard/pull/62


scripts/vendor
```diff
@@ -1,12 +0,0 @@
-#!/bin/bash
```
dfarrell07, line 1: CI seems to be missing this file

skitt: It won’t once the parent PR is merged in shipyard.

dfarrell07: Ah, okay.

mkolesnik: This script is called by scripts/entry hence the CI fails regardless of shipyard's state.

Now should or should it not be called by scripts/entry is a different question :)
My guess is since you now have ` $(TARGETS): vendor/modules.txt` then it should be fine as it's handled by make, so you can remove that call from the entry script

skitt: Ah yes, thanks. In fact I should add the `vendor` handling to submariner and share the `entry` script...

mkolesnik: I guess you mean add it to shipyard as an "sh" script in the shared folder..? :)

Makefile
```diff
@@ -1,43 +1,37 @@
...
-	DAPPER_ENV="OPERATOR_IMAGE" ./.dapper -m bind $(SCRIPTS_DIR)/deploy.sh --globalnet $(globalnet) --deploytool $(deploytool)
 
 e2e: deploy
-	./.dapper -m bind ./scripts/kind-e2e/e2e.sh --status $(status) --deploytool $(deploytool)
+	./scripts/kind-e2e/e2e.sh --status $(status) --deploytool $(deploytool)
 
-$(TARGETS): .dapper vendor/modules.txt
-	./.dapper -m bind $@ --build_debug $(build_debug)
+$(TARGETS): vendor/modules.txt
+	./scripts/$@ --status $(status) --k8s_version $(k8s_version) --deploytool $(deploytool) --globalnet $(globalnet) --build_debug $(build_debug)
```
mkolesnik, line 18:
```suggestion
	./scripts/$@ --build_debug $(build_debug)
```

No need for the flags, only build_debug

skitt: Done, thanks!

Issue #308, opened 2020-01-29
Expose endpoint connection status and information on every submariner engine/gateway


mangelajo:
May be using a healthz server https://github.com/submariner-io/coastguard/blob/master/pkg/healthz/healthz.go or similar

We could expose, per remote endpoint:
   * State (connecting/connected/error?...)
   * IPs used for the connection
   * Nat status
   * Latency
   * Bandwidth
   * Local & Remote IPs of the connection

PR #485, opened 2020-04-01
Define the status structures for a connection list

mangelajo:
Related-Issue: #308

### DIFF
--- a/pkg/cable/status.go
+++ b/pkg/cable/status.go
@@ -1,0 +1,21 @@
+package cable
+
+type ConnectionStatus string
+
+const (
+	Established     ConnectionStatus = "ESTABLISHED"
+	Connecting      ConnectionStatus = "CONNECTING"
+	ConnectionError ConnectionStatus = "ERROR"
+	Unreachable     ConnectionStatus = "UNREACHABLE"
+	AuthError       ConnectionStatus = "AUTH ERROR"
+)
+
+type Connection struct {
+	CableName     string           `json:"cableName"`
+	ClusterID     string           `json:"clusterID"`
+	Status        ConnectionStatus `json:"status"`
+	RemoteIP      string           `json:"remoteIP"`
+	LocalIP       string           `json:"localIP"`
+	CableDriver   string           `json:"cableDriver"`
+	RemoteSubnets []string         `json:"remoteSubnets"`
+}
--- a/pkg/cable/status.go
+++ b/pkg/cable/status.go
@@ -10,8 +10,8 @@ const (
 	ConnectionError ConnectionStatus = "ERROR"
 )
 
-// Connection structure represents the status of an existing cable connection,
-// it includes details the endpoint details, and it's meant to be exposed
+// Connection structure represents the state of an existing cable connection
+// including the Endpoint details. It is meant to be exposed
 // through an HTTP API on the submariner-engine.
 type Connection struct {
 	Status        ConnectionStatus `json:"status"`
--- a/pkg/cable/status.go
+++ b/pkg/cable/status.go
@@ -1,21 +1,20 @@
 package cable
 
+import v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+
 type ConnectionStatus string
 
 const (
-	Established     ConnectionStatus = "ESTABLISHED"
+	Connected       ConnectionStatus = "CONNECTED"
 	Connecting      ConnectionStatus = "CONNECTING"
 	ConnectionError ConnectionStatus = "ERROR"
-	Unreachable     ConnectionStatus = "UNREACHABLE"
-	AuthError       ConnectionStatus = "AUTH ERROR"
 )
 
+// Connection structure represents the status of an existing cable connection,
+// it includes details the endpoint details, and it's meant to be exposed
+// through an HTTP API on the submariner-engine.
 type Connection struct {
-	CableName     string           `json:"cableName"`
-	ClusterID     string           `json:"clusterID"`
 	Status        ConnectionStatus `json:"status"`
-	RemoteIP      string           `json:"remoteIP"`
-	LocalIP       string           `json:"localIP"`
-	CableDriver   string           `json:"cableDriver"`
-	RemoteSubnets []string         `json:"remoteSubnets"`
+	StatusMessage string           `json:"statusMessage"`
+	Endpoint      v1.EndpointSpec  `json:"endpoint"`
 }
--- a/pkg/cableengine/cableengine.go
+++ b/pkg/cableengine/cableengine.go
@@ -30,6 +30,8 @@ type Engine interface {
 	// RemoveCable disconnects the Engine from the given remote endpoint. Upon completion.
 	// remote Pods and Service may not be accessible any more.
 	RemoveCable(remote types.SubmarinerEndpoint) error
+	// ListCableConnections returns a list of cable connection, and the related status
+	ListCableConnections() (*[]cable.Connection, error)
 }
 
 type engine struct {
@@ -116,4 +118,8 @@ func (i *engine) RemoveCable(endpoint types.SubmarinerEndpoint) error {
 
 	klog.Infof("Successfully removed Endpoint cable %q", endpoint.Spec.CableName)
 	return nil
+}
+
+func (i *engine) ListCableConnections() (*[]cable.Connection, error) {
+	return nil, fmt.Errorf("Not implemented yet")
 }


pkg/cableengine/cableengine.go
```diff
@@ -117,3 +119,7 @@ func (i *engine) RemoveCable(endpoint types.SubmarinerEndpoint) error {
 	klog.Infof("Successfully removed Endpoint cable %q", endpoint.Spec.CableName)
 	return nil
 }
+
+func (i *engine) ListCableConnections() (*[]cable.Connection, error) {
+	return nil, fmt.Errorf("Not implemented yet")
```
tpantelis, line 124: Will this also require a new method on the `cable.Driver` interface?I see there's a `GetActiveConnections` method but that takes a clusterID.

mangelajo: Of course, I was planning to make that happen during the real implementation, This was just defining the high level API and structures (which will also be used for reporting)

I was considering the driver part as part of the implementation that we can discuss later.

I can add it here too, no harm :-)

mangelajo:
@roytman @deanlorenz ^

pkg/cable/status.go
```diff
@@ -0,0 +1,21 @@
...
+
+type Connection struct {
+	CableName     string           `json:"cableName"`
+	ClusterID     string           `json:"clusterID"`
+	Status        ConnectionStatus `json:"status"`
+	RemoteIP      string           `json:"remoteIP"`
+	LocalIP       string           `json:"localIP"`
+	CableDriver   string           `json:"cableDriver"`
+	RemoteSubnets []string         `json:"remoteSubnets"`
+}
```
deanlorenz, line 21: Beyond status, do we expect these attributes to be different from the endpoint spec?

pkg/cable/status.go
```diff
@@ -0,0 +1,21 @@
...
+
+type ConnectionStatus string
+
+const (
+	Established     ConnectionStatus = "ESTABLISHED"
+	Connecting      ConnectionStatus = "CONNECTING"
+	ConnectionError ConnectionStatus = "ERROR"
+	Unreachable     ConnectionStatus = "UNREACHABLE"
+	AuthError       ConnectionStatus = "AUTH ERROR"
+)
```
deanlorenz, line 11: Are these supposed to be computed during connect (i.e., by errors reported) or supposed to reflect some health check?
I believe wiregaurd does not report any authentication errors -- packets with bad keys are silently dropped.
Wireguard has some liveness metrics -- last handsake, TxRx bytes -- however these are updated only if someone actually uses the connection. , Perhaps we need a health check probe, e.g., a ping between the GWs

mangelajo: Can you investigate the AUTH ERROR part? I hope we can find a minimal common denominator. But I think we must find a way to at least identify "AUTH ERROR", "CONNECTING", "CONNECTED" status. As those should be very basic.

Later on I think we can also add ping metrics to the connection, that'd be very helpful to calculate transit latencies, but we don't need to define all from the start.

pkg/cable/status.go
mangelajo: No In fact, we could add the endpoint spec struct as an attribute itself and probably make things simpler.

The idea is that we want to expose this structure in an API, and then the operator can look at the API of it's own active gateway engine and report the list of connections as a status field

mangelajo:
@deanlorenz thanks a lot for reviewing, I'll have an eye on the wireguard side to help figure out the MCD.

pkg/cable/status.go
deanlorenz: 1. Do we need to do anything if the data does not match the endpoint spec?
2. Most fields, even if stored in the driver, are just a copy of the endpoint -- so mismatch is just an error. Only IPs, and subnets are part of the peer information that actually comes from the wg device. 
3. Going forward we should probably encode the publickey in cable name

pkg/cable/status.go
deanlorenz: I ran a few tests and checked again. 

- This is all done in the kernel. The only way to get the status is to turn on dynamic debug in debugfs and fetch the messages with -- I don't think we should do that. debug provides a lot of info, but is too verbose.
```
drivers/usb/host/xhci-mem.c:1136 [xhci_hcd]xhci_setup_addressable_virt_dev =_ "FIXME xHCI doesn't support wireless speeds\012"
drivers/net/ethernet/intel/e1000/e1000_hw.c:4203 [e1000]e1000_write_eeprom_microwire =_ "EEPROM Write did not complete\012"
/var/lib/dkms/wireguard/0.0.20200318/build/noise.c:825 [wireguard]wg_noise_handshake_begin_session =_ "%s: Keypair %llu created for peer %llu\012"
/var/lib/dkms/wireguard/0.0.20200318/build/noise.c:127 [wireguard]keypair_free_kref =_ "%s: Keypair %llu destroyed for peer %llu\012"
/var/lib/dkms/wireguard/0.0.20200318/build/device.c:377 [wireguard]wg_newlink =_ "%s: Interface created\012"
/var/lib/dkms/wireguard/0.0.20200318/build/device.c:257 [wireguard]wg_destruct =_ "%s: Interface deleted\012"
/var/lib/dkms/wireguard/0.0.20200318/build/device.c:155 [wireguard]wg_xmit =_ "%s: No valid endpoint has been configured or discovered for peer %llu\012"
/var/lib/dkms/wireguard/0.0.20200318/build/device.c:147 [wireguard]wg_xmit =_ "%s: No peer has allowed IPs matching %pI6\012"
/var/lib/dkms/wireguard/0.0.20200318/build/device.c:144 [wireguard]wg_xmit =_ "%s: No peer has allowed IPs matching %pI4\012"
/var/lib/dkms/wireguard/0.0.20200318/build/device.c:135 [wireguard]wg_xmit =_ "%s: Invalid IP packet\012"
/var/lib/dkms/wireguard/0.0.20200318/build/peer.c:215 [wireguard]kref_release =_ "%s: Peer %llu (%pISpfsc) destroyed\012"
/var/lib/dkms/wireguard/0.0.20200318/build/peer.c:68 [wireguard]wg_peer_create =_ "%s: Peer %llu created\012"
/var/lib/dkms/wireguard/0.0.20200318/build/timers.c:129 [wireguard]wg_queued_expired_zero_key_material =_ "%s: Zeroing out all keys for peer %llu (%pISpfsc), since we haven't received a new one in %d seconds\012"
/var/lib/dkms/wireguard/0.0.20200318/build/timers.c:97 [wireguard]wg_expired_new_handshake =_ "%s: Retrying handshake with peer %llu (%pISpfsc) because we stopped hearing back after %d seconds\012"
/var/lib/dkms/wireguard/0.0.20200318/build/timers.c:68 [wireguard]wg_expired_retransmit_handshake =_ "%s: Handshake for peer %llu (%pISpfsc) did not complete after %d seconds, retrying (try %d)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/timers.c:49 [wireguard]wg_expired_retransmit_handshake =_ "%s: Handshake for peer %llu (%pISpfsc) did not complete after %d attempts, giving up\012"
/var/lib/dkms/wireguard/0.0.20200318/build/send.c:237 [wireguard]wg_packet_send_keepalive =_ "%s: Sending keepalive packet to peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/send.c:118 [wireguard]wg_packet_send_handshake_cookie =_ "%s: Sending cookie response for denied handshake message for %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/send.c:93 [wireguard]wg_packet_send_handshake_response =_ "%s: Sending handshake response to peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/send.c:33 [wireguard]wg_packet_send_handshake_initiation =_ "%s: Sending handshake initiation to peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:583 [wireguard]wg_packet_receive =_ "%s: Dropping handshake packet from %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:487 [wireguard]wg_packet_rx_poll =_ "%s: Packet has invalid nonce %llu (max %llu)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:449 [wireguard]wg_packet_consume_data_done =_ "%s: Packet has incorrect size from peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:443 [wireguard]wg_packet_consume_data_done =_ "%s: Packet is neither ipv4 nor ipv6 from peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:437 [wireguard]wg_packet_consume_data_done =_ "%s: Packet has unallowed src IP (%pISc) from peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:428 [wireguard]wg_packet_consume_data_done =_ "%s: Failed to give packet to userspace from peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:370 [wireguard]wg_packet_consume_data_done =_ "%s: Receiving keepalive packet from peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:183 [wireguard]wg_receive_handshake_packet =_ "%s: Receiving handshake response from peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:177 [wireguard]wg_receive_handshake_packet =_ "%s: Invalid handshake response from %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:161 [wireguard]wg_receive_handshake_packet =_ "%s: Receiving handshake initiation from peer %llu (%pISpfsc)\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:155 [wireguard]wg_receive_handshake_packet =_ "%s: Invalid handshake initiation from %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:138 [wireguard]wg_receive_handshake_packet =_ "%s: Invalid MAC of handshake, dropping packet from %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/receive.c:114 [wireguard]wg_receive_handshake_packet =_ "%s: Receiving cookie response from %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/socket.c:156 [wireguard]send6 =_ "%s: Avoiding routing loop to %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/socket.c:150 [wireguard]send6 =_ "%s: No route to %pISpfsc, error %d\012"
/var/lib/dkms/wireguard/0.0.20200318/build/socket.c:83 [wireguard]send4 =_ "%s: Avoiding routing loop to %pISpfsc\012"
/var/lib/dkms/wireguard/0.0.20200318/build/socket.c:77 [wireguard]send4 =_ "%s: No route to %pISpfsc, error %d\012"
/var/lib/dkms/wireguard/0.0.20200318/build/cookie.c:231 [wireguard]wg_cookie_message_consume =_ "%s: Could not decrypt invalid cookie response\012"
```

- Wireguard driver cannot cause authentication errors, but address misconfigurations can always happen -- that is, pubkey is the identity, so if there is a wrong privkey it is always someone malicious. Connection errors can occur -- blocked ports, bad remote IP, wrong pubkey, etc. -- I will not call these AUTH ERROR -- we should perhaps indicate when the peer info is different from the peer endpoint spec (i.e., do a deep compare when asked to get connections).  

- I ran a test with forced failures (like using IP out of range or using a wrong key) just shows as stale last handshake -- looks exactly like an unused connection.

- I still think the last handshake time is the best bet. But it needs a ping.

pkg/cable/status.go
mangelajo: I don't think the pk would fit in the cable name, but we could include it in the backend config map.

mangelajo:
@deanlorenz @tpantelis @skitt can you look at the new version of the patch and vote please?

deanlorenz:
@mangelajo looks good to me. 
Will refresh status on ListConnections.
For WG will check last handshake is not stale + compare current to desired endpoint

pkg/cable/status.go
```diff
@@ -0,0 +1,20 @@
...
+
+type ConnectionStatus string
+
+const (
+	Connected       ConnectionStatus = "CONNECTED"
+	Connecting      ConnectionStatus = "CONNECTING"
+	ConnectionError ConnectionStatus = "ERROR"
+)
+
+// Connection structure represents the status of an existing cable connection,
```
tpantelis, line 13:
```suggestion
// Connection structure represents the state of an existing cable connection
```

pkg/cable/status.go
```diff
@@ -0,0 +1,20 @@
...
+// it includes details the endpoint details, and it's meant to be exposed
```
tpantelis, line 14:
```suggestion
// including the Endpoint details. It is meant to be exposed
```

PR #488, opened 2020-04-02
Allow running just the e2e tests

mangelajo:
This is useful when working on the e2e tests, you can deploy once
and run e2e-tests many times.

```
make e2e status=keep
... modify your e2e tests here ...
make e2e-tests
```


mangelajo:
@mkolesnik I have proposed an extended version of this, because -o deploy doesn't work in my system.

Also the new proposal brings new features and it's cleaner.

https://github.com/submariner-io/submariner/pull/510

PR #490, opened 2020-04-07
Use latest image as cache to skip rebuilding layers

mkolesnik:
OS and other base layers don't change that often, use them from the
latest image instead of rebuilding them each time.


mkolesnik:
Seems to be working well, see #491 for testing it picks up the changes

mkolesnik:
Closing in favor of #498

PR #487, opened 2020-04-02
Remove status flag from E2E

mkolesnik:
This flag made sense when everything was inside the e2e.sh script,
however since we migrated to shipyard which has distinct stages for each
operation, it doesn't make any sense to keep supporting this flag.

PR #492, opened 2020-04-07
Don't make build package explicitly in travis

mkolesnik:
Since the makefile already specifies the dependency tree, `make e2e`
will always perform these targets anyway.
Specifying them explicitly in travis, due to all targets being run
inside dapper, causes it to extranously perform these steps before the
e2e, thus it actually builds and packages everything twice.
Since the whole process takes a few minutes, removing these unnecessary
steps will shorten the CI run time quite significantly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ git:
 jobs:
   include:
   - env: CMD="make test validate"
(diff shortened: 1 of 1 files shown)


mangelajo:
Conflict now, I think you fixed it on the previous patch already

mkolesnik:
This was already fixed by #487

PR #496, opened 2020-04-08
Avoid package conflict (script/directory)

skitt:
“package” exists both as a directory and a script (and Make target),
which causes problems when it’s not defined as a real Make target.
This renames the script to “images” to avoid problems.


skitt:
> Might need to update the docs here: https://submariner-io.github.io/en/contributing/building_testing/#building-engine-routeagent-and-globalnet-container-images

Done, thanks, https://github.com/submariner-io/submariner-website/pull/76

PR #501, opened 2020-04-10
Move connection types to the submariner.io API

skitt:
We want to expose the connection status in the operator status. The
initial plan was to make the status available over HTTP, and query
that from the operator; however, the operator only updates its status
in the reconciliation loop, and apart from that the status is static.

This means that operator status updates require watching Kubernetes
objects which reflect the state encapsulated in the operator status.
To enable this, we can expose the status types in Submariner alongside
the existing Cluster and Endpoint types; to avoid package import
loops, we move everything there, and adjust the existing cable engine
and drivers accordingly.


mangelajo:
https://travis-ci.com/github/submariner-io/submariner/jobs/318192563#L510

why did this fail on CI?, it passes locally for me.

mangelajo:
Ah, the return order is not deterministic, you fixed that for one of the cases, and not the other... investigating.

mangelajo:
Added another commit on top to handle the unit test failures which happen sometimes, and cleaning up tests.

mangelajo:
@skitt I started a follow up here to report the status: https://github.com/submariner-io/submariner/pull/502

still in WIP status, will finish it tomorrow along with the CRD definitions to test it

PR #502, opened 2020-04-13
Implement Gateway status updates

mangelajo:


main.go
```diff
@@ -90,39 +92,40 @@ func main() {
...
-		}
+	if err != nil {
+		klog.Fatalf("Error creating local endpoint object from %#v: %v", submSpec, err)
+	}
 
-		cableEngine, err := cableengine.NewEngine(localSubnets, localCluster, localEndpoint)
-		if err != nil {
-			klog.Fatalf("Fatal error occurred creating engine: %v", err)
-		}
+	cableEngine, err := cableengine.NewEngine(localSubnets, localCluster, localEndpoint, submarinerClient,
```
mangelajo, line 122: We move this creation ahead in time, so ... the loop reconciling the gateway status will also
work for the PASSIVE gateway.

The driver initialization does not happen until "StartEngine" though.

mangelajo:
ok, removed the workaround after https://github.com/submariner-io/submariner/pull/506 was merged

mangelajo:
E2E tests being handled here: https://github.com/submariner-io/submariner/pull/508

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil && !errors.IsNotFound(err) {
+		klog.Errorf("trying to read existing gateway from k8s api: %s", err)
+		return
+	}
+
+	gwClient := i.clientset.SubmarinerV1().Gateways(i.namespace)
+
+	if err != nil && errors.IsNotFound(err) {
+		klog.V(log.TRACE).Infof("Gateway object needs creation: %+v", gatewayObj)
+		_, err = gwClient.Create(gatewayObj)
```
sridhargaddam, line 47: Does this patch depend on any other PR?

I tried this PR locally and see that following error continuously repeated.
 I0417 09:12:46.798166       1 syncgatewaystatus.go:46] Gateway object needs creation: &{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:cluster2-worker GenerateName: Namespace: SelfLink: U │
│ ID: ResourceVersion: Generation:0 CreationTimestamp:0001-01-01 00:00:00 +0000 UTC DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[ha-status:active] Annotations:map[up │
│ ate-timestamp:1587114766[] OwnerReferences:[] Initializers:nil Finalizers:[] ClusterName:} Status:{Version:dev HAStatus:ACTIVE LocalEndpoint:{ClusterID:cluster2 CableName:submariner-cabl │
│ e-cluster2-172-17-0-8 Hostname:cluster2-worker Subnets:[100.92.0.0/16 10.242.0.0/16] PrivateIP:172.17.0.8 PublicIP: NATEnabled:false Backend:strongswan BackendConfig:map[]} Connections:[ │
│ {Status:CONNECTED StatusMessage:Connected to 172.17.0.5:4500 - encryption alg=AES_GCM_16, keysize=128 rekey-time=12786 Endpoint:{ClusterID:cluster3 CableName:submariner-cable-cluster3-17 │
│ 2-17-0-5 Hostname:cluster3-worker Subnets:[100.93.0.0/16 10.243.0.0/16] PrivateIP:172.17.0.5 PublicIP: NATEnabled:false Backend:strongswan BackendConfig:map[]}}]}}                        │
│ E0417 09:12:46.798660       1 syncgatewaystatus.go:49] creating Gateway object: the server could not find the requested resource (post gateways.submariner.io)                             │
│ I0417 09:12:51.799029       1 syncgatewaystatus.go:27] Running syncGatewayStatus()                                                                                                         │
│ I0417 09:12:51.802367       1 syncgatewaystatus.go:159] generateGatewayObject: {TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:cluster2-worker GenerateName: Namespace: SelfLink: UID: Reso │
│ urceVersion: Generation:0 CreationTimestamp:0001-01-01 00:00:00 +0000 UTC DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[ha-status:active] Annotations:map[update-tim │
│ stamp:1587114771[] OwnerReferences:[] Initializers:nil Finalizers:[] ClusterName:} Status:{Version:dev HAStatus:ACTIVE LocalEndpoint:{ClusterID:cluster2 CableName:submariner-cable-cluste │
│ r2-172-17-0-8 Hostname:cluster2-worker Subnets:[100.92.0.0/16 10.242.0.0/16] PrivateIP:172.17.0.8 PublicIP: NATEnabled:false Backend:strongswan BackendConfig:map[]} Connections:[{Status: │
│ CONNECTED StatusMessage:Connected to 172.17.0.5:4500 - encryption alg=AES_GCM_16, keysize=128 rekey-time=12781 Endpoint:{ClusterID:cluster3 CableName:submariner-cable-cluster3-172-17-0-5 │
│  Hostname:cluster3-worker Subnets:[100.93.0.0/16 10.243.0.0/16] PrivateIP:172.17.0.5 PublicIP: NATEnabled:false Backend:strongswan BackendConfig:map[]}}]}}                                │
│ I0417 09:12:51.804919       1 syncgatewaystatus.go:123] getLastSyncedGateway: &{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name: GenerateName: Namespace: SelfLink: UID: ResourceVersion: Ge │
│ neration:0 CreationTimestamp:0001-01-01 00:00:00 +0000 UTC DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[] Annotations:map[] OwnerReferences:[] Initializers:nil Fin │
│ alizers:[] ClusterName:} Status:{Version: HAStatus: LocalEndpoint:{ClusterID: CableName: Hostname: Subnets:[] PrivateIP: PublicIP: NATEnabled:false Backend: BackendConfig:map[]} Connecti │
│ ons:[]}}

mangelajo: You need to remove your local shipyard image so it will pull the latest one with the latest subctl which includes it.

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: gateways.submariner.io
spec:
  conversion:
    strategy: None
  group: submariner.io
  names:
    kind: Gateway
    listKind: GatewayList
    plural: gateways
    singular: gateway
  scope: Namespaced
  version: v1
  versions:
  - name: v1
    served: true
    storage: true

```

mangelajo: You can apply that CRD manually otherwise :-)

sridhargaddam: Thanks for the clarification. Yes, it works fine after updating the shipyard image.

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+			existingGw.Labels = gatewayObj.Labels
+			existingGw.Annotations = gatewayObj.Annotations
+
+			gw, err := gwClient.Update(existingGw)
+			if err != nil {
+				klog.Errorf("updating Gateway object: %s", err)
+				return
+			} else {
+				klog.V(log.TRACE).Infof("Gateway updated correctly: %+v", gw)
+			}
```
sridhargaddam, line 65: Shall we use retry.RetryOnConflict here?

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+	return nil
+}
+
+func isGatewayStale(gateway v1.Gateway) (bool, error) {
+
+	timestamp, ok := gateway.ObjectMeta.Annotations[updateTimestampAnnotation]
+	if !ok {
+		return true, fmt.Errorf("update-timestamp annotation not found")
```
sridhargaddam, line 108: nit: can you use updateTimestampAnnotation instead of hardcoding the string "update-timestamp" (including below)

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+		return
+	}
+
+	gwClient := i.clientset.SubmarinerV1().Gateways(i.namespace)
+
+	if err != nil && errors.IsNotFound(err) {
+		klog.V(log.TRACE).Infof("Gateway object needs creation: %+v", gatewayObj)
+		_, err = gwClient.Create(gatewayObj)
+		if err != nil {
+			klog.Errorf("creating Gateway object: %s", err)
```
sridhargaddam, line 49: Can you please take a look at the Errorf log messages in the patch and include "error " as part of the message.
i.e., "error creating Gateway object: ...."
I know we are printing the err as part of it. While debugging some problem and looking at a long log message, it would be easy if the error messages starts with "error".

sridhargaddam:
BTW I noticed that logs in gateway node were displayed at TRACE level by default. Pushed the following PR to set it to DEBUG - https://github.com/submariner-io/submariner/pull/509

CC: @tpantelis @skitt @mangelajo

pkg/cableengine/syncgatewaystatus.go
mangelajo: I thought that's an unwanted practice because the log will be already indicative of error, may be I got it wrong?

pkg/cableengine/syncgatewaystatus.go
mangelajo: It's supposed to be only one updater per gateway, a conflict would be indicative of something very very wrong in the system (2 gateways running on the same host?). IMO I think it's better not to retry.

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+			// In this case we don't want to stop the cleanup loop and just log it
+			klog.Errorf("error processing stale gateway: %+v , %s", gw, err)
+		}
+		if stale {
+			err := gwClient.Delete(gw.Name, &metav1.DeleteOptions{})
+			if err != nil {
+				// In this case we don't want to stop the cleanup loop and just log it
+				klog.Errorf("error deleting stale gateway: %+v, %s", gw, err)
+			}
+			klog.Warningf("deleted stale gateway: %s, didn't report for %d seconds",
```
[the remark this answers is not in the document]
mangelajo: It's something that should not normally happen, that's why I chose Warning. IMO it should be higher than info.

pkg/cableengine/syncgatewaystatus.go
mangelajo: makes sense, I'll update it.

pkg/cableengine/syncgatewaystatus.go
mangelajo: E0417 13:29:03.745481       1 syncgatewaystatus.go:49] creating Gateway object: the server could not find the requested resource (post gateways.submariner.io)


Oh you're right, it's prefixed by "E" , but that's not very helpful for human readers

pkg/cableengine/cableengine.go
```diff
@@ -31,37 +33,63 @@ type Engine interface {
...
 }
 
 // NewEngine creates a new Engine for the local cluster
-func NewEngine(localSubnets []string, localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint) (Engine, error) {
-	driver, err := cable.NewDriver(localSubnets, localEndpoint)
-	if err != nil {
-		return nil, err
-	}
+func NewEngine(localSubnets []string, localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint,
+	clientset *submarinerClientset.Clientset, submNamespace string, version string, stopCh <-chan struct{}) (Engine, error) {
```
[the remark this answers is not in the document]
mangelajo: :+1:

pkg/cableengine/cableengine.go
```diff
@@ -31,37 +33,63 @@ type Engine interface {
...
 		localCluster:  localCluster,
 		localEndpoint: localEndpoint,
 		localSubnets:  localSubnets,
-		driver:        driver,
-	}, nil
+		driver:        nil,
+		version:       version,
+	}
+
+	go wait.Until(i.syncGatewayStatus, GatewayUpdateIntervalSeconds*time.Second, stopCh)
```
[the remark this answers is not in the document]
mangelajo: List connections was designed to be consumed by an HTTP exposed API.

But what you're proposing makes sense, more separation of concerns.

pkg/cableengine/cableengine.go
```diff
@@ -31,37 +33,63 @@ type Engine interface {
...
-	return i.driver.Init()
+
+	return nil
+
+}
+
+func (i *engine) startDriver() error {
+	var err error
+
+	if i.driver, err = cable.NewDriver(i.localSubnets, i.localEndpoint); err != nil {
```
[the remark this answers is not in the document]
mangelajo: right. Although I will make the changes you suggested around moving the reporting to a different package object of it's own. I will still need to create the CableEngine before starting it, so the service can have a reference to the pre-created cable engine.

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+				gw.Name, GatewayStaleTimeoutSeconds)
+		}
+	}
+	return nil
+}
+
+func isGatewayStale(gateway v1.Gateway) (bool, error) {
+
+	timestamp, ok := gateway.ObjectMeta.Annotations[updateTimestampAnnotation]
```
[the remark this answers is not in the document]
mangelajo: :+1:

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+		return true, fmt.Errorf("%q annotation not found", updateTimestampAnnotation)
+	}
+
+	timestampInt, err := strconv.ParseInt(timestamp, 10, 64)
+	if err != nil {
+		return true, fmt.Errorf("error parsing update-timestamp: %s", err)
+	}
+	now := time.Now().UTC().Unix()
+
+	return now >= (timestampInt + GatewayStaleTimeoutSeconds), nil
```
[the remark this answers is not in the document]
mangelajo: No, it's a pre-requisite of a running k8s cluster to be NTP time synced, and we use UTC which does not depend of DST time changes.

It's a foundation of any clustered service, and a common methodology to use timestamps.

see https://docs.oracle.com/en/operating-systems/oracle-linux/kubernetes/section_q4v_phr_p2b_ha.html

This kind of stale consideration was also used in openstack-neutron for a long time, still being used.

I'd prefer to avoid the extra complexity while we can.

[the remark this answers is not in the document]
mangelajo: Also see ETCD: https://github.com/etcd-io/etcd/issues/9768   also depends on this fact.

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+
+	gateway := v1.Gateway{
+		Status: v1.GatewayStatus{
+			Version:       i.version,
+			LocalEndpoint: i.localEndpoint.Spec,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        hostName,
```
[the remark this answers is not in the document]
mangelajo: :+1:  makes sense

pkg/cableengine/syncgatewaystatus.go
```diff
@@ -0,0 +1,161 @@
...
+	if i.driver == nil {
+		gateway.Status.HAStatus = v1.HAStatusPassive
+		gateway.SetLabels(map[string]string{"ha-status": "passive"})
+	} else {
+		gateway.Status.HAStatus = v1.HAStatusActive
+		gateway.SetLabels(map[string]string{"ha-status": "active"})
+		connections, err := i.driver.GetConnections()
+		if err != nil {
+			klog.Errorf("getting driver connections: %s", err)
+			return nil, err
```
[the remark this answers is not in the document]
mangelajo: :+1

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,64 @@ func queryAndUpdateGlobalnetStatus() {
...
+			if gw.Status.HAStatus != status {
+				return false, "", fmt.Errorf("Gateway %q showed up but has wrong status %q, expected %q",
+					gw.Name, gw.Status.HAStatus, status)
+			}
+			return true, "", nil
+		})
+	return gw.(*submarinerv1.Gateway)
+}
+
+func (f *Framework) AwaitForGatewayGone(cluster framework.ClusterIndex, name string) {
```
tpantelis, line 95:
```suggestion
func (f *Framework) AwaitForGatewayRemoved(cluster framework.ClusterIndex, name string) {
```

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,64 @@ func queryAndUpdateGlobalnetStatus() {
...
+	gwClient := SubmarinerClients[cluster].SubmarinerV1().Gateways(framework.TestContext.SubmarinerNamespace)
+	framework.AwaitUntil("await gateway gone",
```
tpantelis, line 97:
```suggestion
	framework.AwaitUntil(fmt.Sprintf("await Gateway on %q removed", name),
```

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,64 @@ func queryAndUpdateGlobalnetStatus() {
...
+		func() (interface{}, error) {
+			if _, err := gwClient.Get(name, metav1.GetOptions{}); apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, nil
```
tpantelis, line 102:
```suggestion
			return false, err
```

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,64 @@ func queryAndUpdateGlobalnetStatus() {
...
+		},
+		func(result interface{}) (bool, string, error) {
+			gone := result.(bool)
+			return gone, "", nil
```
tpantelis, line 106:
```suggestion
			return result.(bool), "", nil
```

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,64 @@ func queryAndUpdateGlobalnetStatus() {
...
+}
+
+// gatewayCleanup ensures that only the active gateway node is flagged as gateway node
+//                which could not be after a failed test which left the system on an
+//                unexpected state
+func (f *Framework) gatewayCleanup() {
+	gatewayClient := SubmarinerClients[framework.ClusterA].SubmarinerV1().Gateways(
+		framework.TestContext.SubmarinerNamespace)
+	gwList, err := gatewayClient.List(metav1.ListOptions{LabelSelector: submarinerv1.HAStatusPassiveSelector})
+	Expect(err).NotTo(HaveOccurred())
```
tpantelis, line 118: By querying for `Gateway` with a certain status label for cleanup, it seems we're depending on the code we're trying to test to be correct.

Is such cleanup really necessary?  I realize one test case failure  could could result other test failures but I think that's OK - such tests have initial sanity checks which should make it clear. Also if a bug causes one test failure it may cause others anyway.

test/e2e/gateway/gateway_reporting.go
```diff
@@ -0,0 +1,81 @@
...
+	By(fmt.Sprintf("Sanity check - ensuring there's exactly one gateway node on %q", clusterAName))
+	gatewayNodes := f.FindNodesByGatewayLabel(framework.ClusterA, true)
+	Expect(gatewayNodes).To(HaveLen(1), fmt.Sprintf("Expected only one gateway node on %q", clusterAName))
+
+	By(fmt.Sprintf("Ensuring that the gateway reports as active %q", clusterAName))
+	gatewayClient := subFramework.SubmarinerClients[framework.ClusterA].SubmarinerV1().Gateways(
+		framework.TestContext.SubmarinerNamespace)
+	gwList, err := gatewayClient.List(v1meta.ListOptions{LabelSelector: subv1.HAStatusActiveSelector})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(gwList.Items).To(HaveLen(1))
```
tpantelis, line 44: Can't this call `f.AwaitForGatewayWithStatus`?

test/e2e/framework/framework.go
```diff
@@ -1,12 +1,16 @@
...
 	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
-	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
-	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
```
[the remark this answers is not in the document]
mangelajo: It's project local imports, I'm always setting those last in the list.

test/e2e/framework/framework.go
```diff
@@ -22,7 +26,9 @@ func init() {
 
 // NewFramework creates a test framework.
 func NewFramework(baseName string) *Framework {
-	return &Framework{Framework: framework.NewFramework(baseName)}
+	f := &Framework{Framework: framework.NewFramework(baseName)}
+	framework.AddCleanupAction(f.gatewayCleanup)
```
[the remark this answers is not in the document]
mangelajo: :+1  it should

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,64 @@ func queryAndUpdateGlobalnetStatus() {
 		return true, "", nil
 	})
 }
+
+func (f *Framework) AwaitForGatewayWithStatus(cluster framework.ClusterIndex,
```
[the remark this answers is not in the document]
mangelajo: :+1:

test/e2e/framework/framework.go
mangelajo: I wrote this out of necessity, while developing the E2E tests were leaving my system on inconsistent state.

In the end many e2e tests depend on the underlaying code behaving on a certain way. It's true that if the reporting does not work, the cleanup won't work. I prefer to risk the cleanup not working ok if we break that, but having a cleanup that in most cases will prevent that the behavior of some failed tests will cascade into others.

test/e2e/framework/framework.go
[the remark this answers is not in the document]
skitt: I quite like having the Submariner imports separate from the rest...

test/e2e/gateway/gateway_reporting.go
mangelajo: I also wanted to test that the label selection worked properly, I can separate in an additional test with that specific intent.

pkg/cableengine/syncgatewaystatus.go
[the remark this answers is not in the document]
tpantelis: makes sense.

test/e2e/framework/framework.go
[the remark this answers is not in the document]
tpantelis: That's fine but we should probably standardize on something as I've been keeping them in alphabetical order 😄

test/e2e/gateway/gateway_reporting.go
tpantelis: I had a previous comment in the production code asking why you were duplicating the HA status in a label.  I see now why you did it.  But we don't really need the added complexity of a duplicated label just for test code - we can just filter by the `HAStatus` field.

pkg/cableengine/cableengine.go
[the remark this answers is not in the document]
tpantelis: It would also make it easier to unit test the syncer part with a fake `Engine`.

test/e2e/gateway/gateway_reporting.go
mangelajo: No, also for administrators, they can use the label to get (from cmdline, or API) the active gateway, I think it's a valide and helpful use case.

tpantelis: I don't see why. One can filter on fields as well.  But if labels are easier then just store as a label  but not both although I think it makes more sense as a field.

pkg/cableengine/cableengine.go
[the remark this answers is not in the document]
mangelajo: Done.

pkg/cableengine/cableengine.go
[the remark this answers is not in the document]
mangelajo: Done.

pkg/cableengine/syncgatewaystatus.go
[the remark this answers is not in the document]
mangelajo: Done

pkg/cableengine/syncgatewaystatus.go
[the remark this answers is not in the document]
mangelajo: Done.

pkg/cableengine/syncgatewaystatus.go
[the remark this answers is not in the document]
mangelajo: Done.

test/e2e/framework/framework.go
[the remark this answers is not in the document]
mangelajo: I think I even use goimports -w or something to do it automatically, may be we should add it to the CI chain. Not sure if the local-packages thing is part of the IDE :-?

test/e2e/framework/framework.go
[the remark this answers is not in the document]
mangelajo: I will handle the e2e comments on a separate patch, submitting one now.

pkg/cableengine/cableengine.go
```diff
@@ -66,11 +66,10 @@ func (i *engine) GetLocalEndpoint() *types.SubmarinerEndpoint {
 
 func (i *engine) StartEngine() error {
 
-	klog.Infof("CableEngine controller started, driver: %q", i.driver.GetName())
 	if err := i.startDriver(); err != nil {
 		return err
 	}
-
+	klog.Infof("CableEngine controller started, driver: %q", i.driver.GetName())
```
mangelajo, line 72: This wasn't really a comment, by tom, just fixing a mistake of my previous commit.

pkg/cableengine/cableengine.go
```diff
@@ -121,7 +144,22 @@ func (i *engine) RemoveCable(endpoint types.SubmarinerEndpoint) error {
 	return nil
 }
 
+func (i *engine) GetHAStatus() v1.HAStatus {
+	if i.driver == nil {
```
mangelajo, line 148: locking

pkg/cableengine/cableengine.go
```diff
@@ -121,7 +144,22 @@ func (i *engine) RemoveCable(endpoint types.SubmarinerEndpoint) error {
...
+		// the driver is running from the start, but could be in passive status, or
+		// in active/active.
+		return v1.HAStatusActive
+	}
+}
+
 func (i *engine) ListCableConnections() (*[]v1.Connection, error) {
 
-	return i.driver.GetConnections()
+	if i.driver != nil {
```
mangelajo, line 160: locking/mutex

test/e2e/framework/framework.go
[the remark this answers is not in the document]
mangelajo: Done.

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,64 @@ func queryAndUpdateGlobalnetStatus() {
...
+			}
+			return resGw, err
+		},
+		func(result interface{}) (bool, string, error) {
+			if result == nil {
+				return false, "gateway not found yet", nil
+			}
+			gw := result.(*submarinerv1.Gateway)
+			if gw.Status.HAStatus != status {
+				return false, "", fmt.Errorf("Gateway %q showed up but has wrong status %q, expected %q",
```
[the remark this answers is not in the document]
mangelajo: Done.

test/e2e/framework/framework.go
mangelajo: Done.

test/e2e/framework/framework.go
mangelajo: Done.

test/e2e/framework/framework.go
mangelajo: I prefer to keep it, It's added as self-documentation otherwise the lambda function was very obscure and I thought it was harder to understand the real intention of that bool

test/e2e/gateway/gateway_reporting.go
mangelajo: Ok, my concern on usability is very well handled by the proposal from @Oats87 to make the ha-status printable:

```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: gateways.submariner.io
spec:
  conversion:
    strategy: None
  group: submariner.io
  names:
    kind: Gateway
    listKind: GatewayList
    plural: gateways
    singular: gateway
  scope: Namespaced
  version: v1
  versions:
  - name: v1
    served: true
    storage: true
  additionalPrinterColumns:
  - name: ha-status 
    type: string
    description: High Availability Status of the Gateway 
    JSONPath: .status.haStatus

```

Now:

```
$ kubectl get Gateways -A --context cluster2
NAMESPACE             NAME               HA-STATUS
submariner-operator   cluster2-worker2   active
```
I removed the labels :-)

pkg/apis/submariner.io/v1/types.go
```diff
@@ -90,8 +91,8 @@ type GatewayList struct {
 type HAStatus string
 
 const (
-	HAStatusActive  HAStatus = "ACTIVE"
-	HAStatusPassive HAStatus = "PASSIVE"
+	HAStatusActive  HAStatus = "active"
```
mangelajo, line 94: Those where changed while playing with labels, full upcase was weird compared to other resources.

pkg/cableengine/cableengine.go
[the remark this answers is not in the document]
mangelajo: Locking is now handled in StartEngine. Done.

test/e2e/framework/framework.go
mangelajo: Oh I missed this one, adding another patch sorry.

mangelajo: Done.

mangelajo:
Unrelated globalnet failure, I re-ran that test.

mangelajo:
Thanks for the review @skitt

skitt:
> Thanks for the review @skitt

Not much of a review, but it works for me, and my concerns have been raised and addressed already!

pkg/cableengine/syncer/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+			Name:        localEndpoint.Spec.Hostname,
+			Annotations: map[string]string{updateTimestampAnnotation: strconv.FormatInt(time.Now().UTC().Unix(), 10)}},
+	}
+
+	gateway.Status.HAStatus = i.engine.GetHAStatus()
+
+	connections, err := i.engine.ListCableConnections()
+	if err != nil {
+		gateway.Status.StatusFailure = fmt.Sprintf("error getting driver connections: %s", err)
+		klog.Errorf("error getting driver connections: %s", err)
```
tpantelis, line 171:
```suggestion
		klog.Errorf("Error getting driver connections: %s", err)
```

pkg/cableengine/syncer/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+	"k8s.io/klog"
+
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cableengine"
+	v1typed "github.com/submariner-io/submariner/pkg/client/clientset/versioned/typed/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/log"
+)
+
+type GatewaySyncer struct {
+	sync.Mutex
```
tpantelis, line 22: This mutex is not really needed as no local state is mutated and `syncGatewayStatus` is not called concurrently.

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,122 @@ func queryAndUpdateGlobalnetStatus() {
...
+			if gw.Status.HAStatus != status {
+				return false, "", fmt.Errorf("Gateway %q exists but has wrong status %q, expected %q",
+					gw.Name, gw.Status.HAStatus, status)
+			}
+			return true, "", nil
+		})
+	return gw.(*submarinerv1.Gateway)
+}
+
+func (f *Framework) AwaitForGatewayRemoved(cluster framework.ClusterIndex, name string) {
```
tpantelis, line 95:
```suggestion
func (f *Framework) AwaitGatewayRemoved(cluster framework.ClusterIndex, name string) {
```

test/e2e/redundancy/gateway_failover.go
```diff
@@ -27,9 +37,19 @@ var _ = Describe("[redundancy] Gateway fail-over tests", func() {
 			testGatewayFailOverScenario(f)
 		})
 	})
-
 })
 
+func testBasicGatewayReporting(f *subFramework.Framework) {
+	clusterAName := framework.TestContext.ClusterIDs[framework.ClusterA]
+
+	By(fmt.Sprintf("Ensuring that one gateway reports as active %q", clusterAName))
```
tpantelis, line 45:
```suggestion
	By(fmt.Sprintf("Ensuring that one gateway reports as active on %q", clusterAName))
```

test/e2e/redundancy/gateway_failover.go
```diff
@@ -27,9 +37,19 @@ var _ = Describe("[redundancy] Gateway fail-over tests", func() {
...
+	activeGateways := f.GetGatewaysWithHAStatus(subv1.HAStatusActive)
+	Expect(activeGateways).To(HaveLen(1))
```
tpantelis, line 47: I think this should be wrapped with `framework.AwaitUntil` to avoid intermittent failures if the active was not yet running or written it's status. Perhaps `AwaitActiveGateway`.

test/e2e/framework/framework.go
```diff
@@ -60,3 +66,122 @@ func queryAndUpdateGlobalnetStatus() {
...
+	names := []string{}
+	for _, gw := range gateways {
+		names = append(names, gw.Name)
+	}
+	return names
+}
+
+func (f *Framework) GetGatewaysWithHAStatus(status submarinerv1.HAStatus) []submarinerv1.Gateway {
+
+	gatewayClient := SubmarinerClients[framework.ClusterA].SubmarinerV1().Gateways(
```
tpantelis, line 174: This should really pass in the `ClusterIndex` instead of assuming `ClusterA` to be consistent with other such functions.

test/e2e/redundancy/gateway_failover.go
```diff
@@ -41,12 +61,24 @@ func testEnginePodRestartScenario(f *subFramework.Framework) {
 	enginePod := f.AwaitSubmarinerEnginePod(framework.ClusterA)
 	By(fmt.Sprintf("Found submariner engine pod %q on %q", enginePod.Name, clusterAName))
 
+	By(fmt.Sprintf("Ensuring that the gateway reports as active %q", clusterAName))
```
tpantelis, line 64:
```suggestion
	By(fmt.Sprintf("Ensuring that the gateway reports as active on %q", clusterAName))
```

test/e2e/redundancy/gateway_failover.go
```diff
@@ -41,12 +61,24 @@ func testEnginePodRestartScenario(f *subFramework.Framework) {
...
+	activeGateways := f.GetGatewaysWithHAStatus(subv1.HAStatusActive)
+	Expect(activeGateways).To(HaveLen(1))
+
+	gw := activeGateways[0]
```
tpantelis, line 68:
```suggestion
	activeGateway := activeGateways[0]
```

For consistency ...

test/e2e/redundancy/gateway_failover.go
```diff
@@ -41,12 +61,24 @@ func testEnginePodRestartScenario(f *subFramework.Framework) {
 	enginePod := f.AwaitSubmarinerEnginePod(framework.ClusterA)
 	By(fmt.Sprintf("Found submariner engine pod %q on %q", enginePod.Name, clusterAName))
 
+	By(fmt.Sprintf("Ensuring that the gateway reports as active %q", clusterAName))
+	activeGateways := f.GetGatewaysWithHAStatus(subv1.HAStatusActive)
+	Expect(activeGateways).To(HaveLen(1))
+
```
tpantelis, line 67: We should verify that the found active `Gateway`'s name matches that of the gateway node found above.

test/e2e/redundancy/gateway_failover.go
```diff
@@ -41,12 +61,24 @@ func testEnginePodRestartScenario(f *subFramework.Framework) {
...
+	gw := activeGateways[0]
+	By(fmt.Sprintf("Ensuring that the gateway %q is reporting connections", gw.Name))
+	Expect(gw.Status.Connections).NotTo(BeEmpty())
+	Expect(gw.Status.Connections[0].Status).To(Equal(subv1.Connected))
```
tpantelis, line 71: This could cause an intermittent failure if active not yet reporting connections - it seems this should call `AwaitGatewayFullyConnected`.

test/e2e/redundancy/gateway_failover.go
```diff
@@ -41,12 +61,24 @@ func testEnginePodRestartScenario(f *subFramework.Framework) {
...
+
 	By(fmt.Sprintf("Deleting submariner engine pod %q", enginePod.Name))
 	f.DeletePod(framework.ClusterA, enginePod.Name, framework.TestContext.SubmarinerNamespace)
 
 	newEnginePod := f.AwaitSubmarinerEnginePod(framework.ClusterA)
 	By(fmt.Sprintf("Found new submariner engine pod %q", newEnginePod.Name))
 
+	By(fmt.Sprintf("Waiting for the gateway to be up and connected %q", newEnginePod.Name))
+	f.AwaitGatewayFullyConnected(framework.ClusterA, gw.Name)
```
tpantelis, line 80: This could report fully connected as the last state from the previous engine pod.  Perhaps we should move this after the connectivity tests below - that way we know the gateway is re-connected and should be reporting the new state.

test/e2e/redundancy/gateway_failover.go
```diff
@@ -100,6 +132,13 @@ func testGatewayFailOverScenario(f *subFramework.Framework) {
 	By(fmt.Sprintf("Setting the gateway label for node %q to true", initialNonGatewayNode.Name))
 	f.SetGatewayLabelOnNode(framework.ClusterA, initialNonGatewayNode.Name, true)
 
+	By(fmt.Sprintf("Ensuring that two Gateways become available in cluster %q", clusterAName))
+
+	gwActive := f.AwaitGatewayWithStatus(framework.ClusterA, initialGatewayNode.Name, subv1.HAStatusActive)
+	gwPassive := f.AwaitGatewayWithStatus(framework.ClusterA, initialNonGatewayNode.Name, subv1.HAStatusPassive)
+	Expect(gwActive.Status.Connections).ToNot(BeEmpty(), "The active gateway must have active connections")
```
tpantelis, line 139: As above, it would be safest to call `AwaitGatewayFullyConnected` for the active gateway.

test/e2e/redundancy/gateway_failover.go
tpantelis: Actually there's already `AwaitGatewayWithStatus` although it takes the gateway name which could be obtained by finding the active gateway node. But I realize we're actually duplicating what `testEnginePodRestartScenario` does so why do we need this test?

test/e2e/redundancy/gateway_failover.go
tpantelis: Or better yet, call `AwaitGatewayWithStatus`.

pkg/cableengine/syncer/syncer.go
mangelajo: I'm not sure if the GatewayInterface client can be called concurrently, are we sure about that? that was the reason.

test/e2e/framework/framework.go
mangelajo: oh true!

test/e2e/redundancy/gateway_failover.go
mangelajo: For independence, I wanted an indication that the basic reporting is working if the dataplane is broken. When we have stats then we could identify if it's the dataplane breaking, or the reporting breaking. The reporting would always affect the dataplane tests though.

mangelajo: I think we should do AwaitActiveGateway at least to avoid possible intermittent failures as you said.

test/e2e/redundancy/gateway_failover.go
mangelajo: But I want to verify there's only "1". I can't do that with Await, I can call Await just before this, which would make sense, also would remove any glitches if it's starts as passive and then moves to active.

test/e2e/redundancy/gateway_failover.go
mangelajo: +1

test/e2e/redundancy/gateway_failover.go
mangelajo: Oh, that is true, I was actually wanting to make sure we had connections before starting the dataplane tests.

This will work better/faster when we I implement the cleanup of state at pod exit, I didn't dig into that, but I will do right after 0.3.0-rc1
https://github.com/submariner-io/submariner/issues/518

test/e2e/redundancy/gateway_failover.go
mangelajo: True.

mangelajo:
@tpantelis I opened this to handle comments as a follow-up https://github.com/submariner-io/submariner/issues/519 if that's ok

Currently lighthouse pushed a new ":latest" image, and this will fail when I submit any new changes.

In such way we are able to release 0.3.0-rc1 and at the same time test the changes you're proposing here (changes that make sense and I agree with).

mangelajo:
Wait, may be there's no such chicken-egg, since we don't depend on lighthouse here.. ':D
testing locally and I will send a PR update if it works.

pkg/cableengine/syncer/syncer.go
tpantelis: the clientset? yeah they're  thread-safe - they carry no state - they're just front-ends to the API server.

mangelajo:
The technical difficulty with the lighthouse images does not exist, so it should be fine :) anyway we are talking on slack about merging this and do the follow up PR so we can cut RC1 now

test/e2e/redundancy/gateway_failover.go
tpantelis: yeah I was thinking that too - delete on engine shutdown.

tpantelis:
My comments aren't critical - let's merge this and address in the follow-ups @mangelajo created.

pkg/cableengine/syncer/syncer.go
mangelajo: Done.

pkg/cableengine/syncer/syncer.go
mangelajo: Done.

test/e2e/redundancy/gateway_failover.go
mangelajo: Oh AwaitGateway does not work, because we don't know the name, but we know that we want one. I will add another function that works with any.

test/e2e/framework/framework.go
mangelajo: Done.

test/e2e/redundancy/gateway_failover.go
mangelajo: done

test/e2e/redundancy/gateway_failover.go
mangelajo: Ok, I did it in a slightly different way, let me know what you think

test/e2e/redundancy/gateway_failover.go
mangelajo: Done

test/e2e/redundancy/gateway_failover.go
mangelajo: Done.

test/e2e/redundancy/gateway_failover.go
mangelajo: I added a DeleteGateway function for now, right after the pod deletion, which will ensure that at least the gateway status is re-created, and that it eventually moves to "fully connected"

When I do #518 then we remove this, and probably do a different check (I guess an informer to make sure it's deleted and re-created).

test/e2e/redundancy/gateway_failover.go
mangelajo: Done.

PR #503, opened 2020-04-14
Add reload-images makefile target

mangelajo:
This target will push the images to the localhost registry on the clusters
so the developer can subsequently set imagePullPolicy to always on the desired
resource and force it to reload. Alowing a very quick developer cycle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ e2e: deploy
 	./scripts/kind-e2e/e2e.sh --deploytool $(deploytool)
 
 reload-images: build images
-	./scripts/$@ --globalnet true
+	./scripts/$@
 
 $(TARGETS): vendor/modules.txt
 	./scripts/$@ --build_debug $(build_debug)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,14 +6,17 @@ ifneq (,$(DAPPER_HOST_ARCH))
...
+TARGETS := $(shell ls -p scripts | grep -v -e / | grep -v reload-images)
 CLUSTERS_ARGS = --cluster_settings $(DAPPER_SOURCE)/scripts/kind-e2e/cluster_settings
 
 clusters: build images
 
 e2e: deploy
 	./scripts/kind-e2e/e2e.sh --deploytool $(deploytool)
 
+reload-images: build images
+	./scripts/$@ --globalnet true
```
sridhargaddam, line 18: I think "--globalnet true" is not required for this target. Can you please check.

mangelajo: Oh true let me fix that 
```suggestion
	./scripts/$@
```

mangelajo:
> Wouldn’t this be more useful in shipyard?

It would, but every project would have it's own different images, which are project-dependent.

We are already using the shipyard includes to build this. I also want to add behavior to let us restart images in next patch, like:

```
make reload-images restart=all
```

```
make reload-images restart=routeagent
```
etc, which again is very project specific.

PR #506, opened 2020-04-15
Register gateway type

skitt:


mangelajo:
Merging since we have another build with this patch which passed: https://travis-ci.com/github/submariner-io/submariner/jobs/319627719

PR #509, opened 2020-04-17
Fix log levels in Submariner Gateway and Globalnet

sridhargaddam:
Recently log levels were updated in Submariner code to use one
of INFO/DEBUG/TRACE. This PR modifies the defaults to use DEBUG
which otherwise were logging at TRACE.


package/submariner-globalnet.sh
```diff
@@ -3,10 +3,12 @@ set -e -x
 
 trap "exit 1" SIGTERM SIGINT
 
+SUBMARINER_VERBOSITY=${SUBMARINER_VERBOSITY:-2}
+
 if [ "${SUBMARINER_DEBUG}" == "true" ]; then
-    DEBUG="-v=9"
+    DEBUG="-v=3"
```
tpantelis, line 9: Shouldn't this be 2 as TRACE is 3?

mangelajo: By default we are making debug = trace, something we talked about at some point during the levels cleanup.

We can introduce another flag if we want SUBMARINER_TRACE ? , we'd need to enable it on the operator/subctl too, but I'm all for it. Otherwise it's not very useful that we make a difference.

tpantelis: I'm confused - this PR summary says "This PR modifies the defaults to use DEBUG
which otherwise were logging at TRACE." which I take to mean we shouldn't be printing TRACE when the _debug_  flag is set. Previously **v** was set to 9 which of course encompassed 3.  So I'm not clear on what this PR is actually fixing/changing? Or am I missing something.

sridhargaddam: Okay, the description might be a bit confusing. This PR fixes the following.

1. The problem with the current code is that even when SUBMARINER_DEBUG is set to false, v was configured as 4 in submariner-globalnet.sh and submariner.sh files. Because of this, globalnetController and SMGateway PODs were logging TRACE logs (as the updated logging uses value of 3 for TRACE). 

2. In submariner-route-agent.sh file, when SUBMARINER_DEBUG is false, *v* is set to 1 because of which only INFO logs were logged in route-agent-pods making it extremely difficult to debug any issues with default configuration.

So this PR fixes this behavior by modifying the default deployment to use v=2 (aka DEBUG) and when SUBMARINER_DEBUG is set to true, v=3 (aka TRACE). 

I understand that name of SUBMARINER_DEBUG is a bit confusing and I'm all for modifying it to SUBMARINER_TRACE (or something that is more appropriate). IMO we can handle those changes in a separate PR and unblock this PR to fix the current codebase.

tpantelis: Thanks for the explanation. So we set to DEBUG (2) by default and TRACE when the flag is set to get more output. Sounds fine.  Longer term I want to make the debug/verbosity level configurable via a `ConfigMap` so we can (hopefully) set it on the fly w/o restart (inn which case we can remove SUBMARINER_DEBUG flag here).

sridhargaddam: Configmap sounds like a nice idea, thanks.

PR #510, opened 2020-04-17
Allow running only the e2e part, and focusing on specific tests

mangelajo:
This PR addes the new e2e-tests target, as well as a focus env var which allows filtering e2e tests to be ran.
```
make e2e-tests focus=gateway
```

Will run only the e2e tests tagged as [gateway]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 build_debug ?= false
 restart ?= all
+focus ?= .\*
 
 ifneq (,$(DAPPER_HOST_ARCH))
 
@@ -13,7 +14,7 @@ CLUSTERS_ARGS = --cluster_settings $(DAPPER_SOURCE)/scripts/kind-e2e/cluster_set
 clusters: build images
 
 e2e: # internally depends on deploy target, which will execute only if not already deployed
-	./scripts/kind-e2e/e2e.sh --deploytool $(deploytool)
+	./scripts/kind-e2e/e2e.sh --deploytool $(deploytool) --focus $(focus)
 
 reload-images: build images
 	./scripts/$@ --restart $(restart)
--- a/scripts/kind-e2e/README.md
+++ b/scripts/kind-e2e/README.md
@@ -86,6 +86,28 @@ manually, then run:
 make reload-images restart=none
 ```
 
+#### Re-running e2e tests after your code changes
+Once your kind virtual clusters and submariner are deployed, you can re-run e2e just by repeating the `make e2e` call.
+Deployment and install will be avoided, hence the existing environment will remain as is.
+If you need to load new container images, please check `make reload-images` from the previous section.
+
+In case you want to re-run just the tests:
+```bash
+make e2e
+```
+
+In case you want to use updated images and re-run the tests:
+```bash
+make reload-images e2e
+```
+
+You can focus the e2e tests on specific tags by using the focus makefile env. The following
+example runs only the tests tagged as [redundancy].
+
+```bash
+make e2e focus=redundancy
+```
+
 #### Cleanup
 At any time you can run a cleanup command that will remove all the kind clusters.
 
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -4,11 +4,13 @@
 
 source /usr/share/shflags/shflags
 DEFINE_string 'deploytool' 'operator' 'Tool to use for deploying (operator/helm)'
+DEFINE_string 'focus' '.*' 'Ginkgo focus for the E2E tests'
 FLAGS "$@" || exit $?
 eval set -- "${FLAGS_ARGV}"
 
 deploytool="${FLAGS_deploytool}"
-echo "Running with: deploytool=${deploytool}"
+focus="${FLAGS_focus}"
+echo "Running with: deploytool=${deploytool} focus=${focus}"
 
 set -em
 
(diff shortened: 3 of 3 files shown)


scripts/kind-e2e/README.md
```diff
@@ -86,6 +86,19 @@ manually, then run:
 make reload-images restart=none
 ```
 
+#### Re-running e2e tests after your code changes
+You can run the e2e tests without re-deploying everything by using the e2e-tests target.
+```bash
+make e2e-tests
```
skitt, line 92:
```suggestion
make e2e-test
```

scripts/kind-e2e/README.md
```diff
@@ -86,6 +86,19 @@ manually, then run:
...
+```
+
+And you can focus the e2e tests on specific tags by using the focus makefile env. The following
+example runs only the tests tagged as [gateway].
+
+```bash
+make e2e-tests focus=gateway
```
skitt, line 99:
```suggestion
make e2e-test focus=gateway
```

scripts/kind-e2e/README.md
```diff
@@ -86,6 +86,19 @@ manually, then run:
...
 ```
 
+#### Re-running e2e tests after your code changes
+You can run the e2e tests without re-deploying everything by using the e2e-tests target.
+```bash
+make e2e-test
+```
+
+And you can focus the e2e tests on specific tags by using the focus makefile env. The following
+example runs only the tests tagged as [gateway].
```
tpantelis, line 96: They're actually tagged as _[redundancy]_, not _[gateway]_ 😄

mangelajo: I was referencing the new tests I was writing but let's use redundancy instead as as we talked about, they could end up being the same..

mangelajo, line 96:
```suggestion
example runs only the tests tagged as [redundancy].
```

scripts/kind-e2e/README.md
```diff
@@ -86,6 +86,19 @@ manually, then run:
...
+
+```bash
+make e2e-test focus=gateway
```
mangelajo, line 99:
```suggestion
make e2e-test focus=redundancy
```

Makefile
```diff
@@ -12,8 +13,10 @@ CLUSTERS_ARGS = --cluster_settings $(DAPPER_SOURCE)/scripts/kind-e2e/cluster_set
 
 clusters: build images
 
-e2e: deploy
-	./scripts/kind-e2e/e2e.sh --deploytool $(deploytool)
+e2e: deploy e2e-test
```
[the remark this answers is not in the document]
mangelajo: I think that whatever we propose ideally should be uniform across repositories, so contributors don't need to think how things are done when moving around.

I'm going to split the focus part to a separate patch while we keep thinking, but we need to find a solution soon, I find hard to develop if I can't have a quick devel cycle (1 modify, 2 compile, 3 test...) :-)

scripts/kind-e2e/README.md
```diff
@@ -86,6 +86,27 @@ manually, then run:
 make reload-images restart=none
 ```
 
+#### Re-running e2e tests after your code changes
+Once your kind virtual clusters and submariner are deployed once you can run e2e
+just by repeating the e2e make call. Deployment and install will be avoided, if
+if you need to load new container images, please check `make reload-images` from
+the previous section.
```
mkolesnik, line 93:
```suggestion
#### Re-running e2e tests after your code changes
Once your kind virtual clusters and submariner are deployed, you can re-run e2e just by repeating the `make e2e` call.
Deployment and install will be avoided, hence the existing environment will remain as is.
If you need to load new container images, please check `make reload-images` from the previous section.
```

scripts/kind-e2e/README.md
```diff
@@ -86,6 +86,27 @@ manually, then run:
...
+
+```bash
```
mkolesnik, line 95:
```suggestion
In case you want to re-run just the tests:
```bash
```

scripts/kind-e2e/README.md
```diff
@@ -86,6 +86,27 @@ manually, then run:
...
+make e2e
+```
+
+```bash
```
mkolesnik, line 99:
```suggestion
In case you want to use updated images and re-run the tests:
```bash
```

scripts/kind-e2e/e2e.sh
```diff
@@ -48,9 +50,14 @@ function test_with_e2e_tests {
...
     cd ${DAPPER_SOURCE}/test/e2e
 
+    if [[ "${focus}" != "all" ]]; then
+      GINKGO_FOCUS="-ginkgo.focus \[${focus}\]"
+    fi
+
     go test -v -args -ginkgo.v -ginkgo.randomizeAllSpecs \
         -submariner-namespace $SUBM_NS -dp-context cluster2 -dp-context cluster3 -dp-context cluster1 \
         -ginkgo.noColor -ginkgo.reportPassed \
+        ${GINKGO_FOCUS} \
```
mkolesnik, line 60: If you go with default `.*` then you can just always pass `-ginkgo.focus ${focus}`

scripts/kind-e2e/e2e.sh
```diff
@@ -4,11 +4,13 @@
 
 source /usr/share/shflags/shflags
 DEFINE_string 'deploytool' 'operator' 'Tool to use for deploying (operator/helm)'
+DEFINE_string 'focus' 'all' 'Ginkgo focus for the E2E tests'
```
mkolesnik, line 7: Since focus is a regex you can set the default to '.*' and it should work always if you pass $focus as is to -ginkgo.focus

mangelajo:
@mkolesnik I tried locally, and I prefer not to apply your suggestion.

I was trying to target tags \[xxxx\] which are a bit nasty to type manually, and if you don't scope with tags it will grab almost anything because it uses all the combinations of texts in the describes/whens, etc...

mangelajo:
ah Wait but I can do --ginkgo.focus "\[${focus}\]" 
trying

mangelajo:
> Please remember to squash this when merging.

Thanks Skitt, I "squashed@home" for just in case

Makefile
```diff
@@ -1,5 +1,6 @@
 build_debug ?= false
 restart ?= all
+focus ?= .\*
```
mkolesnik, line 3:
```suggestion
focus ?= '.*'
```
nit: if you use quotes it should work as well

PR #511, opened 2020-04-19
Remove load_deploytool

mkolesnik:
Since SUBM_NS is now declared in utils in shipyard, no need for this
copied code and this flag.

Also removed cleanup method which was left undeleted by mistake.


Makefile
```diff
@@ -13,7 +13,7 @@ CLUSTERS_ARGS = --cluster_settings $(DAPPER_SOURCE)/scripts/kind-e2e/cluster_set
 clusters: build images
 
 e2e: # internally depends on deploy target, which will execute only if not already deployed
-	./scripts/kind-e2e/e2e.sh --deploytool $(deploytool)
+	./scripts/kind-e2e/e2e.sh
```
mangelajo, line 16: Don't we need to pass the flags down ? how's it working now that we are calling deploy from the kind e2e? :?

mangelajo: Hmm, I see it was working for globalnet that we don't pass, how? :?

mkolesnik: it's magic of make :)
it's passed since it's still declared in .travis.yaml

mangelajo: But .. hmmm I see this:   - env: CMD="make e2e deploytool=helm globalnet=true"

How are those passed down anymore if we remove the flags when calling e2e.sh ?

mangelajo: Looking at CI it's clearly working, so I'll keep the questions offline ;)

PR #512, opened 2020-04-19
Deploy "lazily" so that e2e can be re run

mkolesnik:
This allows consecutively re running the e2e tests without going through
the (reenrant, yet still time consuming) deploy process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ CLUSTERS_ARGS = --cluster_settings $(DAPPER_SOURCE)/scripts/kind-e2e/cluster_set
 
 clusters: build images
 
-e2e:
+e2e: # internally depends on deploy target, which will execute only if not already deployed
 	./scripts/kind-e2e/e2e.sh --deploytool $(deploytool)
 
 reload-images: build images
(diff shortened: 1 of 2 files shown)


scripts/kind-e2e/e2e.sh
```diff
@@ -34,6 +34,15 @@ function load_deploytool() {
     . $deploy_lib
 }
 
+function deploy_env() {
```
mangelajo, line 37: nit: may be call it deploy_env_once ?

Makefile
```diff
@@ -11,7 +11,7 @@ CLUSTERS_ARGS = --cluster_settings $(DAPPER_SOURCE)/scripts/kind-e2e/cluster_set
 
 clusters: build images
 
-e2e: deploy
+e2e:
```
mangelajo, line 14:
```suggestion
e2e: # internally depends on deploy target, which will execute only if not already deployed
```

scripts/kind-e2e/e2e.sh
mangelajo, line 37:
```suggestion
function deploy_env_once() {
```

scripts/kind-e2e/e2e.sh
```diff
@@ -56,6 +65,7 @@ declare_kubeconfig
 
 load_deploytool
 
+deploy_env
```
mangelajo, line 68:
```suggestion
deploy_env_once
```

PR #515, opened 2020-04-21
add Wireguard support for connection status

deanlorenz:
Implement GetConnections
Update README.md --install linux-headers

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,4 +1,4 @@
-FROM quay.io/submariner/shipyard-dapper-base
+FROM quay.io/submariner/shipyard-dapper-base:b21b6fd
 
 ENV DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD TRAVIS_COMMIT CLUSTERS_ARGS DEPLOY_ARGS" \
     DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner DAPPER_DOCKER_SOCKET=true
--- a/pkg/cable/wireguard/driver.go
+++ b/pkg/cable/wireguard/driver.go
@@ -29,10 +29,16 @@ const (
 	// PublicKey is name (key) of publicKey entry in back-end map
 	PublicKey = "publicKey"
 
-	// KeepAliveInterval specifies 1/3 timeout for announcing a connection ERROR
+	// KeepAliveInterval to use for wg peers
 	KeepAliveInterval = 10 * time.Second
 
+	// handshakeTimeout is maximal time from handshake a connections is still considered connected
+	handshakeTimeout = 2 * time.Minute + 10 * time.Second
+
 	cableDriverName = "wireguard"
+	receiveBytes    = "ReceiveBytes"   // for peer connection status
+	transmitBytes   = "TransmitBytes"  // for peer connection status
+	lastChecked     = "LastChecked" // for connection peer status
 )
 
 func init() {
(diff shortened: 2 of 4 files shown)


deanlorenz:
@mangelajo I updated the code to use the Tx/Rx byte count and not rely only on last handshake. I did some WG testing and it seems handshake is not updated on idle connection, even with persistent keepalive turned on. 
I removed almost all of the TODOs. The ones still there are not can be left for future enhancements.

pkg/cable/wireguard/driver.go
```diff
@@ -291,6 +299,15 @@ func (w *wireguard) DisconnectFromEndpoint(remoteEndpoint types.SubmarinerEndpoi
 }
 
 func (w *wireguard) GetActiveConnections(clusterID string) ([]string, error) {
+
+	//klog.V(log.DEBUG).Infof("WG STATUS for cluster %s", clusterID)
```
mangelajo, line 303: I think it's better not to keep commented code in the codebase, we can always have this PR for reference if you want to look at this code later for implementation.

Does it work without the GetActiveConnections?, I guess your driver is already idempotent, we could do the same in the ipsec versions and remove that method

pkg/cable/wireguard/driver.go
```diff
@@ -134,26 +148,38 @@ func (w *wireguard) ConnectToEndpoint(remoteEndpoint types.SubmarinerEndpoint) (
...
 		_ = w.removePeer(&oldKey)
 
 		delete(w.peers, remoteEndpoint.Spec.ClusterID)
+		delete(w.connections, remoteEndpoint.Spec.ClusterID)
 	}
 
+	// create connection, overwrite existing connection
+	connection := v1.NewConnection(remoteEndpoint.Spec)
+	klog.V(log.DEBUG).Infof("Adding connection for cluster %s, %v", remoteEndpoint.Spec.ClusterID, connection)
+	w.connections[remoteEndpoint.Spec.ClusterID] = connection
```
tpantelis, line 178: We add the `connections` entry here but an intervening error would elide the entry added to the `peers` map below on L212. Maintaining one map as discussed above avoids inconsistency.

pkg/cable/wireguard/getconnections.go
```diff
@@ -2,10 +2,115 @@ package wireguard
...
+
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	for _, p := range d.Peers {
+		key := p.PublicKey
+		cid, err := w.clusterIDByPeer(&key)
+		if err != nil {
+			klog.Warningf("Found unknown peer with key %s, removing", key)
+			if err2 := w.removePeer(&key); err2 != nil {
```
tpantelis, line 29: The purpose of this function is to retrieve current connection info thus it doesn't seem appropriate to mutate internal state as a side effect.

Perhaps the approach here should be the opposite, ie iterate the cached `Connection`s and retrieve its peer from `d.Peers` and , if not found, then report that fact in the returned `Connection`. I don't know how an orphaned connection could occur or if it would be a bug somewhere

pkg/cable/wireguard/driver.go
```diff
@@ -134,26 +148,38 @@ func (w *wireguard) ConnectToEndpoint(remoteEndpoint types.SubmarinerEndpoint) (
...
-	// delete old peers for ClusterID
+	// delete or update old peers for ClusterID
 	oldKey, found := w.peers[remoteEndpoint.Spec.ClusterID]
 	if found {
 		if oldKey.String() == remoteKey.String() {
-			//TODO check that peer config has not changed (eg allowedIPs)
-			klog.V(log.DEBUG).Infof("Skipping update of existing peer key %s", oldKey)
-			return ip, nil
+			// keys match, existing peer, so lets update status
+			if err := w.updatePeerStatus(remoteKey, remoteEndpoint.Spec.ClusterID); err == nil {
```
tpantelis, line 161: Why do we update the `Connection` info here as that's done on `GetConnections` anyway?

pkg/cable/wireguard/driver.go
deanlorenz: Removed. 
WG cable does not use GetActiveConnections.

pkg/cable/wireguard/getconnections.go
```diff
@@ -2,10 +2,115 @@ package wireguard
...
+				klog.Errorf("Could not delete WireGuard peer with key %s, ignoring: %v", key, err)
```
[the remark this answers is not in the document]
deanlorenz: done

pkg/cable/wireguard/driver.go
```diff
@@ -36,7 +48,8 @@ func init() {
 type wireguard struct {
 	localSubnets  []net.IPNet
 	localEndpoint types.SubmarinerEndpoint
-	peers         map[string]wgtypes.Key // clusterID -> publicKey
+	peers         map[string]wgtypes.Key    // clusterID -> publicKey
+	connections   map[string]*v1.Connection // clusterID -> publicKey
```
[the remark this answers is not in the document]
deanlorenz: done. key saved in connection

pkg/cable/wireguard/driver.go
deanlorenz: agree

pkg/cable/wireguard/getconnections.go
deanlorenz: The source of truth should be the wg network device. 
No connection would be returned for submariner endpoints without a wg peer.
wg peers without a submariner endpoint indicate an error or someone changing things externally. Although I agree that this may not be appropriate for a status retrieval function, I think it is better to remove the wg peer.

pkg/cable/wireguard/driver.go
deanlorenz: This is not a must, if we assume GetConnections is called periodically.
The status relies on Tx/Rx counts from last check. 
ConnectToEndpoint for existing peers is called periodically so it is an opportunity to refresh Tx/Rx counters.

pkg/cable/wireguard/getconnections.go
deanlorenz: I think this applies to both ipsec and wg -- 
there are two "orphan" cases (1) a submariner endpoint without a corresponding cable connection (2) a cable connection without a submariner endpoint. For (1), do we return only existing cable connections or do we create a connection for each submariner endpoint, but with ERROR status -- I implemented the former. For (2), there is no other place where this is checked so either we report an error or delete the orphan cable connections -- I implemented the latter.

deanlorenz:
@tpantelis , @mangelajo

main.go
```diff
@@ -53,9 +53,9 @@ type leaderConfig struct {
 
 const (
 	leadershipConfigEnvPrefix = "leadership"
-	defaultLeaseDuration      = 5 // In Seconds
-	defaultRenewDeadline      = 3 // In Seconds
-	defaultRetryPeriod        = 2 // In Seconds
+	defaultLeaseDuration      = 10 // In Seconds
+	defaultRenewDeadline      = 5  // In Seconds
+	defaultRetryPeriod        = 2  // In Seconds
```
tpantelis, line 58: These changes were made by another PR that was merged - not sure how they ended up here...

deanlorenz: no idea... will reset and remerge

deanlorenz: @tpantelis rebase fixed it

Dockerfile.dapper
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/submariner/shipyard-dapper-base
+FROM quay.io/submariner/shipyard-dapper-base:b21b6fd
```
mangelajo, line 1: This is agreed because the latest shipyard image was breaking subm e2e

mangelajo:
@tpantelis ping ^ ;)

Issue #519, opened 2020-04-22
Handle review comments for Gateway status update before 0.3.0


mangelajo:
The comments are here:

https://github.com/submariner-io/submariner/pull/502/files

PR #521, opened 2020-04-23
Update redundancy e2e tests with improvements from code review

mangelajo:
Closes-Issue: #519

### DIFF
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -76,7 +76,7 @@ type GatewayStatus struct {
 	Version       string       `json:"version"`
 	HAStatus      HAStatus     `json:"haStatus"`
 	LocalEndpoint EndpointSpec `json:"localEndpoint"`
-	StatusFailure string       `json:"statusMessage"`
+	StatusFailure string       `json:"statusFailure"`
 	Connections   []Connection `json:"connections"`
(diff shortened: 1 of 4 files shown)


mangelajo:
Lol, I'm uncovering a deployment difference with helm, in helm we don't setup cluster1 dataplane, and we have a SubmarinerClient to that one where "Gateway CRD" is not registered.
(I touch it during the cleanup, where now I iterate to all clusters, for just in case tests will be modifying gateways in such clusters).

That difference will not play well either with the add/remove clusters cases , which now are disabled.
@tpantelis  ^

tpantelis:
> Lol, I'm uncovering a deployment difference with helm, in helm we don't setup cluster1 dataplane, and we have a SubmarinerClient to that one where "Gateway CRD" is not registered.
> (I touch it during the cleanup, where now I iterate to all clusters, for just in case tests will be modifying gateways in such clusters).
> 
> That difference will not play well either with the add/remove clusters cases , which now are disabled.
> @tpantelis ^
> 
> https://travis-ci.com/github/submariner-io/submariner/jobs/322569999#L2846

I guess helm has been left behind a bit 😄

mangelajo:
I will update the patch so that we account for the possibility of the CRD not existing in one of the clusters, then later on we need to update helm deploy, I'll do that.

PR #523, opened 2020-04-24
Remove gateway entry when gateway pod exits

mangelajo:
Without this patch an active gateway pod could die and not
be respawn (no available passive node...) and in the meanwhile
nobody would remove the Gateway entry, appearing still reported
as active, connected etc...

### DIFF
--- a/main.go
+++ b/main.go
@@ -133,7 +133,7 @@ func main() {
 
 	cableEngineSyncer.Run(stopCh)
 
-	start := func(context.Context) {
+	becameLeader := func(context.Context) {
 		klog.Info("Creating the tunnel controller")
 
 		tunnelController := tunnel.NewController(submSpec.Namespace, cableEngine, kubeClient, submarinerClient,
@@ -204,10 +204,16 @@ func main() {
 	eventBroadcaster.StartLogging(klog.V(log.DEBUG).Infof)
 	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "submariner-controller"})
 
-	startLeaderElection(leClient, recorder, start)
+	lostLeader := func() {
+		cableEngineSyncer.CleanupGatewayEntry()
+		klog.Fatalf("Leader election lost, shutting down")
+	}
+
+	startLeaderElection(leClient, recorder, becameLeader, lostLeader)
 }
 
-func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder record.EventRecorder, run func(ctx context.Context)) {
+func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder record.EventRecorder,
+	run func(ctx context.Context), end func()) {
 	gwLeadershipConfig := leaderConfig{}
 
 	err := envconfig.Process(leadershipConfigEnvPrefix, &gwLeadershipConfig)
@@ -267,9 +273,7 @@ func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder rec
 		RetryPeriod:   time.Duration(gwLeadershipConfig.RetryPeriod) * time.Second,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: run,
-			OnStoppedLeading: func() {
-				klog.Fatalf("Leader election lost")
-			},
+			OnStoppedLeading: end,
 		},
 	})
 }
--- a/pkg/cableengine/syncer/syncer.go
+++ b/pkg/cableengine/syncer/syncer.go
@@ -39,7 +39,11 @@ func NewGatewaySyncer(engine cableengine.Engine, client v1typed.GatewayInterface
 }
 
 func (s *GatewaySyncer) Run(stopCh <-chan struct{}) {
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -204,10 +204,11 @@ func main() {
 	eventBroadcaster.StartLogging(klog.V(log.DEBUG).Infof)
 	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "submariner-controller"})
 
-	startLeaderElection(leClient, recorder, start)
+	startLeaderElection(leClient, recorder, start, cableEngineSyncer.CleanupOwnEntry)
```
mangelajo, line 207: the new "end" function will not force an exit of the process, we need to do that.

pkg/cableengine/syncer/syncer.go
```diff
@@ -171,3 +175,15 @@ func (i *GatewaySyncer) generateGatewayObject() (*v1.Gateway, error) {
 	klog.V(log.TRACE).Infof("generateGatewayObject: %+v", gateway)
 	return &gateway, nil
 }
+
+// CleanupOwnEntry removes this Gateway entry from the k8s API, it does not
+// propagate error up because it's a termination function that we also provide externally
+func (s *GatewaySyncer) CleanupOwnEntry() {
+	hostName := s.engine.GetLocalEndpoint().Spec.Hostname
+	err := s.client.Delete(hostName, &metav1.DeleteOptions{})
+	if errors.IsNotFound(err) {
```
tpantelis, line 184: We should check and report any error.

mangelajo: Ah, correct, we are missing the other errors.

mangelajo: Actually what I tried to do initially was : `!errors.IsNotFound(err) ` 

not logging the "not found" which means it was already deleted...

test/e2e/redundancy/gateway_failover.go
```diff
@@ -132,11 +132,20 @@ func testGatewayFailOverScenario(f *subFramework.Framework) {
...
+
+	deleteCh := subFramework.GetDeletionChannel(gwInformer)
+
 	// Set the gateway label for the active gateway worker node to false so the submariner engine pod will be
 	// terminated.
 	By(fmt.Sprintf("Setting the gateway label for node %q to false", initialGatewayNode.Name))
 	f.SetGatewayLabelOnNode(framework.ClusterA, initialGatewayNode.Name, false)
 
+	By(fmt.Sprintf("Verifying that the gateway %q was deleted", initialGatewayNode.Name))
+	Eventually(deleteCh, 10.0).Should(Receive(Equal(initialGatewayNode.Name)))
```
tpantelis, line 147: Perhaps we should use the same general timeout as `AwaitUntil`.

test/e2e/framework/framework.go
```diff
@@ -223,3 +225,32 @@ func (f *Framework) DeleteGateway(cluster framework.ClusterIndex, name string) {
...
+	informerFactory := externalversions.NewSharedInformerFactory(SubmarinerClients[cluster], 0)
+	informer := informerFactory.Submariner().V1().Gateways().Informer()
+	go informer.Run(stopCh)
+	Expect(cache.WaitForCacheSync(stopCh, informer.HasSynced)).To(BeTrue())
+	return informer, stopCh
+}
+
+func GetDeletionChannel(informer cache.SharedIndexInformer) chan string {
+	dh := simpleDeletionHandler{
+		deletionChannel: make(chan string),
```
tpantelis, line 240: Perhaps we should add a capacity to the channel, something high like 100, otherwise it's a one-shot.

test/e2e/framework/framework.go
```diff
@@ -223,3 +225,32 @@ func (f *Framework) DeleteGateway(cluster framework.ClusterIndex, name string) {
...
+	}
+	informer.AddEventHandler(dh)
```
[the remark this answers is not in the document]
mangelajo: But how do I keep the reference to the channel then?

test/e2e/redundancy/gateway_failover.go
mangelajo: :+1:

test/e2e/framework/framework.go
mangelajo: :+1: 

This use case is just 1 that we will be looking for, but just in case,

pkg/cableengine/syncer/syncer.go
```diff
@@ -171,3 +175,15 @@ func (i *GatewaySyncer) generateGatewayObject() (*v1.Gateway, error) {
 	klog.V(log.TRACE).Infof("generateGatewayObject: %+v", gateway)
 	return &gateway, nil
 }
+
+// CleanupOwnEntry removes this Gateway entry from the k8s API, it does not
+// propagate error up because it's a termination function that we also provide externally
+func (s *GatewaySyncer) CleanupOwnEntry() {
```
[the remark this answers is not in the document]
mangelajo: :+1:

Issue #529, opened 2020-04-30
Fix PANIC in Submariner seen on LH builds in CI


sridhargaddam:
Lighthouse builds are failing with the following error.

E0430 07:33:06.632456       1 runtime.go:69] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
/go/src/github.com/submariner-io/submariner/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:76
/go/src/github.com/submariner-io/submariner/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65
/go/src/github.com/submariner-io/submariner/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51
/usr/lib/golang/src/runtime/panic.go:679
/usr/lib/golang/src/runtime/panic.go:199
/usr/lib/golang/src/runtime/signal_unix.go:394
/go/src/github.com/submariner-io/submariner/pkg/routeagent/controllers/route/driver.go:75
/go/src/github.com/submariner-io/submariner/vendor/k8s.io/client-go/util/retry/util.go:64
/go/src/github.com/submariner-io/submariner/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:203
/go/src/github.com/submariner-io/submariner/vendor/k8s.io/client-go/util/retry/util.go:63
/go/src/github.com/submariner-io/submariner/pkg/routeagent/controllers/route/driver.go:66
/go/src/github.com/submariner-io/submariner/pkg/routeagent/controllers/route/route.go:220
/go/src/github.com/submariner-io/submariner/pkg/routeagent/main.go:94
/usr/lib/golang/src/runtime/asm_amd64.s:1357
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xf8193c]

goroutine 33 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/src/github.com/submariner-io/submariner/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:58 +0x105
panic(0x1137020, 0x1c7e440)
	/usr/lib/golang/src/runtime/panic.go:679 +0x1b2
[...]

PR #530, opened 2020-04-30
Return error if CNIIface is not discovered

sridhargaddam:
This PR returns an appropriate error message if we are unable to determine
the CNI Interface on the host at an early stage. Otherwise, the errors could
manifest at later stage making it difficult to understand what is going wrong.

Related to# https://github.com/submariner-io/submariner/issues/529

### DIFF
--- a/pkg/routeagent/controllers/route/driver.go
+++ b/pkg/routeagent/controllers/route/driver.go
@@ -12,24 +12,21 @@ import (
 	"github.com/submariner-io/submariner/pkg/log"
 )
 
-func discoverCNIInterface(clusterCIDR string) *cniInterface {
+func discoverCNIInterface(clusterCIDR string) (*cniInterface, error) {
 	_, clusterNetwork, err := net.ParseCIDR(clusterCIDR)
 	if err != nil {
-		klog.Errorf("Unable to ParseCIDR %q : %v", clusterCIDR, err)
-		return nil
+		return nil, fmt.Errorf("unable to ParseCIDR %q : %v", clusterCIDR, err)
 	}
 
 	hostInterfaces, err := net.Interfaces()
 	if err != nil {
-		klog.Errorf("net.Interfaces() returned error : %v", err)
-		return nil
+		return nil, fmt.Errorf("net.Interfaces() returned error : %v", err)
 	}
 
 	for _, iface := range hostInterfaces {
 		addrs, err := iface.Addrs()
 		if err != nil {
-			klog.Errorf("For interface %q, iface.Addrs returned error: %v", iface.Name, err)
-			return nil
+			return nil, fmt.Errorf("for interface %q, iface.Addrs returned error: %v", iface.Name, err)
 		}
 
 		for i := range addrs {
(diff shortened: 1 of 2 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -173,17 +173,20 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
...
+	cniIface, err := discoverCNIInterface(r.localClusterCidr[0])
+	if err != nil {
 		// Configure CNI Specific changes
+		r.cniIface = cniIface
 		err := toggleCNISpecificConfiguration(r.cniIface.name)
 		if err != nil {
 			return fmt.Errorf("toggleCNISpecificConfiguration returned error. %v", err)
 		}
+	} else {
+		return fmt.Errorf("discoverCNIInterface returned error. %v", err)
```
mkolesnik, line 185: Shouldn't this be inside the `if err != nil {` check?

pkg/routeagent/controllers/route/route.go
```diff
@@ -173,17 +173,20 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
 		return fmt.Errorf("failed to wait for caches to sync")
 	}
 
-	r.cniIface = discoverCNIInterface(r.localClusterCidr[0])
-	if r.cniIface != nil {
+	cniIface, err := discoverCNIInterface(r.localClusterCidr[0])
+	if err != nil {
```
vthapar, line 177: Think you reversed the check. It should be if err == nil or if r.cniIface != nil.

pkg/routeagent/controllers/route/route.go
```diff
@@ -173,17 +173,20 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
...
 		// Configure CNI Specific changes
+		r.cniIface = cniIface
```
vthapar, line 179: If err != nil cniIface will always be nil. Next line will panic.

pkg/routeagent/controllers/route/route.go
mangelajo: right

pkg/routeagent/controllers/route/driver.go
```diff
@@ -44,12 +41,12 @@ func discoverCNIInterface(clusterCIDR string) *cniInterface {
...
 					klog.V(log.DEBUG).Infof("Found CNI Interface %q that has IP %q from ClusterCIDR %q",
 						iface.Name, ipAddr.String(), clusterCIDR)
-					return &cniInterface{ipAddress: ipAddr.String(), name: iface.Name}
+					return &cniInterface{ipAddress: ipAddr.String(), name: iface.Name}, nil
 				}
 			}
 		}
 	}
-	return nil
+	return nil, fmt.Errorf("unable to find CNI Interface on the host which has IP from %q", clusterCIDR)
```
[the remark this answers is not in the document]
sridhargaddam: I'm afraid no. This is an unexpected case. We ideally expect to discover the CNIInterface on the host. In the rare case, if it's missing, we want to treat that as an error (instead of PANIC) and continue to function with some missing functionality.

Issue #532, opened 2020-04-30
Categorize intermittent failures in CI


dfarrell07:
It seems that we have some sporadic failures in CI, and that they are related to non-Submariner parts of the CI deployment. However, to really see the pattern and resole it, we need to document the rare-ish failures. Use this issue collect logs from sporadic CI failures.

For example:

Repo: submariner-io/submariner
Job: `CMD="make e2e" CLUSTERS_ARGS="--globalnet" DEPLOY_ARGS="${CLUSTERS_ARGS} --deploytool helm"`
Failure summary: E2E connection tests timeout
Logs: https://pastebin.com/0AWQt1pp

- [ ] Globalnet sub-issue: https://github.com/submariner-io/submariner/issues/544

dfarrell07:
Repo: submariner-io/submariner
Job: CMD="make e2e" CLUSTERS_ARGS="--globalnet" DEPLOY_ARGS="${CLUSTERS_ARGS} --deploytool helm"
Failure summary: Gateway fail-over test failure
Logs: https://pastebin.com/qxdzusfd

dfarrell07:
Repo: submariner-io/submariner
Job: `CMD="make e2e" RELEASE=true`
Failure summary: Gateway fail-over tests
Logs: https://github.com/submariner-io/submariner/pull/526#issuecomment-621361178

Repo: submariner-io/submariner
Job: `CMD="make e2e" CLUSTERS_ARGS="--globalnet true" DEPLOY_ARGS="${CLUSTERS_ARGS} --deploytool helm"`
Failure summary: Travis timeout at 10m no output
Logs: https://github.com/submariner-io/submariner/pull/526#issuecomment-621361178

dfarrell07:
Repo: Shipyard
Job: Undocumented
Failure summary: Weave deployment failure
Logs: https://github.com/submariner-io/shipyard/pull/122#issuecomment-620573349

dfarrell07:
Repo: Lighthouse
Job: `CMD="make e2e deploytool=helm"`
Failure summary: Timeout after 10m in output, in Weave section of deploy
Logs: https://pastebin.com/A6iBzNGq

Repo: Lighthouse
Job: `CMD="make e2e delpoytool=operator" RELEASE=true`
Failure summary: Failure while running clusters.sh, just after Weave deploy
Logs: https://pastebin.com/7rYT0rPU

dfarrell07:
Repo: Submariner
[...]

PR #535, opened 2020-04-30
Add validation in github actions, including artifact collection

mangelajo:


.github/workflows/validation.yml
```diff
@@ -0,0 +1,54 @@
...
+      - 'v**'
+
+name: validation
+jobs:
+  validate:
+    name: Linting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@master
```
mkolesnik, line 18: I guess GH knows to check out the correct ref? Since it says `master` here..

dfarrell07:
Seems like you had a flaky failure. Retriggered the job and documented here: https://github.com/submariner-io/submariner/issues/532#issuecomment-621947343

tpantelis:
Sorry I'm not familiar with github actions.  Can you provide some detail in the PR description as to the purpose and how it's related to the travis jobs?

.github/workflows/validation.yml
mangelajo: @master is the version of "actions/checkout" action, not ours. It's smart about which one to checkout (pull request, or pull...)

https://github.com/actions/checkout#usage

I must admit github has done a great job with the actions stuff :)

mangelajo:
Ok, based on the conversation we had the other day during the team meeting, do you guys think we could merge this? :)

.github/workflows/validation.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Create artifacts directory
+        run: mkdir artifacts
+
+      - name: Run validation
+        run: make test 2>&1 | tee artifacts/unittest.log
```
mkolesnik, line 32: Should you `set -o pipefail`?
Usually it's off in which case the command won't necessarily fail in case `make test` fails, since the `tee` is the last thing to return so it's return code will be considered (unless you've set pipefail beforehand)

mangelajo: oh true!

Issue #540, opened 2020-05-04
Submariner Gateway POD is being restarted sometimes...


sridhargaddam:
On a VSphere environment, it was seen that after leaving the setup for few days (4 in this example), Submariner Gateway POD was restarted a few times.

We have to identify the reasons why Gateway POD is restarted. Incase its done by K8s, then we have to see if we can do something to reduce/optimize it.

![9](https://user-images.githubusercontent.com/6574656/80948444-b7f28780-8e0f-11ea-9a33-9eeffdebcb51.png)

Note: Currently we have a known [issue](https://github.com/submariner-io/submariner/issues/386) where datapath will be disrupted (as IPSEC tunnel is terminated) when GW POD is restarted.

mangelajo:
Did you look at the status of the daemonset?, It should provide a record of the issues?

kubectl describe Daemonset submariner-gateaway -n submariner-operator

mangelajo:
Adding you as assignee as you are the only one with vspheres :D

sridhargaddam:
> kubectl describe Daemonset submariner-gateaway -n submariner-operator

Yes, I had a look, but could not find anything interesting.

[sgaddam@localhost mcm-cluster1]$  kubectl describe Daemonset submariner-gateway -n submariner-operator 
Name:           submariner-gateway
Selector:       app=submariner-engine
Node-Selector:  submariner.io/gateway=true
Labels:         app=submariner-engine
                component=engine
Annotations:    deprecated.daemonset.template.generation: 1
Desired Number of Nodes Scheduled: 1
Current Number of Nodes Scheduled: 1
Number of Nodes Scheduled with Up-to-date Pods: 1
Number of Nodes Scheduled with Available Pods: 1
Number of Nodes Misscheduled: 0
Pods Status:  1 Running / 0 Waiting / 0 Succeeded / 0 Failed
Pod Template:
  Labels:           app=submariner-engine
  Service Account:  submariner-operator
  Containers:
   submariner:
    Image:      quay.io/submariner/submariner:0.3.0-rc1
    Port:       <none>
    Host Port:  <none>
    Command:
      submariner.sh
    Environment:
      SUBMARINER_NAMESPACE:        submariner-operator
      <SNIP>
      CE_IPSEC_NATTPORT:           4500
    Mounts:                        <none>
  Volumes:                         <none>
Events:                            <none>
[sgaddam@localhost mcm-cluster1]$

sridhargaddam:
[...]

PR #542, opened 2020-05-04
Tweak leader election timings

sridhargaddam:
On some platforms, it was seen that the active leader node sometimes (not
frequent though) is unable to renew the lease for leaderLock and because
of this Gateway POD is getting restarted. In one of the earlier PRs, we
reduced these intervals to improve the failover detection. In this PR we
are increasing the interval slightly until we find a proper solution to
handle HA.

Related to# https://github.com/submariner-io/submariner/issues/540

### DIFF
--- a/main.go
+++ b/main.go
@@ -53,9 +53,9 @@ type leaderConfig struct {
 
 const (
 	leadershipConfigEnvPrefix = "leadership"
-	defaultLeaseDuration      = 5 // In Seconds
-	defaultRenewDeadline      = 3 // In Seconds
-	defaultRetryPeriod        = 2 // In Seconds
+	defaultLeaseDuration      = 10 // In Seconds
+	defaultRenewDeadline      = 5  // In Seconds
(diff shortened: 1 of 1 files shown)


mangelajo:
@sridhargaddam code validation complained, I think it's tabs vs spaces, not sure.

sridhargaddam:
> @sridhargaddam code validation complained, I think it's tabs vs spaces, not sure.

Thanks @mangelajo. Updated the PR.

sridhargaddam:
> May be also keep low the renew deadline?

I thought about it, but it might add more load to the already constrained API server. But if you prefer to have low values, I'm fine. Let me know.

sridhargaddam:
> > May be also keep low the renew deadline?
> 
> I thought about it, but it might add more load to the already constrained API server. But if you prefer to have low values, I'm fine. Let me know.

I had a quick look at leaderelection values and I think it makes sense to lower the values.

Issue #555, opened 2020-05-08
Running make e2e on OSX fails with make path issue


mangelajo:
```
[submariner]$ trap chown -R 501:20 . exit
[submariner]$ mkdir -p bin dist output
[submariner]$ /Applications/Xcode.app/Contents/Developer/usr/bin/make e2e
./scripts/entry: line 13: /Applications/Xcode.app/Contents/Developer/usr/bin/make: No such file or directory
[submariner]$ /Applications/Xcode.app/Contents/Developer/usr/bin/make e2e

```

This is because we pass down $(MAKE) to be ran into dapper. We need to use make.

PR #556, opened 2020-05-08
Don't pass the make binary path down to dapper

mangelajo:
The host could be different from linux or have a weird path for make.

Fixes: #555

### DIFF
--- a/Makefile.dapper
+++ b/Makefile.dapper
@@ -11,7 +11,7 @@
 	@mv .dapper.tmp .dapper
 
 %: .dapper
-	./.dapper -m bind make $@ $(MAKEFLAGS)
+	+./.dapper -m bind make $@ $(MAKEFLAGS)
 
 shell: .dapper
 	./.dapper -m bind -s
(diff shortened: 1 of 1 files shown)


Makefile.dapper
```diff
@@ -11,7 +11,7 @@
 	@mv .dapper.tmp .dapper
 
 %: .dapper
-	./.dapper -m bind $(MAKE) $@ $(MAKEFLAGS)
+	./.dapper -m bind make $@ $(MAKEFLAGS)
```
skitt, line 14:
```suggestion
	+./.dapper -m bind make $@ $(MAKEFLAGS)
```

The `+` is necessary if we don’t use the `MAKE` variable, so that `make -n` and the like will continue working. See [How the `MAKE` Variable Works](https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html) for details.

mangelajo: What's that black + magic? :)

mangelajo:
I'm testing with "+" on OSX.

mangelajo:
it works on osx

Issue #552, opened 2020-05-07
WireGuard doesn't seem to be working with globalnet


mkolesnik:
Running in this combination, it seems the E2E tests are failing

See also https://github.com/submariner-io/submariner/pull/551/checks?check_run_id=652184597

The logs are attached as well - 
[logs_13.zip](https://github.com/submariner-io/submariner/files/4591708/logs_13.zip)

mangelajo:
ping @deanlorenz @roytman  ^

deanlorenz:
ACK. Will look into it.
Probably has to do with finding the local interface ip -- with globalnet there is no local interface with a global IP, so no interface with an allowed IP.

deanlorenz:
[c3.3.log](https://github.com/submariner-io/submariner/files/4608701/c3.3.log)
driver is being called with empty public key. 
remove EP called before init

deanlorenz:
[e2e-tcpdump-dp-gn.log](https://github.com/submariner-io/submariner/files/4612279/e2e-tcpdump-dp-gn.log)
[e2e-tests-dp-gn.log](https://github.com/submariner-io/submariner/files/4612280/e2e-tests-dp-gn.log)

- e2e tests with focus=dataplane-globalnet
- 5 of 6 tests pass, WG config and routes look as expected
- Failed test does not get SNAT to GN so gets dropped and times-out -- *seems to be a problem in GN IPT*
- redundancy test also causes GN-DP failures (see log in previous comment) -- engine breaks the cable driver with corrupt endpoint entries

mangelajo:
The issue was fixed by https://github.com/submariner-io/submariner/pull/562

PR #562, opened 2020-05-13
Wireguard: better localIP discovery search

deanlorenz:
**Add the Cluster Spec as input during driver creation:**
- Cross-cluster traffic from pods on host network are assigned the source address of the WG interface, according to the routing rules; therefore, this should be an allowed IP
- The driver looks for a valid address within the local node interfaces; however, with GN, the allowed IPs (`localSubnets`) passed to driver are GN CIDRs and do not match any local interface IP
- Cluster CIDR is now used to look for a valid IP if not found in `localSubnets`
- GN will SNAT this address to a the GN CIDR, which is an allowed IP for the WG tunnel

**Misc**
- For better robustness, a FW mark was added to WG interface to block any circular routing, in case there is a routing rule back to WG device 
- More debug logging.

### DIFF
--- a/pkg/cable/driver.go
+++ b/pkg/cable/driver.go
@@ -33,7 +33,7 @@ type Driver interface {
 }
 
 // Function prototype to create a new driver
-type DriverCreateFunc func(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (Driver, error)
+type DriverCreateFunc func(localSubnets []string, localEndpoint types.SubmarinerEndpoint, localCluster types.SubmarinerCluster) (Driver, error)
 
 // Static map of supported drivers
 var drivers = map[string]DriverCreateFunc{}
@@ -50,12 +50,12 @@ func AddDriver(name string, driverCreate DriverCreateFunc) {
 }
 
 // Returns a new driver according the required Backend
-func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (Driver, error) {
+func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint, localCluster types.SubmarinerCluster) (Driver, error) {
 	driverCreate, ok := drivers[localEndpoint.Spec.Backend]
 	if !ok {
 		return nil, fmt.Errorf("unsupported cable type %s", localEndpoint.Spec.Backend)
 	}
-	return driverCreate(localSubnets, localEndpoint)
+	return driverCreate(localSubnets, localEndpoint, localCluster)
 }
 
 // Sets the default cable driver name, if it is not specified by user.
--- a/pkg/cable/wireguard/driver.go
+++ b/pkg/cable/wireguard/driver.go
@@ -133,7 +133,7 @@ func (w *wireguard) Init() error {
 	}
 	d, err := w.client.Device(DefaultDeviceName)
 	if err != nil {
-		return fmt.Errorf("wgctrl cannot find WireGaurd device: %v", err)
+		return fmt.Errorf("wgctrl cannot find WireGuard device: %v", err)
 	}
 	k, err := keyFromSpec(&w.localEndpoint.Spec)
 	if err != nil {
@@ -245,7 +245,11 @@ func (w *wireguard) ConnectToEndpoint(remoteEndpoint types.SubmarinerEndpoint) (
 			Table:     routingTable,
 		}
 		if err = netlink.RouteAdd(&route); err != nil {
-			return "", fmt.Errorf("failed to add route %s: %v", route, err)
+			klog.V(log.DEBUG).Infof("adding route %s", route)
+			if !os.IsExist(err) {
+				return "", fmt.Errorf("failed to add route %s: %v", route, err)
+			}
+			klog.V(log.DEBUG).Infof("route already exists")
 		}
 	}
 
--- a/pkg/cable/wireguard/driver.go
+++ b/pkg/cable/wireguard/driver.go
@@ -7,14 +7,17 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/sys/unix"
+
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 
 	"github.com/submariner-io/submariner/pkg/cable"
 	"github.com/submariner-io/submariner/pkg/log"
+	"github.com/submariner-io/submariner/pkg/types"
 	"github.com/vishvananda/netlink"
+
 	"k8s.io/klog"
 
-	"github.com/submariner-io/submariner/pkg/types"
 	"golang.zx2c4.com/wireguard/wgctrl"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 )
@@ -37,6 +40,7 @@ const (
 
 	//TODO generalize cleanStrongswanRoutingTable, for now, must use 220
 	routingTable = 220
+	fwMark       = 5871
 
 	cableDriverName = "wireguard"
 	receiveBytes    = "ReceiveBytes"  // for peer connection status
@@ -49,7 +53,6 @@ func init() {
 }
 
 type wireguard struct {
-	localSubnets  []net.IPNet
 	localEndpoint types.SubmarinerEndpoint
 	connections   map[string]*v1.Connection // clusterID -> remote ep connection
 	mutex         sync.Mutex
@@ -58,7 +61,7 @@ type wireguard struct {
 }
 
 // NewDriver creates a new WireGuard driver
-func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (cable.Driver, error) {
+func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint, localCluster types.SubmarinerCluster) (cable.Driver, error) {
 	var err error
 
 	w := wireguard{
@@ -66,7 +69,7 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
 		localEndpoint: localEndpoint,
 	}
 
-	if err = w.setWGLink(localSubnets); err != nil {
+	if err = w.setWGLink(localSubnets, localCluster); err != nil {
 		return nil, fmt.Errorf("failed to setup WireGuard link: %v", err)
 	}
 
@@ -111,15 +114,42 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
 		return nil, fmt.Errorf("failed to configure WireGuard device: %v", err)
 	}
 
-	klog.V(log.DEBUG).Infof("Initialized WireGuard %s with publicKey %s", DefaultDeviceName, pub)
+	klog.V(log.DEBUG).Infof("Created WireGuard %s with publicKey %s", DefaultDeviceName, pub)
 	return &w, nil
 }
 
 func (w *wireguard) Init() error {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	klog.V(log.DEBUG).Infof("Initializing WireGuard device for cluster %s", w.localEndpoint.Spec.ClusterID)
+
+	if len(w.connections) != 0 {
+		return fmt.Errorf("cannot initialize with existing connections: %+v", w.connections)
+	}
+	a, err := netlink.AddrList(w.link, unix.AF_INET)
+	if err != nil || len(a) == 0 {
+		return fmt.Errorf("cannot get WireGuard link address: %v", err)
+	}
+	d, err := w.client.Device(DefaultDeviceName)
+	if err != nil {
+		return fmt.Errorf("wgctrl cannot find WireGaurd device: %v", err)
+	}
+	k, err := keyFromSpec(&w.localEndpoint.Spec)
+	if err != nil {
+		return fmt.Errorf("endpoint is missing public key %s: %v", d.PublicKey, err)
+	}
+	if k.String() != d.PublicKey.String() {
+		return fmt.Errorf("endpoint public key %s is different from device key %s", k, d.PublicKey)
+	}
+
 	// ip link set $DefaultDeviceName up
 	if err := netlink.LinkSetUp(w.link); err != nil {
 		return fmt.Errorf("failed to bring up WireGuard device: %v", err)
 	}
+
+	klog.V(log.DEBUG).Infof("WireGuard device %s, is up on %s, listening on port :%d, with key %s",
+		w.link.Attrs().Name, a[0], d.ListenPort, d.PublicKey)
 	return nil
 }
 
@@ -288,7 +318,7 @@ func (w *wireguard) GetActiveConnections(clusterID string) ([]string, error) {
 }
 
 // Create new wg link and assign addr from local subnets
-func (w *wireguard) setWGLink(localSubnets []string) error {
+func (w *wireguard) setWGLink(localSubnets []string, localCluster types.SubmarinerCluster) error {
 	// create routing table
 	if err := setRoutingTable(); err != nil {
 		return fmt.Errorf("failed to create routing table: %v", err)
@@ -315,9 +345,7 @@ func (w *wireguard) setWGLink(localSubnets []string) error {
 		return fmt.Errorf("failed to add WireGuard device: %v", err)
 	}
 
-	// parse localSubnets and get internal address
-	w.localSubnets = parseSubnets(localSubnets)
-	ip, err := discoverInternalIP(w.localSubnets)
+	ip, err := w.discoverInternalIP(localSubnets, localCluster)
 	if err != nil {
 		klog.Errorf("Error while attempting to discover internal IP: %v", err)
 	}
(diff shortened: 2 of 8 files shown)


deanlorenz:
#552

deanlorenz:
The missing public key issue is not fixed yet.
It does not occur when testing after cleanup. 
The problem seems to be related to synchronizing a newly created local EP before the driver is initialized -- the updates to backendConfig are not propagated to the EP resource.

pkg/cable/driver.go
```diff
@@ -50,12 +50,12 @@ func AddDriver(name string, driverCreate DriverCreateFunc) {
 }
 
 // Returns a new driver according the required Backend
-func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (Driver, error) {
-	driverCreate, ok := drivers[localEndpoint.Spec.Backend]
+func NewDriver(subnets []string, endpoint types.SubmarinerEndpoint, cluster types.SubmarinerCluster) (Driver, error) {
```
mangelajo, line 53: nit: can we keep this localSubnets / localEndpoints like in the drivers so they are uniform?
```suggestion
func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint, cluster types.SubmarinerCluster) (Driver, error) {
```

pkg/cable/driver.go
```diff
@@ -50,12 +50,12 @@ func AddDriver(name string, driverCreate DriverCreateFunc) {
...
+	driverCreate, ok := drivers[endpoint.Spec.Backend]
```
mangelajo, line 54:
```suggestion
	driverCreate, ok := drivers[localEndpoint.Spec.Backend]
```

pkg/cable/driver.go
```diff
@@ -50,12 +50,12 @@ func AddDriver(name string, driverCreate DriverCreateFunc) {
...
 	if !ok {
-		return nil, fmt.Errorf("unsupported cable type %s", localEndpoint.Spec.Backend)
+		return nil, fmt.Errorf("unsupported cable type %s", endpoint.Spec.Backend)
```
mangelajo, line 56:
```suggestion
		return nil, fmt.Errorf("unsupported cable type %s", localEndpoint.Spec.Backend)
```

pkg/cable/driver.go
```diff
@@ -50,12 +50,12 @@ func AddDriver(name string, driverCreate DriverCreateFunc) {
...
 	}
-	return driverCreate(localSubnets, localEndpoint)
+	return driverCreate(subnets, endpoint, cluster)
```
mangelajo, line 58:
```suggestion
	return driverCreate(localSubnets, localEndpoint, cluster)
```

pkg/cable/wireguard/driver.go
```diff
@@ -58,15 +61,15 @@ type wireguard struct {
 }
 
 // NewDriver creates a new WireGuard driver
-func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (cable.Driver, error) {
+func NewDriver(subnets []string, endpoint types.SubmarinerEndpoint, cluster types.SubmarinerCluster) (cable.Driver, error) {
```
mangelajo, line 64: localSubnets/localEndpoint?
```suggestion
func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint, cluster types.SubmarinerCluster) (cable.Driver, error) {
```

pkg/cable/wireguard/driver.go
```diff
@@ -58,15 +61,15 @@ type wireguard struct {
...
 	var err error
 
 	w := wireguard{
 		connections:   make(map[string]*v1.Connection),
-		localEndpoint: localEndpoint,
+		localEndpoint: endpoint,
```
mangelajo, line 69:
```suggestion
		localEndpoint: localEndpoint,
```

pkg/cable/wireguard/driver.go
```diff
@@ -58,15 +61,15 @@ type wireguard struct {
...
 	}
 
-	if err = w.setWGLink(localSubnets); err != nil {
+	if err = w.setWGLink(subnets, cluster); err != nil {
```
mangelajo, line 72:
```suggestion
	if err = w.setWGLink(localSubnets, cluster); err != nil {
```

pkg/cable/wireguard/driver.go
```diff
@@ -92,10 +95,10 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
 		return nil, fmt.Errorf("error generating private key: %v", err)
 	}
 	pub = priv.PublicKey()
-	if localEndpoint.Spec.BackendConfig == nil {
-		localEndpoint.Spec.BackendConfig = make(map[string]string)
+	if endpoint.Spec.BackendConfig == nil {
```
mangelajo, line 98:
```suggestion
	if endpoint.Spec.BackendConfig == nil {
```
```suggestion
	if localEndpoint.Spec.BackendConfig == nil {
```

pkg/cable/wireguard/driver.go
```diff
@@ -92,10 +95,10 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
...
+		endpoint.Spec.BackendConfig = make(map[string]string)
```
mangelajo, line 99:
```suggestion
		localEndpoint.Spec.BackendConfig = make(map[string]string)
```

pkg/cable/wireguard/driver.go
```diff
@@ -92,10 +95,10 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
...
 	}
-	localEndpoint.Spec.BackendConfig[PublicKey] = pub.String()
+	endpoint.Spec.BackendConfig[PublicKey] = pub.String()
```
mangelajo, line 101:
```suggestion
	localEndpoint.Spec.BackendConfig[PublicKey] = pub.String()
```

pkg/cable/wireguard/driver.go
deanlorenz: DONE

pkg/cable/wireguard/driver.go
deanlorenz: DONE

pkg/cable/wireguard/driver.go
deanlorenz: DONE

pkg/cable/driver.go
deanlorenz: DONE

pkg/cable/driver.go
deanlorenz: DONE

pkg/cable/driver.go
deanlorenz: DONE

pkg/cable/driver.go
deanlorenz: DONE

pkg/cable/wireguard/driver.go
deanlorenz: DONE

pkg/cable/wireguard/driver.go
deanlorenz: DONE

pkg/cable/wireguard/driver.go
deanlorenz: DONE

deanlorenz:
@sridhargaddam @mangelajo Changed the commit message. Hopefully, it is clearer now

pkg/cable/wireguard/driver.go
```diff
@@ -111,15 +114,42 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
...
+	if len(w.connections) != 0 {
+		return fmt.Errorf("cannot initialize with existing connections: %+v", w.connections)
+	}
+	a, err := netlink.AddrList(w.link, unix.AF_INET)
+	if err != nil || len(a) == 0 {
+		return fmt.Errorf("cannot get WireGuard link address: %v", err)
+	}
+	d, err := w.client.Device(DefaultDeviceName)
+	if err != nil {
+		return fmt.Errorf("wgctrl cannot find WireGaurd device: %v", err)
```
sridhargaddam, line 136: nit: Wireguard

pkg/cable/wireguard/driver.go
```diff
@@ -111,15 +114,42 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
...
-	klog.V(log.DEBUG).Infof("Initialized WireGuard %s with publicKey %s", DefaultDeviceName, pub)
+	klog.V(log.DEBUG).Infof("Created WireGuard %s with publicKey %s", DefaultDeviceName, pub)
 	return &w, nil
 }
 
 func (w *wireguard) Init() error {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	klog.V(log.DEBUG).Infof("Initializing WireGuard device for cluster %s", w.localEndpoint.Spec.ClusterID)
```
sridhargaddam, line 125: With IPsec, we have seen that sometimes the corporate edge routers/firewall configuration blocks the default UDP ports 4500/501. So, we are supporting non-standard ports during subctl join operation. In wireguard, I see that you are using DefaultListenPort (aka 5871). If we encounter such a situation with wireguard, does wireguard support listening/connecting on user-defined ports?

pkg/cable/wireguard/driver.go
```diff
@@ -215,7 +245,11 @@ func (w *wireguard) ConnectToEndpoint(remoteEndpoint types.SubmarinerEndpoint) (
 			Table:     routingTable,
 		}
 		if err = netlink.RouteAdd(&route); err != nil {
-			return "", fmt.Errorf("failed to add route %s: %v", route, err)
+			klog.V(log.DEBUG).Infof("adding route %s", route)
+			if !os.IsExist(err) {
+				return "", fmt.Errorf("failed to add route %s: %v", route, err)
+			}
+			klog.V(log.DEBUG).Infof("route already exists")
```
sridhargaddam, line 252: It would be better to add some additional info about the route (like peerNetwork). Will help us during debugging when multiple things happen in parallel.

pkg/cable/wireguard/driver.go
```diff
@@ -111,15 +114,42 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
...
+	}
+	a, err := netlink.AddrList(w.link, unix.AF_INET)
+	if err != nil || len(a) == 0 {
+		return fmt.Errorf("cannot get WireGuard link address: %v", err)
+	}
+	d, err := w.client.Device(DefaultDeviceName)
+	if err != nil {
+		return fmt.Errorf("wgctrl cannot find WireGaurd device: %v", err)
+	}
+	k, err := keyFromSpec(&w.localEndpoint.Spec)
```
sridhargaddam, line 138: A question just for my understanding: After wireguard successfully sets up the tunnels to remote clusters, if the Submariner Gateway Pod is restarted, will the tunnel be destroyed as soon as GatewayPod is terminated?

pkg/cable/wireguard/driver.go
```diff
@@ -370,6 +402,27 @@ func discoverInternalIP(cidrs []net.IPNet) (string, error) {
 	return "", nil
 }
 
+// look at local subnets, cluster cidrs, and vxlan subnets
+func (w *wireguard) discoverInternalIP(localSubnets []string, localCluster types.SubmarinerCluster) (string, error) {
+	cidrs := parseSubnets(localSubnets)
+	ip, err := matchLocalIP(cidrs)
```
sridhargaddam, line 408: Can we use the discoverCNIInterface API here by moving it to utils?

pkg/cable/wireguard/driver.go
```diff
@@ -370,6 +402,27 @@ func discoverInternalIP(cidrs []net.IPNet) (string, error) {
...
+	if ip != "" {
+		return ip, nil
+	}
+	if err != nil {
+		return "", err
+	}
+
+	// no match, fallback to cluster cidr
```
sridhargaddam, line 416: Can you please explain what we are trying to do here?

dfarrell07:
Thanks for the patch @deanlorenz! Can we [squash the commits](https://submariner-io.github.io/contributing/code-review/) and refactor the resulting commit message to follow [our/K8s guidelines](https://submariner-io.github.io/contributing/code-review/)?

pkg/cable/wireguard/driver.go
```diff
@@ -37,6 +40,7 @@ const (
 
 	//TODO generalize cleanStrongswanRoutingTable, for now, must use 220
 	routingTable = 220
+	fwMark       = 5871
```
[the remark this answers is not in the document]
deanlorenz: any unused mark should be good.

pkg/cable/wireguard/driver.go
deanlorenz: WG can use any UDP port for each endpoint

pkg/cable/wireguard/driver.go
deanlorenz: done

pkg/cable/wireguard/driver.go
```diff
@@ -111,15 +114,42 @@ func NewDriver(localSubnets []string, localEndpoint types.SubmarinerEndpoint) (c
...
 func (w *wireguard) Init() error {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	klog.V(log.DEBUG).Infof("Initializing WireGuard device for cluster %s", w.localEndpoint.Spec.ClusterID)
+
+	if len(w.connections) != 0 {
+		return fmt.Errorf("cannot initialize with existing connections: %+v", w.connections)
+	}
+	a, err := netlink.AddrList(w.link, unix.AF_INET)
```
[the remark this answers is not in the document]
deanlorenz: Everything already setup before in `NewDriver()`
`Init()` only does ifup (+sanity checks, some debug info)
address is as describe on top commit comment

pkg/cable/wireguard/driver.go
deanlorenz: it is logged on the previous line

pkg/cable/wireguard/driver.go
deanlorenz: - WG device remains, but UDP tunnels will timeout and fade away.
- The routes will be removed by RA -- it flushes table 220. 
- Once the other GW comes up, all peers will have the allowed IPs for this cluster pointing only to the new GW -- so no traffic from old GW would be accepted. 
- If a new gateway is restarted on the same node then WG device is destroyed and recreated.

(longer term it may make sense to create a pluggable fwk to call cable libraries for cleanup, by RA or perhaps operator)

pkg/cable/wireguard/driver.go
deanlorenz: +1 for having `discoverCNIInterface` in utils. This function is a bit different -- it looks first for a `localSubnet` CIDR and then for cluster CIDR. It uses `matchLocalIP()` which is perhaps more similar. 
As we discussed with @mangelajo, longer term, the local tunnel address (both GN and non-GN cases) should ideally be decided before the cable driver is started and pass it as parameter. This would allow all iptable rules to use the address (eg for GN SNAT). Also, we should consider using `ip route get <some cluster IP>` to figure out the source address used by CNI.

pkg/cable/wireguard/driver.go
deanlorenz: See PR commit comment.

pkg/cable/wireguard/driver.go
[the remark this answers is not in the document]
sridhargaddam: Ok, thanks.

pkg/cable/wireguard/driver.go
sridhargaddam: So you can probably think of adding support for specifying custom port in future for WG, similar to IPsec.

pkg/cable/wireguard/driver.go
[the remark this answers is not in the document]
sridhargaddam: Okay, I see that you are doing it in Line 352.

Please note that we will not assign an IPaddress from the GlobalCIDR to any of the Linux interfaces. 
The globalCIDR (and the associated IPs) are virtual. I have a feeling that the initial search you are doing in API (discoverInternalIP) for localsubnets (which will be globalCIDR when Globalnet is used and a list of Pod/Service CIDRs with Vanilla Submariner) can be removed and the API (discoverInternalIP) can simply search for ClusterCIDR. I'm not much familiar with Wireguard, so you can decide what is correct.

pkg/cable/wireguard/driver.go
sridhargaddam: > * WG device remains, but UDP tunnels will timeout and fade away.
> * The routes will be removed by RA -- it flushes table 220.
> * Once the other GW comes up, all peers will have the allowed IPs for this cluster pointing only to the new GW -- so no traffic from old GW would be accepted.
> * If a new gateway is restarted on the same node then WG device is destroyed and recreated.

Okay, the behavior is same as IPsec for now where datapath disruption would happen. BTW, we are trying to avoid this with IPsec and I'm looking into it.
> 
> (longer term it may make sense to create a pluggable fwk to call cable libraries for cleanup, by RA or perhaps operator)

+1

pkg/cable/wireguard/driver.go
sridhargaddam: Yup. We can do any such changes in a separate PR.

pkg/cable/wireguard/driver.go
sridhargaddam: Ack, as I mentioned above, I have a feeling that you can all-together remove the search via localsubnets and simply search for an interface that has IPAddress from clusterCIDR.

PR #565, opened 2020-05-14
Use E2E from Shipyard

mkolesnik:
Since it's now available, use the E2E script from Shipyard.

Also eliminated kind-e2e folder since it's no longer needed, and folded
the sections from the guide into the testing guide (which needs to be
ported to the website, but thats out of scope for this change).


dfarrell07:
Flaky failure on one job. Captured the details here: https://github.com/submariner-io/submariner/issues/532#issuecomment-628685503

docs/testing.md
```diff
@@ -164,8 +171,54 @@ the `deploytool` flag to the standard `make` commands.
...
+make reload-images e2e
+```
+
+You can focus the e2e tests on specific tags by using the focus makefile env. The following
+example runs only the tests tagged as [redundancy].
+
+```bash
+make e2e focus=redundancy
+```
+
```
mangelajo, line 219: Probably we want to remove this and link to the website.

mkolesnik: Yeah I wrote regarding that in the original commit message "which needs to be
ported to the website, but thats out of scope for this change"

PR #569, opened 2020-05-15
Fix cabel-driver setting for helm deployments

roytman:


mkolesnik:
Hi Alexey, the omission of helm testing was intentional since we don't want a huge matrix of tests. Even though GH Actions has more parallel slots (20 vs 5 in Travis), we still need to keep a reasonable amount of jobs.

I suggest that we can add one Helm + WG combination if you feel testing with Helm is of importance (as we already test WG with Operator).

roytman:
Hi Mike, no problem, the WG helm deployment is not important, and can be put off. 
But this PR fixes the helm deployment. 
The parameter `'--cable-driver ${{ matrix.cable_driver }}'` format is wrong for helm, it should be `'--set cable-driver=${{ matrix.cable_driver }}'`

.github/workflows/e2e.yaml
```diff
@@ -49,7 +49,8 @@ jobs:
           CLUSTERS_ARGS: ${{ matrix.globalnet }}
           DEPLOY_ARGS: ${{ matrix.globalnet }} --deploytool ${{ matrix.deploytool }}
         run: |
-          [[ -z "${{ matrix.cable_driver }}" ]] || DEPLOY_ARGS+=" --deploytool_submariner_args '--cable-driver ${{ matrix.cable_driver }}'"
+          [[ -z "${{ matrix.cable_driver }}" ]] || [[ ${{ matrix.deploytool }} == "helm" ]] || DEPLOY_ARGS+=" --deploytool_submariner_args '--cable-driver ${{ matrix.cable_driver }}'"
+          [[ -z "${{ matrix.cable_driver }}" ]] || [[ ${{ matrix.deploytool }} == "operator" ]] || DEPLOY_ARGS+=" --deploytool_submariner_args '--set cable-driver=${{ matrix.cable_driver }}'"
```
mangelajo, line 53: The conditions are confusing the way the are written, I know about boolean logic but it's hard for humans :D, I propose something like this
```suggestion
         if [[ -z "${{ matrix.cable_driver }}" ]]; then
              [[ ${{ matrix.deploytool }} == "operator" ]] && DEPLOY_ARGS+=" --deploytool_submariner_args '--cable-driver ${{ matrix.cable_driver }}'"
              [[ ${{ matrix.deploytool }} == "helm" ]] && DEPLOY_ARGS+=" --deploytool_submariner_args '--set cable-driver=${{ matrix.cable_driver }}'"
         fi 
```

mangelajo: Even better with a case/switch in bash?

roytman: as I remember I tried it, and it did work, let's see now.

mkolesnik:
I'm addressing this as part of #613 I suggest we continue with that one (also supports libreswan) and close this one

roytman:
based on @mkolesnik [comment](https://github.com/submariner-io/submariner/pull/569#pullrequestreview-423542148), I'm closing the PR.

Issue #545, opened 2020-05-05
Detect overlapping CIDRs in Submariner


sridhargaddam:
Currently overlapping global/Pod/Service CIDRs validation for clusters is done in subctl, but when clusters are deployed with helm we do not have any mechanism to alert the user.
We can enhance Submariner to detect this overlap and fail fast with an appropriate error message.

mangelajo:
I think this is more  to go on the operator.

sridhargaddam:
> I think this is more to go on the operator.

No @mangelajo. The idea is to handle this in the Submariner Globalnet Code itself. If GN Controller detects that globalCIDRs of multiple clusters are overlapping, it can return FATAL and fail early.

mangelajo:
Please don’t fail fatal :-) may be we can mark the new overlapping cluster and use some sort of reporting to make it evident.

Let’s think of production clusters . If the admin makes a mistake, let’s try to report and minimize the problems. If we fail fatal we will kill the GN connectivity of new spawning pods. 

I’d try to detect and mark the new cluster that overlaps. The use some mean of reporting this to make it evident.

Enviado desde mi iPhone

sridhargaddam:
> Please don’t fail fatal :-) may be we can mark the new overlapping cluster and use some sort of reporting to make it evident.

Ack. What is the best place to report this issue? Gateway Status?

> Let’s think of production clusters . If the admin makes a mistake, let’s try to report and minimize the problems. If we fail fatal we will kill the GN connectivity of new spawning pods.

This detection happens when the GN Pod (or Submariner Gateway Pod?) is booting up in **a newly joined cluster**. So, what do you mean by "we will kill the GN connectivity of new spawning pods"? We cannot support GN globalIP use-case if the globalCIDRs overlap across the cluster. 

> I’d try to detect and mark the new cluster that overlaps. The use some mean of reporting this to make it evident. 

Sure, if you have any ideas/suggestions on where we can report this, we can do the necessary changes instead of returning FATAL.

Enviado desde mi iPhone
> […](#)
> El 5 may 2020, a las 16:41, Sridhar Gaddam ***@***.***> escribió: ﻿ I think this is more to go on the operator. No @mangelajo. The idea is to handle this in the Submariner Globalnet Code itself. If GN Controller detects that globalCIDRs of multiple clusters are overlapping, it can return FATAL and fail early. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

[...]

PR #570, opened 2020-05-15
Validate CIDRs before making changes to the Cluster

sridhargaddam:
Submariner supports clusters with Overlapping CIDRs (via Globalnet) as well as
clusters with non-overlapping CIDRs. However, if administrator accidentally tries
deploying submariner with wrong configuration, Submariner should log a warning
message and avoid making changes in the local cluster which otherwise can disrupt
the functionality of the local cluster.

This PR specifically validates the following and logs an appropriate warning.

1. In a Vanilla Submariner deployment, if administrator is trying to join clusters
   with Overlapping Pod/Service CIDRs.
   Note: For such clusters, Submariner Globalnet should be used.

2. In a Globalnet deployment, if GlobalCIDRs are overlapping across the clusters.

3. In a Globalnet deployment, if localCluster Pod/Service CIDRs overlap with globalCIDR
   of other clusters.

Fixes Issue# https://github.com/submariner-io/submariner/issues/545

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -200,7 +200,10 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
 	// Program iptables rules for traffic destined to all the remote cluster CIDRs
 	for _, endpoint := range endpoints.Items {
 		if endpoint.Spec.ClusterID != r.clusterID {
-			r.checkOverlappingSubnets(endpoint.Spec.Subnets)
+			if r.overlappingSubnets(endpoint.Spec.Subnets) {
+				// Skip processing the endpoint when CIDRs overlap
+				continue
+			}
 			r.updateIptableRulesForInterclusterTraffic(endpoint.Spec.Subnets)
 		}
 	}
@@ -261,7 +264,7 @@ func (r *Controller) runPodWorker() {
 	}
 }
 
-func (r *Controller) checkOverlappingSubnets(remoteSubnets []string) {
+func (r *Controller) overlappingSubnets(remoteSubnets []string) bool {
 	// If the remoteSubnets [*] overlap with local cluster Pod/Service CIDRs we
 	// should not update the IPTable rules on the host, as it will disrupt the
 	// functionality of the local cluster. So, lets validate that subnets do not
@@ -270,7 +273,6 @@ func (r *Controller) checkOverlappingSubnets(remoteSubnets []string) {
 	// [*] Note: In a non-GlobalNet deployment, remoteSubnets will be a list of
 	// Pod/Service CIDRs, whereas in a GlobalNet deployment, it will be a list of
 	// globalCIDRs allocated to the clusters.
-
 	for _, serviceCidr := range r.localServiceCidr {
 		overlap, err := util.IsOverlappingCIDR(remoteSubnets, serviceCidr)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -260,6 +264,40 @@ func (r *Controller) runPodWorker() {
...
+		}
+	}
+
+	for _, podCidr := range r.localClusterCidr {
+		overlap, err := util.IsOverlappingCIDR(remoteSubnets, podCidr)
+		if err != nil {
+			klog.Warningf("unable to validate overlapping Pod CIDR: %s", err)
+		}
+		if overlap {
+			klog.Errorf("Local Pod CIDR %q, overlaps with remote cluster %s", podCidr, err)
```
vthapar, line 294: Will we hit this in case of globalnet?

pkg/routeagent/controllers/route/route.go
```diff
@@ -260,6 +264,40 @@ func (r *Controller) runPodWorker() {
 	}
 }
 
+func (r *Controller) overlappingSubnets(remoteSubnets []string) bool {
```
vthapar, line 267: This looks like an operation that would get costly with scale, so hoping this code is only hit when we're coming up.

pkg/routeagent/controllers/route/route.go
```diff
@@ -260,6 +264,40 @@ func (r *Controller) runPodWorker() {
...
+	// Pod/Service CIDRs, whereas in a GlobalNet deployment, it will be a list of
+	// globalCIDRs allocated to the clusters.
+	for _, serviceCidr := range r.localServiceCidr {
+		overlap, err := util.IsOverlappingCIDR(remoteSubnets, serviceCidr)
+		if err != nil {
+			// Ideally this case will never hit, as the subnets are valid CIDRs
+			klog.Warningf("unable to validate overlapping Service CIDR: %s", err)
+		}
+		if overlap {
+			klog.Errorf("Local Service CIDR %q, overlaps with remote cluster %s", serviceCidr, err)
```
vthapar, line 283: Would be good to add something about enabling globalnet to enable such a use case.

pkg/routeagent/controllers/route/route.go
sridhargaddam: This will be validated when the Submariner route agent is coming up and also at runtime when a new cluster is added.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Although we are currently logging this message, there is a separate issue to report the status via Operator (or for user to consume). I feel that the current log message is good enough because we would also have the following situation (an example)

We have three clusters - all non-overlapping CIDRs and user accidentally tries to join a fourth cluster that overlaps with one of the existing clusters. The objective of this PR is to avoid making changes to local cluster that disrupts the cluster connectivity. Whether user wants to delete submariner from all the existing clusters and then redeploy with GlobalNet or if he wants to update the CIDRs of fourth cluster (to avoid overlaps) is up to him.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Yes, this will validate the third scenario mentioned in the commit message.

"In a Globalnet deployment, if localCluster Pod/Service CIDRs overlap with globalCIDR
of other clusters."

Issue #579, opened 2020-05-19
Datastore interface in Submariner is ignoring the colorCodes and returns all the clusters


sridhargaddam:
**What happened**:
As per the datastore interface, WatchClusters and WatchEndpoints should only return the clusters/endpoints that we are interested in. However, in the current implementation [*], these APIs simply ignore the color code thereby returning all the clusters/endpoints.

[*] Current implementation of WatchClusters https://github.com/submariner-io/submariner/blob/1aba088eb78062f22850f847eea1befc187bb0b0/pkg/datastore/kubernetes/kubernetes.go#L109

type Datastore interface {
...
// Watches all clusters and calls the passed in function on cluster change
WatchClusters(ctx context.Context, selfClusterID string, colorCodes []string, onClusterChange OnClusterChange) error

// Performs a watch of all endpoints and calls the passed in function based on information
WatchEndpoints(ctx context.Context, selfClusterID string, colorCodes []string, onEndpointChange OnEndpointChange) error
...
}

**What you expected to happen**:
Ideally the interface should return only the clusters that match the respective colorCodes.

tpantelis:
I expect the `Datastore` interface to go away when we migrate to the admiral syncer API. We can implement color code support on top of that.

PR #581, opened 2020-05-20
WIP: Handle colorCodes in datastore interface

sridhargaddam:
As per the datastore interface, WatchClusters and WatchEndpoints should only return the
clusters/endpoints that we are interested in. However, in the current implementation,
these APIs simply ignore the color code thereby returning all the clusters/endpoints.
This PR fixes the issue.

Resolves Issue# https://github.com/submariner-io/submariner/issues/579

### DIFF
--- a/pkg/datastore/kubernetes/kubernetes.go
+++ b/pkg/datastore/kubernetes/kubernetes.go
@@ -106,6 +106,34 @@ func (k *Datastore) GetEndpoints(clusterID string) ([]types.SubmarinerEndpoint, 
 	return endpoints, nil
 }
 
+func (k *Datastore) GetClusters(colorCodes []string) ([]types.SubmarinerCluster, error) {
+	k8sClusters, err := k.client.SubmarinerV1().Clusters(k.remoteNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	clusters := []types.SubmarinerCluster{}
+
+	for _, cluster := range k8sClusters.Items {
+		for _, colorCode := range colorCodes {
+			if colorCodeMatches(colorCode, cluster.Spec.ColorCodes) {
+				clusters = append(clusters, types.SubmarinerCluster{Spec: cluster.Spec})
+				break
+			}
+		}
+	}
+	return clusters, nil
+}
+
+func colorCodeMatches(item string, colorCodes []string) bool {
+	for _, colorCode := range colorCodes {
+		if colorCode == item {
+			return true
+		}
+	}
+	return false
+}
+
 func (k *Datastore) WatchClusters(ctx context.Context, selfClusterID string, colorCodes []string, onClusterChange datastore.OnClusterChange) error {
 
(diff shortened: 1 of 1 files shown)


mangelajo:
Wow @sridhargaddam That was fast!! :)

@oats87 This came from an internal use case we were asked about, @sridhargaddam tested it and found it wasn't actually working.

If we find no urgency I propose we deprecate the feature (as it wasn't really working), and we design something that allows connectivity in the form of 1:N, which color codes don't allow.

I had something like ConnectivityPolicy in mind, something we can install in the broker to define connectivity in non all<->all, and even support hub & spoke models at some point.

Oats87:
@mangelajo I can definitely see the merit behind connectivity in the form of `1:N`, although I wonder if the concept of colorCodes could actually be enhanced, further moving the concept of color to something like `ColorLeader` or `ColorFollower`, where `ColorLeader` clusters will be connected to by `ColorFollower` clusters, and `ColorFollower` clusters won't connect to each other. This model would then allow for `m:n` connectivity, past `1:n`

Also, the names are of course subject to change. ConnectivityPolicy would be good as well, but I struggle to (in the current moment) think of a rule syntax that would be flexible enough to be useful while still being simple enough to use.

pkg/datastore/kubernetes/kubernetes.go
```diff
@@ -106,6 +106,34 @@ func (k *Datastore) GetEndpoints(clusterID string) ([]types.SubmarinerEndpoint,
...
+	if err != nil {
+		return nil, err
+	}
+
+	clusters := []types.SubmarinerCluster{}
+
+	for _, cluster := range k8sClusters.Items {
+		for _, colorCode := range colorCodes {
+			if colorCodeMatches(colorCode, cluster.Spec.ColorCodes) {
+				clusters = append(clusters, types.SubmarinerCluster{Spec: cluster.Spec})
```
Oats87, line 120: Is there a possibility to end up with duplicate clusters in the slice here?

sridhargaddam: I think no @Oats87. As we are using 'break' in the inner for loop once there is a match (i.e., we iterate over a cluster and when we find that it has atleast one color code, we add it to the list and stop checking further color codes). 

PS: This was a quick fix. There is some duplication which we can remove by moving them to APIs.

pkg/datastore/kubernetes/kubernetes.go
```diff
@@ -211,10 +254,27 @@ func (k *Datastore) WatchEndpoints(ctx context.Context, selfClusterID string, co
...
+				}
+
+				klog.V(log.DEBUG).Infof("Got endpoints : %#v", endpoints)
+				for _, endpoint := range endpoints {
+					if selfClusterID != endpoint.Spec.ClusterID {
+						utilruntime.HandleError(onEndpointChange(&types.SubmarinerEndpoint{
+							Spec: endpoint.Spec,
+						}, false))
+					}
+				}
```
tpantelis, line 277: Why do we notify `onEndpointChange` of `Endpoints` from all matching clusters instead of just the one that was added?

PR #586, opened 2020-05-21
temporary put off WG installation 

roytman:


.github/workflows/e2e.yaml
```diff
@@ -27,14 +27,6 @@ jobs:
...
         uses: actions/checkout@master
 
-      - name: Install WireGuard module (On WireGuard tests)
-        if: matrix.cable_driver == 'wireguard'
-        run: |
-          sudo add-apt-repository -y ppa:wireguard/wireguard # add WireGuard support
-          sudo apt-get update
-          sudo apt-get install linux-headers-`uname -r` wireguard -y
-          sudo modprobe wireguard
-
```
mkolesnik, line 37: There's actually no need to remove this since it'll only happen in case `cable_driver` has been set to `wireguard` which won't happen since you removed it from the matrix

roytman: Hi Mike, correct, returned it back. Locks like this PR should work together with your PR https://github.com/submariner-io/submariner/pull/587

mkolesnik:
Closing in favor of #589

PR #589, opened 2020-05-21
Temporary fix of the WireGuard module installation

roytman:


roytman:
The culprit is commit 2c64605b590e ("net: Fix CONFIG_NET_CLS_ACT=n and
CONFIG_NFT_FWD_NETDEV={y, m} build") in the kernel.
Apparently, Ubuntu picked it up for azure-1022, but it didn't made it to
azure-1020.
Unfortunately, GitHub uses Ubuntu 16.04 with kernel 5.3.0-1020-azure.
The provided h-file is a temporal workaroud for dkms until github/azure update their kernel .

queueing.h
```diff
@@ -0,0 +1,203 @@
+/* SPDX-License-Identifier: GPL-2.0 */
```
mkolesnik, line 1: Instead of adding this file in repo, I think it's better to download it from the WG repo during installation, e.g. fetch it from https://git.zx2c4.com/wireguard-linux/plain/drivers/net/wireguard/queueing.h

.github/workflows/e2e.yaml
```diff
@@ -31,8 +31,11 @@ jobs:
         if: matrix.cable_driver == 'wireguard'
         run: |
           sudo add-apt-repository -y ppa:wireguard/wireguard # add WireGuard support
-          sudo apt-get update
-          sudo apt-get install linux-headers-`uname -r` wireguard -y
+          sudo apt update
+          # temporary solution for Ubuntu 16.04 azure-1020
+          sudo apt-get install linux-headers-$(uname -r) wireguard  -y || true # because wireguard installation it will fail
+          sudo cp queueing.h  /usr/src/wireguard-1.0.20200506/queueing.h
+          sudo dpkg-reconfigure wireguard-dkms
```
mkolesnik, line 38:
```suggestion
          if ! sudo apt-get install linux-headers-$(uname -r) wireguard  -y; then
              sudo curl -L https://git.zx2c4.com/wireguard-linux/plain/drivers/net/wireguard/queueing.h -o /usr/src/wireguard-1.0.20200506/queueing.h
              sudo dpkg-reconfigure wireguard-dkms
          fi
```

Perhaps something like this?

queueing.h
roytman: Mike, thanks for the suggestion. Indeed it is better to download the file.

roytman:
@mangelajo , I squashed the commits, and changed the PR title.

PR #600, opened 2020-05-27
Release management proposal

mangelajo:


docs/release-management.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+## merge of the release yaml
+
+* A script will run in Github actions, executing step by step the tagging, and watching
+of the multiple components.
+
+It will retry builds if necessary.
+
+* For the operator repository, a new commit will be automatically pushed with the versions.go
```
mkolesnik, line 80: But then you're not using the commit-id you originally had, you'll have another commit id since it's a new patch..

mangelajo: Yeah, that is fine, we would know about it internally, and release on the new.

mangelajo: The problem would be if something merges in the meanwhile... but hopefully we can control ourselves not to merge stuff.. and we can add a verification that our versions patch is being applied on top of the requested commit id

mangelajo: But it's true, the release.yaml would not match with that...commit id, with it's itching, not sure if we can fix that somehow. May be inject those versions as env variables and have some defaults? The problem I see with that is that then.. if somebody tries to develop with subctl they need to use the right env vars... it would be tricky.

We could have subctl may be download the manifest?, or just assume that the versions are all equal to his own version by default?

mkolesnik: I think a more sensible approach would be to bump the versions POST the release, so that the master always had the next version (as is done in java maven builds, for example).

This way the hashes dont change and overall the flow is a bit simpler I guess..

mangelajo: Hmm, that'd make sense, I like it, I will update the document.

mkolesnik: Cool, in fact the automated release process can bump these version and propose that as a PR to the necessary places :)

docs/release-management.md
```diff
@@ -0,0 +1,86 @@
+# Proposal for release management in the submariner-io project
+
+## Introduction
+
+Releasing submariner contains numerous manual [steps](https://submariner-io.github.io/contributing/release-process/)
+that need to be executed carefully in order accross multiple projects,
+waiting for CI executions to properly happen.
```
tpantelis, line 7:
```suggestion
Releasing submariner consists of several manual [steps](https://submariner-io.github.io/contributing/release-process/)
that need to be carefully executed in order across multiple projects.
```

docs/release-management.md
```diff
@@ -0,0 +1,86 @@
...
+
+All this process could be automated as described in the following sections.
```
tpantelis, line 9:
```suggestion
This process can be automated as described in the following sections.
```

docs/release-management.md
```diff
@@ -0,0 +1,86 @@
...
+
+## What's a release
+
+   A submariner release is the sum of all components, tagged on a specific set of versions,
+   that works, and it's tested together.
```
tpantelis, line 14:
```suggestion
   that work and are tested together.
```

docs/release-management.md
```diff
@@ -0,0 +1,86 @@
...
+   * subctl
+   * submariner-operator
+   * submariner core (submariner engine, route agent, globalnet ipam)
+   * lighthouse (lighthouse agent, lighthouse coredns)
+   
+  
+## A release repository
+
+   A release repository (submariner-io/releases) could be used to track and automate
+   all the release process.
```
tpantelis, line 27:
```suggestion
   A release repository (submariner-io/releases) can be used to track and automate
   the whole release process.
```

docs/release-management.md
```diff
@@ -0,0 +1,86 @@
...
+   
+   Yaml files could be used inside this repository to control the definition of a release,
```
tpantelis, line 29:
```suggestion
   Yaml files in this repository ccan be used to control the definition of a release, 
```

docs/release-management.md
```diff
@@ -0,0 +1,86 @@
...
+   
+   ```
+      releases/0.5-<name>/0.5.0-<name>.yaml
+      releases/0.5-<name>/0.5.1-bugfix for....yaml
+      ...
+   ```
+
+## A release yaml
+
+   A release yaml could have the following format:
```
tpantelis, line 39:
```suggestion
   A release yaml can have the following format:
```

docs/release-management.md
```diff
@@ -0,0 +1,86 @@
...
+The verification step would:
+* Verify the format of the yaml file
+* Verify that passed CI runs exist for the commit-ids proposed for the components
+* Run E2E for the combination of commit-ids with helm & operator (with lighthouse enabled)
+
+
+## merge of the release yaml
+
+* A script will run in Github actions, executing step by step the tagging, and watching
+of the multiple components.
```
tpantelis, line 76:
```suggestion
* A script will run in Github actions, executing step by step the tagging and verification
of the multiple components.
```

docs/release-management.md
mangelajo: I'll update and let's review

docs/release-management.md
```diff
@@ -0,0 +1,114 @@
...
+      - shipyard: <commit-id>
+
+   ```
+
+## verification of the release yaml
+
+The verification step would:
+* Verify the format of the yaml file
+* Verify that passed CI runs exist for the commit-ids proposed for the components
+* Verify that submariner-operator dependencies to projects in go.mod are pointing
```
mangelajo, line 70: One alternative is  that all are versions are aligned, so for a 0.5.0 release of the operator we expect all the other to be the same, ... hence we  don't need a versions.go file.

That's more aligned with the yaml format we are defining.

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+## A release repository
+
+   A release repository (submariner-io/releases) can be used to track and automate
+   the whole release process.
+   
+   Yaml files in this repository can be used to control the definition of a release, 
+   
+   ```
+      releases/0.5-<name>/0.5.0-<name>.yaml
+      releases/0.5-<name>/0.5.1-bugfix for....yaml
```
mkolesnik, line 32: BTW space in filenames is evil :P

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+   A submariner release is the sum of all components, tagged on a specific set of versions,
+   that work and are tested together.
+   
+   The components include:
+   
+   * subctl
+   * submariner-operator
+   * submariner core (submariner engine, route agent, globalnet ipam)
+   * lighthouse (lighthouse agent, lighthouse coredns)
```
mkolesnik, line 20: Do you want to have Shipyard on the list? You mention it later so it can be listed here with a note that it's the base env for building the other projects.

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+   ```
+
+## A release yaml
+
+   A release yaml can have the following format:
+   
+   ```yaml
+    ---
+    version: 0.5.0
+    name: the mighty 0.5 release!
```
mkolesnik, line 43: This name becomes the release name in GH? :)

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+      - submariner-charts: <commit-id>
+      - shipyard: <commit-id>
+      - admiral: <commit-id>
+
+   ```
+
+## verification of the release yaml
+
+The verification step would:
+* Verify the format of the yaml file
```
mkolesnik, line 69: Not sure if you intend to do it later, but it would be good to clarify what is expected (ie are you just expecting a valid yaml or are you expecting certain fields to be set to certain values?

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Verify that passed CI runs exist for the commit-ids proposed for the components
```
mkolesnik, line 71: This is also vague and needs more clarification, e.g are you expecting all jobs that ran to have passed?
Now that we have a lot of CI, and some flaky jobs, perhaps we don't want all CI but just certain jobs (or even certain jobs per repo).

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Verify that submariner-operator dependencies to projects in go.mod are pointing
+  to the proposed submariner, lighthouse, and shipyard releases, because this affects
+  the behaviour of `subctl verify` to be up-to-date with the e2e tests of those repositories.
+  Otherwise propose a PR for anything missing and stop.
+
+* Verify that projects have Dockerfile.dapper pinned to the specific shipyard release,
+  otherwise send a PR to the projects which didn't pin the image yet and stop.
```
mkolesnik, line 79: I wouldn't do it here, I'd have another tool (not necessarily a bot) do this.

To me seems that just complaining is fine

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+The verification step would:
+* Verify the format of the yaml file
+
+* Verify that passed CI runs exist for the commit-ids proposed for the components
+
+* Verify that submariner-operator dependencies to projects in go.mod are pointing
+  to the proposed submariner, lighthouse, and shipyard releases, because this affects
+  the behaviour of `subctl verify` to be up-to-date with the e2e tests of those repositories.
+  Otherwise propose a PR for anything missing and stop.
```
mkolesnik, line 76: IDK if this should also propose a PR, since the release would have to be updated anyway after the changes were made.
For starters I'd just fail the release and explain what's the problem.

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Verify that projects have Dockerfile.dapper pinned to the specific shipyard release,
+  otherwise send a PR to the projects which didn't pin the image yet and stop.
+
+* Verify that submariner/lighthouse are using the specific admiral commit ID, otherwise send
+  a PR to submariner/lighthouse importing the specific commit ID, and stop.
```
mkolesnik, line 82: Again, IDK if I'd send a PR, for starters lets have it just fail and complain.

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Run E2E for the combination of commit-ids with helm & operator (with lighthouse enabled)
```
mkolesnik, line 84: This is conditional on the previous verifications passing?

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+## merge of the release yaml
+
+* A script will run in Github actions, executing step by step the tagging and verification
+of the multiple components.
+
+It will retry builds if necessary.
+
+* Shipyard will be tagged for reference
```
mkolesnik, line 93: Don't see a point to do this, since we already expect all repos to be pinned to something which would require the SHA (or tag) to exist before the release.

Also, don't you mean we would tag all the repositories (including Shipyard if we want)?

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+
+* The release notes will be added to the website, proposed as a PR.
+
+* Shipyard Dockerfile.dapper will be unpinned from all repositories in master via PR.
```
mkolesnik, line 97: Coolness :)

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+## merge of the release yaml
+
+* A script will run in Github actions, executing step by step the tagging and verification
+of the multiple components.
+
+It will retry builds if necessary.
+
+* Shipyard will be tagged for reference
+
+* The release notes will be added to the website, proposed as a PR.
```
mkolesnik, line 95: Would it also create a release of subctl?

docs/release-management.md
```diff
@@ -0,0 +1,139 @@
...
+* `-X github.com/submariner-io/submariner-operator/pkg/versions.OperatorVersion=latest`
+
+### references pinned at compilation
+
+#### For a master build
+
+For the [devel bleeding edge](https://github.com/submariner-io/submariner-operator/releases/tag/devel)
+release channel, every component will be pinned at compilation time to the last successfully
+published image of each component (race conditions apart we assume this is what we tested
+in E2E).
```
mkolesnik, line 121: Is this version for reference only, or does it affect the image the subctl actually deploys?

docs/release-management.md
mangelajo: Correct, that'd be the idea.

docs/release-management.md
mangelajo: I mean valid yaml, with the fields we need.

docs/release-management.md
mangelajo: Good point, probably we need a global config (another yaml?) that specifies what jobs are we looking for exactly, like github does.

docs/release-management.md
mangelajo: Makes sense, we can improve later.

docs/release-management.md
mangelajo: Hmmm, it's a lot of manual job. I remember sweating to go through all those changes in repos, it would be nice to automate. If you want we can document something like this as ideas for later work items. (same for L76) ...

May be it could be a completely separate process, or even just some bash/go scripts (like what I made for unpinning).

docs/release-management.md
mangelajo: I think automating those parts are also important, they are probably the biggest gross of the work last time. Probably we should have a big plan, and define iterative work items.

docs/release-management.md
mangelajo: I was assuming everything was conditional on the previous.

docs/release-management.md
```diff
@@ -0,0 +1,159 @@
...
+#### For a master build
+
+For the [devel bleeding edge](https://github.com/submariner-io/submariner-operator/releases/tag/devel)
+release channel, every component will be pinned at compilation time to the last successfully
+published image of each component (race conditions apart we assume this is what we tested
+in E2E). 
+
+* `-X github.com/submariner-io/submariner-operator/pkg/versions.LighthouseVersion=${LH_IMAGE_ID}`
+* `-X github.com/submariner-io/submariner-operator/pkg/versions.SubmarinerVersion=${SUBM_IMAGE_ID}`
+* `-X github.com/submariner-io/submariner-operator/pkg/versions.OperatorVersion=${OPERATOR_IMAGE_ID}`
```
mangelajo, line 125: We won't have control over the specific versions anymore, so I say let's use just "devel" here.

docs/release-management.md
mkolesnik: Yeah that would be the simplest

docs/release-management.md
```diff
@@ -0,0 +1,159 @@
...
+* Verify that submariner/lighthouse are using the specific admiral commit ID, otherwise stop.
+
+## merge of the release yaml
+
+* A script will run in Github actions, executing step by step the tagging and verification
+of the multiple components.
+
+It will retry builds if necessary.
+
+* Shipyard & Admiral will be tagged for reference
```
mkolesnik, line 93: I don't think this would make sense, as we're stating that we're expecting them to be "ready" already before the release.

i.e.
```
Verify that submariner/lighthouse are using the specific admiral commit ID, otherwise stop.
```
If we tag admiral, dependabot will sync it automatically otherwise someone will have to manually sync.

```
Verify that projects have Dockerfile.dapper pinned to the specific shipyard release.
```
Unless you want them pinned to an arbitrary image version (e.g. `v0.5.0-8-gdf0d54`) beforehand, it would be simpler to have them point to a more coherent tag.

Perhaps release can be a 2 phase approach, one would be a "pre-release" stage upon which these are tagged with -preX (X would be some number that we determine either manually or auto incremented), and then automation could pin everything to that version.

Then the actual "release" where everything else that's in this doc make sense.

We could add this in the future of course, doesnt have to happen now, but this seems like a more coherent way of doing it unless you want to have still some manual steps involved in the process (besides the writing of the release yaml of course).

mangelajo: I have tried dependabot in the past and didn't work, may be because I didn't use a "final" release and was 0.5.0-pre0, may be we should start releasing 0.x.0 0.x.1 0.x.2 for admiral by the end of a sprint.

tpantelis:
bump

mangelajo:
> Can we do this as a part of https://submariner.io/contributing/release-process/, instead of adding parallel docs here?

This is more like an enhancement proposal of how things would work, when in place we can move to https://submariner.io/contributing/release-process/. May be for now we could move this instead to the "enhacements" repository?

nyechiel:
@mangelajo can you move this into the enhacements repo? I think it's in a pretty good shape, just need some final clean-ups.

nyechiel:
@mangelajo @mkolesnik  what should we do we this one?

mangelajo:
> @mangelajo @mkolesnik what should we do we this one

We can probably close it now. It's what releases repo implements mostly :)

Issue #553, opened 2020-05-07
Globalnet Controller should skip annotating master nodes with globalip


sridhargaddam:

mangelajo:
I'm not sure if this is necessarily a bug in some cases (where the master nodes are also used as workers?)

sridhargaddam:
> I'm not sure if this is necessarily a bug in some cases (where the master nodes are also used as workers?)

Okay, basically what happens is that submariner-route-agent that runs ONLY on the worker node annotates the respective node with the cniInterfaceIP annotation. Globalnet controller has a listener for nodes and will wait for this annotation before it can program rules for HostNetwork to remoteCluster support. It was seen that since master nodes were constantly getting re-queued as the annotation was missing. The idea is that GN controller can check if the node is capable of scheduling pods (aka worker node) and only then re-queue the node if annotation is missing.

sridhargaddam:
This is addressed

PR #603, opened 2020-05-27
Globalnet Controller to skip annotating master nodes with globalIp

sridhargaddam:
In order to support hostNetwork to remoteCluster connectivity, globalnet needs
the CNIInterfaceIP of all the worker nodes. The CNIInterfaceIP for the nodes is
annotated by submariner-route-agent. However, submariner-route-agent runs ONLY
on the worker nodes and not on the master nodes. Globalnet controller has a
listener for nodes and will wait for this annotation before it can program rules
for HostNetwork to remoteCluster support. In the current code, the master nodes
are constantly getting re-queued as the annotation will never be added on such
nodes.

In this PR we check if the node is capable of scheduling pods (aka worker node)
and only process (or re-queue) such nodes.

Fixes Issue: https://github.com/submariner-io/submariner/issues/553

### DIFF
--- a/pkg/globalnet/controllers/ipam/constants.go
+++ b/pkg/globalnet/controllers/ipam/constants.go
@@ -25,6 +25,8 @@ const (
 	kubeProxyServiceChainPrefix = "KUBE-SVC-"
 	kubeProxyServiceChainName   = "KUBE-SERVICES"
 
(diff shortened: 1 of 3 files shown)


sridhargaddam:
This PR is ready for review :)

PR #605, opened 2020-05-27
Remove separate subnet arguments

skitt:
The cable driver currently takes separate arguments for subnets and
endpoint, but the endpoint includes the subnet information. This patch
removes the separate subnets argument.


mangelajo:
From a refactor point of view makes sense, and @deanlorenz confirmed on slack that this may work for WG.

PR #606, opened 2020-05-28
Update the wireguard azure fix to be more flexible

mangelajo:
The azure Ubuntu Image has broken headers for wireguard, and as they change versions our patch needs to be updated.

This makes the workaround a little bit more flexible until the image is fixed.


mangelajo:
@deanlorenz @roytman  ^

PR #608, opened 2020-05-29
Add a debug engine, minimise the engine image

skitt:
Installing the packages we want, then force-removing the packages we
don’t need, turns out to be somewhat brittle.

Instead:

* build an image with everything, and use that as a debug image (we
  can add tools to that as necessary);
* build a builder image with a script capable of analysing binary
  dependencies;
* use that in a multi-stage build to only keep the files we really
  need.

This results in a 475 MiB debug image and an 83 MiB engine image, with
support for both strongSwan and Libreswan.


skitt:
I messed up the cherry-pick, this depends on #107.

mangelajo:
This is very cool, I need to read it more carefully but sounds like a good plan.

skitt:
> Thanks, however I don't think this PR makes much sense..

Think of it as an experiment only, I’m not particularly attached to it.

> First, the image size is ~250MB which is mostly from the base fedora image (192MB in fact), so it's not that huge in comparison.
> 
> Second, this is super cherry-picky and I don't see how sane it is to support in the long term, generally we have DNF/RPM exactly for managing the dependencies. Right now it might work, but what happens if something changes internally and an so is renames, or introduced?

The .sos are mostly tracked automatically. The NSS and PKCS stuff is annoying, and I’m not entirely sure how to address it properly...

> Also, I didn't see any additional debug tools, but generally didn't we want to provide a sidecar container to do that? Perhaps we can use our `nettest` image (or netshoot) and have a command in subctl to attach it to a given cluster/pod?

Yes, that’s what I keep saying too, but others remain unconvinced.

> Fun fact: I removed the `rpm -e` line from the original subm dockerfile, and the whole image is just 7MB bigger, just sayin'..

Weird, I thought the size reduction was larger than that when you added the SELinux policy removal!

mkolesnik:
> Think of it as an experiment only, I’m not particularly attached to it.

As an experiment it's quite interesting :)

> 
> Yes, that’s what I keep saying too, but others remain unconvinced.
> 

I opened a request to add it - https://github.com/submariner-io/submariner-operator/issues/441

> 
> Weird, I thought the size reduction was larger than that when you added the SELinux policy removal!

It was in shipyard, because it was actually installing them selinux libraries because of the other things installed there (probably docker and stuff), but here it wasn't (perhaps the base image has it, i didnt check).

skitt:
> > Yes, that’s what I keep saying too, but others remain unconvinced.
> 
> I opened a request to add it - [submariner-io/submariner-operator#441](https://github.com/submariner-io/submariner-operator/issues/441)

That one’s going to be interesting to implement ;-) (not starting the pod, but connecting the terminal to it).

> > Weird, I thought the size reduction was larger than that when you added the SELinux policy removal!
> 
> It was in shipyard, because it was actually installing them selinux libraries because of the other things installed there (probably docker and stuff), but here it wasn't (perhaps the base image has it, i didnt check).

Ah, duh, so it was!

Issue #596, opened 2020-05-26
Handle Gateway Migration edge cases in Globalnet


sridhargaddam:
**What happened**:
When we enable redundancy/gateway-migration tests in e2e tests as part of CI, we are seeing some instabilities in Globalnet jobs. 

**What you expected to happen**:
If there are any edge cases that are not handled they should be properly handled to ensure that CI jobs are stable.

PR #610, opened 2020-05-30
Cleanup globalnet rules when globalnet pod is migrated

sridhargaddam:
When the submariner.io/gateway label is deleted from the node, k8s deletes
the submariner engine and also the submariner globalnet pod from that node.
It was seen that when the globalnet pod is terminated, it was only deleting
SUBMARINER-GN-MARK chain and not the other chains like SUBMARINER-GN-INGRESS
and SUBMARINER-GN-EGRESS. This is causing random failures in CI for the
globalnet jobs.

In this PR we handle this by ensuring that when globalnet pod is terminated,
it will cleanup the associated iptable rules that were programmed on that node.

Fixes Issue: https://github.com/submariner-io/submariner/issues/596

### DIFF
--- a/pkg/globalnet/controllers/ipam/gatewaymonitor.go
+++ b/pkg/globalnet/controllers/ipam/gatewaymonitor.go
@@ -77,6 +77,7 @@ func NewGatewayMonitor(spec *SubmarinerIpamControllerSpecification, cfg *rest.Co
 
 func (i *GatewayMonitor) Run(stopCh <-chan struct{}) error {
 	defer utilruntime.HandleCrash()
+	defer ClearGlobalNetChains(i.ipt)
 
 	klog.Info("Starting GatewayMonitor to monitor the active Gateway node in the cluster.")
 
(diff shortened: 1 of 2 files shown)


pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -77,6 +77,7 @@ func NewGatewayMonitor(spec *SubmarinerIpamControllerSpecification, cfg *rest.Co
 
 func (i *GatewayMonitor) Run(stopCh <-chan struct{}) error {
 	defer utilruntime.HandleCrash()
+	defer ClearGlobalNetChains(i.ipt)
```
mangelajo, line 80: I wonder if we may add it right after L95.

I'm not sure if HandleCrash could fire before and stop the process? (not sure how it works)

Just a question.

sridhargaddam: Thanks for reviewing @mangelajo 

When multiple defer() calls are used, they are executed in Last-in-first-out order.
Also, I did some stress/monkey-testing with a script that was removing the label from the active gateway node and configuring it on a new node in a while loop with a small delay in between. I found that using defer was more reliable (aka doing the necessary cleanup) than including it at L95. Hence moved the cleanup to L80 and used defer.

PR #612, opened 2020-06-02
Fix the Wireguard workaround

skitt:
Wireguard now fails to build because a recent commit in the kernel
(https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2c64605b590edad)
moved various skb helpers to skbuff.h, and changed Wireguard’s
queueing.h to use them. The kernel we get in the test setup doesn’t
have this change, so the current Wireguard module package no longer
builds.

This patch changes the Wireguard setup to check for the change in
skbuff.h, and apply a revert patch to queueing.h if it isn’t present.


roytman:
Thanks @skitt !

PR #613, opened 2020-06-02
Add `using=` make flag

mkolesnik:
Added a flag that can be sent to make to pass additional options, e.g
`using="globalnet, wireguard, helm"`

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -126,13 +126,18 @@ To run the tests simply execute the following:
 make e2e
 ```
 
-To test with a specific k8s version, an additional **version** parameter can be passed to **make e2e** command:
-```bash
-make ci e2e version=1.14.1
-```
+To test with optional functionality, use the `using=` flag to enable the extra options.
 
-Full list of supported k8s versions can found on [kind release page] page. We are using kind vesion 0.6.1.
-Default **version** is 1.14.6.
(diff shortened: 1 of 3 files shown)


docs/testing.md
```diff
@@ -126,14 +126,19 @@ To run the tests simply execute the following:
...
+Currently, these options are supported:
+* **globalnet:** To deploy with globalnet (and overlapping IPs)
+* **helm:** To deploy using helm instead of the operator.
+* **libreswan:** To use libreswan to establish connectivity.
+* **wireguard:** To use wireguard to establish connectivity.
+
+For example:
 ```bash
-make ci e2e version=1.14.1
+make e2e using="globalnet, wireguard, helm"
```
skitt, line 139: Since you’ve gone to the trouble of supporting commas, I think it’s worth using them in the docs to make the declaration simpler (without quotes):

```suggestion
make e2e using=globalnet,wireguard,helm
```

mkolesnik: I preferred this style because otherwise it's easy to make a mistake, e.g. `make e2e using=globalnet, wireguard, helm` will only really consider globalnet and fail but only after the e2e has run.

But perhaps as I usually say we shouldn't babysit the users too much :)

PR #615, opened 2020-06-02
Avoid relying on skbuff.h

skitt:
The Azure kernel headers apparently don't have skbuff.h in the
expected place, so instead of checking that, look explicitly for
version -1020- of the kernel. This avoids failing the builds on kernel
1022 which includes an updated skbuff.h and doesn't need the Wireguard
patch.


mangelajo:
@deanlorenz @roytman ^

PR #616, opened 2020-06-03
Don't try to shrink submariner image

mkolesnik:
The DNF layer adds about 50MB (out of which several are due to
distrosync).
Removing stuff actually doesn't clean much since most of these packages
were already there in the base image, so it saves around 7MB but makes
the whole thing needlessly complicated.

We can live with a few extra MB but a much simpler Dockerfile


mkolesnik:
FYI I also tried UPX and it saved a measly few MB (since I also had to install find and upx itself) so it wasn't worth it either in this case.

The base image is 192MB, if we can use a smaller one then our overall image size will be smaller also.

PR #619, opened 2020-06-03
Fix NAT on Libreswan

skitt:
For NAT connections, the local side always uses the private IP
address, not the public IP address, and the routed subnets always use
the private IP addresses. The public IP address is only used for the
remote end of the connection.


mangelajo:
ok libreswan without GN passed.

mangelajo:
Yikes, again same failure for globalnet+libreswan. hmm

PR #620, opened 2020-06-03
Use the host iptables in the engine

skitt:
Like the route agent and globalnet containers, the engine container
needs to use the host iptables.

This requires a corresponding change in the deployment settings (in
the operator and Helm charts).

### DIFF
--- a/package/submariner.sh
+++ b/package/submariner.sh
@@ -20,4 +20,34 @@ sysctl -w net.ipv4.conf.all.send_redirects=0
 
 export PATH=$PATH:/usr/libexec/strongswan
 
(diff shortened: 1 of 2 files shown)


package/submariner.sh
```diff
@@ -20,4 +20,34 @@ sysctl -w net.ipv4.conf.all.send_redirects=0
 
 export PATH=$PATH:/usr/libexec/strongswan
 
+function find_iptables_on_host() {
```
mangelajo, line 23: We probably can extract that common bit with route-agent later on a .sh which both could use.

PR #622, opened 2020-06-04
Libreswan: add explicit authentication and encaps

skitt:
The identifier needs to match the expected authentication, rather than
the apparent IP address, so set it explicitly to the private address.

NAT on AWS at least also needs encapsulation, which doesn’t get
automatically set up, so enforce it when NAT is requested.


mangelajo:
Thank you @skitt !

PR #628, opened 2020-06-07
support pre-shared key in wireguard

deanlorenz:
@mangelajo 

- Add PSK support (environment variable: `CE_IPSEC_PSK`)
- Allow changing WG port (environment variable: `CE_IPSEC_NATTPORT`, default 5871)
- updated README

**Limitations / TODO**

- Using the ipsec env prefix and env variable names
  common options should probably use a common submariner env prefix

- Should probably move the env parsing out of driver and use common code for all cables

- UDP listen port should be part of the endpoint configuration -- allow endpoints to use different ports

### DIFF
--- a/pkg/cable/wireguard/README.md
+++ b/pkg/cable/wireguard/README.md
@@ -1,4 +1,4 @@
-# WireGuard Cable Driver (WIP)
+# WireGuard Cable Driver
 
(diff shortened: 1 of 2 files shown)


mangelajo:
I completely agree on the limitations/TODO here.

mangelajo:
@deanlorenz can you give me a hand opening bugs for the TODO items you mentioned?, I think it's something we need to fix in submariner design.

deanlorenz:
https://github.com/submariner-io/submariner/issues/633



Regards,
--Dean

PR #632, opened 2020-06-08
Wait for Pluto to start

skitt:
In the function which starts the IKE daemon, wait for it to start (up
to 5s) before returning. This helps avoid races when starting the
gateway; if we don't wait, the daemon starts alongside the gateway,
and in some cases we end up trying to set tunnels up before the daemon
is ready, resulting in delays (the tunnels aren't set up until the
next resync round).

### DIFF
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"strings"
 	"syscall"
+	"time"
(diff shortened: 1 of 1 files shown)


mangelajo:
> LGTM
> 
> Are 5 secs sufficient? or shall we give it some more time?

It's normally very quick, just a process start.

mangelajo:
We can always bump it if that's not enough.

PR #637, opened 2020-06-09
Less scripts, use Shipyard

mkolesnik:
This chain of commits removes a lot of the scripts and uses Shipyard components instead.

For now, some CI jobs will fail since they depend on https://github.com/submariner-io/shipyard/pull/204
They should be re-run once that patch merges


mkolesnik:
It's easiest to go commit by commit to see what has been done, but the final result is not that complicated either :)

mangelajo:
:mage_man:

dfarrell07:
> Unit testing is sad: https://github.com/submariner-io/submariner/pull/637/checks?check_run_id=754139357#step:4:63

Just needed a re-run after the Shipyard PR landed, passing now. 

There are two jobs "not reporting" but actually green. 🤷‍♂️

Issue #649, opened 2020-06-15
Broker connections mustn’t assume the SA trust chain is necessary


skitt:
**What happened**:

When attempting to connect to a broker cluster set up with Let’s Encrypt certificates, connections fail with “x509: certificate signed by unknown authority”.

**What you expected to happen**:

The clients should be able to connect; the CA is in the system’s trust store.

**How to reproduce it (as minimally and precisely as possible)**:

Set up a cluster with API endpoints using certificates signed by a public CA, then deploy Submariner using that as the broker.

**Anything else we need to know?**:

See also https://github.com/submariner-io/submariner-operator/issues/451

**Environment**:
- Submariner version (use `subctl version`): 0.4-rc2
- Kubernetes version (use `kubectl version`): 1.17.1
- Cloud provider or hardware configuration: private
- OS (e.g: `cat /etc/os-release`): private
- Kernel (e.g. `uname -a`): private
- Install tools: `subctl`
- Network plugin and version (if this is a network-related bug): N/A
- Others: this is on OpenShift 4.4

skitt:
I’m working on fixing this.

PR #650, opened 2020-06-15
Check whether the CA trust chain is needed

skitt:
The trust anchor stored in the broker information reflects the trust
chain of the token, not necessarily the trust chain of the API
endpoint; but we're using it for the latter in all cases.

This changes NewDatastore to check whether the clientset can be used,
by retrieving the endpoints. This allows the configuration to be
verified, and switched to use the stored trust anchor if necessary.

Fixes: #649

### DIFF
--- a/pkg/datastore/kubernetes/kubernetes.go
+++ b/pkg/datastore/kubernetes/kubernetes.go
@@ -2,8 +2,10 @@ package kubernetes
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/base64"
 	"fmt"
+	"net/url"
 	"reflect"
 	"time"
 
@@ -42,6 +44,9 @@ type datastoreSpecification struct {
 	Ca              string
 }
 
+// defaultPrivate indicates whether to start with a private TLS chain
+var defaultPrivate bool = false
+
 func NewDatastore(thisClusterID string, stopCh <-chan struct{}) (datastore.Datastore, error) {
 	k8sSpec := datastoreSpecification{}
 
@@ -54,25 +59,7 @@ func NewDatastore(thisClusterID string, stopCh <-chan struct{}) (datastore.Datas
 
 	klog.V(log.DEBUG).Infof("Rendered API server host: %q", host)
 
-	tlsClientConfig := rest.TLSClientConfig{}
-	if k8sSpec.Insecure {
-		tlsClientConfig.Insecure = true
-	} else {
-		caDecoded, err := base64.StdEncoding.DecodeString(k8sSpec.Ca)
-		if err != nil {
-			return nil, fmt.Errorf("error decoding CA data: %v", err)
-		}
-		tlsClientConfig.CAData = caDecoded
-	}
-
-	k8sClientConfig := rest.Config{
-		// TODO: switch to using cluster DNS.
-		Host:            host,
-		TLSClientConfig: tlsClientConfig,
-		BearerToken:     k8sSpec.APIServerToken,
-	}
-
-	submarinerClient, err := submarinerClientset.NewForConfig(&k8sClientConfig)
+	submarinerClient, err := getSubmarinerClientset(host, &k8sSpec)
 	if err != nil {
 		return nil, fmt.Errorf("error building submariner clientset: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/datastore/kubernetes/kubernetes.go
```diff
@@ -87,6 +71,63 @@ func NewDatastore(thisClusterID string, stopCh <-chan struct{}) (datastore.Datas
 	}, nil
 }
 
+func getSubmarinerClientset(host string, k8sSpec *datastoreSpecification) (*submarinerClientset.Clientset, error) {
+	if k8sSpec.Insecure {
+		// Shortcut
```
tpantelis, line 76: `getSubmarinerRESTConfig` also checks `k8sSpec.Insecure` so I'm not sure what this shortcut is for - it seems redundant ...

skitt: The comment could be clearer! The point of the shortcut is to avoid calling getSubmarinerRESTConfig twice, and testing the connection, when we’ve asked for an insecure connection. I could drop the shortcut entirely, and then we’d request an insecure connection, test it, and if it fails because of an unknown CA (which it won’t), request a new one. Actually now that I spell it out, given that we’ll never get the expected failure in the insecure case, I might as well drop the test.

skitt:
> This code will be going away soon when we switch to use the admiral syncer API. Can you make these changes in _https://github.com/submariner-io/admiral/blob/master/pkg/syncer/broker/config.go_ instead?

I’d rather not make the changes in Admiral _instead_ — I’ll do them _in addition_ (I need this to work sooner rather than later).

tpantelis:
> > This code will be going away soon when we switch to use the admiral syncer API. Can you make these changes in _https://github.com/submariner-io/admiral/blob/master/pkg/syncer/broker/config.go_ instead?
> 
> I’d rather not make the changes in Admiral _instead_ — I’ll do them _in addition_ (I’d like to get this working sooner rather than later).

OK - that's fine - I see this is for a POC.

pkg/datastore/kubernetes/kubernetes.go
```diff
@@ -87,6 +71,55 @@ func NewDatastore(thisClusterID string, stopCh <-chan struct{}) (datastore.Datas
...
+	_, err = clientset.SubmarinerV1().Endpoints(k8sSpec.RemoteNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		if urlError, ok := err.(*url.Error); ok {
+			if _, ok := urlError.Unwrap().(x509.UnknownAuthorityError); ok {
+				// Certificate error, try with the trust chain
+				config, err = getSubmarinerRESTConfig(host, true, k8sSpec)
+				if err != nil {
+					return nil, err
+				}
+				return submarinerClientset.NewForConfig(config)
```
tpantelis, line 96: Should we do the check on L87 again with the new clientset?

skitt: Yes, I thought of that too overnight — the function should ensure the result is consistent, whatever path is taken.

pkg/datastore/kubernetes/kubernetes.go
```diff
@@ -87,6 +71,56 @@ func NewDatastore(thisClusterID string, stopCh <-chan struct{}) (datastore.Datas
 	}, nil
 }
 
+func getSubmarinerClientset(host string, k8sSpec *datastoreSpecification) (*submarinerClientset.Clientset, error) {
+	// We need to check whether we need the provided CA (for a secure connection)
+	// For an insecure connection, both cases are equivalent, but we’ll never get the X.509 error
+	// First, try without specifying anything (using the system trust store)
+	clientset, err := getAndCheckSubmarinerClientset(host, false, k8sSpec)
```
mangelajo, line 78: I'd cache this once learned, you could use a bool variable, initialized to "false"

clientset, err := getAndCheckSubmarinerClientset(host, usePrivateCA, k8sSpec) 

if ... failed CA ... {
     clientset, err = getAndCheckSubmarinerClientset(host, !usePrivateCA, k8sSpec)
     if err == nil {
           usePrivateCA = !usePrivateCA
     }
}

mangelajo: I'm thinking of failure modes where we get disconnected from the API, may be system overloaded... we every time we try to connect again we would be imposing extra retries even if we already know what's the right certificate chain to use.

skitt: Done.

PR #654, opened 2020-06-16
Remove phpapi datastore implementation

tpantelis:
This is not used and it was agreed in the u/s call a while ago
to remove it.


mangelajo:
This was discussed multiple times in U/S meeting, and it was agreed to remove.

Issue #635, opened 2020-06-09
libreswan:  HostNetworking on GW to remote pod not on gateway


mangelajo:
**What happened**:

Test fails: "[Fail] [dataplane] Basic TCP connectivity tests across clusters without discovery when a pod with HostNetworking connects via TCP to a remote pod when the pod is on a gateway and the remote pod is not on a gateway [It] should have sent the expected data from the pod to the other pod 
/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/shipyard/test/e2e/framework/network_pods.go:139
"

**What you expected to happen**:

Test to pass

**How to reproduce it (as minimally and precisely as possible)**:

Deploy 2 openshift clusters, install submariner, run:
```
subctl verify --only connectivity kubeconfig1 kubeconfig2
```

**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`):
subctl version: dev
built from git ref: v0.4.0-rc2
- Cloud provider or hardware configuration: ocp 4.6 on AWS
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`): `4.18.0-147.8.1.el8_1.x86_64`

sridhargaddam:
@mangelajo @skitt 
Okay, I had a look at the setup and the issue is that from the sourcePod (with HostNetworking), traffic is making its way to the destinationPod on remoteCluster. 
However, the source-ip of the pod is the IPAddress of the Host and **NOT** the CNIInterface IP. Because of this, the return traffic is not able to make its way to the source cluster (which is the expected behavior).
The same use-case is working fine in KIND because KIND setup uses a flat network for both the clusters.

I can think of the following two ways to address this issue:

**Solution1:**
On the Gateway Node, add an explicit route to support HostNetworking.
ip route add <remote Pod/SvcCIDR> dev eth0 proto static src <localCNIIfaceIP>
Note: we can also use a specific table like 220, if required.

**Solution2:**
On the Gateway Node, use ipMark to mark remote traffic (i.e., remote Pod/Svc CIDRs)
i.e., iptables -t nat <submariner-chain> -d remoteCidr -j MARK --set-mark <submarinerMark>
[...]

PR #656, opened 2020-06-17
libreswan: HostNetworking on GW to remote cluster support

sridhargaddam:
This PR implements the necessary support in route-agent code to enable
remote cluster connectivity from a Pod on GatewayNode with HostNetworking
for libreswan cable driver.

Along with this, it also fixes an issue with Strongswan (inc' globalnet)
where the sourceIP of such traffic was not properly configured to the
CNIInterfaceIP of the GatewayNode.

Fixes issue: https://github.com/submariner-io/submariner/issues/635

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -58,6 +58,7 @@ type Controller struct {
 	localClusterCidr []string
 	localServiceCidr []string
 	remoteSubnets    *util.StringSet
+	routeCacheGWNode *util.StringSet
 
 	gwVxLanMutex *sync.Mutex
 	vxlanDevice  *vxLanIface
@@ -113,6 +114,19 @@ const (
 	// supporting connectivity from HostNetwork to remoteClusters.
 	// [#] interface on the node that has an IPAddress from the clusterCIDR
 	CniInterfaceIp = "submariner.io/cniIfaceIp"
+
+	// To support connectivity for Pods with HostNetworking on the GatewayNode, we program
+	// certain routing rules in table 150. As part of these routes, we set the source-ip of
+	// the egress traffic to the corresponding CNIInterfaceIP on that host.
+	RouteAgentHostNetworkTableID = 150
+)
+
+type Operation int
+
+const (
+	AddRoute Operation = iota
+	DeleteRoute
+	FlushRouteTable
 )
 
 func newRateLimiter() workqueue.RateLimiter {
@@ -132,6 +146,7 @@ func NewController(clusterID string, ClusterCidr []string, ServiceCidr []string,
 		defaultHostIface:       link,
 		isGatewayNode:          false,
 		remoteSubnets:          util.NewStringSet(),
+		routeCacheGWNode:       util.NewStringSet(),
 		remoteVTEPs:            util.NewStringSet(),
 		endpointsSynced:        config.EndpointInformer.Informer().HasSynced,
 		smRouteAgentPodsSynced: config.PodInformer.Informer().HasSynced,
@@ -305,6 +320,43 @@ func (r *Controller) updateIptableRulesForInterclusterTraffic(inputCidrBlocks []
 			err := r.programIptableRulesForInterClusterTraffic(inputCidrBlock)
 			if err != nil {
 				klog.Errorf("Failed to program iptable rule. %v", err)
+			}
+		}
+	}
+}
+
+func (r *Controller) updateRoutingRulesForHostNetworkSupport(inputCidrBlocks []string, operation Operation) {
+	if operation == FlushRouteTable {
+		r.routeCacheGWNode.DeleteAll()
+		cmd := exec.Command("/sbin/ip", "r", "flush", "table", strconv.Itoa(RouteAgentHostNetworkTableID))
+		if err := cmd.Run(); err != nil {
+			// We can safely ignore this error, as this table
+			// won't exist in most nodes (only gateway nodes)
+			return
+		}
+	} else {
+		// These routing rules are required ONLY on the Gateway Node.
+		// On the non-Gateway nodes, we use iptable rules to support this use-case.
+		if r.isGatewayNode && r.cniIface != nil {
+			switch operation {
+			case AddRoute:
+				for _, inputCidrBlock := range inputCidrBlocks {
+					if !r.routeCacheGWNode.Contains(inputCidrBlock) {
+						r.routeCacheGWNode.Add(inputCidrBlock)
+						if err := r.configureRoute(inputCidrBlock, operation); err != nil {
+							klog.Errorf("Failed to add route for HostNetwork support on the Gateway node. %v", err)
+						}
+					}
+				}
+			case DeleteRoute:
+				for _, inputCidrBlock := range inputCidrBlocks {
+					if r.routeCacheGWNode.Contains(inputCidrBlock) {
+						r.routeCacheGWNode.Delete(inputCidrBlock)
+						if err := r.configureRoute(inputCidrBlock, operation); err != nil {
+							klog.Errorf("Failed to delete route for HostNetwork support on the Gateway node. %v", err)
+						}
+					}
+				}
 			}
 		}
 	}
@@ -521,6 +573,12 @@ func (r *Controller) processNextEndpoint() bool {
 			klog.Infof("Updating iptable rules for remote cluster's Endpoint %q with subnets %v",
 				endpoint.Name, endpoint.Spec.Subnets)
 			r.updateIptableRulesForInterclusterTraffic(endpoint.Spec.Subnets)
+
+			r.gwVxLanMutex.Lock()
+			// Add routes to the new endpoint on the GatewayNode.
+			r.updateRoutingRulesForHostNetworkSupport(endpoint.Spec.Subnets, AddRoute)
+			r.gwVxLanMutex.Unlock()
+
 			r.endpointWorkqueue.Forget(obj)
 			return nil
 		}
@@ -548,6 +606,15 @@ func (r *Controller) processNextEndpoint() bool {
 				klog.Fatalf("Unable to create VxLAN interface on gateway node (%s): %v", hostname, err)
 			}
 
+			err = r.configureIPRule(AddRoute)
+			if err != nil {
+				klog.Errorf("Unable to add ip rule to table %d on Gateway node %s: %v",
+					RouteAgentHostNetworkTableID, hostname, err)
+			}
+
+			// On the GatewayNode, add routes for the remoteSubnets
+			r.updateRoutingRulesForHostNetworkSupport(r.remoteSubnets.Elements(), AddRoute)
+
 			r.endpointWorkqueue.Forget(obj)
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -826,3 +908,55 @@ func (r *Controller) reconcileRoutes(vxlanGw net.IP) error {
...
+		return fmt.Errorf("error parsing cidr block %s: %v", remoteSubnet, err)
+	}
+	route := netlink.Route{
+		Dst:       dst,
+		Src:       src,
+		Scope:     unix.RT_SCOPE_LINK,
+		LinkIndex: r.defaultHostIface.Index,
+		Protocol:  4,
+		Table:     RouteAgentHostNetworkTableID,
+	}
```
deanlorenz, line 947:
```suggestion
	indx := r.defaultHostIface.Index
	if wg, err := net.InterfaceByName(wireguard.DefaultDeviceName); err == nil {
		indx = wg.Index
	}
	route := netlink.Route{
		Dst:       dst,
		Src:       src,
		Scope:     unix.RT_SCOPE_LINK,
		LinkIndex: indx,
		Protocol:  4,
		Table:     RouteAgentHostNetworkTableID,
	}
```

sridhargaddam:
> Added code to use wireguard i/f index if exists.
> Can probably generalize if we use the same i/f name for all cables that need them.
> Note: also need to remove all the wireguard routing code.

Thanks for reviewing and validating the PR with wireguard drive @deanlorenz.

pkg/routeagent/controllers/route/route.go
```diff
@@ -310,6 +326,43 @@ func (r *Controller) updateIptableRulesForInterclusterTraffic(inputCidrBlocks []
 	}
 }
 
+func (r *Controller) updateRoutingRulesForHostNetworkSupport(inputCidrBlocks []string, operation Operation) {
+	if operation == FlushRouteTable {
+		r.routeCacheGWNode.DeleteAll()
+		cmd := exec.Command("/sbin/ip", "r", "flush", "table", strconv.Itoa(RouteAgentHostNetworkTableID))
+		if err := cmd.Run(); err != nil {
+			// We can safely ignore this error, as this table
+			// won't exist in most nodes (only gateway nodes)
```
tpantelis, line 335: I think we should at least log it to debug. Also can we distinguish between the table non-existent and another failure?

pkg/routeagent/controllers/route/route.go
```diff
@@ -310,6 +326,43 @@ func (r *Controller) updateIptableRulesForInterclusterTraffic(inputCidrBlocks []
...
+			return
+		}
+	} else {
+		// These routing rules are required ONLY on the Gateway Node.
+		// On the non-Gateway nodes, we use iptable rules to support this use-case.
+		if r.isGatewayNode && r.cniIface != nil {
```
tpantelis, line 341: This can be combined with L338 to reduce indentation.

pkg/routeagent/controllers/route/route.go
```diff
@@ -310,6 +326,43 @@ func (r *Controller) updateIptableRulesForInterclusterTraffic(inputCidrBlocks []
...
+			switch operation {
+			case AddRoute:
+				for _, inputCidrBlock := range inputCidrBlocks {
+					if !r.routeCacheGWNode.Contains(inputCidrBlock) {
+						r.routeCacheGWNode.Add(inputCidrBlock)
```
tpantelis, line 346:
```suggestion
					if r.routeCacheGWNode.Add(inputCidrBlock) {
```

pkg/routeagent/controllers/route/route.go
```diff
@@ -310,6 +326,43 @@ func (r *Controller) updateIptableRulesForInterclusterTraffic(inputCidrBlocks []
...
+						if err := r.configureRoute(inputCidrBlock, operation); err != nil {
+							klog.Errorf("Failed to add route for HostNetwork support on the Gateway node. %v", err)
+						}
+					}
+				}
+			case DeleteRoute:
+				for _, inputCidrBlock := range inputCidrBlocks {
+					if r.routeCacheGWNode.Contains(inputCidrBlock) {
+						r.routeCacheGWNode.Delete(inputCidrBlock)
```
tpantelis, line 355:
```suggestion
					if r.routeCacheGWNode.Delete(inputCidrBlock) {
```

pkg/routeagent/controllers/route/route.go
```diff
@@ -310,6 +326,43 @@ func (r *Controller) updateIptableRulesForInterclusterTraffic(inputCidrBlocks []
...
+		// These routing rules are required ONLY on the Gateway Node.
+		// On the non-Gateway nodes, we use iptable rules to support this use-case.
+		if r.isGatewayNode && r.cniIface != nil {
+			switch operation {
+			case AddRoute:
+				for _, inputCidrBlock := range inputCidrBlocks {
+					if !r.routeCacheGWNode.Contains(inputCidrBlock) {
+						r.routeCacheGWNode.Add(inputCidrBlock)
+						if err := r.configureRoute(inputCidrBlock, operation); err != nil {
+							klog.Errorf("Failed to add route for HostNetwork support on the Gateway node. %v", err)
```
tpantelis, line 348:
```suggestion
							klog.Errorf("Failed to add route %q for HostNetwork support on the Gateway node: %v", inputCidrBlock , err)
```

Perhaps it will help troubleshooting to output the route it was trying to add.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Actually, we cannot do anything and we will not have any benefit if we log the error message. So, it's okay to ignore it.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
sridhargaddam: thats a good point.

pkg/routeagent/controllers/route/route.go
sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
tpantelis: I assume it  could fail for another reason. In my experience it's better to log an error rather than silence it.  Ideally if we can distinguish table non-existent error but at least log it to debug.

sridhargaddam: Ok

pkg/routeagent/controllers/route/route.go
```diff
@@ -826,3 +916,62 @@ func (r *Controller) reconcileRoutes(vxlanGw net.IP) error {
...
+
+func (r *Controller) configureRoute(remoteSubnet string, operation Operation) error {
+	src := net.ParseIP(r.cniIface.ipAddress)
+	_, dst, err := net.ParseCIDR(remoteSubnet)
+	if err != nil {
+		return fmt.Errorf("error parsing cidr block %s: %v", remoteSubnet, err)
+	}
+
+	ifaceIndex := r.defaultHostIface.Index
+	if r.localCableDriver == "wireguard" {
```
mangelajo, line 950:
```suggestion
        // TODO: Add support for this in the CableDrivers themselves.
	if r.localCableDriver == "wireguard" {
```

pkg/routeagent/controllers/route/route.go
```diff
@@ -826,3 +916,62 @@ func (r *Controller) reconcileRoutes(vxlanGw net.IP) error {
...
+		if wg, err := net.InterfaceByName(wireguard.DefaultDeviceName); err == nil {
```
mangelajo, line 951:
```suggestion

		if wg, err := net.InterfaceByName(wireguard.DefaultDeviceName); err == nil {
			klog.Errorf("Wireguard interface %s not found", wireguard.DefaultDeviceName)
```

pkg/routeagent/controllers/route/route.go
```diff
@@ -548,6 +611,18 @@ func (r *Controller) processNextEndpoint() bool {
...
 
+			err = r.configureIPRule(AddRoute)
+			if err != nil {
+				klog.Errorf("Unable to add ip rule to table %d on Gateway node %s: %v",
+					RouteAgentHostNetworkTableID, hostname, err)
+			}
+
+			if r.localCableDriver == "" {
+				r.localCableDriver = endpoint.Spec.Backend
+			}
```
tpantelis, line 622: Should this also be checked/set when on a non-gateway node as well?

mangelajo: For non-gateway we don't use that later, it wouldn't hurt to do always if we did.

tpantelis: Sorry not sure  follow.  `localCableDriver` is used by `configureRoute` which is called by `updateRoutingRulesForHostNetworkSupport ` on all endpoint-processing paths.

sridhargaddam: Basically, we need the cableDriver used in the cluster. Although currently we do not support a mix and match of cableDrivers, its a good idea to only depend on what is the localDriver on the cluster. This information, as you pointed out, will be used on the GatewayNode to program routing rules to remoteSubnets.

On nonGateway nodes, we call the API updateRoutingRulesForHostNetworkSupport with FlushRouteTable which does not require the driver information (i.e., driver specific interface).

pkg/routeagent/controllers/route/route.go
sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
sridhargaddam: Done

pkg/routeagent/controllers/route/route.go
sridhargaddam: BTW we are doing this before the controller is initialized in L227, so no need to do it again on non-gateway nodes.

pkg/routeagent/controllers/route/route.go
```diff
@@ -826,3 +916,65 @@ func (r *Controller) reconcileRoutes(vxlanGw net.IP) error {
...
+
+func (r *Controller) configureRoute(remoteSubnet string, operation Operation) error {
+	src := net.ParseIP(r.cniIface.ipAddress)
+	_, dst, err := net.ParseCIDR(remoteSubnet)
+	if err != nil {
+		return fmt.Errorf("error parsing cidr block %s: %v", remoteSubnet, err)
+	}
+
+	ifaceIndex := r.defaultHostIface.Index
+	// TODO: Add support for this in the CableDrivers themselves.
```
tpantelis, line 950: Perhaps we could have the cable driver optionally publish its interface index via the Endpoint's BackendConfig.

mangelajo: It's going to be more complicated, we will need implication from the cable driver in some cases, for example. Some cable drivers (ip on ip, ... etc) could need multiple interfaces, one per remote cluster. That would mean that we need to decide about the interface based on the remote cluster.

PR #657, opened 2020-06-17
Use shared image recipe from Shipyard

mkolesnik:
We've added a shared recipe to build images in Shipyard[1] so use that
instead of redefining it.

[1] https://github.com/submariner-io/shipyard/pull/212

Right now this PR will fail on CI until the one from Shipyard is merged.


tpantelis:
This was already done by https://github.com/submariner-io/submariner/pull/672.

PR #658, opened 2020-06-17
Remove routing from WireGuard cable driver

deanlorenz:
Remove routing (now in RouteAgent per #656)
Change interface name to be `submariner`. Should consider making this user configurable.

WireGuard driver now assumes that RouteAgent adds routes to CIDRs in remote clusters which are forwarded to `submariner` interface. The driver only adds those CIDRs to `allowed-IPs`. RouteAgent is also expected to flush the routes if the Gateway moves to another node. The `submariner` interface is still created by the driver on startup. Any existing interface is first removed, so any existing routes would be removed too.


sridhargaddam:
The tests were failing for an unrelated issue. Retriggered the tests.

Issue #642, opened 2020-06-11
libreswan: NAT-T does not work when one side has no public IP.


mangelajo:
**What happened**:

NAT traversal from an on-prem deployment to a public cloud doesn't work when the on-prem deployment has no public/floating-ip.

**What you expected to happen**:

To work, as in strongswan.

**How to reproduce it (as minimally and precisely as possible)**:

Deploy with libreswan cabledriver, and check the gateway connection list with
kubectl get Gateways -A -o yaml

**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

mangelajo:
@manosnoam be aware of this when you are testing ^ let me know if you experience it during your testing with libreswan and if you can provide more details.

manosnoam:
I've installed subctl 0.4.0 RC2 with Libreswan, and I see HA is passive on Cluster B (cluster on-premise OSP):

```
Name:         default-cl1-9vgwj-worker-bq8rt
11:25:28 Namespace:    submariner-operator
11:25:28 Labels:       <none>
11:25:28 Annotations:  update-timestamp: 1592209423
11:25:28 API Version:  submariner.io/v1
11:25:28 Kind:         Gateway
11:25:28 Metadata:
11:25:28   Creation Timestamp:  2020-06-15T08:23:44Z
11:25:28   Generation:          1
11:25:28   Resource Version:    2834031
11:25:28   Self Link:           /apis/submariner.io/v1/namespaces/submariner-operator/gateways/default-cl1-9vgwj-worker-bq8rt
11:25:28   UID:                 c5a419dd-ab09-44fb-9665-689ad0aca164
11:25:28 Status:
11:25:28   Connections:
11:25:28   Ha Status:  passive
11:25:28   Local Endpoint:
11:25:28     Backend:      libreswan
11:25:28     cable_name:   submariner-cable-nmanos-cluster-b-10-166-0-26
11:25:28     cluster_id:   nmanos-cluster-b
11:25:28     Hostname:     default-cl1-9vgwj-worker-bq8rt
11:25:28     nat_enabled:  true
11:25:28     private_ip:   10.166.0.26
11:25:28     public_ip:    66.187.233.202
[...]

PR #659, opened 2020-06-17
libreswan: do not replicate connections in list

mangelajo:
GetActiveConnections from cable drivers is supposed to return the
list of created cables instead of the connected cables.

If we only return connected ones connections will eventually get
replicated.

Related-Issue: #642


pkg/cable/libreswan/libreswan.go
```diff
@@ -150,15 +150,9 @@ func (i *libreswan) refreshConnectionStatus() error {
 
 // GetActiveConnections returns an array of all the active connections for the given cluster.
 func (i *libreswan) GetActiveConnections(clusterID string) ([]string, error) {
-	if err := i.refreshConnectionStatus(); err != nil {
-		return []string{}, err
-	}
-
```
sridhargaddam, line 156: @mangelajo, just for info - Is there an issue if we call refreshConnectionStatus?

PR #660, opened 2020-06-17
List the supported cable driver types

skitt:
... when an unknown type is specified.

### DIFF
--- a/pkg/cable/driver.go
+++ b/pkg/cable/driver.go
@@ -2,6 +2,7 @@ package cable
 
 import (
 	"fmt"
+	"strings"
 
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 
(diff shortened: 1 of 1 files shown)


pkg/cable/driver.go
```diff
@@ -53,7 +54,14 @@ func AddDriver(name string, driverCreate DriverCreateFunc) {
...
 	driverCreate, ok := drivers[localEndpoint.Spec.Backend]
 	if !ok {
-		return nil, fmt.Errorf("unsupported cable type %s", localEndpoint.Spec.Backend)
+		var driverList strings.Builder
+		for driver := range drivers {
+			if driverList.Len() > 0 {
+				driverList.WriteString(", ")
+			}
+			driverList.WriteString(driver)
+		}
```
tpantelis, line 63: This is fine - I wasn't aware of `strings.Builder` - but simpler to use `strings.Join`?

skitt: I couldn’t figure out a way to use `Join` on a map...

tpantelis: ahh - sorry I thought it was an array.  It would be nice if there were built-in map utilities like Keys(), Values() etc like Java has. Maybe there is somewhere ...

PR #663, opened 2020-06-17
libreswan: initiate connections asynchronously and retry whacks (stricter err checking)

mangelajo:
Connection initiation is now done asynchronously, so whack --initiate
won't keep waiting if the other side isn't ready (eliminating potential
timing issues).

If whack fails we retry at least 3 times (as we did for strongswan vici
connections).

Fail hard on any whack exit error, which will result on a later retry
because the endpoint is enqueued for retry by the cable engine.

Closes-Issue: #642

### DIFF
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -201,7 +201,7 @@ func whack(args ...string) error {
 			break
 		}
 
-		klog.Errorf("error %v whacking with args: %v", err, args)
+		klog.Warningf("error %v whacking with args: %v", err, args)
 		time.Sleep(1 * time.Second)
 
 	}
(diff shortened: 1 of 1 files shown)


mangelajo:
ok, the libreswan checks are all fine ':D

pkg/cable/libreswan/libreswan.go
```diff
@@ -186,6 +187,31 @@ func extractSubnets(endpoint subv1.EndpointSpec) []string {
...
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		klog.V(log.TRACE).Infof("Whacking with %v", args)
+
+		if err = cmd.Run(); err == nil {
+			break
+		}
+
+		klog.Errorf("error %v whacking with args: %v", err, args)
```
tpantelis, line 204:
```suggestion
		klog.V(log.DEBUG).Infof("Error whacking with args: %v - retrying: %v", args, err)
```

mangelajo: Even if it's retriable, it's an error, it's not supposed to happen.

May be we should keep it as warning instead? (middle ground ;D )

tpantelis: yeah warning sounds good - it's not an actual error until we're out of retries.

mangelajo:
Thanks!! :) ... /me prays to quay...

PR #664, opened 2020-06-18
Update the member request Issue template with DCO

nyechiel:
Fixes https://github.com/submariner-io/submariner-website/issues/181

Signed-off-by: Nir Yechiel <nyechiel@redhat.com>

### DIFF
--- a/.github/ISSUE_TEMPLATE/membership.md
+++ b/.github/ISSUE_TEMPLATE/membership.md
@@ -13,6 +13,7 @@ e.g. (at)example_user
 ### Requirements
 - [ ] I have reviewed the community membership guidelines (https://submariner-io.github.io/en/contributing/community-membership/)
 - [ ] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
+- [ ] I have reviewed the Developer Certificate Of Origin (https://submariner-io.github.io/contributing/community-membership/#new-contributors) and sign-off my commits
 - [ ] I have subscribed to the submariner-dev e-mail list (https://groups.google.com/forum/#!forum/submariner-dev)
 - [ ] I have multiple contributions to Submariner that meet the requirements listed in the community membership guidelines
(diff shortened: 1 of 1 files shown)


skitt:
I’m not sure this is appropriate; signing off commits as per the DCO is an action to be taken on every commit, it doesn’t have much to do with membership... Put another way, we need to ask all contributors to sign off their commits as per the DCO, not only members.

nyechiel:
Thanks for the feedback, @skitt. The intent was to cover the general case with https://github.com/submariner-io/submariner-website/pull/182, and use this one as an extra check. We can't control all contributors, but we do have more control over members. It's not either/or.

skitt:
> Thanks for the feedback, @skitt. The intent was to cover the general case with [submariner-io/submariner-website#182](https://github.com/submariner-io/submariner-website/pull/182), and use this one as an extra check. We can't control all contributors, but we do have more control over members. It's not either/or. That said, I agree that this is not a membership item per-se.

My point is that this isn’t an attribute that should be assigned to contributors (“signs off their commits”), it’s an attribute that should be assigned to commits (“is signed off”). We mustn’t merge commits which aren’t signed off; we can’t control contributors, but we can control contributions.

nyechiel:
> > Thanks for the feedback, @skitt. The intent was to cover the general case with [submariner-io/submariner-website#182](https://github.com/submariner-io/submariner-website/pull/182), and use this one as an extra check. We can't control all contributors, but we do have more control over members. It's not either/or. That said, I agree that this is not a membership item per-se.
> 
> My point is that this isn’t an attribute that should be assigned to contributors (“signs off their commits”), it’s an attribute that should be assigned to commits (“is signed off”). We mustn’t merge commits which aren’t signed off; we can’t control contributors, but we can control contributions.

Ack. FYI - @dfarrell07  just posted https://github.com/submariner-io/submariner-website/pull/184 which seems to be the appropriate way of verifying this going forward. I am going to abandon this change now.

PR #667, opened 2020-06-19
Cache Go modules

skitt:
See
https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows
for details.


skitt:
I’m not sure this will work as-is with Dapper, let’s see what happens...

tpantelis:
bump

PR #687, opened 2020-06-26
Update Gateway StatusFailure when ListConnections fails

tpantelis:
The code intended to update `StatusFailure` but instead returned the error thus eliding the update.
    
Also handle nil from `ListConnections` to be safe.

This PR also adds unit tests for the cable engine `Syncer` and corresponding changes in a series of small commits.

### DIFF
--- a/pkg/cableengine/syncer/syncer.go
+++ b/pkg/cableengine/syncer/syncer.go
@@ -23,8 +23,9 @@ type GatewaySyncer struct {
 	version string
 }
 
-const GatewayUpdateIntervalSeconds = 5
-const GatewayStaleTimeoutSeconds = GatewayUpdateIntervalSeconds * 3
+var GatewayUpdateInterval = 5 * time.Second
+var GatewayStaleTimeout = time.Duration(GatewayUpdateInterval.Seconds()) * 3 * time.Second
+
 const updateTimestampAnnotation = "update-timestamp"
 
 // NewEngine creates a new Engine for the local cluster
@@ -40,7 +41,7 @@ func NewGatewaySyncer(engine cableengine.Engine, client v1typed.GatewayInterface
 
 func (s *GatewaySyncer) Run(stopCh <-chan struct{}) {
 	go func() {
-		wait.Until(s.syncGatewayStatus, GatewayUpdateIntervalSeconds*time.Second, stopCh)
+		wait.Until(s.syncGatewayStatus, GatewayUpdateInterval, stopCh)
 		s.CleanupGatewayEntry()
 	}()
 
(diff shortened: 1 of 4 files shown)


pkg/cableengine/syncer/syncer.go
```diff
@@ -23,8 +23,9 @@ type GatewaySyncer struct {
 	version string
 }
 
-const GatewayUpdateIntervalSeconds = 5
-const GatewayStaleTimeoutSeconds = GatewayUpdateIntervalSeconds * 3
+var GatewayUpdateInterval = 5 * time.Second
+var GatewayStaleTimeout = time.Duration(GatewayUpdateInterval.Seconds()) * 3 * time.Second
```
mangelajo, line 28: Isn't it simpler to keep L27 as *3 ? (it would be time unit independent)

pkg/cableengine/syncer/syncer.go
```diff
@@ -129,11 +129,6 @@ func (i *GatewaySyncer) cleanupStaleGatewayEntries(localGatewayName string) erro
 }
 
 func isGatewayStale(gateway v1.Gateway) (bool, error) {
-
-	if gateway.ObjectMeta.Annotations == nil {
```
mangelajo, line 127: Nice, I didn't know this :)

pkg/cableengine/syncer/syncer.go
```diff
@@ -53,7 +54,7 @@ func (i *GatewaySyncer) syncGatewayStatus() {
 
 	gatewayObj, err := i.generateGatewayObject()
 	if err != nil {
-		klog.Errorf("Error generating Gateway object: %s", err)
+		utilruntime.HandleError(fmt.Errorf("Error generating Gateway object: %s", err))
```
mangelajo, line 57: Nit:

In our case HandleError doesn't buy much:
   * checks for nil (we already need to do that for return)
   * does error backoff if errors happen <1ms , but this function is time gated.

pkg/cableengine/syncer/syncer.go
```diff
@@ -160,11 +160,14 @@ func (i *GatewaySyncer) generateGatewayObject() (*v1.Gateway, error) {
 
 	connections, err := i.engine.ListCableConnections()
 	if err != nil {
-		gateway.Status.StatusFailure = fmt.Sprintf("Error getting driver connections: %s", err)
-		klog.Errorf("error getting driver connections: %s", err)
-		return nil, err
+		gateway.Status.StatusFailure = fmt.Sprintf("Error retrieving driver connections: %s", err)
```
mangelajo, line 163: :+1:

pkg/cableengine/syncer/syncer.go
tpantelis: yeah - you're right -I made that more difficult - not sure what I was thinking 😄  - I'll change it

pkg/cableengine/syncer/syncer.go
tpantelis: Using `HandleError` gives the unit tests a mechanism to verify something when testing error paths. Also from the docs "_Invoking this method is preferable to logging the error - the default behavior is to log but the errors may be sent to a remote server for analysis."_ .  This is used in other places in the submariner code base as well by the original author(s).

PR #693, opened 2020-06-30
Add tunnel controller tests

tpantelis:
Preceded by some minor cleanup commits to simplify the tests.

### DIFF
--- a/pkg/controllers/tunnel/tunnel_test.go
+++ b/pkg/controllers/tunnel/tunnel_test.go
@@ -1,0 +1,144 @@
+package tunnel_test
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	fakeEngine "github.com/submariner-io/submariner/pkg/cableengine/fake"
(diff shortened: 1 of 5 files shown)


pkg/controllers/tunnel/tunnel_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog"
+)
+
+const (
+	namespace = "submariner"
+)
+
+var _ = Describe("", func() {
+	klog.InitFlags(nil)
```
skitt, line 27: Shouldn’t this be an `init()` function?

pkg/controllers/tunnel/tunnel_test.go
```diff
@@ -0,0 +1,144 @@
...
+})
+
+var _ = Describe("", func() {
```
skitt, line 30:
```suggestion
var _ = Describe("managing a tunnel", func() {
```

pkg/controllers/tunnel/tunnel_test.go
tpantelis: This has been brought up in other PRs 😄 It fails if inside an `init` function or inside the main Test function. I forget the exact reason why.

pkg/controllers/tunnel/tunnel_test.go
tpantelis: I made it "Managing tunnels".

PR #695, opened 2020-07-01
Increase e2e operation timeout

sridhargaddam:

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -41,6 +41,7 @@ func beforeSuite() {
 	}
 
 	queryAndUpdateGlobalnetStatus()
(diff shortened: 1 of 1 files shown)


test/e2e/framework/framework.go
```diff
@@ -41,6 +41,7 @@ func beforeSuite() {
 	}
 
 	queryAndUpdateGlobalnetStatus()
+	framework.TestContext.OperationTimeout = 200
```
tpantelis, line 44: This would override a setting on the command line. Why not make this the default in shipyard?

sridhargaddam: Sure @tpantelis. I pushed the following PR in shipyard - https://github.com/submariner-io/shipyard/pull/250

PR #698, opened 2020-07-01
Add golangci-lint config file, add linters

dfarrell07:
Explicitly declare linters to for golangci-lint to enable. This will
keep our checks more consistent, but also is required for adding
non-standard linters.

This adds linters:

* bodyclose: checks whether HTTP response body is closed successfully
* depguard: Go linter that checks if package imports are in a list of
  acceptable packages
* gocyclo: Computes and checks the cyclomatic complexity of functions
* gofmt: Gofmt checks whether code was gofmt-ed. By default this tool
  runs with -s option to check for code simplification
* goimports: Goimports does everything that gofmt does. Additionally it
  checks unused imports
* nakedret: Finds naked returns in functions greater than a specified
  function length
* unconvert: Remove unnecessary type conversions

Other linters from the golangci-lint example configuration that seem
desirable but currently fail are included but left commented out. We
should enable them as we fix violations in future changes.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,37 @@
+linters-settings:
+  gocyclo:
+    min-complexity: 20
+linters:
+  disable-all: true
+  enable:
+    - bodyclose
+    - deadcode
+    - depguard
+    #- dupl
+    - errcheck
+    #- funlen
+    #- gochecknoinits
+    #- goconst
+    #- gocritic
+    - gocyclo
+    - gofmt
+    - goimports
+    #- golint
+    #- gosec
+    - gosimple
+    - govet
+    - ineffassign
+    #- interfacer
+    #- lll
+    #- misspell
+    - nakedret
+    #- scopelint
+    - staticcheck
+    - structcheck
+    #- stylecheck
+    - typecheck
+    - unconvert
+    #- unparam
+    - unused
+    - varcheck
+    #- whitespace


tpantelis:
@dfarrell07 What are the default linters that are run when none are specified?

tpantelis:
> @dfarrell07 What are the default linters that are run when none are specified?

I found it:

_golangci-lint help linters_
Enabled by default linters:
**deadcode**: Finds unused code [fast: true, auto-fix: false]
**errcheck**: Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases [fast: true, auto-fix: false]
**gosimple** (megacheck): Linter for Go source code that specializes in simplifying a code [fast: true, auto-fix: false]
**govet** (vet, vetshadow): Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string [fast: true, auto-fix: false]
**ineffassign**: Detects when assignments to existing variables are not used [fast: true, auto-fix: false]
**staticcheck** (megacheck): Staticcheck is a go vet on steroids, applying a ton of static analysis checks [fast: true, auto-fix: false]
**structcheck**: Finds unused struct fields [fast: true, auto-fix: false]
**typecheck**: Like the front-end of a Go compiler, parses and type-checks Go code [fast: true, auto-fix: false]
unused (megacheck): Checks Go code for unused constants, variables, functions and types [fast: false, auto-fix: false]
**varcheck**: Finds unused global variables and constants [fast: true, auto-fix: false]

.golangci.yml
```diff
@@ -0,0 +1,34 @@
...
+    - bodyclose
+    - deadcode
+    - depguard
+    #- dupl
+    - errcheck
+    #- funlen
+    #- gochecknoinits
+    #- goconst
+    #- gocritic
+    - gocyclo
```
tpantelis, line 13: Based on my past experiences with cyclomatic complexity and considering some of the rather large functions in our code base (eg `processNextEndpoint` and `reconcileRoutes` in _route.go_), I'm surprised this passes - luckily the default `min-complexity` threshold of 30 is rather large 😄 If you set it to < 20, we start getting failures.  We can enable it although it would be more useful with a lower threshold - I think 20 would be reasonable. `reconcileRoutes` is at 19 so two more  branches would require refactoring but that's not a bad thing 😄 

```
linters-settings:
  gocyclo:
    min-complexity: 20
```

dfarrell07: Good feedback @tpantelis, thanks! I reduced the threshold to 20 in the latest version.

dfarrell07:
> @dfarrell07 What are the default linters that are run when none are specified?

I opened https://github.com/submariner-io/shipyard/pull/241 to clarify this in the future.

PR #708, opened 2020-07-06
Remove unnecessary use of Sprintf

dfarrell07:
The latest release of golangci-lint (1.28.0) now flags this as an error
via the newer version of the gosimple linter. Fix before upgrading the
version in Shipyard to keep the verify job here passing.


dfarrell07:
Seems like there's some temporary flakiness in the Fedora image pull.

PR #709, opened 2020-07-06
Avoid exposing range pointers

skitt:
When looping over a range, if a value variable is declared, the range
can be copied before processing; this means that passing a pointer to
a range value can cause issues. Using the index to address slice items
avoids a copy and means that pointers to slice items are canonical.

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -792,15 +792,15 @@ func (r *Controller) cleanVxSubmarinerRoutes() {
 		klog.Errorf("Unable to cleanup routes, error retrieving routes on the link %s: %v", VxLANIface, err)
 		return
 	}
-	for _, route := range currentRouteList {
-		klog.V(log.DEBUG).Infof("Processing route %v", route)
-		if route.Dst == nil || route.Gw == nil {
+	for i := range currentRouteList {
+		klog.V(log.DEBUG).Infof("Processing route %v", currentRouteList[i])
+		if currentRouteList[i].Dst == nil || currentRouteList[i].Gw == nil {
 			klog.V(log.DEBUG).Infof("Found nil gw or dst")
 		} else {
-			if r.remoteSubnets.Contains(route.Dst.String()) {
-				klog.V(log.DEBUG).Infof("Removing route %s", route.String())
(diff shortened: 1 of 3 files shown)


dfarrell07:
Testing with golangci-lint 1.28.0, after these fixes the linter continues on with some additional errors.

```
pkg/routeagent/controllers/route/route.go:838:34: G601: Implicit memory aliasing in for loop. (gosec)
		if err = netlink.XfrmPolicyDel(&xfrmPolicy); err != nil {
		                               ^
pkg/routeagent/controllers/route/route.go:870:31: G601: Implicit memory aliasing in for loop. (gosec)
				if err = netlink.RouteDel(&route); err != nil {
```

I'd like to figure out how to configure it to print every error, or at least more of them, and fix that in Shipyard.

skitt:
Thanks, I’ve added a few more, and Submariner now passes validation with 1.28 (with your `Sprintf` patch).

pkg/routeagent/controllers/route/route.go
```diff
@@ -792,15 +792,15 @@ func (r *Controller) cleanVxSubmarinerRoutes() {
...
+		klog.V(log.DEBUG).Infof("Processing route %v", currentRouteList[i])
+		if currentRouteList[i].Dst == nil || currentRouteList[i].Gw == nil {
 			klog.V(log.DEBUG).Infof("Found nil gw or dst")
 		} else {
-			if r.remoteSubnets.Contains(route.Dst.String()) {
-				klog.V(log.DEBUG).Infof("Removing route %s", route.String())
-				if err = netlink.RouteDel(&route); err != nil {
-					klog.Errorf("Error removing route %s: %v", route.String(), err)
+			if r.remoteSubnets.Contains(currentRouteList[i].Dst.String()) {
+				klog.V(log.DEBUG).Infof("Removing route %s", currentRouteList[i].String())
```
tpantelis, line 801:
```suggestion
				klog.V(log.DEBUG).Infof("Removing route %s", currentRouteList[i])
```

I know the call to `String` was pre-existing but it would be nice to remove it since you already touched the line 😄 Also on the other lines below.

skitt: Done!

PR #711, opened 2020-07-06
Enable maligned memory linting, fix issue

dfarrell07:
Saves a few bytes of memory by aligning structures with themselves
better, without requiring padding.


dfarrell07:
This will fail before https://github.com/submariner-io/shipyard/pull/245 is merged and propogated.

PR #715, opened 2020-07-09
Libreswan: simplify connection removal

skitt:
Comparing the index and the length isn't particularly useful, and
confuses linters (and reviewers).


skitt:
@mangelajo I can’t request a review without messaging people ;-).

pkg/cable/libreswan/libreswan.go
```diff
@@ -320,11 +320,14 @@ func (i *libreswan) DisconnectFromEndpoint(endpoint types.SubmarinerEndpoint) er
...
+				// Removing the only connection
 				return []subv1.Connection{}
 			} else if j == 0 {
+				// Removing the first connection
 				return connections[j+1:]
 			} else if j == len(connections)-1 {
+				// Removing the last connection
 				return connections[:j]
 			}
 			return append(connections[:j], connections[j+1:]...)
```
tpantelis, line 333: I think simpler way to do this:

```suggestion
			copy(connections[j:], connections[j+1:])
			return connections[:len(connections)-1]
```

skitt: Thanks, this is much better!

PR #716, opened 2020-07-09
Enable unparam unused param linting, fix error

dfarrell07:

### DIFF
--- a/pkg/cableengine/syncer/syncer.go
+++ b/pkg/cableengine/syncer/syncer.go
@@ -51,11 +51,7 @@ func (s *GatewaySyncer) Run(stopCh <-chan struct{}) {
 func (i *GatewaySyncer) syncGatewayStatus() {
 	klog.V(log.TRACE).Info("Running Gateway status sync")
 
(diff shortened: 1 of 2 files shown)


pkg/cableengine/syncer/syncer.go
```diff
@@ -169,7 +169,7 @@ func (i *GatewaySyncer) generateGatewayObject() (*v1.Gateway, error) {
 	}
 
 	klog.V(log.TRACE).Infof("Generated Gateway object: %+v", gateway)
-	return &gateway, nil
+	return &gateway, err
```
tpantelis, line 172: We want to return nil here - we only use `err` to set `StatusFailure` on L162. This causes a unit test to fail (yay for regression tests 😄 ).  We could ignore the linting error or  set `err` to nil after L163 to silence the error. However since we now always return nil error, we might as well remove it from the return signature. If any error occurred generating the object, we'd want to report that in the `StatusFailure` anyway.

dfarrell07: Done in the latest, thanks @tpantelis!

PR #718, opened 2020-07-10
Update shipyard version in go.mod

tpantelis:


tpantelis:
The 2 job failures are caused by "_manifest for registry.fedoraproject.org/fedora-minimal:32 not found: manifest unknown: manifest unknown_" - seems random, intermittent.

Issue #694, opened 2020-07-01
Shutdown path to be Optimized in Submariner Engine and route-agent


sridhargaddam:
**What happened**:
As part of this PR [https://github.com/submariner-io/submariner-operator/pull/445], we configured terminationGracePeriodSeconds to 10 secs. However, it was seen that on some occasions SM Engine Pod taking longer than 10 secs, so K8s is terminating the Pod with SIGKILL. Have to investigate and understand if we can improve the behavior.

**What you expected to happen**:
The shutdown path should be optimized.

**How to reproduce it (as minimally and precisely as possible)**:
Monitor the time it takes for Engine Pod to restart when the Pod is terminated. We can also check the behavior by removing the Gateway label from the node and see the amount of time its taking for the Engine Pod to terminate.

PR #727, opened 2020-07-19
Fix shutdown path in Submariner Gateway Pod

sridhargaddam:
In the current code, it was seen that GatewayPod was not handling the
shutdown signal properly and was finally getting terminated by K8s
after the terminationGracePeriod interval is exhaused. This is because
leaderelection.RunOrDie does not return. This PR fixes it by running
the leaderelection in a Goroutine and using stopCh to wait for shutdown
signal.

Partially fixes: https://github.com/submariner-io/submariner/issues/694

### DIFF
--- a/main.go
+++ b/main.go
@@ -193,7 +193,10 @@ func main() {
 		klog.Fatalf("Leader election lost, shutting down")
 	}
 
-	startLeaderElection(leClient, recorder, becameLeader, lostLeader)
+	go startLeaderElection(leClient, recorder, becameLeader, lostLeader)
+	<-stopCh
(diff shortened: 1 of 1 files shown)


tpantelis:
@sridhargaddam  I assume this was b/c `leaderelection.RunOrDie` never returned - can you update the PR summary with a brief explanation as to why?

sridhargaddam:
> @sridhargaddam I assume this was b/c `leaderelection.RunOrDie` never returned - can you update the PR summary with a brief explanation as to why?

Yes @tpantelis `leaderelection.RunOrDie` does not return. Updated the commit message, please take a look. Thanks for reviewing.

Issue #726, opened 2020-07-19
Globalnet pod in error state after joining clusters


manosnoam:
After joining clusters, Globalnet pod status is _ERROR_ or _CrashLoopBackOff_.
This is a sporadic error, which happens once in a while on a new Submariner deployments:

https://qe-jenkins-csb-skynet.cloud.paas.psi.redhat.com/job/Submariner-OSP-AWS/755/Test-Report/
https://qe-jenkins-csb-skynet.cloud.paas.psi.redhat.com/job/debug_job/36/Test-Report/

**What happened**:

Deploying and joining clusters works as expected:

```
$ subctl deploy-broker  --service-discovery --kubecontext nmanos-cluster-a --globalnet

 • Deploying broker  ...
 ✓ Deploying broker
 • Creating broker-info.subm file  ...
 ✓ Creating broker-info.subm file
 ✓ Backed up previous broker-info.subm to broker-info.subm.2020-07-19T08_21_08+03_00
 ⚠ Reusing IPsec PSK from existing broker-info.subm

$ subctl join --clusterid nmanos-cluster-a   ./broker-info.subm    --ikeport 501 --nattport 4501 --version devel --subm-debug

* ./broker-info.subm says broker is at: https://api.nmanos-cluster-a.devcluster.openshift.com:6443
* There are 1 labeled nodes in the cluster:
  - ip-10-166-73-131.us-east-2.compute.internal
 • Discovering network details  ...
    Discovered network details:
        Network plugin:  OpenShift
        Service CIDRs:   [100.96.0.0/16]
        Cluster CIDRs:   [10.252.0.0/14]
 ✓ Discovering network details
 • Discovering multi cluster details  ...
 • Validating Globalnet configurations  ...
 ✓ Validating Globalnet configurations
 • Assigning Globalnet IPs  ...
 ✓ Assigning Globalnet IPs
 ✓ Allocated GlobalCIDR: 169.254.0.0/19
 ✓ Discovering multi cluster details
 • Deploying the Submariner operator  ...
 ✓ Deploying the Submariner operator
 ✓ Created operator CRDs
 ✓ Created operator namespace: submariner-operator
 ✓ Created operator service account and role
 ✓ Updated the privileged SCC
 ✓ Created lighthouse CRDs
 ✓ Created lighthouse service account and role
 ✓ Updated the privileged SCC
 ✓ Created Lighthouse service accounts and roles
 ✓ Deployed the operator successfully
 • Creating SA for cluster  ...
[...]

PR #735, opened 2020-07-21
During initialization, if clusterCRD is missing, retry few times

sridhargaddam:
During the Submariner installation phase (aka when join command is issued),
typically all the Submariner pods would be initialized in parallel.
Submariner-gateway Pod, being one of them, is responsible for creating the
clusterCRD and the creation of this CRD would normally take a few seconds.
In the meantime, if GlobalnetPod tries to read the clusterCRD and finds it
missing, it should retry the operation few times before giving up. This PR
implements this change.

Fixes issue: https://github.com/submariner-io/submariner/issues/726

### DIFF
--- a/pkg/globalnet/main.go
+++ b/pkg/globalnet/main.go
@@ -4,8 +4,10 @@ import (
 	"flag"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/kelseyhightower/envconfig"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/globalnet/controllers/ipam"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/globalnet/main.go
```diff
@@ -45,7 +47,16 @@ func main() {
...
-	localCluster, err := submarinerClient.SubmarinerV1().Clusters(ipamSpec.Namespace).Get(ipamSpec.ClusterID, metav1.GetOptions{})
+	var localCluster *submarinerv1.Cluster
+	// During installation, sometimes creation of clusterCRD by submariner-gateway-pod would take few secs.
+	for i := 0; i < 5; i++ {
+		localCluster, err = submarinerClient.SubmarinerV1().Clusters(ipamSpec.Namespace).Get(ipamSpec.ClusterID, metav1.GetOptions{})
+		if err == nil {
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
```
tpantelis, line 58: Could 5 sec be fragile? Can we wait indefinitely - perhaps use a simple one-time informer with just an AddFunc and no queue?

sridhargaddam: @tpantelis: As you know, this can happen ONLY during the very first installation of Submariner and what we have seen is that this does **not** happen frequently. TBH, I've never seen it with the KIND setup (both helm and operator deployments). IMHO we can have this 5 secs retry and if we continue to see the issue even after this change, we can revisit in future. I'm personally inclined with 5 seconds retry rather than indefinite as it will be an indication for the users too that globalnetPod went into an error - otherwise, they have to look at the logs to figure out.

tpantelis: ok

PR #741, opened 2020-07-27
Use the main Fedora image

skitt:
... to avoid issues with the Fedora registry. Given that the libreswan
package pulls in Python, the practical differences between the minimal
and main images aren’t significant for us anyway.


dfarrell07:
Thanks @skitt (and glad to have you back/hope you had a good break)!

> /bin/sh: microdnf: command not found

PR #742, opened 2020-07-27
Enable wsl positive whitespace linting, fix errors

dfarrell07:


tpantelis:
The exclusions seem reasonable. Let's see what others think.

dfarrell07:
> I wonder if IDEs can be configured to auto-honor this

I know golangci-lint has IDE integrations. Are Go IDEs are auto-creating code that violates these rules?

> or if this could become a hassle to contribution, or if we can provide some sort of script or make target which would fix this.

I'd expect people to run `make verify` as they work on their PR, which would flag it. As far as auto-fixing it, that seems like a contribution we'd want to make to golangci-lint. Currently wsl auto-fix isn't supported (can see in `golangci-lint linters`).

mangelajo:
I propose we go ahead, merge it... but keep an eye on how it reacts with people IDEs, or if people has issues tuning to it, we can always consider reverting any of the lintings, would that make sense @dfarrell07  ?

dfarrell07:
> I propose we go ahead, merge it... but keep an eye on how it reacts with people IDEs, or if people has issues tuning to it, we can always consider reverting any of the lintings, would that make sense @dfarrell07 ?

Yep, makes sense to me! I'm happy to continue changing the rules/linters as people get some time with them, see if they like the details.

Issue #683, opened 2020-06-25
Libreswan cable driver does not work with non-standard IPSEC ports


mangelajo:
**What happened**:

When specifying non standard ipsec ports libreswan/pluto will target the standard 500 port / 4500..
```
12:05:50.405473 IP ip-10-166-76-227.us-east-2.compute.internal.501 > x.x.x.x.500: isakmp: parent_sa ikev2_init[I]
```

The connection of the gateways will not succeed (`kubectl get Gateways -A -o yaml`)

(see 501 in source but 500 in dst)

**What you expected to happen**:

(501 in source and **501** in dst)

**How to reproduce it (as minimally and precisely as possible)**:

Deploy submariner with the libreswan cabledriver and non-standard ports:

`subctl join broker-config.subm --cable-driver libreswan --ikeport 501 --nattport 4501

**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`):
subctl version: v0.4.0
built from git ref: v0.4.0
- Install tools: subctl

letoams:
With libreswan 3.33 (soon to be released) there will be the option for leftikeport= and rightikeport=
And whack will be able to specify: ipsec whack --name conn [...] --ikeport XXX  --to  [...]  --ikeport YYY    
The code for this is now in the git main branch (last commit via https://github.com/libreswan/libreswan/commit/e1cebcc60d6a3993ff0816e87291420a8ecab5cd

mangelajo:
Added this one #746 to allow us to rebuild the submariner image with libreswan from git on fedora32

mangelajo:
Ok I'm testing and this is what I see:
`
bash-5.0# tcpdump -i eth0 port 501 or port 4501 or port 500 or port 4501
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
14:17:54.400083 IP cluster2-worker.4501 > ec2-18-224-183-91.us-east-2.compute.amazonaws.com.4501: UDP, length 1
14:17:54.400207 IP cluster2-worker.4501 > ec2-18-219-82-148.us-east-2.compute.amazonaws.com.4501: UDP, length 1
14:17:54.400257 IP cluster2-worker.4501 > ec2-18-224-183-91.us-east-2.compute.amazonaws.com.4501: UDP, length 1
[...]

PR #747, opened 2020-08-04
Libreswan 4.00 ikeports

mangelajo:
Related-Issue: #683


mangelajo:
Still not fully tested

Probably I should remove the true/false and just keep this patch ready for when libreswan is bumped.

mangelajo:
> If we want to avoid tying this to a specific Libreswan package upgrade, we could check the version at runtime: `ipsec --version` outputs a string of the form `Linux Libreswan 3.32 (netkey) on 5.7.7-100.fc31.x86_64`.

Yeah, althought there is some problem with the libreswan logic to generate the packages (at least from git), it generates 3.30-xxxxxx-nnn 

I tried to do auto-detection based on the pluto output, but the exit error in that case is asynchronouseon the go func... and it didn't work.

skitt:
> > If we want to avoid tying this to a specific Libreswan package upgrade, we could check the version at runtime: `ipsec --version` outputs a string of the form `Linux Libreswan 3.32 (netkey) on 5.7.7-100.fc31.x86_64`.
> 
> Yeah, althought there is some problem with the libreswan logic to generate the packages (at least from git), it generates 3.30-xxxxxx-nnn

Ah, yes, the tags aren’t on the main branch after 3.30 :-/.

> I tried to do auto-detection based on the pluto output, but the exit error in that case is asynchronouseon the go func... and it didn't work.

How about `ipsec pluto --help` and `ipsec whack --help`, and looking for `ikeport` in the output (on stderr)?

mangelajo:
> > > If we want to avoid tying this to a specific Libreswan package upgrade, we could check the version at runtime: `ipsec --version` outputs a string of the form `Linux Libreswan 3.32 (netkey) on 5.7.7-100.fc31.x86_64`.
> > 
> > 
> > Yeah, althought there is some problem with the libreswan logic to generate the packages (at least from git), it generates 3.30-xxxxxx-nnn
> 
> Ah, yes, the tags aren’t on the main branch after 3.30 :-/.
> 
> > I tried to do auto-detection based on the pluto output, but the exit error in that case is asynchronouseon the go func... and it didn't work.
> 
> How about `ipsec pluto --help` and `ipsec whack --help`, and looking for `ikeport` in the output (on stderr)?

I think it's simpler if we just wait for the release, although the transition will be less seamless but we will have 2 benefits:
1) Knowing exactly when we got the new libreswan
2) The code is simpler

As you said it's actually easy to do if we look at help.

mangelajo:
@skitt If you think seamless transition is better I can handle it too.. I'm just not sure ':D

skitt:
Test PR in https://github.com/submariner-io/submariner/pull/904

pkg/cable/libreswan/libreswan.go
```diff
@@ -262,13 +262,15 @@ func (i *libreswan) ConnectToEndpoint(endpoint types.SubmarinerEndpoint) (string
 				args = append(args, "--id", localEndpointIdentifier)
 				args = append(args, "--host", localEndpointIP)
 				args = append(args, "--client", leftSubnets[lsi])
+				args = append(args, "--ikeport", i.ipSecNATTPort)
```
skitt, line 265: Doesn’t this need to be handled differently for NAT and non-NAT?

```suggestion
if endpoint.Spec.NATEnabled {
    args = append(args, "--ikeport", i.ipSecNATTPort)
} else {
    args = append(args, "--ikeport", i.ipSecPort)
}
```

and same again for the right-hand side.

skitt (changes requested):

pkg/cable/libreswan/libreswan.go
mangelajo: Yes,  that makes sense.

Honestly I guess it doesn't really matter if we always go over a single port and remove the concept of nat/non-nat port ? 

I believe IKEv2 can do both over 4500 , may be all over "ipSecPort" and that's it?

skitt: Yup, e2e works over 4500 in non-NAT mode.

mangelajo:
Let's handle the bits here as part of : https://github.com/submariner-io/submariner/pull/904

Issue #749, opened 2020-08-05
AWS: Route-agent pod gets into Error state 


sridhargaddam:
**What happened**:
On an AWS Setup, when Submariner is deployed, one of the routeagent pods get into Error state with the following log message.

_Unable to create VxLAN interface on gateway node (ip-10-0-69-158): failed to add FDB entry on the Gateway Node vxlan iface invalid ipAddress (<nil>) supplied._

**How to reproduce it (as minimally and precisely as possible)**:
This is not reproduced all the time and is a bit random.

PR #750, opened 2020-08-05
Fix error in routeagent-pod while configuring vxLan tunnels

sridhargaddam:
Currently, while routeagent is running, it queries the list of
other routeagent pods in the cluster to get their ip-addresses.
This info is necessary while creating the VxLAN tunnels in the
cluster. In one of the clusters, it was seen that when routeagent
pods are queried, the pod.Status.PodIP was empty. The empty check
was not handled in the code and this was leading the routeagent
pod not coming up on the Gateway node. This PR fixes it.

Fixes issue: https://github.com/submariner-io/submariner/issues/749

### DIFF
--- a/pkg/routeagent/controllers/route/route.go
+++ b/pkg/routeagent/controllers/route/route.go
@@ -243,7 +243,11 @@ func (r *Controller) Run(stopCh <-chan struct{}) error {
 
 	for index, pod := range podList.Items {
 		klog.V(log.DEBUG).Infof("In %s, podIP of submariner-route-agent[%d] is %s", r.clusterID, index, pod.Status.PodIP)
-		r.populateRemoteVtepIps(pod.Status.PodIP)
+		// If Pod is not yet assigned with an ip-address it will be processed subsequently during pod update event.
+		if pod.Status.PodIP != "" {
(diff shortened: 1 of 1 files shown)


pkg/routeagent/controllers/route/route.go
```diff
@@ -516,6 +520,11 @@ func (r *Controller) processNextPod() bool {
 			return fmt.Errorf("error retrieving submariner-route-agent pod object %s: %v", name, err)
 		}
 
+		if pod.Status.PodIP == "" {
+			r.podWorkqueue.Forget(obj)
+			return fmt.Errorf("submariner-route-agent pod %q does not have an ip-address yet", name)
```
mangelajo, line 525: May be we should consider failing hard Fatalf if this happens more than X times so the route-agent is re-created ?

sridhargaddam: IMHO this is not required as K8s would assign the IPAddress to the Pod in few ms/secs. This will be seen only when the route-agent pods are coming up.

mangelajo: Ok, I guess that would be a broken k8s, may be it's too defensive.

Issue #721, opened 2020-07-14
Datapath goes down when globalnet controller pod is restarted.


sridhargaddam:
**What happened**:

In the current implementation of Globalnet, when the Globalnet Pod is restarted, the datapath fails as Globalnet controller flushes the IPTable chains. The chains are repopulated once the Globalnet Pod comes up. 

**What you expected to happen**:
Ideally, it would be good if the datapath is not disrupted when the Globalnet Pod is restarted.

**How to reproduce it (as minimally and precisely as possible)**:
In a stable environment that includes Globalnet controller, establish a connection from a Pod in one cluster to a GlobalIP of remoteService. Then restart the Globalnet Pod, the connection will be disrupted.

PR #762, opened 2020-08-18
Preserve datapath when GlobalnetPod is restarted

sridhargaddam:
In the current implementation of Globalnet, when the Globalnet Pod
is restarted, the datapath goes down as Globalnet controller flushes
the IPTable chains that support inter-cluster connectivity.

In this PR we move the cleanup to RouteAgent Pod which does the cleanup
(aka flushing of iptable chains) when there is a Gateway migration.

This PR also moves the IPTable Chain names to routeAgent code to avoid
cyclic dependency and to have all the chain names at a single place.

Fixes Issue: https://github.com/submariner-io/submariner/issues/721

### DIFF
--- a/pkg/globalnet/cleanup/gn_cleanup.go
+++ b/pkg/globalnet/cleanup/gn_cleanup.go
@@ -21,21 +21,21 @@ const (
 
 type GlobalnetStatus int
 
-func GlobalnetCleanupHandler(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) []cleanup.Handler {
+func GetGlobalnetCleanupHandlers(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) []cleanup.Handler {
 	return []cleanup.Handler{
-		NewGlobalnetCleanupHandler(clusterID, objectNamespace, submarinerClientSet),
+		newCleanupGlobalnetRules(clusterID, objectNamespace, submarinerClientSet),
 	}
 }
 
-type CleanupGlobalnetRules struct {
+type cleanupGlobalnetRules struct {
 	clusterID       string
 	objectNamespace string
 	clientSet       clientset.Interface
 	globalnetStatus GlobalnetStatus
 }
 
-func NewGlobalnetCleanupHandler(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) cleanup.Handler {
-	return &CleanupGlobalnetRules{
+func newCleanupGlobalnetRules(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) cleanup.Handler {
+	return &cleanupGlobalnetRules{
 		clusterID:       clusterID,
 		objectNamespace: objectNamespace,
 		clientSet:       submarinerClientSet,
@@ -43,15 +43,15 @@ func NewGlobalnetCleanupHandler(clusterID, objectNamespace string, submarinerCli
 	}
 }
 
-func (gn *CleanupGlobalnetRules) GetName() string {
-	return "Globalnet cleanup handler"
+func (gn *cleanupGlobalnetRules) GetName() string {
+	return "Globalnet rules cleanup handler"
 }
 
-func (gn *CleanupGlobalnetRules) NonGatewayCleanup() error {
+func (gn *cleanupGlobalnetRules) NonGatewayCleanup() error {
 	return nil
 }
 
-func (gn *CleanupGlobalnetRules) GatewayToNonGatewayTransition() error {
+func (gn *cleanupGlobalnetRules) GatewayToNonGatewayTransition() error {
 	if gn.globalnetStatus == GN_Status_Not_Verified {
 		localCluster, err := gn.clientSet.SubmarinerV1().Clusters(gn.objectNamespace).Get(gn.clusterID, metav1.GetOptions{})
 		if err != nil {
--- a/pkg/globalnet/cleanup/gn_cleanup.go
+++ b/pkg/globalnet/cleanup/gn_cleanup.go
@@ -1,0 +1,90 @@
+package cleanup
+
+import (
+	"fmt"
+
+	"github.com/coreos/go-iptables/iptables"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	clientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
+	"github.com/submariner-io/submariner/pkg/routeagent/cleanup"
+	"github.com/submariner-io/submariner/pkg/routeagent/constants"
+
+	"k8s.io/klog"
+)
+
+const (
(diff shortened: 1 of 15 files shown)


sridhargaddam:
This PR is now ready for review. Appreciate your review comments, thanks.

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -95,3 +96,43 @@ func (r *Controller) programIptableRulesForInterClusterTraffic(remoteCidrBlock s
...
+		localCluster, err := r.submarinerClientSet.SubmarinerV1().Clusters(r.objectNamespace).Get(r.clusterID, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("error while retrieving the local ClusterInfo: %v", err)
+		}
+
+		if localCluster.Spec.GlobalCIDR != nil && len(localCluster.Spec.GlobalCIDR) > 0 {
+			r.globalnetStatus = GN_Enabled
+		} else {
+			r.globalnetStatus = GN_Disabled
+		}
```
tpantelis, line 111: So `globalnetStatus` will only be evaluated once. Could the `GlobalCIDR` on the `Cluster` ever change? If so, is it assumed that would entail  a route-agent restart? Either way it doesn't seem like we really need this optimization to chance it since it's executed on local gateway EP change which should be infrequent.

sridhargaddam: AFAICT we do not support modifying globalnetStatus at runtime. IMHO having this would be helpful (as we can avoid querying cluster object) since it will be periodically invoked every 60 secs.

sridhargaddam (commented):

tpantelis (approved):

pkg/routeagent/controllers/route/iptables.go
```diff
@@ -95,3 +96,43 @@ func (r *Controller) programIptableRulesForInterClusterTraffic(remoteCidrBlock s
 
 	return nil
 }
+
+func (r *Controller) clearGlobalnetChains() error {
+	if r.globalnetStatus == GN_Status_Not_Verified {
```
mangelajo, line 101: The code seems in general good, but I propose an improvement.

Here we are moving the cleanup logic back into the route agent, which is what we want. But remember that we have technical debt still for this and ipsec too that we wanted to cleanup. We are adding more.

Why don't we design a simple interface for cleanups, and have a cleanup object in globalnet of which we can create an instance here,and just call it at the right hook points.

That way, we could later do the same for the different cable drivers. We'd load the "cleanup" interface of the cable driver, and use it.

We could even have a list of those objects, and when we call the hook for cleanup all execute. We could have a single interface with multiple events (of when a cleanup could need to happen...
- GatewayStarted, GatewayStopped, RouteAgentStopped, stuff like that. (I'm enumerating some random ones which may be make no sense).

mangelajo (changes requested):
Code seems fine, but I'm suggesting a restructure to start tackling the cleanup tech-debt we already had.

pkg/routeagent/controllers/route/iptables.go
sridhargaddam: Thanks for your review comments @mangelajo. I've updated the code, please take a look.

mangelajo (approved):
Super clean @sridhargaddam thanks!

mangelajo:
I'm asking to @tpantelis to re-review since the proposals made the code change considerably, although the low level is the same, the design is different, I believe @tpantelis will like it though.

pkg/globalnet/cleanup/gn_cleanup.go
```diff
@@ -0,0 +1,90 @@
...
+
+const (
+	GN_Status_Not_Verified = iota
+	GN_Enabled
+	GN_Disabled
+)
+
+type GlobalnetStatus int
+
+func GlobalnetCleanupHandler(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) []cleanup.Handler {
```
tpantelis, line 24:
```suggestion
func GetGlobalnetCleanupHandlers(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) []cleanup.Handler {
```

pkg/globalnet/cleanup/gn_cleanup.go
```diff
@@ -0,0 +1,90 @@
...
+	return []cleanup.Handler{
+		NewGlobalnetCleanupHandler(clusterID, objectNamespace, submarinerClientSet),
+	}
+}
+
+type CleanupGlobalnetRules struct {
```
tpantelis, line 30:
```suggestion
type cleanupGlobalnetRules struct {
```

pkg/globalnet/cleanup/gn_cleanup.go
```diff
@@ -0,0 +1,90 @@
...
+	clusterID       string
+	objectNamespace string
+	clientSet       clientset.Interface
+	globalnetStatus GlobalnetStatus
+}
+
+func NewGlobalnetCleanupHandler(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) cleanup.Handler {
```
tpantelis, line 37:
```suggestion
func newCleanupGlobalnetRules(clusterID, objectNamespace string, submarinerClientSet clientset.Interface) cleanup.Handler {
```

pkg/globalnet/cleanup/gn_cleanup.go
```diff
@@ -0,0 +1,90 @@
...
+	return &CleanupGlobalnetRules{
+		clusterID:       clusterID,
+		objectNamespace: objectNamespace,
+		clientSet:       submarinerClientSet,
+		globalnetStatus: GN_Status_Not_Verified,
+	}
+}
+
+func (gn *CleanupGlobalnetRules) GetName() string {
+	return "Globalnet cleanup handler"
```
tpantelis, line 47:
```suggestion
	return "Globalnet rules cleanup handler"
```

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -157,6 +154,20 @@ func (i *GatewayMonitor) processNextEndpoint() bool {
 			i.endpointWorkqueue.Forget(obj)
 
 			return nil
+		} else {
```
tpantelis, line 157: No need for else clause and extra indentation.

tpantelis (commented):
Just some minor style nits.

pkg/globalnet/cleanup/gn_cleanup.go
sridhargaddam: Done

sridhargaddam (commented):
Thanks for reviewing @mangelajo @tpantelis

tpantelis (approved):

PR #763, opened 2020-08-18
Add config for Git commit message linting

dfarrell07:
Allow commit messages with only a title, no body. Will be consumed by
Shipyard's validate script.


dfarrell07:
We're currently pinned to a version of Shipyard without https://github.com/submariner-io/shipyard/pull/281. This will fail until we go back to using the latest version.

mangelajo (approved):
+1 when we go back to shipyard:devel

PR #773, opened 2020-08-25
Remove a possible race condition during  WireGuard driver initialization 

roytman:
check https://github.com/submariner-io/submariner-operator/issues/461#issuecomment-680021600

### DIFF
--- a/main.go
+++ b/main.go
@@ -157,17 +157,13 @@ func main() {
 
 		submarinerInformerFactory.Start(stopCh)
 
+		if err = cableEngine.StartEngine(); err != nil {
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
       matrix:
         globalnet: ['', 'globalnet']
         deploytool: ['operator', 'helm']
-        cable_driver: ['', 'wireguard', 'libreswan']
```
dfarrell07: Can we squash these into one commit, so we don't have this intermediary step of removing other cable drivers in the git history?

mangelajo (approved):

PR #782, opened 2020-09-01
Reorganize GHA jobs/workflows

dfarrell07:
Per discussions on Automation Sync, reorganize the GitHub Action
Workflows and Jobs to make more sense.

"Validate" is too vague of a term, and it's not clear why some types of
linting are currently covered by it while others (or even e2e and unit)
aren't.

Use workflows linting, unit, e2e, and periodic. Group all linting jobs
under linting Workflow, getting rid of the validate Workflow.


.github/workflows/linting.yml
```diff
@@ -0,0 +1,79 @@
...
+        uses: actions/checkout@v1
+
+      - name: Run markdown-link-check
+        uses: gaurav-nelson/github-action-markdown-link-check@v1
+        with:
+          config-file: ".markdownlinkcheck.json"
+          check-modified-files-only: "yes"
+
+  validate:
+    name: Go and MD Linting
```
dfarrell07, line 60: I'd like to split these up into two different make targets and jobs (golangci-lint and markdownlint).

dfarrell07:
@mangelajo Can you update the required GHA config here?

mangelajo:
> @mangelajo Can you update the required GHA config here?

Ah I see, we can do that when this is merged. An admin will need to force merge, since the required will be different from the new... let's update once merged.

mangelajo:
The org looks good, an idea for improvement:

Do you think there'd be a way to have always the deployment method first on the list here:

image


so "globalnet" does not get in front of operator or helm

dfarrell07:
> The org looks good, an idea for improvement:
> 
> Do you think there'd be a way to have always the deployment method first on the list here:
> 
> image
> 
> so "globalnet" does not get in front of operator or helm

I'd love to be able to control the order it shows jobs, but I have no idea how. @mangelajo

mangelajo:
> > The org looks good, an idea for improvement:
> > Do you think there'd be a way to have always the deployment method first on the list here:
> > image
> > so "globalnet" does not get in front of operator or helm
> 
> I'd love to be able to control the order it shows jobs, but I have no idea how. @mangelajo

I think you just need to invert L18 & L19 here: https://github.com/submariner-io/submariner/blob/master/.github/workflows/e2e.yml#L18

dfarrell07:
> I think you just need to invert L18 & L19 here: https://github.com/submariner-io/submariner/blob/master/.github/workflows/e2e.yml#L18

I inverted them, how's that?

mangelajo (approved):
Looks much better, thanks

skitt (approved):

.github/workflows/e2e.yml
```diff
@@ -1,22 +1,26 @@
 ---
+name: End to End Tests
+
 on:
   pull_request:
+    branches:
+      - '**'
```
mkolesnik, line 7: Why is this needed?

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: tim-actions/get-pr-commits@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: DCO Check
+        uses: tim-actions/dco@master
+        with:
+          commits: ${{ steps.get-pr-commits.outputs.commits }}
+
+  gitlint:
+    name: Gitlint
```
mkolesnik, line 39:
```suggestion
    name: Commit Message(s)
```
Perhaps it's clearer the way it was, or just 'Commit Messages' or something.

Describe they 'why' not the 'what' ;)

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@master
+      - name: Run validation
+        run: make codegen
+
+  dco:
+    runs-on: ubuntu-latest
+    name: DCO Check
```
mkolesnik, line 26:
```suggestion
    name: DCO in Commit Message(s)
```

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: actions/checkout@v1
+
+      - name: Run markdown-link-check
+        uses: gaurav-nelson/github-action-markdown-link-check@v1
+        with:
+          config-file: ".markdownlinkcheck.json"
+          check-modified-files-only: "yes"
+
+  validate:
+    name: Go and MD Linting
```
mkolesnik, line 61:
```suggestion
    name: Go and Markdown
```

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
...
+  gitlint:
+    name: Gitlint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+      - run: make gitlint
+
+  markdown-link-check:
```
mkolesnik, line 47: I wonder if this can be a `make` target instead of a GHA only validation

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
...
+    name: Go and MD Linting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@master
+      - name: Run golangci-lint
+        run: make validate
+
+  yaml-lint:
+    name: YAML Linting
```
mkolesnik, line 70:
```suggestion
    name: YAML
```

.github/workflows/periodic.yml
```diff
@@ -1,13 +1,19 @@
...
+name: Periodic
+
 on:
   schedule:
     - cron: "0 0 * * 0"
+  push:
+    branches:
+      - master
+    tags:
+      - 'v**'
```
mkolesnik, line 11: Why would periodic run on push?
I don't think this makes sense, and looks like it could even confuse

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
...
+
+on:
+  pull_request:
+    branches:
+      - '**'
+  push:
+    branches:
+      - master
+    tags:
+      - 'v**'
```
mkolesnik, line 12: Is there a benefit to run these on push even?
They already run on each PR, so most of them should be fine on push.

Perhaps it only makes sense to run those that could've changed if the PR was sitting a whjile before push (like the MD links since that's affected by 3rd party sources)

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
...
+
+jobs:
+  code-gen:
+    name: Submariner K8s API Code Generation
```
mkolesnik, line 16: Now that I think of it, is this really a validation?

This generates the clientset and such, but what's the validation exactly?

Do we expect it to fail or something?

Or are we expecting git state to be clean after running it (because there's no such check currently here so it won't matter if anything changed or not)?

.github/workflows/linting.yml
```diff
@@ -0,0 +1,80 @@
+---
+name: Linting
+
+on:
+  pull_request:
+    branches:
+      - '**'
```
mkolesnik, line 7: Why is this needed?

mkolesnik (changes requested):
Thanks Daniel, I have a few comments and some questions.

I know some might be seem a bit nit-picky, but since the idea is to get things in order IMHO they make sense :)

.github/workflows/e2e.yml
dfarrell07: I was just trying to make it consistent, but you're right it's likely better to make it consistent by removing vs adding.

.github/workflows/linting.yml
dfarrell07: Good idea, done.

.github/workflows/linting.yml
dfarrell07: Done

.github/workflows/linting.yml
dfarrell07: Done

.github/workflows/linting.yml
dfarrell07: Done

.github/workflows/linting.yml
dfarrell07: Removed

.github/workflows/linting.yml
dfarrell07: If I understand this triggering correctly, it's only running when pushing tags. To me it makes sense to run absolutely everything when we're tagging.

.github/workflows/linting.yml
skitt: I think it’s useful to make sure the code generation process works (we’ve broken it in the past).

dfarrell07: Good questions, I'm not sure, I've basically carried that job over into this workflow as-is.

.github/workflows/linting.yml
dfarrell07: The main issue is that we want this check in the submariner-website repo, and that repo doesn't have the Shipyard shared infra.

mangelajo: I'd make it a make target if we can at some point (but I'd say that may be a separate PR while we keep the same name for the linting task), even if we use the GHA-only for the website, or... we can add dapper to the website if we want though..

I'm sometimes miss the opportunity to run that validation locally before sending into the outside woruld

.github/workflows/linting.yml
mangelajo: Yeah, I wondered that too, but I couldn't find a better category. We only want to validate that API code gen still works with our code.

.github/workflows/periodic.yml
mangelajo: Right, that part is confusing I agree, for the push it's not periodic anymore. So it would show as "Periodic" on a push.

Could it be part of Validation and that's it?

mangelajo: I think we have Validation/Markdown links for that, Id' leave push out of this. It's perfectly fine to have just periodic (no need to check the push anymore)

.github/workflows/linting.yml
dfarrell07: Yeah, adding the Shipyard infra to all the repos would be nice in the longer-term. As you said, it would be nice to always have the option to run everything locally.

.github/workflows/periodic.yml
dfarrell07: Okay, I removed the push-based triggering here.

mangelajo (approved):

.github/workflows/linting.yml
mkolesnik: I marked lines 6-10 which are run on every push not just tags

Except perhaps MD links (which is 3rd party effected) I don't think there's a benefit to re-run all linting on the push as it should've already passed in the PR

Even for MD links there's limited value to run on push since if the job actually fails you won't be able to change the commit and you'll have to fix it in a PR anyhow.

.github/workflows/linting.yml
mkolesnik: OK then let's go a step further and assert after the command ran that the git tree is clean since otherwise you'd need to manually regenerate the API

mkolesnik: > Good questions, I'm not sure, I've basically carried that job over into this workflow as-is.

Yeah I know just want to make sure it makes sense :)

mkolesnik (changes requested):

.github/workflows/linting.yml
dfarrell07: Is there a situation where we push something other than a tag? I guess we don't have automation to prevent that, but maybe we should. I thought all commits should go through PRs, never direct pushes.

.github/workflows/linting.yml
dfarrell07: > OK then let's go a step further and assert after the command ran that the git tree is clean since otherwise you'd need to manually regenerate the API

Maybe you'd like to propose that in a different PR? It's not really in-scope for "reorganizing GHA jobs/workflows" I think.

.github/workflows/linting.yml
dfarrell07: I guess https://docs.github.com/en/github/administering-a-repository/about-branch-restrictions for master and https://docs.github.com/en/github/administering-a-repository/enabling-required-status-checks for all checks, with "Include admins", would get us part-way there. As I understand, pushes would be rejected to master if all required checks don't pass.

We could also add a job that raises an Issue to report direct pushes, like the example here: https://github.com/marketplace/actions/create-an-issue#usage.

mkolesnik: I think you missed how GHA works, for every PR that's merged a "push" event is created which runs the tests you specify in the file that is listening to the push.
If the push was specifying only to listen to tags then it would only trigger if a tag was pushed, but as it's written here it'll trigger for any push to `master` (i.e. every PR merged to master) and for any tag starting with `v`.

For example you can see here the merge appearing as a push event:
image

.github/workflows/linting.yml
mkolesnik: Yeah sure we can do it in a follow up to make this check more robust

.github/workflows/linting.yml
dfarrell07: So you're saying it's running twice on the same code for normal PRs - once for the PR check and then again when the PR is actually merged? It seems like that would be mentioned in https://docs.github.com/en/actions/reference/events-that-trigger-workflows#push.

I understand that this will trigger the jobs on both pushes to master and PRs, and that's what I'm saying is desirable above. The jobs triggering on pushes is required for us to enforce these checks pass for direct pushes to master for all users.

dfarrell07: Even if they jobs re-run when PRs are merged, is that a bad thing? Don't we want tests job that directly parallel release jobs, not just verify PRs?

.github/workflows/linting.yml
dfarrell07: Raised https://github.com/submariner-io/submariner/issues/809 to track.

.github/workflows/linting.yml
dfarrell07: For example, one could imagine that a PR could pass checks that the same code, after "Rebase and Merge" on top of a logically-related change, would fail. We want a job to catch that imho.

mkolesnik: > So you're saying it's running twice on the same code for normal PRs - once for the PR check and then again when the PR is actually merged?

Yes at least twice (it's also re run when you update the PR)

> 
> I understand that this will trigger the jobs on both commit/tag pushes to master and PRs, and that's what I'm saying is desirable above. The jobs triggering on pushes is required for us to enforce these checks pass for direct pushes to master for all users.

AFAIK direct pushes are protected by GH policy so this is not a concern.

> Even if the jobs re-run when PRs are merged, is that a bad thing? Don't we want test jobs that directly parallel release jobs, not just verify PRs?

Not sure what you mean, can you please clarify?

> For example, one could imagine that a PR could pass checks that the same code, after "Rebase and Merge" on top of a logically-related change, would fail. We want a job to catch that imho.

While for unit tests and E2E tests that would be a desirable action since even if the code "merges nicely" it could still break something after rebase, I don't think the same can be said of linting.
I don't see how suddenly a DCO check would fail or MarkDown would stop linting after such a scenario, if the rebase was clean.
I guess the only exception is if you change the linting behavior itself and merge that before the patch, which would be quite the corner case and I dont think makes sense to run these checks on merge just for that.

Originally many of these checks didnt use to run on merge, and now they would, this needs to be justified since every job run we add takes time and resources which couldve been spent on other jobs (not to mention downloading docker images needlessly) and hence I raised this point about the linting.

One more thing to consider - these jobs (that run after push) aren't gating so if they fail the merge won't be affected (as it has already happened), so it's really up to an observant maintainer to notice if they even fail.

dfarrell07: > AFAIK direct pushes are protected by GH policy so this is not a concern.

Right, but only via the branch protection and required status checks I've been talking about. So we have:

1. Branch protection for master, so only admins can directly push.
2. Required status checks for master, so admins can only directly push if required checks pass.

So to prevent admins from pushing commits that don't pass required checks, those checks have to create statuses on pushes.

> One more thing to consider - these jobs (that run after push) aren't gating so if they fail the merge won't be affected (as it has already happened), so it's really up to an observant maintainer to notice if they even fail.

That's not true with the configuration I'm describing:

> You won't be able to push local changes to a protected branch until all required status checks pass. Instead, you'll receive an error message similar to the following:
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: error: Required status check "ci-build" is failing

https://docs.github.com/en/github/administering-a-repository/about-required-status-checks#troubleshooting-required-status-checks

dfarrell07: We talked about this quite a bit on the automation sync just now, summarizing:

* Required status checks aren't reported for push events, only PRs, which invalidates much of what I was thinking above.
* There is config for requiring branches (and so also statues) to be up-to-date before merging, which does most/all of the work that I was looking for.
* With the config now on this repo (should be copied to other repos), I think all my concerns about statues applying to admins are addressed.
* We want to cause all jobs to run on tags, but only release jobs to run on pushes to master.

dfarrell07: Okay, I think it's fixed in the latest. Thanks @mkolesnik!

.github/workflows/periodic.yml
```diff
@@ -1,13 +1,18 @@
...
-name: Markdown Link Checks
+name: Periodic
+
 on:
   schedule:
     - cron: "0 0 * * 0"
+  push:
+    tags:
+      - 'v**'
+
```
mkolesnik, line 10:
```suggestion
```

Otherwise you'll get "periodic" checks on each tag push (which would be confusing)

mkolesnik (changes requested):

.github/workflows/periodic.yml
dfarrell07: I thought we agreed we wanted to run periodic checks on tag pushes, for example so we get the "all files" version of the MD links job, and also so we have all job logs for releases if we add additional periodic tests (like perf).

dfarrell07: Removed

mkolesnik (approved):

mangelajo (approved):
Thanks

.github/workflows/periodic.yml
```diff
@@ -1,13 +1,14 @@
...
-name: Markdown Link Checks
+name: Periodic
+
 on:
   schedule:
     - cron: "0 0 * * 0"
 
 jobs:
   markdown-link-check-periodic:
+    name: Markdown Links (all files)
```
dfarrell07, line 10: I clarified the difference between the PR-triggered/modified-files-only and periodic/all-files jobs, per discussions with @mangelajo and @mkolesnik on Slack.

mangelajo:
I changed the required job settings but they don't reflect here, seeing that the new test names pass I'll proceed to merge, forcing as administrator.

mangelajo:
> I changed the required job settings but they don't reflect here, seeing that the new test names pass I'll proceed to merge, forcing as administrator.

Oh, no need to force, they updated now.

dfarrell07:
Awesome, thanks @mangelajo, thanks @mkolesnik!

PR #797, opened 2020-09-10
Scan generated images using Anchore

skitt:


.github/workflows/imagescan.yaml
```diff
@@ -0,0 +1,26 @@
+name: Security checks
```
dfarrell07, line 1: There is also gosec scanning going on elsewhere, so maybe we can use a name that's more specifically about container image scanning.

.github/workflows/imagescan.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Build the Docker image
+      run: make bin/submariner-engine package/.image.submariner
+    - uses: anchore/scan-action@v1
+      with:
+        image-reference: "quay.io/submariner/submariner:devel"
+        dockerfile-path: "package/Dockerfile.submariner"
+        fail-build: true
+        acs-report-enable: true
+        #acs-report-severity-cutoff: "Medium"
+    - name: anchore inline scan JSON results
```
dfarrell07, line 19: Maybe we can capitalize the first letters of these names, as proposed in https://github.com/submariner-io/submariner/pull/782.

dfarrell07 (changes requested):

PR #810, opened 2020-09-15
Merge the package owners, align entries

skitt:
package/* is associated with mkolesnik, /package/* with skitt, this
merges both entries so that shared ownership actually works.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,11 +3,10 @@
 /scripts/ 		@mkolesnik
 Makefile* 		@mkolesnik
 Dockerfile.dapper 	@mkolesnik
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
I think the package owner overlap is no longer there? But the file could still use some alignment help imho.

dfarrell07 (changes requested):
I think the package owner overlap is no longer there? But the file could still use some alignment help imho.

mangelajo:
It was handled here: https://github.com/submariner-io/submariner/pull/812/files

Issue #809, opened 2020-09-15
Assert git tree is clean after API generation


dfarrell07:
Per discussions on code review, the API code generation job should be enhanced to assert that the git tree is clean after API generation.

> OK then let's go a step further and assert after the command ran that the git tree is clean since otherwise you'd need to manually regenerate the API

_Originally posted by @mkolesnik in https://github.com/submariner-io/submariner/pull/782#discussion_r487858692_

PR #814, opened 2020-09-17
Fail codegen "validation" if it caused a change

mkolesnik:
If the repo is dirty after running the validation, fail as it indicates
that the API changed and needs to ge updated.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -13,6 +13,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Run codegen
         run: make codegen
+      - name: Make sure there's no pending changes
+        run: git add -A && git diff --staged --exit-code
 
   dco:
     name: DCO in Commit Message(s)
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Run codegen
         run: make codegen
-      - name: Make sure there's no pending changes
+      - name: Verify generated code matches committed code
         run: git add -A && git diff --staged --exit-code
 
   dco:


.github/workflows/linting.yml
```diff
@@ -16,6 +16,8 @@ jobs:
         uses: actions/checkout@master
       - name: Run validation
         run: make codegen
+      - name: Make sure there's no pending changes
+        run: output=$(git status --porcelain) && [ -z "$output" ]
```
skitt, line 20:
```suggestion
        run: git diff --exit-code
```

This will show the differences in the logs too.

skitt (changes requested):

.github/workflows/linting.yml
mkolesnik: Problem is that `git diff` doesnt count untracked files, so if such appear we'll miss them, while `git status` does

pkg/apis/submariner.io/v1/types.go
```diff
@@ -46,6 +46,7 @@ type EndpointSpec struct {
 	ClusterID     string            `json:"cluster_id"`
 	CableName     string            `json:"cable_name"`
 	Hostname      string            `json:"hostname"`
+	Hostname2     string            `json:"hostname2"`
```
dfarrell07, line 49: I think these type changes shouldn't be a part of this commit?

dfarrell07 (changes requested):

.github/workflows/linting.yml
skitt: Good point, so

```suggestion
        run: git add -A && git diff --staged --exit-code
```

pkg/apis/submariner.io/v1/types.go
skitt: I imagine the point here was to prove that the test works ;-). (But yes, they’ll need to be removed before the merge.)

dfarrell07:
Closes: https://github.com/submariner-io/submariner/issues/809

dfarrell07:
Maybe we can finish this one off @mkolesnik?

.github/workflows/linting.yml
```diff
@@ -13,6 +13,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Run codegen
         run: make codegen
+      - name: Make sure there's no pending changes
```
dfarrell07, line 16:
```suggestion
      - name: Verify generated code matches committed code
```

Sorry for the nit, started with avoiding "there's"->"there are".

dfarrell07 (approved):

mangelajo (approved):

skitt (approved):

Issue #777, opened 2020-08-27
API 'describe Gateway' does not show Gateway crashing and cable-driver failure


manosnoam:
When there was an error on the Gateway pod (due to libreswan failure, as described in #776),
The `describe Gateway` API was not displaying the actual status of the cable driver, while there was an ongoing failure on the Gateway (keep on crashing).

**What happened**:

```
 $ oc describe Gateway -n submariner-operator

 gateway_info="Name:         ip-10-0-74-54
 Namespace:    submariner-operator
 Labels:       <none>
 Annotations:  update-timestamp: 1598481496
 API Version:  submariner.io/v1
 Kind:         Gateway
 Metadata:
   Creation Timestamp:  2020-08-26T22:35:15Z
   Generation:          1
   Managed Fields:
     API Version:  submariner.io/v1
     Fields Type:  FieldsV1
     fieldsV1:
       f:metadata:
         f:annotations:
           .:
           f:update-timestamp:
       f:status:
         .:
         f:connections:
         f:haStatus:
         f:localEndpoint:
           .:
           f:backend:
           f:cable_name:
           f:cluster_id:
           f:hostname:
           f:nat_enabled:
           f:private_ip:
           f:public_ip:
           f:subnets:
         f:statusFailure:
         f:version:
     Manager:         submariner-engine
     Operation:       Update
     Time:            2020-08-26T22:38:16Z
   Resource Version:  28849
   Self Link:         /apis/submariner.io/v1/namespaces/submariner-operator/gateways/ip-10-0-74-54
   UID:               af4f3164-5274-42b0-be3f-cbb2796505d6
 Status:
   Connections:
   Ha Status:  passive
   Local Endpoint:
     Backend:      libreswan
     cable_name:   submariner-cable-nmanos-cluster-a-10-0-74-54
     cluster_id:   nmanos-cluster-a
     Hostname:     ip-10-0-74-54
     nat_enabled:  true
     private_ip:   10.0.74.54
     public_ip:    18.221.13.96
     Subnets:
       172.30.0.0/16
       10.128.0.0/14
   Status Failure:  
   Version:         v0.5.0-9-ga1627f5
 Events:            <none>"
 Name:         ip-10-0-74-54
 Namespace:    submariner-operator
 Labels:       <none>
 Annotations:  update-timestamp: 1598481496
 API Version:  submariner.io/v1
 Kind:         Gateway
 Metadata:
   Creation Timestamp:  2020-08-26T22:35:15Z
[...]

PR #819, opened 2020-09-17
Report startup errors in the Gateway StatusFailure

tpantelis:
Issue #777 

Added a `SetGatewayStatusError` method to the `Gateway` syncer that creates/updates the `Gateway` resource with a specific `StatusFailure`. This is called prior to exiting via `Fatalf`.

This doesn't include some bootstrap failures that prevent the syncer from being created, like failure to create the kube client.

Eliminated a couple error paths that never actually result in an error to simplify:

 - `util.GetLocalCluster` - this function simply copies fields so I eliminated it altogether and inlined it.

 - `cableengine.NewEngine` - removed the error return - any errors should be reported by `StartEngine`..


skitt (approved):

main.go
```diff
@@ -92,9 +94,15 @@ func main() {
...
-		klog.Fatalf("Error creating local cluster object from %#v: %v", submSpec, err)
+	localCluster := types.SubmarinerCluster{
+		ID: submSpec.ClusterID,
+		Spec: subv1.ClusterSpec{
+			ClusterID:   submSpec.ClusterID,
+			ColorCodes:  submSpec.ColorCodes,
+			ServiceCIDR: submSpec.ServiceCidr,
+			ClusterCIDR: submSpec.ClusterCidr,
+			GlobalCIDR:  submSpec.GlobalCidr,
+		},
```
mangelajo, line 105: Just for having a shorter main, I'd keep it on a function, may be `localCluster := GetClusterFromSpec(submSpec)`

of course, without return of error which made no sense.

main.go
```diff
@@ -268,4 +277,12 @@ func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder res
 			OnStoppedLeading: end,
 		},
 	})
+
+	return nil
+}
+
+func fatal(gwSyncer *syncer.GatewaySyncer, format string, args ...interface{}) {
```
mangelajo, line 284: I think this approach is great :+1:

pkg/cableengine/cableengine.go
```diff
@@ -48,14 +48,12 @@ type engine struct {
...
-func NewEngine(localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint) (Engine, error) {
-	i := engine{
+func NewEngine(localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint) Engine {
+	return &engine{
 		localCluster:  localCluster,
 		localEndpoint: localEndpoint,
 		driver:        nil,
 	}
-
-	return &i, nil
```
mangelajo, line 58: why did we end up with this? very weird...

pkg/cableengine/syncer/syncer.go
```diff
@@ -49,6 +52,21 @@ func (s *GatewaySyncer) Run(stopCh <-chan struct{}) {
 }
 
 func (i *GatewaySyncer) syncGatewayStatus() {
+	i.Lock()
+	defer i.Unlock()
+
+	i.syncGatewayStatusSafe()
```
mangelajo, line 58: :+1:

mangelajo (changes requested):
Just a small nit, but the change in general makes it 10 times more robust.

main.go
tpantelis: Done

main.go
```diff
@@ -92,10 +94,7 @@ func main() {
 
 	klog.Info("Creating the cable engine")
 
-	localCluster, err := util.GetLocalCluster(submSpec)
-	if err != nil {
-		klog.Fatalf("Error creating local cluster object from %#v: %v", submSpec, err)
-	}
+	localCluster := submarinerClusteFrom(&submSpec)
```
skitt, line 97:
```suggestion
	localCluster := submarinerClusterFrom(&submSpec)
```

main.go
```diff
@@ -198,18 +193,36 @@ func main() {
...
+		if err = startLeaderElection(leClient, recorder, becameLeader, lostLeader); err != nil {
+			fatal(cableEngineSyncer, "Error starting leader election: %v", err)
+		}
+	}()
+
 	<-stopCh
 	klog.Info("All controllers stopped or exited. Stopping main loop")
 }
 
+func submarinerClusteFrom(submSpec *types.SubmarinerSpecification) types.SubmarinerCluster {
```
skitt, line 206:
```suggestion
func submarinerClusterFrom(submSpec *types.SubmarinerSpecification) types.SubmarinerCluster {
```

skitt (changes requested):
Typo ;-).

main.go
tpantelis: > Typo ;-).

I had just woke up and didn't have my readers on 😄

skitt (approved):

mangelajo (approved):

PR #822, opened 2020-09-21
Annotate to declare the Gateway print column

skitt:
So that we can generate the Gateway CRD automatically, annotate the
Gateway type with the appropriate directive to generate the additional
printer column.

### DIFF
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -65,6 +65,7 @@ type EndpointList struct {
 // +genclient
 // +genclient:noStatus
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:JSONPath=".status.haStatus",name="HA Status",description="High availability status of the Gateway",type="string"
 
 type Gateway struct {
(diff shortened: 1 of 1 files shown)


pkg/apis/submariner.io/v1/types.go
```diff
@@ -65,6 +65,7 @@ type EndpointList struct {
 // +genclient
 // +genclient:noStatus
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:JSONPath=".status.haStatus",name="ha-status",description="High availability status of the Gateway",type="string"
```
tpantelis, line 68:
```suggestion
// +kubebuilder:printcolumn:JSONPath=".status.haStatus",name="HA Status",description="High availability status of the Gateway",type="string"
```

skitt: I wondered about that; our current CRD uses `ha-status`, so I just copied that, but it would make sense to change it.

tpantelis: yeah this becomes the user-visible column title.

tpantelis (approved):

mangelajo (approved):

PR #829, opened 2020-09-23
Remove the cluster owned tag from submariner sg

mangelajo:
We should not set the submariner security group as cluster-owned,
otherwise the LoadBalancer controller stops working, because that's
not supported.

```text
Warning  SyncLoadBalancerFailed  16s (x4 over 53s)  service-controller
         Error syncing load balancer: failed to ensure load balancer:
         Multiple tagged security groups found for instance i-03fcabdc39f37f25c;
         ensure only the k8s security group is tagged; the tagged groups were
         sg-06cf365995bf20f3f(majopela-cluster-a-09-b7ch5-submariner-gw-sg)
         sg-0c4abc1fcf12d5c3b(terraform-20200922090409222000000002)
```


nyechiel (approved):

sridhargaddam (approved):
LGTM, thanks @mangelajo

dfarrell07 (approved):

skitt (approved):

tpantelis (approved):

PR #832, opened 2020-09-24
Change framework e2e test to allow testing with only 1 cluster

maayanf24:
Needed to enable latency and throughput intra-cluster tests
https://github.com/submariner-io/submariner-operator/issues/706


mangelajo (approved):
Correct, It doesn't really matter where we get this info from, and ClusterA will always be there.

tpantelis (approved):

PR #843, opened 2020-09-30
Expose Prometheus metrics

skitt:
This adds a Prometheus handler on port 8080, on /metrics (as
expected). There are no Submariner-specific metrics yet, but the Go
Prometheus client exposes a number of useful metrics as-is.


dfarrell07 (dismissed):

dfarrell07 (approved):

main.go
```diff
@@ -216,6 +224,20 @@ func submarinerClusterFrom(submSpec *types.SubmarinerSpecification) types.Submar
...
 }
 
+func startHttpServer() *http.Server {
+	srv := &http.Server{Addr: ":8080"}
+
+	http.Handle("/metrics", promhttp.Handler())
+
+	go func() {
+		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
+			klog.Fatalf("Error starting metrics server: %v", err)
```
tpantelis, line 234: We should call `fatal` to set the Gateway status.

main.go
```diff
@@ -201,6 +205,10 @@ func main() {
 
 	<-stopCh
 	klog.Info("All controllers stopped or exited. Stopping main loop")
+
+	if err := httpServer.Shutdown(context.TODO()); err != nil {
+		klog.Fatalf("Error shutting down metrics HTTP server: %v", err)
```
tpantelis, line 210: Should this be fatal since we're shutting down anyway?

main.go
skitt: Or perhaps just log the error and continue running — the gateway can run without metrics...

sridhargaddam (approved):

tpantelis (approved):

PR #846, opened 2020-10-01
Track RX/TX bytes in the Libreswan cable driver

skitt:


pkg/cable/libreswan/libreswan.go
```diff
@@ -118,10 +135,23 @@ func (i *libreswan) refreshConnectionStatus() error {
...
-			activeConnections.Add(components[1])
+		matches := trafficStatusRE.FindStringSubmatch(line)
+		if matches != nil {
+			activeConnections.Add(matches[1])
+
+			inBytes, err := strconv.Atoi(matches[2])
+			if err != nil {
+				klog.V(log.DEBUG).Infof("Invalid inBytes in whack output line: %q", line)
+			} else {
+				libreswanRxGauge.Set(float64(inBytes))
```
mangelajo, line 146: Won't we have a different line for endpoint/subnet? we may need to make aggregations I think?

pkg/cable/libreswan/libreswan.go
```diff
@@ -24,8 +27,19 @@ const (
 	cableDriverName = "libreswan"
 )
 
+// The metrics are gauges because we want to set the absolute value
```
mangelajo, line 30: I think we should not expose cable-driver specific gauges, 

Ideally we should create an interface on the cable-drivers, so we can expose tx_bytes / rx_bytes per remote endpoint, we could label with the cable driver may be.


I think it's OK if currently until we create such interface we publish with a generic name. rx_bytes, tx_bytes.  But I'd expected this to be per remote cluster endpoint, instead of a global counter.

mangelajo (changes requested):
Nice, but some comments.

pkg/cable/libreswan/libreswan.go
skitt: Yes, I need to sum the values.

pkg/cable/libreswan/libreswan.go
skitt: I’m following the rule that metrics should be gathered as close as possible to the point of origin. The idea here is to provide metrics to see how they can be integrated, and how they compare with those from WireGuard once we get them.

Ultimately I think the common interface may come from the metrics themselves, not necessarily the cable driver interface. That is to say, we should define the metrics we want, and then let the cable drivers implement them.

And yes, this should be per remote cluster endpoint, with appropriate labels...

pkg/cable/libreswan/libreswan.go
```diff
@@ -118,10 +135,23 @@ func (i *libreswan) refreshConnectionStatus() error {
...
-		components := strings.Split(line, "\"")
-		if len(components) == 3 {
-			activeConnections.Add(components[1])
+		matches := trafficStatusRE.FindStringSubmatch(line)
+		if matches != nil {
+			activeConnections.Add(matches[1])
+
+			inBytes, err := strconv.Atoi(matches[2])
+			if err != nil {
+				klog.V(log.DEBUG).Infof("Invalid inBytes in whack output line: %q", line)
```
tpantelis, line 144: Warning?

pkg/cable/libreswan/libreswan.go
mangelajo: Yes, but... I still believe it makes more sense to keep a single name
rx_bytes, tx_bytes, regardless of the driver

One driver will emit metrics with one label, other with a different label. Can that be done?

skitt: Yes, that’s the plan.

pkg/cable/libreswan/libreswan.go
skitt: Fixed!

tpantelis (approved):

skitt:
@mangelajo could you re-review this? I’ve tried to address your comments.

dfarrell07 (approved):

skitt:
> @mangelajo could you re-review this? I’ve tried to address your comments.

@mangelajo ping?

mangelajo (approved):

Issue #241, opened 2019-12-03
prep_for_subm.sh flag to accept terraform actions


manosnoam:
Please add '--accept' or similar flag as user input for [prep_for_subm.sh](https://github.com/submariner-io/submariner/blob/master/tools/openshift/ocp-ipi-aws/prep_for_subm.sh) script, to accept Terraform actions, without user interact.

manosnoam:
Still relevant.

PR #851, opened 2020-10-05
Fixes for prep_for_subm.sh

manosnoam:
- Fix https://github.com/submariner-io/submariner/issues/241
- Fix download_ocp_ipi_aws_tool()
- Fix several variables assignments

Signed-off-by: Noam Manos <nmanos@redhat.com>

### DIFF
--- a/tools/openshift/ocp-ipi-aws/prep_for_subm.sh
+++ b/tools/openshift/ocp-ipi-aws/prep_for_subm.sh
@@ -1,78 +1,84 @@
-#!/bin/sh
+#!/bin/bash
 set -e
-OCP_INS_DIR="$(realpath ${1:-.})"
+
+# Get OCP installer path, or use current directory
+OCP_INS_DIR="$(realpath -- ${1:-.})"
 METADATA_JSON="${OCP_INS_DIR}/metadata.json"
+
+# Get Terraform apply options, for the rest of args (if given)
+if (( $# > 1 )) ; then
+  shift
+  TERRAFORM_ARGS=("$@") # e.g. -auto-approve -lock-timeout=3m
+fi
+
+# Set Github parameters
 GITHUB_BRANCH="${GITHUB_BRANCH:-master}"
 GITHUB_USERFORK="${GITHUB_USERFORK:-submariner-io}"
-GITHUB_ZIP="${GITHUB_ZIP:-https://github.com/$GITHUB_USERFORK/submariner/archive/$GITHUB_BRANCH.zip}"
+GITHUB_ARCHIVE="https://github.com/$GITHUB_USERFORK/submariner/archive/$GITHUB_BRANCH.tar.gz"
+
+
+# Functions
 
 req_install() {
-        if ! command -v $1 >/dev/null 2>&1; then
-	    echo "$1 is required by this tool, please install $1" >&2
-	    exit 1
-	fi
+  if ! command -v $1 >/dev/null 2>&1; then
+    echo "$1 is required by this tool, please install $1" >&2
+    exit 2
+  fi
 }
 
 download_ocp_ipi_aws_tool() {
-	TMP=$(mktemp -d)
-	IPI_AWS="submariner-${GITHUB_BRANCH}/tools/openshift/ocp-ipi-aws"
-	curl -L "${GITHUB_ZIP}" --output $TMP/submariner.zip
-	unzip $TMP/submariner.zip "$IPI_AWS/*" -d $TMP
-	cp -rfp $TMP/$IPI_AWS .
-	rm -rf $TMP
+  echo "Downloading $GITHUB_ARCHIVE and extracting 'ocp-ipi-aws' tool"
+  wget -O - $GITHUB_ARCHIVE | tar xz --strip=4 "submariner-${GITHUB_BRANCH}/tools/openshift/ocp-ipi-aws"
 }
 
-# check pre-requisites
-
-for cmd in unzip terraform oc aws; do
-	req_install $cmd
-done
-
-# check parameters
-
+# Check parameters
 if [[ ! -d $OCP_INS_DIR ]] || [[ ! -f $METADATA_JSON ]]; then
-	echo "please provide a valid openshift installer directory for this script to work"  >&2
-	echo "usage:" >&2
-	echo "   prep_for_subm.sh ocp4-install-dir" >&2
-	echo ""	>&2
-	exit 2
+  echo "Please provide a valid OpenShift installation directory as the first argument." >&2
+  echo "Usage:" >&2
+  echo "   $0 <ocp-install-path> [optional terraform apply arguments]" >&2
+  echo "" >&2
+  exit 1
 fi
 
-cd $OCP_INS_DIR
+# Check pre-requisites
+for cmd in wget terraform oc aws; do
+  req_install $cmd
+done
 
-INFRA_ID=$(egrep -o -E '\"infraID\":\"([^\"]*)\"' metadata.json | cut -d: -f2 | tr -d \")
-REGION=$(egrep -o -E '\"region\":\"([^\"]*)\"' metadata.json | cut -d: -f2 | tr -d \")
+
+# Main
+
+INFRA_ID=$(egrep -o -E '\"infraID\":\"([^\"]*)\"' $METADATA_JSON | cut -d: -f2 | tr -d \")
+REGION=$(egrep -o -E '\"region\":\"([^\"]*)\"' $METADATA_JSON | cut -d: -f2 | tr -d \")
 
 echo infraID: $INFRA_ID
 echo region: $REGION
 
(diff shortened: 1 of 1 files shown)


tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,84 @@
 #!/bin/sh
 set -e
+
+# Script args
+OCP_INS_DIR="$(realpath ${1:-.})"
+(( $# == 1 )) || shift && TERRAFORM_ARGS=$@ # e.g. -auto-approve -lock-timeout=3m
+
+# Script args
 OCP_INS_DIR="$(realpath ${1:-.})"
```
skitt, line 9: This ends up setting `OCP_INS_DIR` twice, is that intentional?

skitt (changes requested):

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
manosnoam: @Stephen Kitt 
right, a duplicate paste...  will fix it

mangelajo (commented):
It looks good noam, only @skitt comment

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,80 @@
...
-	unzip $TMP/submariner.zip "$IPI_AWS/*" -d $TMP
-	cp -rfp $TMP/$IPI_AWS .
-	rm -rf $TMP
+  echo "Downloading $GITHUB_ARCHIVE and extracting 'ocp-ipi-aws' tool"
+  wget -O - $GITHUB_ARCHIVE | tar xz --strip=4 "submariner-${GITHUB_BRANCH}/tools/openshift/ocp-ipi-aws"
 }
 
 # check pre-requisites
 
 for cmd in unzip terraform oc aws; do
```
mangelajo, line 29:
```suggestion
for cmd in terraform oc aws; do
```

mangelajo: You don't need unzip anymore with your change.

mangelajo (approved):

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,80 @@
 #!/bin/sh
 set -e
+
+# Script args
 OCP_INS_DIR="$(realpath ${1:-.})"
+(( $# == 1 )) || shift && TERRAFORM_ARGS=$@ # e.g. -auto-approve -lock-timeout=3m
```
skitt, line 6: What is this supposed to do? There are a few problems:

* if the number of arguments is 0 (which is allowed, based on the `OCP_INS_DIR` line above), `shift` will run and bail out;
* if the number of arguments is 1, `TERRAFORM_ARGS` will be set to the arguments, *i.e.* the directory.

This will also fail if the user runs `prep_for_subm.sh -auto-approve`, expecting the installation directory to be set to `.` and `TERRAFORM_ARGS` to be set to `-auto-approve`.

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,80 @@
...
-	rm -rf $TMP
+  echo "Downloading $GITHUB_ARCHIVE and extracting 'ocp-ipi-aws' tool"
+  wget -O - $GITHUB_ARCHIVE | tar xz --strip=4 "submariner-${GITHUB_BRANCH}/tools/openshift/ocp-ipi-aws"
 }
 
 # check pre-requisites
 
-for cmd in unzip terraform oc aws; do
-	req_install $cmd
+for cmd in terraform oc aws; do
```
skitt, line 29: This could check for `wget` too, since that’s not installed by default everywhere:

```suggestion
for cmd in wget terraform oc aws; do
```

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,80 @@
 #!/bin/sh
```
skitt, line 1: You’re adding commands which rely on this being bash, so the shebang should change accordingly:

```suggestion
#!/bin/bash
```

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
skitt: Regarding `TERRAFORM_ARGS` specifically, you should use an array to preserve the separate arguments:

```suggestion
(( $# == 1 )) || shift && TERRAFORM_ARGS=("$@") # e.g. -auto-approve -lock-timeout=3m
```

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,80 @@
...
+if [[ ! -d ocp-ipi-aws-prep ]]; then
+  download_ocp_ipi_aws_tool
+fi
 
 sed -i "s/\"cluster_id\"/\"$INFRA_ID\"/g" main.tf
 sed -i "s/\"aws_region\"/\"$REGION\"/g" main.tf
 
 terraform init
-terraform apply
+terraform apply $TERRAFORM_ARGS
```
skitt, line 67: Assuming this is an array (see my comment above), this should then be

```suggestion
terraform apply "${TERRAFORM_ARGS[@]}"
```

skitt (changes requested):

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
manosnoam: Thanks @skitt, I fixed it.

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,85 @@
...
 
 if [[ ! -d $OCP_INS_DIR ]] || [[ ! -f $METADATA_JSON ]]; then
-	echo "please provide a valid openshift installer directory for this script to work"  >&2
-	echo "usage:" >&2
-	echo "   prep_for_subm.sh ocp4-install-dir" >&2
-	echo ""	>&2
-	exit 2
+  echo "please provide a valid openshift installer directory for this script to work"  >&2
+  echo "usage:" >&2
+  echo "   prep_for_subm.sh ocp4-install-dir" >&2
```
SteveMattar, line 42:
```suggestion
  echo "   prep_for_subm.sh <ocp4_install_dir> [additional_terraform_args...]" >&2
```

SteveMattar (changes requested):
I understand the TERRAFORM_ARGS are optional but could you still update the usage? it is worth to let the user be aware of it

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
```diff
@@ -1,82 +1,85 @@
-#!/bin/sh
+#!/bin/bash
 set -e
-OCP_INS_DIR="$(realpath ${1:-.})"
+
+# Get OCP installer path, or use current directory
+OCP_INS_DIR="$(realpath ${1:-.} 2>/dev/null || :)"
```
skitt, line 5: This is still broken; how about

```suggestion
OCP_INS_DIR="$(realpath -- ${1:-.})"
if [ ! -d "${OCP_INS_DIR}" ]; then
  echo Please specify the OpenShift installation directory as the first argument, for example
  echo $0 ${PWD} $*
  exit 1
fi
```

skitt (changes requested):

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
manosnoam: Can you show this broken output you get ?

Also, the problem with using realpath without `2>/dev/null || :`
that it can fail before that condition you've suggested. e,g:
```
OCP_INS_DIR="$(realpath -- "/no/dir")"
realpath: ‘/no/dir’: No such file or directory
```

skitt: We could fix that by using `|| echo .` but, thinking about it a bit more, I think failing here is fine: the error message is clear, the user will know that they provided an invalid directory.

manosnoam: Fixed as you suggested, thanks!

tools/openshift/ocp-ipi-aws/prep_for_subm.sh
manosnoam: Fixed as you suggested, thanks!

SteveMattar (approved):

skitt (approved):

Issue #375, opened 2020-02-25
Migrate controllers to use admiral APIs


tpantelis:
Specifically, the `ResourceSyncer` and `ResourceWatcher` that eliminate  boilerplate and code duplication.

tpantelis:
> This issue has been automatically marked as stale because it has not had activity for 120 days. It will be closed if no further activity occurs. Please make a comment if this issue/pr is still valid. Thank you for your contributions.

Still valid dude.

tpantelis:
bump

tpantelis:
There's controller in globalnet that should be converted.

PR #852, opened 2020-10-05
Simplify/Refactor tunnel controller to use ResourceSyncer

tpantelis:
The tunnel controller doesn't actually sync a k8s resource but it does transform an Endpoint to a tunnel connection and thus can use the `ResourceSyncer` infrastructure to eliminate a lot of informer/event handler boilerplate.

Relatee to #375

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kelseyhightower/envconfig"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/submariner-io/admiral/pkg/log"
+	admUtil "github.com/submariner-io/admiral/pkg/util"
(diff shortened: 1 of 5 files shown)


skitt (approved):

main.go
```diff
@@ -129,11 +131,24 @@ func main() {
 
 	cableEngineSyncer.Run(stopCh)
 
+	err = subv1.AddToScheme(scheme.Scheme)
+	if err != nil {
+		fatal(cableEngineSyncer, "Error adding submariner types to the scheme: %v", err)
```
sridhargaddam, line 136: nit: Any specific reason for not using "klog.Fatal" in the new code?

sridhargaddam (approved):

main.go
tpantelis: `fatal` cals `klog.Fatal`.

PR #856, opened 2020-10-07
Fix Gateway connection status for Libreswan cable driver

sridhargaddam:
Occasionally, with libreswan driver, we see that Connection status
for remote clusters in Gateway object shows as connecting eventhough
the datapath continues to work fine. In Libreswan driver, Gateway status
is populated by quering the IPsec SAs from the host and validating the
connection names.

On debugging this issue, it was seen that the connections are all
loaded but some of them were in "prospective erouted" state and not yet
active. In this PR, instead of checking for all the connections associated
to a remote cluster, we just check if atleast one of the connections is
established and mark the tunnel status accordingly.

Fixes issue: https://github.com/submariner-io/submariner-operator/issues/719

### DIFF
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -131,28 +131,23 @@ func (i *libreswan) refreshConnectionStatus() error {
 		return errors.WithMessage(err, "error waiting for whack")
 	}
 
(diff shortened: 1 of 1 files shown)


skitt (approved):

mangelajo (approved):
Let's merge for now, and let's consult with Paul

PR #878, opened 2020-10-14
Fix prep_for_subm.sh to not ignore local ocp-ipi-aws yamls

manosnoam:
Fix prep_for_subm.sh to not ignore local ocp-ipi-aws yamls
    
- Fix for #880
- Safer way to replace values of 'cluster_id' and 'aws_region' keys
- Change directory to $OCP_INS_DIR before downloading ocp-ipi-aws

Signed-off-by: Noam Manos <nmanos@redhat.com>

### DIFF
--- a/tools/openshift/ocp-ipi-aws/prep_for_subm.sh
+++ b/tools/openshift/ocp-ipi-aws/prep_for_subm.sh
@@ -64,15 +64,14 @@ if [[ -z "$REGION" ]]; then
   exit 4
 fi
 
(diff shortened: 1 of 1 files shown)


mangelajo (changes requested):
This does not work. It overwrites the machine set yaml, as well as it shares the terraform state between two clusters. Try to follow one of our QuickStart guides with your patch and two aws clusters and you will see the issue

manosnoam:
Abandon this PR in favor of https://github.com/submariner-io/submariner/pull/881

PR #887, opened 2020-10-19
Alert user the supported terraform version in prep_for_subm.sh

sridhargaddam:
Currently, while preparing the AWS cluster for Submariner as
part of prep_for_subm.sh script, we only support terraform
version 12.x

In this PR, we display an appropriate error message if users
are trying to run the script with a different version of terraform.

### DIFF
--- a/tools/openshift/ocp-ipi-aws/versions.tf
+++ b/tools/openshift/ocp-ipi-aws/versions.tf
@@ -1,0 +1,3 @@
+terraform {
+    required_version = ">= 0.12, < 0.13"
+}
--- a/tools/openshift/ocp-ipi-aws/versions.tf
+++ b/tools/openshift/ocp-ipi-aws/versions.tf
@@ -1,3 +1,3 @@
 terraform {
-    required_version = ">= 0.12, < 0.13"
+    required_version = ">= 0.12, <= 0.12.12"
 }


manosnoam (approved):

mangelajo:
Actually we believe it has to be <=0.12.3 if we can check that

sridhargaddam:
> Actually we believe it has to be <=0.12.3 if we can check that

I remember using v 0.12.12 (or something very close to that) and it worked.

tpantelis (approved):

mangelajo:
> > Actually we believe it has to be <=0.12.3 if we can check that
> 
> I remember using v 0.12.12 (or something very close to that) and it worked.

I know noam had issues about duplicate security groups in later versions @manosnoam do you remember which one?

mangelajo (approved):

tpantelis:
We can squash and merge to eliminate the DCO check failure.

nyechiel:
@manosnoam what's your experience with the Terraform version? I ran some tests earlier this week, and **v0.12.3** is the only version that really worked smoothly for me.

manosnoam:
For me 0.12.2 works

sridhargaddam:
@nyechiel @manosnoam I just tried terraform 0.12.12 and it works fine :)

Issue #857, opened 2020-10-07
Create the initial architecture/interface for the network-plugin drivers.


mangelajo:
**What would you like to be added**:

A network-plugin driver architecture for the routeagent

**Why is this needed**:

Supporting non-iptables kubeproxy based network-plugins requires integration to the
network-plugin or the network technology being used.

See:  https://github.com/submariner-io/enhancements/pull/11
Related-Issue: https://github.com/submariner-io/submariner/issues/778

Issue #778, opened 2020-08-30
Epic: OVN support


nyechiel:
Submariner is not compatible with OVN-based deployments at the moment. The request is to add support for interconnecting clusters that use the OVN (ovn-kubernetes) CNI plugin.

- [x] Submariner-OVN exploration: https://github.com/submariner-io/submariner/issues/793
- [x] Document the design via an enhancement proposal: https://github.com/submariner-io/enhancements/pull/11
- [x] Shipyard support for deploying OVN https://github.com/submariner-io/shipyard/issues/335
- [x] Improve the submariner-operator to pass down the network-plugin type down to the route-agent https://github.com/submariner-io/submariner-operator/issues/736
- [x] Improve the submariner-operator to start the network-plugin-syncer for plugins that need it (OVN for now).
- [x] Improve the submariner-operator to gather OVN credentials (for NB-DB and SB-DB and pass them down to the syncer)
- [x] Create the initial architecture/interface for the network-plugin drivers. https://github.com/submariner-io/submariner/issues/857
- [x] Convert the existing implementation into a "kube-proxy iptables" driver which will be the default. https://github.com/submariner-io/submariner/issues/858
- [x] Create the OVN Driver  https://github.com/submariner-io/submariner/issues/859
- [x] Create the network-plugin-syncer for OVN https://github.com/submariner-io/submariner/pull/943
- [ ] Globalnet support https://github.com/submariner-io/submariner/issues/383 
- [ ] Move all ovn handling to go-ovn, contribute any missing parts to go-ovn https://github.com/submariner-io/submariner/issues/1100 and fix go-ovn limitations for #1101 
- [ ] no destroy-recreate objects in ovn under updates https://github.com/submariner-io/submariner/issues/1101
- [x] Any documentation changes necessary regarding the architecture, or ovn details in http://submariner.io https://github.com/submariner-io/submariner-website/issues/304

nyechiel:
@mangelajo do you see a reason to keep this Epic open? At this point, we can probably track the remaining gaps individually IMHO.

skitt:
There is still some work to be done here.

PR #891, opened 2020-10-21
Event interface for route & sync agent

mangelajo:
This commit implements the event interface described in the OVN
support EPIC. This interface will be used afterwards for refactoring
the route agent, and building the network-plugin-sync agent.

Resolves-Issue: #857
Related-Issue: #778


pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/submariner-io/admiral/pkg/log"
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/events"
+)
+
+type LoggerHandler struct {
+	events.EventHandlerBase
+}
+
+// GetName returns the name of the event handler
```
tpantelis, line 16: We don't need to duplicate the interface method docs here also it can become out-of-date. Same below.

pkg/events/events.go
```diff
@@ -0,0 +1,120 @@
+package events
+
```
tpantelis, line 2: I think we should rename this file to _handler.go_.

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+
+// GetNetworkPlugin ... logging driver says: "all"
+func (l *LoggerHandler) GetNetworkPlugin() string {
+	return events.AnyNetworkPlugin
+}
+
+// TransitionToNonGateway is called when there is a transition from GatewayNode to NonGateway on the
+// This will be called only when the transition happens and not periodically.
+func (l *LoggerHandler) TransitionToNonGateway() error {
+	klog.V(log.DEBUG).Info("The current node stopped being Gateway")
```
tpantelis, line 29:
```suggestion
	klog.V(log.DEBUG).Info("The current node is no longer a Gateway")
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+// TransitionToGateway is called when there is a transition from NonGatewayNode to GatewayNode.
+// This will be called only when the transition happens and not periodically.
+func (l *LoggerHandler) TransitionToGateway() error {
+	klog.V(log.DEBUG).Info("The current node became an active Gateway")
```
tpantelis, line 36:
```suggestion
	klog.V(log.DEBUG).Info("The current node has become a Gateway")
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+func (l *LoggerHandler) LocalEndpointUpdated(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A local Endpoint has been discovered on node %q with IP %q",
+		endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
+	return nil
+}
+
+// LocalEndpointRemoved is called when the local endpoint is removed.
+func (l *LoggerHandler) LocalEndpointRemoved(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A local Endpoint has been removed on node %q with IP %q",
+		endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
```
tpantelis, line 57:
```suggestion
	klog.V(log.DEBUG).Infof("The Endpoint for the local cluster has been removed: %#v", endpoint)
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+func (l *LoggerHandler) LocalEndpointCreated(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A new local Endpoint has been discovered for active Gateway on node %q with IP %q",
+		endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
+	return nil
+}
+
+// LocalEndpointUpdated is called when the local endpoint is updated.
+func (l *LoggerHandler) LocalEndpointUpdated(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A local Endpoint has been discovered on node %q with IP %q",
+		endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
```
tpantelis, line 50:
```suggestion
	klog.V(log.DEBUG).Infof("The Endpoint for the local cluster has been updated: %#v", endpoint)
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+// This will be called only when the transition happens and not periodically.
+func (l *LoggerHandler) TransitionToGateway() error {
+	klog.V(log.DEBUG).Info("The current node became an active Gateway")
+	return nil
+}
+
+// LocalEndpointCreated is called when the endpoint for the localCluster is created.
+func (l *LoggerHandler) LocalEndpointCreated(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A new local Endpoint has been discovered for active Gateway on node %q with IP %q",
+		endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
```
tpantelis, line 43:
```suggestion
	klog.V(log.DEBUG).Infof("A new Endpoint for the local cluster has been created: %#v", endpoint)
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+func (l *LoggerHandler) LocalEndpointRemoved(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A local Endpoint has been removed on node %q with IP %q",
+		endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
+	return nil
+}
+
+// RemoteEndpointCreated is called when an endpoint associated with a remote cluster is created.
+func (l *LoggerHandler) RemoteEndpointCreated(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A remote Endpoint has been discovered on cluster %q node %q with IP %q",
+		endpoint.Spec.ClusterID, endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
```
tpantelis, line 64:
```suggestion
        klog.V(log.DEBUG).Infof("A new Endpoint for remote cluster %q has been created: %#v", endpoint.Spec.ClusterID, endpoint)
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+// RemoteEndpointUpdated is called when an endpoint associated with a remote cluster is updated.
+func (l *LoggerHandler) RemoteEndpointUpdated(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A remote Endpoint has been updated on cluster %q node %q with IP %q",
+		endpoint.Spec.ClusterID, endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
```
tpantelis, line 71:
```suggestion
	klog.V(log.DEBUG).Infof("A new Endpoint for remote cluster %q has been updated: %#v", endpoint.Spec.ClusterID, endpoint)
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+// RemoteEndpointRemoved is called when an endpoint associated with a remote cluster is removed
+func (l *LoggerHandler) RemoteEndpointRemoved(endpoint submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("A remote Endpoint has been deleted on cluster %q node %q with IP %q",
+		endpoint.Spec.ClusterID, endpoint.Spec.Hostname, endpoint.Spec.PrivateIP)
```
tpantelis, line 78:
```suggestion
	klog.V(log.DEBUG).Infof("A new Endpoint for remote cluster %q has been removed: %#v", endpoint.Spec.ClusterID, endpoint)
```

pkg/events/events.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	k8sV1 "k8s.io/api/core/v1"
+
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+)
+
+const AnyNetworkPlugin = "any"
+
+type EventHandler interface {
```
tpantelis, line 11:
```suggestion
type Handler interface {
```
I also suggest to rename the package to `event` so references read `event.Handler` which I think is cleaner and avoids stutter.

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	k8sV1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+
+	"github.com/submariner-io/admiral/pkg/log"
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/events"
+)
+
+type LoggerHandler struct {
```
tpantelis, line 12:
```suggestion
type Handler struct {
```

Also rename package to `logger` so  references become `logger.Handler`. Also rename file to _handler.go_.

pkg/events/registry.go
```diff
@@ -0,0 +1,37 @@
+package events
+
+import (
+	"k8s.io/klog"
+)
+
+type EventRegistry struct {
```
tpantelis, line 7:
```suggestion
type Registry struct {
```

pkg/events/registry.go
```diff
@@ -0,0 +1,37 @@
...
+	eventHandlers map[string]EventHandler
+	name          string
+}
+
+func NewRegistry(name string) EventRegistry {
+	return EventRegistry{
+		name: name,
```
tpantelis, line 14: `eventHandlers` also needs to be initialized.

pkg/events/registry.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+// Adds a supported driver, prints a fatal error in the case of double registration
+func (er *EventRegistry) AddDriver(name string, eventHandler EventHandler) {
```
tpantelis, line 19:
```suggestion
// AddHandler adds an EventHandler. If one already exists with the handler's name, a fatal error is logged.
func (er *EventRegistry) AddHandler(eventHandler EventHandler) {
```

I assume `name` and `eventHandler.GetName()` should be the same.

pkg/events/registry.go
```diff
@@ -0,0 +1,37 @@
...
+	if er.eventHandlers[name] != nil {
+		klog.Fatalf("Multiple event handlers with the same name %q registered on the %q registry ", name,
+			er.name)
+	}
```
tpantelis, line 23: We also need to add `eventHandler` to the map.

```suggestion
        name := eventHandler.GetName()
	if er.eventHandlers[name] != nil {
		klog.Fatalf("An event handler with name %q already exists for registry %q", name, er.name)
	}
	
	er.eventHandlers[name] = eventHandler
```

pkg/events/registry.go
```diff
@@ -0,0 +1,37 @@
+package events
+
+import (
+	"k8s.io/klog"
+)
+
+type EventRegistry struct {
+	eventHandlers map[string]EventHandler
```
tpantelis, line 8: Can we assume handlers will be registered on startup by the same thread so we don't have to worry about synchronization?

pkg/events/registry.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Adds a supported driver, prints a fatal error in the case of double registration
+func (er *EventRegistry) AddDriver(name string, eventHandler EventHandler) {
+	if er.eventHandlers[name] != nil {
+		klog.Fatalf("Multiple event handlers with the same name %q registered on the %q registry ", name,
+			er.name)
+	}
+}
+
+func (er *EventRegistry) GetHandlers(networkPlugin string) []EventHandler {
```
tpantelis, line 26:
```suggestion
func (er *EventRegistry) GetHandlersFor(networkPlugin string) []EventHandler {
```

tpantelis (changes requested):
Where will the EventHandlers be defined and used? The event package/module seems specific to some module and should exist under that module to provide additional context/meaning, eg under `routeagent` or `cableengine`.  I mean the interface doesn't seem general enough to stand on  its own.

pkg/events/events.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Init initializes the driver with any state it needs.
+	// 	    Called once during startup, to let the plugin configure any necessary settings on the
+	//	    system once the route-agent is ready
+	Init() error
+
+	// GetName returns the name of the event handler
+	GetName() string
+
+	// GetNetworkPlugin ... logging driver says: "all"
```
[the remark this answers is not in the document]
mangelajo: It's defined on the enhancement proposal. We need to support k8s network plugins. It's a parameter that will need to be configured on the route-agent and the neworkplugin-sync agent.

Those event handlers need to define which network plugin they support, or are supposed to be called for..

pkg/events/registry.go
mangelajo: Right, This part of the registry I think it's probably better handled in a separate PR, with it's own unit testing and all..

pkg/events/registry.go
mangelajo: Yes, that's the plan, but very good point. May be it's a fact that we may need to document.

pkg/events/registry.go
mangelajo: :+1:

pkg/events/events.go
mangelajo: You have a much better sense of making uniform names for packages/modules than I do :-) thanks

pkg/events/events.go
```diff
@@ -0,0 +1,120 @@
...
+	// TransitionToNonGateway is called when there is a transition from GatewayNode to NonGateway on the
+	// This will be called only when the transition happens and not periodically.
+	TransitionToNonGateway() error
+
+	// TransitionToGateway is called when there is a transition from NonGatewayNode to GatewayNode.
+	// This will be called only when the transition happens and not periodically.
+	TransitionToGateway() error
+
+	// LocalEndpointCreated is called when the endpoint for the localCluster is created.
+	LocalEndpointCreated(endpoint submV1.Endpoint) error
```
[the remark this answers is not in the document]
mangelajo: Applied to all objects

pkg/events/events.go
```diff
@@ -0,0 +1,120 @@
...
+	// NodeUpdated indicates when a node has been updated in the cluster
+	NodeUpdated(node k8sV1.Node) error
+
+	// NodeRemoved indicates when a node has been removed from the cluster
+	NodeRemoved(node k8sV1.Node) error
+}
+
+// Base structure and methods for event handlers, we define all the event handling
+// methods considered to be optional
+type EventHandlerBase struct {
```
mangelajo, line 67: I renamed this one to HandlerBase following the above changes.

pkg/events/logger_handler/logger_handler.go
mangelajo: oh, this is part of this commit, yikes, I didn't handle this part. I'll do

pkg/events/logger_handler/logger_handler.go
mangelajo: :+1:

pkg/events/logger_handler/logger_handler.go
mangelajo: We didn't want to make it overly verbose, but .. honestly it's not an event that happens a lot

mangelajo:
@tpantelis changes applied

pkg/events/events.go
```diff
@@ -0,0 +1,120 @@
...
+	// GetNetworkPlugin ... logging driver says: "all"
+	GetNetworkPlugin() string
+
+	// Stop is called once during stop to let the network plugin cleanup anything necessary during
+	// at stop if at all. If the uninstall flag is true, the handler must cleanup everything because
+	// Submariner is being uninstalled from the system.
+	Stop(uninstall bool) error
+
+	// TransitionToNonGateway is called when there is a transition from GatewayNode to NonGateway on the
+	// This will be called only when the transition happens and not periodically.
```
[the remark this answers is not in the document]
tpantelis: Sorry I'm not clear on this, specifically the phrase "in each node".  What exactly triggers this event? Is it for the node on which the handler's process is running? Or is it for any node that transitions, in which case shouldn't the method pass identifying node information?

pkg/events/events.go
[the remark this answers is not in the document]
tpantelis: ok.  How about "_GetNetworkPlugin returns the kubernetes network plugin that this handler supports._"?

pkg/events/events.go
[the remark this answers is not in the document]
mangelajo: I rephrased your suggestion a bit, because what we want to say is... during than transition on the exact node the handler is running..., not on other nodes.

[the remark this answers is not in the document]
tpantelis: OK then I think it should be phrased as "TransitionToNonGateway is called once for each transition of the local node from Gateway to a non-Gateway.".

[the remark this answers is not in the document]
mangelajo: Done

pkg/events/handler.go
```diff
@@ -0,0 +1,114 @@
+package events
```
tpantelis, line 1:
```suggestion
package event
```

pkg/events/logger_handler/logger_handler.go
```diff
@@ -0,0 +1,83 @@
+package logger_handler
```
tpantelis, line 1: Also rename file to _handler.go_.

```suggestion
package logger
```

tpantelis:
> Where will the EventHandlers be defined and used? The event package/module seems specific to some module and should exist under that module to provide additional context/meaning, eg under `routeagent` or `cableengine`. I mean the interface doesn't seem general enough to stand on its own.

@mangelajo not sure if you saw this comment.

mangelajo:
> > Where will the EventHandlers be defined and used? The event package/module seems specific to some module and should exist under that module to provide additional context/meaning, eg under `routeagent` or `cableengine`. I mean the interface doesn't seem general enough to stand on its own.
> 
> @mangelajo not sure if you saw this comment.

Oh no sorry.

It was intentional to have it in this neutral place. The intent is that these event handlers will be created specifically for:

1) networkplugin-sync agent, 
2) route-agent

The plan is that those, at start, will create a registry of handlers and populate them.
Then we ask the registry to give us a controller, that will register himself with the different informers, and workqueues... and throw event to those handlers.

But, an important detail is that, (except for the logger), the route-agent & networkplugin sync will have their own event handler implementations, for example for ovn, so there will be one ovn handler for the route-agent (living under the routeagent tree), and there will be one ovn handler for the networkplugin-sync (living under the networkplugin-sync tree).

Both agents, at least in our design seemed to share the same set of events that needed to be handled, https://github.com/submariner-io/enhancements/blob/master/submariner/pluggable-network-plugin-and-ovn.md#synchronizer-implementation

I didn't want to put everything on one pull request because it then becomes a little hard to review. But we can keep up this conversation over the other patches.

Let me address the last comments, and let me know if this plan names sense ^

pkg/event/logger/logger.go
```diff
@@ -0,0 +1,83 @@
+package logger
+
```
tpantelis, line 2: Sorry not trying to nit - I mentioned this before - can we rename the file to handler.go for consistency?

mangelajo: Sure, sorry for missing some of your comments Tom.

tpantelis (approved):

pkg/event/logger/handler.go
```diff
@@ -0,0 +1,83 @@
...
+	klog.V(log.DEBUG).Infof("The Endpoint for the local cluster has been updated: %#v", endpoint)
+	return nil
+}
+
+func (l *Handler) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
+	klog.V(log.DEBUG).Infof("The Endpoint for the local cluster has been removed: %#v", endpoint)
+	return nil
+}
+
+func (l *Handler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
```
dfarrell07, line 49: The maximum line length is 140c, you don't need to wrap these.

dfarrell07 (approved):

PR #893, opened 2020-10-22
Event controller

mangelajo:
this PR needs the base changed, but submariner bot seems broken.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,11 +46,15 @@ bin/submariner-route-agent: vendor/modules.txt $(shell find pkg/routeagent)
 bin/submariner-globalnet: vendor/modules.txt $(shell find pkg/globalnet)
 	${SCRIPTS_DIR}/compile.sh $@ ./pkg/globalnet/main.go $(BUILD_ARGS)
 
-build: bin/submariner-engine bin/submariner-route-agent bin/submariner-globalnet
+bin/submariner-networkplugin-syncer: vendor/modules.txt $(shell find pkg/networkplugin-syncer)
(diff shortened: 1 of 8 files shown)


pkg/events/controller/controller.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/kelseyhightower/envconfig"
+
+	k8sInformersV1 "k8s.io/client-go/informers/core/v1"
+
+	submarinerInformers "github.com/submariner-io/submariner/pkg/client/informers/externalversions"
+	submarinerInformersV1 "github.com/submariner-io/submariner/pkg/client/informers/externalversions/submariner.io/v1"
+)
+
+type submarinerEventControllerSpecification struct {
```
tpantelis, line 15:
```suggestion
type specification struct {
```

pkg/events/controller/controller.go
```diff
@@ -0,0 +1,49 @@
...
+	Namespace   string
+}
+
+
+type SubmarinerEventController struct {
```
tpantelis, line 20:
```suggestion
type Controller struct {
```

pkg/events/controller/controller.go
```diff
@@ -0,0 +1,49 @@
+package controller
+
```
tpantelis, line 2: I don't think this needs to be in its own package. If we move it up a directory then it would be referenced as `event.Controller` rather than `controller.Controller`.

mangelajo:
Closing this one in favor of this https://github.com/submariner-io/submariner/pull/908

PR #901, opened 2020-10-26
Eventhandler registry

mangelajo:
This registry will be used later by the networkplugin-syncer or the route-agent.
    
### DIFF
--- a/pkg/event/event_suite_test.go
+++ b/pkg/event/event_suite_test.go
@@ -1,0 +1,140 @@
+package event_test
+
+import (
+	"testing"
+
+	k8sV1 "k8s.io/api/core/v1"
+
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/event"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestEventPackage(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Event suite")
+}
+
+type TestEvent struct {
+	Name      string
+	Handler   string
+	Parameter interface{}
+}
+
+type TestHandler struct {
+	event.HandlerBase
+	Name          string
+	NetworkPlugin string
+	FailOnEvent   error
+	Events        chan TestEvent
+	Initialized   bool
+}
+
+var allTestEvents chan TestEvent = make(chan TestEvent, 1000)
+
+func cleanupTestEvents() {
+	allTestEvents = make(chan TestEvent, 1000)
+}
+
+func NewTestHandler(name, networkPlugin string) *TestHandler {
+	return &TestHandler{
+		Name:          name,
+		NetworkPlugin: networkPlugin,
+		Events:        make(chan TestEvent, 100),
+		Initialized:   false,
+	}
+}
+
+func (t *TestHandler) addEvent(eventName string, param interface{}) error {
+	if t.FailOnEvent == nil {
+		ev := TestEvent{
+			Name:      eventName,
+			Parameter: param,
+		}
+
+		t.Events <- ev
+
+		// On the global channel we also include the handler name
+		// so we can verify ordering
+		ev.Handler = t.Name
+		allTestEvents <- ev
+	}
+
+	return t.FailOnEvent
+}
+func (t *TestHandler) Init() error {
+	t.Initialized = true
+
+	return nil
+}
+
+func (t *TestHandler) GetName() string {
+	return t.Name
+}
+
+func (t *TestHandler) GetNetworkPlugin() string {
+	return t.NetworkPlugin
+}
+
+const evTransitionToNonGateway = "TransitionToNonGateway"
+const evTransitionToGateway = "TransitionToGateway"
+const evLocalEndpointCreated = "LocalEndpointCreated"
+const evLocalEndpointUpdated = "LocalEndpointUpdated"
+const evLocalEndpointRemoved = "LocalEndpointRemoved"
+const evRemoteEndpointCreated = "RemoteEndpointCreated"
+const evRemoteEndpointUpdated = "RemoteEndpointUpdated"
+const evRemoteEndpointRemoved = "RemoteEndpointRemoved"
+const evNodeCreated = "NodeCreated"
+const evNodeUpdated = "NodeUpdated"
+const evNodeRemoved = "NodeRemoved"
+const evStop = "Stop"
+
+func (t *TestHandler) Stop(uninstall bool) error {
+	return t.addEvent(evStop, uninstall)
+}
+
+func (t *TestHandler) TransitionToNonGateway() error {
+	return t.addEvent(evTransitionToNonGateway, nil)
+}
+
+func (t *TestHandler) TransitionToGateway() error {
+	return t.addEvent(evTransitionToGateway, nil)
+}
+
+func (t *TestHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
+	return t.addEvent(evLocalEndpointCreated, endpoint)
+}
+
+func (t *TestHandler) LocalEndpointUpdated(endpoint *submV1.Endpoint) error {
+	return t.addEvent(evLocalEndpointUpdated, endpoint)
+}
+
+func (t *TestHandler) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
+	return t.addEvent(evLocalEndpointRemoved, endpoint)
+}
+
+func (t *TestHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	return t.addEvent(evRemoteEndpointCreated, endpoint)
+}
+
+func (t *TestHandler) RemoteEndpointUpdated(endpoint *submV1.Endpoint) error {
+	return t.addEvent(evRemoteEndpointUpdated, endpoint)
+}
+
+func (t *TestHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	return t.addEvent(evRemoteEndpointRemoved, endpoint)
+}
+
+func (t *TestHandler) NodeCreated(node *k8sV1.Node) error {
+	return t.addEvent(evNodeCreated, node)
+}
+
+func (t *TestHandler) NodeUpdated(node *k8sV1.Node) error {
+	return t.addEvent(evNodeUpdated, node)
+}
+
+func (t *TestHandler) NodeRemoved(node *k8sV1.Node) error {
+	return t.addEvent(evNodeRemoved, node)
+}
--- a/pkg/event/handler.go
+++ b/pkg/event/handler.go
@@ -47,8 +47,8 @@ type Handler interface {
 	// RemoteEndpointRemoved is called when an endpoint associated with a remote cluster is removed
 	RemoteEndpointRemoved(endpoint *submV1.Endpoint) error
 
-	// NodeAdded indicates when a node has been added to the cluster
-	NodeAdded(node *k8sV1.Node) error
+	// NodeCreated indicates when a node has been added to the cluster
+	NodeCreated(node *k8sV1.Node) error
 
 	// NodeUpdated indicates when a node has been updated in the cluster
 	NodeUpdated(node *k8sV1.Node) error
@@ -101,7 +101,7 @@ func (ev *HandlerBase) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 	return nil
 }
 
-func (ev *HandlerBase) NodeAdded(node *k8sV1.Node) error {
+func (ev *HandlerBase) NodeCreated(node *k8sV1.Node) error {
 	return nil
 }
 
--- a/pkg/event/logger/handler.go
+++ b/pkg/event/logger/handler.go
@@ -13,6 +13,10 @@ type Handler struct {
 	event.HandlerBase
 }
 
+func NewHandler() event.Handler {
+	return &Handler{}
+}
+
 func (l *Handler) GetName() string {
 	return "logger"
 }
@@ -64,7 +68,7 @@ func (l *Handler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 	return nil
 }
 
-func (l *Handler) NodeAdded(node *k8sV1.Node) error {
+func (l *Handler) NodeCreated(node *k8sV1.Node) error {
 	klog.V(log.DEBUG).Infof("A Node with name %q and addresses %#v has been added to the cluster",
 		node.Name, node.Status.Addresses)
 	return nil
--- a/pkg/event/registry.go
+++ b/pkg/event/registry.go
@@ -1,0 +1,144 @@
+package event
+
+import (
+	"github.com/pkg/errors"
+
+	k8sV1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog"
+
+	"github.com/submariner-io/admiral/pkg/log"
+
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+)
+
+type Registry struct {
+	name          string
+	networkPlugin string
+	eventHandlers []Handler
+}
+
+// NewRegistry creates a new registry with the given name,  typically referencing the owner, to manage event
+// Handlers that match the given networkPlugin name.
+func NewRegistry(name, networkPlugin string) Registry {
+	return Registry{
+		name:          name,
+		networkPlugin: networkPlugin,
+		eventHandlers: []Handler{},
+	}
+}
+
+func (er *Registry) addHandler(eventHandler Handler) error {
+	evNetworkPlugin := eventHandler.GetNetworkPlugin()
+
+	if evNetworkPlugin == AnyNetworkPlugin || evNetworkPlugin == er.networkPlugin {
+		if err := eventHandler.Init(); err != nil {
+			return errors.Wrapf(err, "Event handler %q failed to initialize", eventHandler.GetName())
+		}
+
+		er.eventHandlers = append(er.eventHandlers, eventHandler)
+		klog.Infof("Event handler %q added to registry %q.", eventHandler.GetName(), er.name)
+	} else {
+		klog.V(log.DEBUG).Infof("Event handler %q ignored for registry %q.", eventHandler.GetName(), er.name)
+	}
+
+	return nil
+}
+
+// AddHandlers adds the given event Handlers whose associated network plugin matches the network plugin
+// associated with this registry. Non-matching Handlers are ignored. Handlers will be called in registration order.
+func (er *Registry) AddHandlers(eventHandlers ...Handler) error {
+	for _, eventHandler := range eventHandlers {
+		err := er.addHandler(eventHandler)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (er *Registry) StopHandlers(uninstall bool) error {
+	return er.invokeHandlers("Stop", func(h Handler) error {
+		return h.Stop(uninstall)
+	})
+}
+
+func (er *Registry) TransitionToNonGateway() error {
+	return er.invokeHandlers("TransitionToNonGateway", func(h Handler) error {
+		return h.TransitionToNonGateway()
+	})
+}
+
+func (er *Registry) TransitionToGateway() error {
+	return er.invokeHandlers("TransitionToGateway", func(h Handler) error {
+		return h.TransitionToGateway()
+	})
+}
+
+func (er *Registry) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
+	return er.invokeHandlers("LocalEndpointCreated", func(h Handler) error {
+		return h.LocalEndpointCreated(endpoint)
+	})
+}
+
+func (er *Registry) LocalEndpointUpdated(endpoint *submV1.Endpoint) error {
+	return er.invokeHandlers("LocalEndpointUpdated", func(h Handler) error {
+		return h.LocalEndpointUpdated(endpoint)
+	})
+}
+
+func (er *Registry) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
+	return er.invokeHandlers("LocalEndpointRemoved", func(h Handler) error {
+		return h.LocalEndpointRemoved(endpoint)
+	})
+}
+
+func (er *Registry) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	return er.invokeHandlers("RemoteEndpointCreated", func(h Handler) error {
+		return h.RemoteEndpointCreated(endpoint)
+	})
+}
+
+func (er *Registry) RemoteEndpointUpdated(endpoint *submV1.Endpoint) error {
+	return er.invokeHandlers("RemoteEndpointUpdated", func(h Handler) error {
+		return h.RemoteEndpointUpdated(endpoint)
+	})
+}
+
+func (er *Registry) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	return er.invokeHandlers("RemoteEndpointRemoved", func(h Handler) error {
+		return h.RemoteEndpointRemoved(endpoint)
+	})
+}
+
+func (er *Registry) NodeCreated(node *k8sV1.Node) error {
+	return er.invokeHandlers("NodeCreated", func(h Handler) error {
+		return h.NodeCreated(node)
+	})
+}
+
+func (er *Registry) NodeUpdated(node *k8sV1.Node) error {
+	return er.invokeHandlers("NodeUpdated", func(h Handler) error {
+		return h.NodeUpdated(node)
+	})
+}
+
+func (er *Registry) NodeRemoved(node *k8sV1.Node) error {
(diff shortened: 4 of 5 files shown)


pkg/event/registry.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (er *Registry) TransitionToNonGateway() error {
+	errors := map[string] error{}
+
+	for _, h := range er.eventHandlers {
+		errors[h.GetName()] = h.TransitionToNonGateway()
+	}
+
+	return flattenErrors(errors, "TransitionToNonGateway")
```
tpantelis, line 48: We can make a generic function and use k8s `NewAggregate` instead of `flattenErrors`, ie

```
k8serrors "k8s.io/apimachinery/pkg/util/errors"
...

func (er *Registry) invokeHandlers(eventName string, invoke func(h Handler) error) error {
	var errs []error

	for _, h := range er.eventHandlers {
		err := invoke(h)
		if err != nil {
			errs = append(errs, errors.Wrapf(err, "%q returned error", h.GetName()))
		}
	}

	return errors.Wrapf(k8serrors.NewAggregate(errs), "%s failed", eventName)
}
```

Then this method becomes:

  ```
  return er.invokeHandlers("TransitionToNonGateway", func(h Handler) error {
		return h.TransitionToNonGateway()
	})
```

File: pkg/event/registry.go [hunk 2]
```diff
@@ -0,0 +1,163 @@
...
+	evNetworkPlugin := eventHandler.GetNetworkPlugin()
+
+	if evNetworkPlugin == AnyNetworkPlugin || evNetworkPlugin == er.networkPlugin {
+		er.eventHandlers = append(er.eventHandlers, eventHandler)
+		klog.Infof("Event handler %q loaded for the %q registry ",
+			eventHandler.GetName(), er.name)
+	} else {
+		klog.V(log.DEBUG).Infof("Event handler %q ignored for the %q registry ",
+			eventHandler.GetName(), er.name)
+	}
```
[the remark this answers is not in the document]
mangelajo: > I'm not clear of the purpose of the `networkPlugin` field defined on the `Registry`. It seems then we'd need a registry of network plugin `Registry`'s? How will handlers know to get the right `Registry` to register with? I'd think there would be one `Registry` for all handlers.

I guess that my intent will be clearer when I finish the unit tests.

The idea was... that .. on the main of the routeagent/networksyncer they will create a registry, and state which network plugin is being used:

```
handlerRegistry := NewRegistry("route-agent", $NETWORK_PLUGIN_FROM_ENV_this_is_given_by_operator)
handlerRegistry.AddHandler(logger.NewHandler())
handlerRegistry.AddHandler(kubeproxy_iptables_handler.NewHandler())
handlerRegistry.AddHandler(ovn_routeagent_handler.NewHandler())
```

Then later on we can create the controller:
```
controller := controller.NewController(hanlderRegistry)
controller.Start(stopCh)
```


But the idea is that the registry already knows on-runtime, which network plugin are we serving for, so it knows which handlers it can discard (dynamically)... instead of keeping them all and returning a list afterwards, I thought we could just optimize that memory and free it as soon as possible. (not that it's a lot.., but won't be used)

pkg/event/registry.go
mangelajo: Yes, that works only for the parameter-less, we'd need one invokeHandlers by type of function, unless there is some go-magic that would allow us to do a more dynamic thing... that would be perfect.

mangelajo: I thought of this, but ... at least we save 1:3 of the code... so it's worth it, I'll update this.

I didn't know we had an error aggregate helper, that's cool

mangelajo: Ahh, no with the lambda function we can handle the parameters, very clever :+1: :heart:

mangelajo: Applied

pkg/event/handler.go
```diff
@@ -47,8 +47,8 @@ type Handler interface {
 	// RemoteEndpointRemoved is called when an endpoint associated with a remote cluster is removed
 	RemoteEndpointRemoved(endpoint *submV1.Endpoint) error
 
-	// NodeAdded indicates when a node has been added to the cluster
-	NodeAdded(node *k8sV1.Node) error
+	// NodeCreated indicates when a node has been added to the cluster
+	NodeCreated(node *k8sV1.Node) error
```
mangelajo, line 51: I detected this being odd respect to the others on the interface, so I updated it. Now all of them are *Created..

pkg/event/registry.go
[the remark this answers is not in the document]
tpantelis: OK - that clarifies it. There will be a singleton registry which will filter out handlers based on process input.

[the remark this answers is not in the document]
mangelajo: Correct, I will be updating this patch with full testing in a second.

pkg/event/logger/handler.go
```diff
@@ -23,61 +27,72 @@ func (l *Handler) GetNetworkPlugin() string {
...
 	klog.V(log.DEBUG).Infof("A Node with name %q and addresses %#v has been updated",
 		node.Name, node.Status.Addresses)
+
 	return nil
 }
 
 func (l *Handler) NodeRemoved(node *k8sV1.Node) error {
 	klog.V(log.DEBUG).Infof("A Node with name %q has been removed",
 		node.Name)
+
```
mangelajo, line 96: unrelated to this patch, my local linter was complaining I will remove those spaces.

mangelajo:
This one is hopefully ready, but submariner-io:z_pr891/mangelajo/event-interface  must be merged first.

mangelajo:
Ok, I updated the patch, this is ready for reviews again

pkg/event/registry_test.go
```diff
@@ -0,0 +1,106 @@
...
+			checkRegistryCall(registry.NodeUpdated(node), handler1, evNodeUpdated, node)
+			checkRegistryCall(registry.NodeRemoved(node), handler1, evNodeRemoved, node)
+			checkRegistryCall(registry.LocalEndpointCreated(endpoint), handler1, evLocalEndpointCreated, endpoint)
+			checkRegistryCall(registry.LocalEndpointUpdated(endpoint), handler1, evLocalEndpointUpdated, endpoint)
+			checkRegistryCall(registry.LocalEndpointRemoved(endpoint), handler1, evLocalEndpointRemoved, endpoint)
+			checkRegistryCall(registry.RemoteEndpointCreated(endpoint), handler1, evRemoteEndpointCreated, endpoint)
+			checkRegistryCall(registry.RemoteEndpointUpdated(endpoint), handler1, evRemoteEndpointUpdated, endpoint)
+			checkRegistryCall(registry.RemoteEndpointRemoved(endpoint), handler1, evRemoteEndpointRemoved, endpoint)
+
+		})
```
tpantelis, line 92: I think we can streamline this a bit:

```
It("should invoke all handlers for all events", func() {
    endpoint := &submV1.Endpoint{ObjectMeta: v1meta.ObjectMeta{Name: "endpoint1"}}
    node := &k8sV1.Node{ObjectMeta: v1meta.ObjectMeta{Name: "node1"}}

    // I only defined some of the events... 
    events := map[TestEvent]func () error {
        {Name: evTransitionToGateway}: registry.TransitionToGateway,
        ....
        {Name: evNodeCreated, Parameter: node}: func() error {return registry.NodeCreated(node)}, 
        ...
        {Name: evLocalEndpointCreated, Parameter: endpoint}: func() error {return registry. LocalEndpointCreated(endpoint)},
        ...
        {Name: evStop, true}: func() error {return registry. StopHandlers(true)} ,
    }

    for ev, f := range events {
        err := f()
        Expect(err).To(Succeed())
        Expect(handler1.Events).Should(Receive(&ev)) 
        Expect(handler2.Events).Should(Receive(&ev)) 
    }
})
```

pkg/event/event_suite_test.go
```diff
@@ -0,0 +1,140 @@
...
+	HandlerNetworkPlugin string
+	Parameter            interface{}
+}
+
+type TestHandler struct {
+	event.HandlerBase
+	Name          string
+	NetworkPlugin string
+	FailOnEvent   error
+	EventList     []TestEvent
```
tpantelis, line 35: I think it would be easier and more flexible to make this `Events chan *TestEvent` instead of an array.

pkg/event/event_suite_test.go
```diff
@@ -0,0 +1,140 @@
...
+func TestRegistryMethods(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Event Registry suite")
+}
+
+type TestEvent struct {
+	Timestamp            int64
+	Name                 string
+	HandlerName          string
+	HandlerNetworkPlugin string
```
tpantelis, line 26: These are unused.

tpantelis (changes requested):

pkg/event/event_suite_test.go
```diff
@@ -0,0 +1,140 @@
...
+const evLocalEndpointCreated = "LocalEndpointCreated"
+const evLocalEndpointUpdated = "LocalEndpointUpdated"
+const evLocalEndpointRemoved = "LocalEndpointRemoved"
+const evRemoteEndpointCreated = "RemoteEndpointCreated"
+const evRemoteEndpointUpdated = "RemoteEndpointUpdated"
+const evRemoteEndpointRemoved = "RemoteEndpointRemoved"
+const evNodeCreated = "NodeCreated"
+const evNodeUpdated = "NodeUpdated"
+const evNodeRemoved = "NodeRemoved"
+const evStop = "Stop"
```
sridhargaddam, line 92: Can we move this to handler.go so that we can reuse them even in registry.go

pkg/event/registry.go
```diff
@@ -0,0 +1,142 @@
...
+		name:          name,
+		networkPlugin: networkPlugin,
+		eventHandlers: []Handler{},
+	}
+}
+
+func (er *Registry) addHandler(eventHandler Handler) {
+	evNetworkPlugin := eventHandler.GetNetworkPlugin()
+
+	if evNetworkPlugin == AnyNetworkPlugin || evNetworkPlugin == er.networkPlugin {
```
sridhargaddam, line 36: Instead of using an empty value for "AnyNetworkPlugin", shall we use "All" (or "Any")?

pkg/event/registry.go
```diff
@@ -0,0 +1,142 @@
...
+		er.eventHandlers = append(er.eventHandlers, eventHandler)
+		klog.Infof("Event handler %q loaded for the %q registry ",
+			eventHandler.GetName(), er.name)
+
+		if err := eventHandler.Init(); err != nil {
+			klog.Fatalf("Event handler %q failed to initialize: %v", eventHandler.GetName(), err)
```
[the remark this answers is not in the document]
sridhargaddam: Also, it might be better to first initialize and only if it's successful, append it to er.eventHandlers (i.e., move this to a line before 37)

pkg/event/registry.go
```diff
@@ -0,0 +1,142 @@
...
+// the runtime network plugin will be ignored, and the ones that work with the runtime network plugin
+// will be stored on the registry, so later method invocations will be dispatched to the handler.
+func (er *Registry) AddHandlers(eventHandlers ...Handler) {
+	for _, eventHandler := range eventHandlers {
+		er.addHandler(eventHandler)
+	}
+}
+
+func (er *Registry) StopHandlers(uninstall bool) error {
+	return er.invokeHandlers("Stop", func(h Handler) error {
```
sridhargaddam, line 60: Instead of event names (aka Stop), it will be better to use constants/macros here as commented above.

pkg/event/registry_test.go
```diff
@@ -0,0 +1,106 @@
...
+	When("Registry is populated with different network plugin handlers", func() {
+		var (
+			ovn, generic *TestHandler
+			registry     event.Registry
+		)
+
+		BeforeEach(func() {
+			registry = event.NewRegistry("test-registry", npGenericKubeproxyIptables)
+			ovn = NewTestHandler("handler-ovn", npOVNKubernetes)
+			generic = NewTestHandler("handler-ovn", npGenericKubeproxyIptables)
```
sridhargaddam, line 27:
```suggestion
			generic = NewTestHandler("handler-kubeproxy", npGenericKubeproxyIptables)
```

pkg/event/registry_test.go
```diff
@@ -0,0 +1,106 @@
...
+			registry.AddHandlers(logger.NewHandler(), ovn, generic)
```
sridhargaddam, line 28: nit: logger.NewHandler() is unused in this testcase.

pkg/event/registry_test.go
```diff
@@ -0,0 +1,106 @@
...
+			handler2 = NewTestHandler("handler2", npGenericKubeproxyIptables)
+			registry.AddHandlers(logger.NewHandler(), handler1, handler2)
+		})
+
+		It("All handlers should be initialized", func() {
+			Expect(handler1.Initialized).To(BeTrue())
+			Expect(handler2.Initialized).To(BeTrue())
+		})
+
+		It("Should call all handlers", func() {
```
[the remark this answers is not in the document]
sridhargaddam: I'm afraid no. In L76, we call checkRegistryCall where we clear the events after every call. In this API, we are ensuring that when an event happens it is delivered to all the registered handlers.

pkg/event/registry_test.go
```diff
@@ -0,0 +1,106 @@
...
+			_ = registry.TransitionToGateway()
+			Expect(handler1.EventList).To(HaveLen(1))
+			Expect(handler2.EventList).To(HaveLen(1))
+			Expect(handler1.EventList[0].Name).Should(Equal(evTransitionToGateway))
+			Expect(handler2.EventList[0].Name).Should(Equal(evTransitionToGateway))
+		})
+
+		It("Should be called in registration order", func() {
```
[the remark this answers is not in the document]
sridhargaddam: IMHO, it might be better to enforce this in the API contract, although its not a major issue if the order is not enforced for the current implementation.

sridhargaddam (changes requested):
Nice work @mangelajo. Some comments.

pkg/event/event_suite_test.go
```diff
@@ -0,0 +1,140 @@
...
+	HandlerName          string
+	HandlerNetworkPlugin string
+	Parameter            interface{}
+}
+
+type TestHandler struct {
+	event.HandlerBase
+	Name          string
+	NetworkPlugin string
+	FailOnEvent   error
```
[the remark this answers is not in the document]
mangelajo: Oh right, I wanted to test the error handling of a single eventhandler. Let me finish those tests.

pkg/event/event_suite_test.go
mangelajo: HandlerNetworkPlugin is used. I'll check the others.

mangelajo: oh, right on the event it's not used.

pkg/event/event_suite_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/submariner-io/submariner/pkg/event"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestRegistryMethods(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Event Registry suite")
+}
```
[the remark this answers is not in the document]
mangelajo: Yeah, we need another name

pkg/event/event_suite_test.go
mangelajo: Hmm, I tend to prefer separating the constants of test & real implementation, so we double check that we are writting them all correct. Also if somebody inadvertedly changes the const on implementation it will fail the test.

pkg/event/registry.go
mangelajo: It was a previous suggestion by @tpantelis, either way works for me.

pkg/event/registry.go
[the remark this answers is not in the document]
mangelajo: Right, we do that everywhere. My thinking was that this kind of failure will be fatal anyway, and we need to stop. So it didn't matter, but for code clarity, when people read it they will see that it can err & exit.

pkg/event/registry.go
mangelajo: I prefer the practice of separating both. In this case I believe we don't really need the constants, because those are not used anywhere else (except for testing...), so we'd only be adding lines. 

Even on the testing side may be it's worth removing the constants into literals to save lines

pkg/event/registry_test.go
mangelajo: I wanted the logger to be exercised for all methods, so if there's something broken about it it will fail.

pkg/event/registry_test.go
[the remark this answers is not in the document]
mangelajo: Yes exactly, this one is ensuring that we call all handlers in the chain, the other one exercises all the calls to make sure they are properly handled.

pkg/event/registry_test.go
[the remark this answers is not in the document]
mangelajo: Yeah, there will be cases where the ordering is necessary, in the end it will be controlled by the creator of the registry. We should document this, it's a good point.

pkg/event/registry_test.go
```diff
@@ -0,0 +1,106 @@
...
+			_ = registry.TransitionToGateway()
+			Expect(handler1.EventList).To(HaveLen(1))
+			Expect(handler2.EventList).To(HaveLen(1))
+			Expect(handler1.EventList[0].Timestamp).To(BeNumerically("<", handler2.EventList[0].Timestamp))
```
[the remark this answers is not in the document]
mangelajo: Well, if we have processors running under 1ns per instruction, and al that can happen on one instruction.
I can change the Timestamp to be a global incremented variable, instead of a global chan. The chan is elegant TBH, but I'd prefer not to spend too much time in making the test more elgant, this is blocking a lot of work ':)

mangelajo (commented):
Thanks for the review guys, I'm applying most of the suggestions and working on the the ones I can't apply directly.

pkg/event/registry.go
mangelajo: No, in testing we can't remove the constants we use them at least twice each.

pkg/event/registry_test.go
mangelajo: Done.

mangelajo:
@tpantelis I finally did the event channels too ;)

pkg/event/event_suite_test.go
mangelajo: Made it a channel in the end.

pkg/event/event_suite_test.go
sridhargaddam: Okay

sridhargaddam (approved):

pkg/event/registry.go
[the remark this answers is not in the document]
tpantelis: It probably is fatal but let the top-level caller decide that. Also it allows the caller(s) to handle fatal appropriately. Remember in the engine we write the GW status on fatal error. Also  propagating the error makes it easier for unit tests.

pkg/event/registry_test.go
tpantelis: I think that should be in a separate test specific to the logger handler.

pkg/event/registry_test.go
```diff
@@ -0,0 +1,108 @@
...
+				Expect(handler2.Events).Should(Receive(Equal(ev)))
+			}
+		})
+
+		It("one handler returns error subsequent handlers should still execute", func() {
+			handler1.FailOnEvent = errors.New("I shall fail!")
+			err := registry.TransitionToGateway()
+			Expect(handler1.Events).ShouldNot(Receive())
+			Expect(handler2.Events).Should(Receive())
+			Expect(err).Should(HaveOccurred())
```
tpantelis, line 105: Seeing the error handling usage now makes me wonder of the registry should return an error at all or just log if a handler fails. One handler could succeed while another fails - the caller has no knowledge of that. How is a caller to handle the returned error other than to log it...

pkg/event/event_suite_test.go
```diff
@@ -0,0 +1,143 @@
...
+func cleanupTestEvents() {
+	allTestEvents = make(chan TestEvent, 1000)
+}
+
+func NewTestHandler(name, networkPlugin string) *TestHandler {
+	return &TestHandler{
+		Name:          name,
+		NetworkPlugin: networkPlugin,
+		Events:        make(chan TestEvent, 100),
+		FailOnEvent:   fmt.Errorf("Test handler not initialized"),
```
tpantelis, line 47: We can remove the initialization here since we're no using it.

pkg/event/event_suite_test.go
```diff
@@ -0,0 +1,143 @@
...
+		// On the global channel we also include the handler name
+		// so we can verify ordering
+		ev.Handler = t.Name
+		allTestEvents <- ev
+	}
+
+	return t.FailOnEvent
+}
+func (t *TestHandler) Init() error {
+	t.FailOnEvent = nil
```
tpantelis, line 70: Remove this line.

pkg/event/registry.go
tpantelis: Typically empty is used for such things to eliminate the chance for a clash with a real value, however unlikely that is.

pkg/event/registry_test.go
```diff
@@ -0,0 +1,108 @@
...
+		var (
+			handler1, handler2 *TestHandler
+			registry           event.Registry
+		)
+
+		BeforeEach(func() {
+			registry = event.NewRegistry("test-registry", npGenericKubeproxyIptables)
+			handler1 = NewTestHandler("handler1", npGenericKubeproxyIptables)
+			handler2 = NewTestHandler("handler2", npGenericKubeproxyIptables)
+			err := registry.AddHandlers(logger.NewHandler(), handler1, handler2)
```
tpantelis, line 57: It's useful to test a "wildcarded" handler but, instead of using the logger handler, I think we should use a test handler so we can verify it.

pkg/event/registry_test.go
```diff
@@ -0,0 +1,108 @@
...
+			Expect(handler2.Initialized).To(BeTrue())
+		})
+
+		It("Should be called in registration order", func() {
+			_ = registry.TransitionToGateway()
+			Expect(allTestEvents).Should(Receive(Equal(TestEvent{Name: evTransitionToGateway, Handler: handler1.Name})))
+			Expect(allTestEvents).Should(Receive(Equal(TestEvent{Name: evTransitionToGateway, Handler: handler2.Name})))
+		})
+
+		It("Should work for all calls", func() {
```
tpantelis, line 73:
```suggestion
		It("should invoke all handlers for all events", func() {
```

pkg/event/registry_test.go
```diff
@@ -0,0 +1,108 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			cleanupTestEvents()
+		})
+
+		It("should initialize all the handlers", func() {
+			Expect(handler1.Initialized).To(BeTrue())
+			Expect(handler2.Initialized).To(BeTrue())
+		})
+
+		It("Should be called in registration order", func() {
```
tpantelis, line 67:
```suggestion
		It("should invoke handlers in registration order", func() {
```

pkg/event/registry_test.go
```diff
@@ -0,0 +1,108 @@
...
+			_ = registry.TransitionToGateway()
+			Expect(allTestEvents).Should(Receive(Equal(TestEvent{Name: evTransitionToGateway, Handler: handler1.Name})))
+			Expect(allTestEvents).Should(Receive(Equal(TestEvent{Name: evTransitionToGateway, Handler: handler2.Name})))
```
tpantelis, line 70: To be complete we should test this for all events. I suggest removing this spec and adding these checks to the spec below.

pkg/event/event_suite_test.go
mangelajo: right.

pkg/event/registry_test.go
mangelajo: I'm not sure what you mean.

My intent was exercising the logging handler to make sure at least it doesnt crash.

Proper testing would mean redirecting the log to somewhere else and checking the output but I didn't think we needed to go so far.

pkg/event/registry_test.go
mangelajo: We can do that, yup

pkg/event/registry_test.go
tpantelis: Irregardless of the logger handler, I meant to add a `handler3 = NewTestHandler("handler3", AnyNetworkPlugin)` and verify it also receives all the events. 

I understand the intent of adding in the logger handler here but perhaps it should be tested separately via it's own test file.

pkg/event/registry_test.go
mangelajo: The events will be coming through a workqueue, we should expect that the handlers are idempotent, and that the events will be retried if any handler fails,

Probably the controller will need to put create/update/delete events in a single workqueue for something like that to work, but that's implementation we can discuss on next PRs.

Let's see how it unfolds/looks on next PRs. It would be equivalent to what we are doing today, we are just changing the shape into something more modular where code will be better organized than today.

pkg/event/registry_test.go
mangelajo: Ahh, that's a good point!. thanks

tpantelis (approved):

Issue #821, opened 2020-09-18
HA: With "cluster to cluster heartbeats" to identify the cluster down and update the lighthouse DNS cache as soon as possible


ming-ddtechcg:
I am running v0.6.1 on the two Openshift 4.4.20 clusters with "--cable-driver libreswan" while they joined together, say "cluster A" and "cluster B".

Deployed the test nginx-unprivileged case from "KIND (Local Environment)" to both clusters and expose services with kubectl and subctl.  Performed nslookup with the FQDN service name in the pod "tmp-shell", and two service IP addresses are returned as round-robin.

The test scenario is that to shut down the cluster A (I bring down all worker nodes) and the service will be routed to the cluster B only:

**the result of the testing to return the cluster B service IP took approximately _25 minutes_ (the timings were more or less sometimes).**

There was the cluster B Gateway describe while the cluster A was down after 10 minutes:
```
$ kubectl describe Gateway -n submariner-operator  
Name:         clusterb-62kfm-worker-0-2pvhw
Namespace:    submariner-operator
Labels:       <none>
Annotations:  update-timestamp: 1600352166
API Version:  submariner.io/v1
Kind:         Gateway
Metadata:
  Creation Timestamp:  2020-09-17T13:06:49Z
  Generation:          2
  Resource Version:    5746231
  Self Link:           /apis/submariner.io/v1/namespaces/submariner-operator/gateways/clusterb-62kfm-worker-0-2pvhw
  UID:                 0886e928-8706-4103-a3cb-5cf1a8390e51
Status:
  Connections:
    Endpoint:
      Backend:      libreswan
      cable_name:   submariner-cable-cluster-clustera-10-5-33-224
      cluster_id:   cluster-a
      Hostname:     worker0.clustera.test.com
      nat_enabled:  false
      private_ip:   10.5.33.224
      public_ip:    
      Subnets:
        172.30.0.0/16
        10.128.0.0/16
    Status:          connected
    Status Message:  
  Ha Status:         active
  Local Endpoint:
    Backend:      libreswan
    cable_name:   submariner-cable-cluster-b-10-5-71-225
    cluster_id:   cluster-c16
    Hostname:     clusterb-62kfm-worker-0-2pvhw
    nat_enabled:  false
    private_ip:   10.5.71.225
    public_ip:    
    Subnets:
      172.31.0.0/16
      10.132.0.0/14
  Status Failure:  
  Version:         v0.6.0
Events:            <none>
```

**Recommendation:**

[...]

PR #902, opened 2020-10-27
Add fields for HealthCheck

aswinsuryan:
*HealthCheck Ip is added in EndpointSpec
*Latency is added in GatewayStatus

Fixes: submariner-io/submariner#821


pkg/apis/submariner.io/v1/types.go
```diff
@@ -77,6 +78,7 @@ type GatewayStatus struct {
 	Version       string       `json:"version"`
 	HAStatus      HAStatus     `json:"haStatus"`
 	LocalEndpoint EndpointSpec `json:"localEndpoint"`
+	Latency       string       `json:"latency"`
```
mangelajo, line 81: May be we can include a composite field where we will report last/min/avg/max/mdev  fields? 

See the output of ping: 
```
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 2.552/2.647/2.837/0.134 ms
```

That gives a more meaningful picture of latency (instead of just instant latency)

mangelajo (changes requested):

pkg/apis/submariner.io/v1/types.go
aswinsuryan: Changed it to a composite field.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -77,10 +79,20 @@ type GatewayStatus struct {
...
 	StatusFailure string       `json:"statusFailure"`
 	Connections   []Connection `json:"connections"`
 }
 
+type LatencySpec struct {
+	Last string `json:"last"`
+	Min  string `json:"min"`
+	Avg  string `json:"avg"`
+	Max  string `json:"max"`
+	Mdev string `json:"mdev"`
```
sridhargaddam, line 93: Does the golang package we plan to use for "ping" provide all this info?

pkg/apis/submariner.io/v1/types.go
```diff
@@ -77,10 +79,20 @@ type GatewayStatus struct {
...
+	// +optional
+	Latency       LatencySpec  `json:"latency"`
 	StatusFailure string       `json:"statusFailure"`
 	Connections   []Connection `json:"connections"`
 }
 
+type LatencySpec struct {
+	Last string `json:"last"`
+	Min  string `json:"min"`
+	Avg  string `json:"avg"`
```
tpantelis, line 91:
```suggestion
	Average  string `json:"average"`
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -43,8 +43,10 @@ type Endpoint struct {
 }
 
 type EndpointSpec struct {
-	ClusterID     string            `json:"cluster_id"`
-	CableName     string            `json:"cable_name"`
+	ClusterID string `json:"cluster_id"`
+	CableName string `json:"cable_name"`
+	// +optional
+	HealthCheckIP string            `json:"health_check_ip,omitempty"`
```
tpantelis, line 49:
```suggestion
	HealthCheckIP string            `json:"healthCheckIP,omitempty"`
```

tpantelis (changes requested):

pkg/apis/submariner.io/v1/types.go
aswinsuryan: Done.

pkg/apis/submariner.io/v1/types.go
aswinsuryan: Done.

pkg/apis/submariner.io/v1/types.go
aswinsuryan: I need to check this. Will update.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -77,10 +79,20 @@ type GatewayStatus struct {
...
 	StatusFailure string       `json:"statusFailure"`
 	Connections   []Connection `json:"connections"`
 }
 
+type LatencySpec struct {
+	Last    string `json:"last"`
+	Min     string `json:"min"`
+	Average string `json:"average"`
+	Max     string `json:"max"`
+	Mdev    string `json:"mdev"`
```
tpantelis, line 93: What is this one? Can we use a more descriptive name?

aswinsuryan: I changed the names of the field, do you think it makes more sense now ? or should I  add comments?

pkg/apis/submariner.io/v1/types.go
aswinsuryan: It seems to be supported in one of the libraries [1] I checked.

[1]https://godoc.org/github.com/sparrc/go-ping

pkg/apis/submariner.io/v1/types.go
tpantelis: "StdDev" looks OK.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -77,10 +79,20 @@ type GatewayStatus struct {
 	Version       string       `json:"version"`
 	HAStatus      HAStatus     `json:"haStatus"`
 	LocalEndpoint EndpointSpec `json:"localEndpoint"`
+	// +optional
+	Latency       LatencySpec  `json:"latency"`
```
sridhargaddam, line 83: Since Latency is for each connection, this should be part of "Connections" struct.

sridhargaddam (changes requested):

pkg/apis/submariner.io/v1/types.go
aswinsuryan: Done

sridhargaddam (dismissed):

pkg/apis/submariner.io/v1/types.go
```diff
@@ -81,6 +83,14 @@ type GatewayStatus struct {
 	Connections   []Connection `json:"connections"`
 }
 
+type LatencySpec struct {
+	LastRtt    string `json:"lastrtt"`
+	MinRtt     string `json:"minrtt"`
+	AverageRtt string `json:"averagertt"`
+	MaxRtt     string `json:"maxrtt"`
+	StdDevRtt  string `json:"stddevrtt"`
```
nyechiel, line 91: "Rtt" should be "RTT" across all of these.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -81,6 +83,14 @@ type GatewayStatus struct {
 	Connections   []Connection `json:"connections"`
 }
 
+type LatencySpec struct {
+	LastRTT    string `json:"lastRTT"`
+	MinRTT     string `json:"minRTT"`
+	AverageRTT string `json:"averageRTT"`
+	MaxRTT     string `json:"maxRTT"`
+	StdDevRTT  string `json:"stddevRTT"`
```
tpantelis, line 91: What format/units will the values be in? Should they be a numeric type? I think we should add some documentation, at least at the struct level.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -100,6 +110,8 @@ type Connection struct {
 	Status        ConnectionStatus `json:"status"`
 	StatusMessage string           `json:"statusMessage"`
 	Endpoint      EndpointSpec     `json:"endpoint"`
+	// +optional
+	Latency LatencySpec `json:"latency"`
```
tpantelis, line 114: If it's optional I think it should be `*LatencySpec `.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -100,6 +110,8 @@ type Connection struct {
 	Status        ConnectionStatus `json:"status"`
 	StatusMessage string           `json:"statusMessage"`
 	Endpoint      EndpointSpec     `json:"endpoint"`
+	// +optional
+	LatencySpec LatencySpec `json:"latencySpec"`
```
tpantelis, line 114: I think the name `Latency` is fine but the type should be `* LatencySpec` since it's optional.

```suggestion
	Latency *LatencySpec `json:"latency,omitempty"`
```

pkg/apis/submariner.io/v1/types.go
tpantelis: @aswinsuryan ^^^

aswinsuryan: I changed it to time.Duration so the unit is now a standard one. Added documentation st the struct level.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -81,6 +84,17 @@ type GatewayStatus struct {
 	Connections   []Connection `json:"connections"`
 }
 
+// LatencySpec describes the round trip time for a packet
+// between the gateway pod of different clusters this cluster
+// is connected.
```
tpantelis, line 89: I assume the units are in nanosec.

```suggestion
// LatencySpec describes the round trip time information in nanoseconds for a packet
// between the gateway pods of two clusters.
```

aswinsuryan: yes it is in nano-seconds

tpantelis (dismissed):

pkg/apis/submariner.io/v1/types.go
```diff
@@ -84,14 +82,14 @@ type GatewayStatus struct {
...
-// LatencySpec describes the round trip time information in nanoseconds for a packet
+// LatencySpec describes the round trip time information in miliseconds for a packet
 // between the gateway pods of two clusters.
 type LatencySpec struct {
-	LastRTT    time.Duration `json:"lastRTT"`
-	MinRTT     time.Duration `json:"minRTT"`
-	AverageRTT time.Duration `json:"averageRTT"`
-	MaxRTT     time.Duration `json:"maxRTT"`
-	StdDevRTT  time.Duration `json:"stddevRTT"`
+	LastRTT    float64 `json:"lastRTT"`
```
tpantelis, line 89: What's the reason for changing these to float and ms units?

tpantelis:
The purpose of this PR is to add fields and it has had several review cycles already. I think the last commit to add implementation n should be in another PR.

pkg/apis/submariner.io/v1/types.go
aswinsuryan: The value in nano seconds seems to be not understandable, like

 Latency:
      Average RTT:   585160
      Last RTT:      403876
      Max RTT:       839461
      Min RTT:       403876
      Stddev RTT:    185170

So to make it similar to normal ping response changed it to float and ms unit

  Latency:
      Average RTT:   0.724
      Last RTT:      0.635
      Max RTT:       0.857
      Min RTT:       0.635
      Stddev RTT:    0.095

aswinsuryan:
> The purpose of this PR is to add fields and it has had several review cycles already. I think the last commit to add implementation n should be in another PR.

SInce I changed the API , I thought it can go together . I will revert to the original one.

pkg/apis/submariner.io/v1/types.go
tpantelis: OK - but it was already approved in the previous form so now need re-review.

tpantelis (approved):

sridhargaddam (approved):

PR #904, opened 2020-10-27
Libreswan 4

skitt:


mangelajo:
Thank you @skitt

mangelajo (approved):

pkg/cable/libreswan/libreswan.go
```diff
@@ -347,8 +349,6 @@ func (i *libreswan) runPluto() error {
 	klog.Info("Starting Pluto")
 
 	args := []string{}
-	args = append(args, "--ikeport", i.ipSecIKEPort)
-	args = append(args, "--natikeport", i.ipSecNATTPort)
```
sridhargaddam, line 351: @skitt We are removing these params from Pluto and now configuring it as part of whack.
However, for whack, we are only configuring "--ikeport". What about "--natikeport"?

skitt: `whack` only has an `--ikeport` parameter. Pluto needed two, because the port was specified for different connection types; the daemon would then choose the appropriate port for each connection based on its type. `whack --ikeport` specifies the port for a specific connection, so there’s no longer a need to distinguish NAT and non-NAT ports.

However I suspect we’re missing something in the non-NAT case, I’ll add a comment on Miguel’s PR.

dfarrell07 (approved):

pkg/cable/libreswan/libreswan.go
sridhargaddam: Okay thanks for clarification @skitt

sridhargaddam (approved):

skitt:
This needs https://github.com/submariner-io/submariner/pull/747 to be fixed.

mangelajo:
> This needs #747 to be fixed.

What do you mean?, Isn't it ok making it part of this PR or the switching commit?

skitt:
> > This needs #747 to be fixed.
> 
> What do you mean?, Isn't it ok making it part of this PR or the switching commit?

I mean this: https://github.com/submariner-io/submariner/pull/747#pullrequestreview-518490763

PR #906, opened 2020-10-27
Fix intermittent E2E GW redundancy failures

tpantelis:
Intermittent failures have been seen during GW cleanup for the engine pod restart test. After deleting the engine pod, we intend to wait for a new pod to spin up but sometimes the previous one hasn't gone away yet so the check prematurely succeeds. On cleanup, we may happen to observe the new pod in the temporary passive state and unintentionally un-label it
thus causing subsequent failures.


sridhargaddam (approved):
LGTM, thanks @tpantelis

skitt (approved):

mangelajo (commented):
Thank you!

PR #908, opened 2020-10-28
Event controller

mangelajo:
The event controller subscribes to the objects it needs to watch
and sends the events to an event handler registry.

This includes a network-plugin-syncer stub as reference.

Resolves-Issue: #857
    

tpantelis (approved):

pkg/event/controller/controller_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+			Eventually(testEvents).Should(Receive(Equal(
+				testing.TestEvent{Handler: testHandlerName, Name: testing.EvNodeCreated, Parameter: node})))
+			Consistently(testEvents).ShouldNot(Receive())
+
+			node.Labels = map[string]string{"labeled-i-am": "i-am"}
+
+			test.UpdateResource(nodes, node)
+
+			Eventually(testEvents).Should(Receive(Equal(
```
sridhargaddam, line 92: @mangelajo cool. So you added a test-case to verify that we are not getting events when labels are updated. Nice.

sridhargaddam (approved):
LGTM

pkg/event/controller/controller_test.go
mangelajo: I actually removed that filter for now. I will be needing labels and annotations for OVN.

We can later add some interface from the controller consumers to let us install specific filters. I hope that sounds ok @sridhargaddam

sridhargaddam: Okay @mangelajo

PR #912, opened 2020-10-29
Modifications to event registry tests

tpantelis:
See commits for details.

### DIFF
--- a/pkg/event/registry_test.go
+++ b/pkg/event/registry_test.go
@@ -54,26 +54,8 @@ var _ = Describe("Event Registry", func() {
 			}
 		})
 
-		It("should invoke all matching handlers of all events", func() {
-			endpoint := &submV1.Endpoint{ObjectMeta: v1meta.ObjectMeta{Name: "endpoint1"}}
-			node := &k8sV1.Node{ObjectMeta: v1meta.ObjectMeta{Name: "node1"}}
-
-			events := map[TestEvent]func() error{
-				{Name: evStop, Parameter: false}:                     func() error { return registry.StopHandlers(false) },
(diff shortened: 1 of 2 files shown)


pkg/event/registry_test.go
```diff
@@ -17,95 +17,91 @@ const npOVNKubernetes = "OVNKubernetes"
...
-				Expect(handler1.Events).Should(Receive(Equal(ev)))
-				Expect(handler2.Events).Should(Receive(Equal(ev)))
-
-				// Handlers should be handled invoked in order of registration
-				ev.Handler = handler1.Name
-				Expect(allTestEvents).Should(Receive(Equal(ev)))
-				ev.Handler = handler2.Name
-				Expect(allTestEvents).Should(Receive(Equal(ev)))
+
+				for _, h := range matchingHandlers {
```
mangelajo, line 62: In this unit testing it's not implicit that we expect that the handlers should be called in order. I'd add a comment at least.

Imagining that someone rewrites the registry (somehow.. with a map...) .. and then the order is broken, so they change this test here.

If we say explicitly somewhere here that the events of the handlers should be received in order ... it'd safer I think

mangelajo (changes requested):
Only a small comment, not sure how to do it best, just comments in the code, or extension of the "It" item.

pkg/event/registry_test.go
tpantelis: yeah - that is already described in the registry `AddHandlers` API - "_Handlers will be called in registration order"_. But I add it to the _It_ text as well.

mangelajo (approved):
Thanks !! :-)

sridhargaddam (approved):

PR #916, opened 2020-11-02
Add CI for CODEOWNERS generation

dfarrell07:
Validate that changes to CODEOWNERS/CODEOWNERS.in are the same generated
by the gen-codeowners script.

### DIFF
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -19,4 +19,4 @@ jobs:
       - name: Run gen-codeowners to rebuild CODEOWNERS file
         run: make CODEOWNERS
       - name: Validate new CODEOWNERS file is the same as tracked by Git
-        run: git diff --exit-code
+        run: git diff --exit-code CODEOWNERS
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -1,0 +1,22 @@
+---
+name: CODEOWNERS
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

.github/workflows/codeowners.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v2
+      - name: Delete current CODEOWNERS file
+        run: rm CODEOWNERS
+      - name: Run gen-codeowners to rebuild CODEOWNERS file
+        run: make CODEOWNERS
+      - name: Validate new CODEOWNERS file is the same as tracked by Git
+        run: git diff --exit-code
```
skitt, line 22:
```suggestion
        run: git diff --exit-code CODEOWNERS
```

to avoid having this trip over other changes?

skitt (changes requested):

.github/workflows/codeowners.yml
dfarrell07: This would have to mean that `make CODEOWNERS` causes some other file to be changed, right? I'm not sure how that would happen, but still good to narrow the check.

dfarrell07:
There is an example of this triggering on `CODEOWNERS.in` changes and passing in https://github.com/submariner-io/submariner/pull/917.

skitt (approved):

Issue #903, opened 2020-10-27
Globalnet pod is terminated after deployment


manosnoam:
It happans once in a while after Submariner 0.7 deployment, where globalnet pod is terminated.
It is later restarted, but we need to understand what is causing this termination.

See:
https://qe-jenkins-csb-skynet.cloud.paas.psi.redhat.com/job/Submariner-OSP-AWS/855/Test-Report/
(pods log collection at the end)

```

NAME                                               READY   STATUS    RESTARTS   AGE   LABELS
pod/submariner-gateway-2ll7r                       1/1     Running   0          14s   app=submariner-engine,controller-revision-hash=7f4bf8884b,pod-template-generation=1
pod/submariner-globalnet-s6pfs                     0/1     Error     1          13s   app=submariner-globalnet,component=globalnet,controller-revision-hash=6964446597,pod-template-generation=1
pod/submariner-lighthouse-agent-7d6db4bb9-xkj7k    1/1     Running   0          13s   app=submariner-lighthouse-agent,component=submariner-lighthouse,pod-template-hash=7d6db4bb9
pod/submariner-lighthouse-coredns-c88f64f5-qqzmw   1/1     Running   0          12s   app=submariner-lighthouse-coredns,component=submariner-lighthouse,pod-template-hash=c88f64f5
pod/submariner-lighthouse-coredns-c88f64f5-z2sh8   1/1     Running   0          11s   app=submariner-lighthouse-coredns,component=submariner-lighthouse,pod-template-hash=c88f64f5
pod/submariner-operator-dcbdf5669-fqt4h            1/1     Running   0          38s   name=submariner-operator,pod-template-hash=dcbdf5669
pod/submariner-routeagent-bz8qx                    1/1     Running   0          14s   app=submariner-routeagent,component=routeagent,controller-revision-hash=677d5775dc,pod-template-generation=1
pod/submariner-routeagent-c6jfb                    1/1     Running   0          14s   app=submariner-routeagent,component=routeagent,controller-revision-hash=677d5775dc,pod-template-generation=1
pod/submariner-routeagent-ckkcw                    1/1     Running   0          14s   app=submariner-routeagent,component=routeagent,controller-revision-hash=677d5775dc,pod-template-generation=1
pod/submariner-routeagent-cz8c4                    1/1     Running   0          14s   app=submariner-routeagent,component=routeagent,controller-revision-hash=677d5775dc,pod-template-generation=1
[...]

PR #918, opened 2020-11-03
Wait for ClusterCRD in Globalnet during initialization

sridhargaddam:
When Submariner is installed in a Cluster, the local ClusterCRD is
created by the submariner-gateway pod. While the submariner-gateway
pod is in the process of initializing itself if Globalnet pod tries
to read the ClusterCRD it will be missing. In this PR, we modify
Globalnet code to retry for 5 mins before giving up.

Fixes issue: https://github.com/submariner-io/submariner/issues/903


mkolesnik (approved):
Technically it's waiting at least 5 minutes :)

nyechiel (approved):

mangelajo (approved):

PR #919, opened 2020-11-03
Move our green testing badges to the top

mangelajo:
I think it's a strong point of this project that we are making a good work with CI
and it's something we can share with the world.


mangelajo:
We need the top level heading first ... it seems @dfarrell07 may be we should consider disabling this check ?(wanted to put the badges up up... ) ;)

dfarrell07:
I agree with the check, I think the file should start with an H1. We could put the badges after the heading and before the ToC, but I'm also okay with the current layout.

mangelajo:
> I agree with the check, I think the file should start with an H1. We could put the badges after the heading and before the ToC, but I'm also okay with the current layout.

I disagree, under the H1 the result is weird from a composition point of view. That bit of the linter is a little bit artificial.

Most github projects that I've seen start by showing the badges. 

Anyway, I can agree that this is a completely subjective matter ;)

dfarrell07:
> Most github projects that I've seen start by showing the badges.

The main k8s/k8s repo starts with H1, then badges.

https://github.com/kubernetes/kubernetes#kubernetes

With HTML, I think the title element should be the first element of the document. This is the parallel in Markdown.

mangelajo:
> > Most github projects that I've seen start by showing the badges.
> 
> The main k8s/k8s repo starts with H1, then badges.
> 
> https://github.com/kubernetes/kubernetes#kubernetes
> 
> With HTML, I think the title element should be the first element of the document. This is the parallel in Markdown.

In HTML that's not mandatory. It's just preference. You can even go without H1 at all and HTML will be equally valid.

dfarrell07:
> > > Most github projects that I've seen start by showing the badges.
> > 
> > 
> > The main k8s/k8s repo starts with H1, then badges.
> > https://github.com/kubernetes/kubernetes#kubernetes
> > With HTML, I think the title element should be the first element of the document. This is the parallel in Markdown.
> 
> In HTML that's not mandatory. It's just preference. You can even go without H1 at all and HTML will be equally valid.

Yeah, of course it's not required, but it's definitely the norm for <h1> to be the first tag in <body>.

https://www.w3schools.com/tags/tag_body.asp

I didn't mean to go down a rabbit hole about HTML vs Markdown though. The key point is that it's not a norm in the K8s ecosystem to put the badges before the H1, it looks fine, and I don't think this aesthetic preference is a compelling reason to not conform with linter defaults.

mangelajo:
> > > > Most github projects that I've seen start by showing the badges.
> > > 
> > > 
> > > The main k8s/k8s repo starts with H1, then badges.
> > > https://github.com/kubernetes/kubernetes#kubernetes
> > > With HTML, I think the title element should be the first element of the document. This is the parallel in Markdown.
> > 
> > 
> > In HTML that's not mandatory. It's just preference. You can even go without H1 at all and HTML will be equally valid.
> 
> Yeah, of course it's not required, but it's definitely the norm for `<h1>` to be the first tag in `<body>`.
> 
> https://www.w3schools.com/tags/tag_body.asp
> 
> I didn't mean to go down a rabbit hole about HTML vs Markdown though. The key point is that it's not a norm in the K8s ecosystem to put the badges before the H1, it looks fine, and I don't think this aesthetic preference is a compelling reason to not conform with linter defaults.

ok, so lets put it at least under H1, let me update the patch

mangelajo:
@dfarrell07 moving it up may be it's something we can do as part of the periodic check badges you were preparing and not this patch.

mangelajo:
https://github.com/submariner-io/submariner/blob/c1f4198748c89d96f254474f145159c9c0a9f6a1/README.md

nyechiel (approved):

dfarrell07 (approved):

tpantelis (approved):

PR #928, opened 2020-11-04
Rename Periodic E2E->Flake Finder, refactor README badges, update automated issue reporting

dfarrell07:


dfarrell07:
This will fail until the PR it's based on is merged.

skitt (approved):

dfarrell07:
This is now a fairly different PR, re-requesting review @skitt.

The MD link linter will fail until this is merged and there is a "Flake Finder" GHA Workflow.

.github/workflows/flake_finder.yml
```diff
@@ -49,3 +49,11 @@ jobs:
           df -h
           free -h
           make post-mortem
+
+      - name: Raise an issue to report flaky test failure
```
mkolesnik, line 53: I don't know how useful it is in a job that's quite affected by 3rd party factors (image repositories and such).

I fear a flood of automated issues which would just create noise or be ignored since they're automated..

dfarrell07: Would this be impacted by 3rd party factors we want to ignore often? It doesn't seem like running once a week, and so far always passing, would create much noise.

Although cranking the frequency up to debug a flaky failure, like we're doing in the operator repo currently, definitely does create a lot of noise. I'm okay with removing this step or keeping it, but the feedback previously was to add it.

tpantelis: It generates emails so folks are already getting notified that way.  I don't think we need to go a step further and create issues.

dfarrell07: Okay, so I'll remove it elsewhere too.

mkolesnik (approved):

tpantelis:
This has a conflict.

dfarrell07:
This needs https://github.com/submariner-io/submariner/pull/1010 to get the link tests passing.

nyechiel (approved):

nyechiel:
@dfarrell07 there are some Markdown link failures here now

dfarrell07:
> @dfarrell07 there are some Markdown link failures here now

@nyechiel They still look like failures that will be fixed by https://github.com/submariner-io/submariner/pull/1010 as noted above.

PR #935, opened 2020-11-10
Fix intermittent event controller test failures

tpantelis:
A race condition with the `registry` field:

```
Write at 0x00c0003a3fa0 by goroutine 9:
  github.com/submariner-io/submariner/pkg/event/controller_test...
    /go/src/github.com/submariner-io/submariner/pkg/event/
       controller/controller_test.go:43


Previous read at 0x00c0003a3fa0 by goroutine 21:
  github.com/submariner-io/submariner/pkg/event.(*Registry).invokeHandlers()
      /go/src/github.com/submariner-io/submariner/pkg/event/registry.go:141
  github.com/submariner-io/submariner/pkg/event.(*Registry).StopHandlers()
      /go/src/github.com/submariner-io/submariner/pkg/event/registry.go:67
```

This is due to the test starting the controller in a Go routine but not waiting for it to complete. There's another intermittent failure related to fake client sets/informers that is also due to the async Start.

To fix both I changed `Start` to not wait on the stopCh. This also aligns with the semantics of Start vs Run.

### DIFF
--- a/pkg/event/controller/controller.go
+++ b/pkg/event/controller/controller.go
@@ -119,23 +119,25 @@ func New(config *Config) (*Controller, error) {
 	return &ctl, nil
 }
 
-// Start will start the controller, and wait until stopCh receives a message
+// Start starts the controller.
 func (c *Controller) Start(stopCh <-chan struct{}) error {
-	klog.Info("Starting Controller")
+	klog.Info("Starting the Event controller...")
 
 	err := c.resourceWatcher.Start(stopCh)
 	if err != nil {
 		return err
 	}
 
-	klog.Info("Event Controller workers started")
-	<-stopCh
-	klog.Info("Event Controller stopping")
+	klog.Info("Event controller started")
+
+	return nil
+}
+
+func (c *Controller) Stop() {
+	klog.Info("Event controller stopping")
 
 	// TODO: Detect if it's an uninstall and invoke StopHandlers with uninstall=true if it's the case
 	if err := c.handlers.StopHandlers(false); err != nil {
 		klog.Warningf("In Event Controller, StopHandlers returned error: %v", err)
 	}
-
-	return nil
 }
--- a/pkg/networkplugin-syncer/main.go
+++ b/pkg/networkplugin-syncer/main.go
@@ -45,6 +45,9 @@ func main() {
 		klog.Fatalf("Error starting controller: %v", err)
 	}
 
+	<-stopCh
+	ctl.Stop()
+
 	klog.Info("All controllers stopped or exited. Stopping submariner-networkplugin-syncer")
 }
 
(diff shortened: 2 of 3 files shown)


dfarrell07 (approved):
Yay for fixing flaky failures! Thanks @tpantelis!

pkg/networkplugin-syncer/main.go
```diff
@@ -45,6 +45,7 @@ func main() {
 		klog.Fatalf("Error starting controller: %v", err)
 	}
 
+	<-stopCh
```
sridhargaddam, line 48: Should we call ctl.Stop() here?

sridhargaddam (dismissed):

pkg/networkplugin-syncer/main.go
mangelajo: I think it'd be better if we leave go-routine from Start waiting on the stop channel, and then calling Stop, that way it's transparent and works with the stopCh right away. (and saves one line/possible mistake on usages).

mangelajo: But good point

pkg/event/controller/controller.go
```diff
@@ -119,23 +119,25 @@ func New(config *Config) (*Controller, error) {
...
 	err := c.resourceWatcher.Start(stopCh)
 	if err != nil {
 		return err
 	}
 
-	klog.Info("Event Controller workers started")
-	<-stopCh
-	klog.Info("Event Controller stopping")
+	klog.Info("Event controller started")
+
```
mangelajo, line 132:
```suggestion
	klog.Info("Event controller started")
        go func() {
            <- stopCh
            c.Stop()
        }
```

mangelajo:
Thanks for looking / catching this @tpantelis

pkg/networkplugin-syncer/main.go
tpantelis: Having Start wait on the stopCh is problematic for UTs. I've run into this in other tests.

tpantelis: I forgot to call Stop here 😄

mangelajo: Yeah I was in doubt about if it could be problematic. Then we need to call it on the main.go before stop. that would be fine too as @sridhargaddam proposed.

pkg/event/controller/controller.go
tpantelis: This is still problematic for the UTs, ie frequent race failures. The tests need to wait for `Stop` to complete. The simplest solution is to explicitly call `Stop` otherwise we need a method and flag or chan to check for stopped or wait for it. We can do the latter but seems like overkill. Also, in the production  code, calling `Stop` explicitly makes it deterministic that all handlers will be stopped.

mangelajo (approved):

sridhargaddam (approved):
LGTM, thanks @tpantelis

PR #938, opened 2020-11-11
Add health checker implementation

aswinsuryan:
*Populate the healthchecker Ip
*Ping the healthchecker IP to check the remote
gateway status
*Update the gateway satus if the ping fails

Fixes: submariner-io/submariner#821

### DIFF
--- a/main.go
+++ b/main.go
@@ -110,13 +110,12 @@ func main() {
 	submSpec.CableDriver = strings.ToLower(submSpec.CableDriver)
 
 	localEndpoint, err := util.GetLocalEndpoint(submSpec.ClusterID, submSpec.CableDriver, nil, submSpec.NatEnabled,
-		localSubnets, util.GetLocalIP())
+		localSubnets, util.GetLocalIP(), submSpec.ClusterCidr)
 
 	if err != nil {
 		klog.Fatalf("Error creating local endpoint object from %#v: %v", submSpec, err)
 	}
 
-	klog.Errorf("The local endpoint is %v", localEndpoint)
 	cableEngine := cableengine.NewEngine(localCluster, localEndpoint)
 
 	err = subv1.AddToScheme(scheme.Scheme)
--- a/main.go
+++ b/main.go
@@ -116,27 +116,34 @@ func main() {
 		klog.Fatalf("Error creating local endpoint object from %#v: %v", submSpec, err)
 	}
 
-	latencyMap := new(sync.Map)
+	klog.Errorf("The local endpoint is %v", localEndpoint)
 	cableEngine := cableengine.NewEngine(localCluster, localEndpoint)
 
-	if len(submSpec.GlobalCidr) == 0 {
-		cableHealthchecker := healthchecker.NewHealthChecker(localEndpoint.Spec.Hostname,
-			submarinerClient.SubmarinerV1().Gateways(submSpec.Namespace), latencyMap)
+	err = subv1.AddToScheme(scheme.Scheme)
+	if err != nil {
+		klog.Errorf("Error adding submariner types to the scheme: %v", err)
+	}
 
-		cableHealthchecker.Run(stopCh)
+	var cableHealthchecker *healthchecker.HealthChecker
+	if len(submSpec.GlobalCidr) == 0 {
+		cableHealthchecker, err = healthchecker.New(&watcher.Config{RestConfig: cfg}, submSpec.Namespace, submSpec.ClusterID)
+		if err != nil {
+			klog.Errorf("Error creating healthChecker: %v", err)
+		}
+
+		err = cableHealthchecker.Start(stopCh)
+
+		if err != nil {
+			klog.Errorf("Error starting healthChecker: %v", err)
+		}
 	}
 
 	cableEngineSyncer := syncer.NewGatewaySyncer(
 		cableEngine,
 		submarinerClient.SubmarinerV1().Gateways(submSpec.Namespace),
-		VERSION, latencyMap)
+		VERSION, cableHealthchecker)
 
 	cableEngineSyncer.Run(stopCh)
-
-	err = subv1.AddToScheme(scheme.Scheme)
-	if err != nil {
-		fatal(cableEngineSyncer, "Error adding submariner types to the scheme: %v", err)
-	}
 
 	becameLeader := func(context.Context) {
 		klog.Info("Creating the datastore syncer")
--- a/main.go
+++ b/main.go
@@ -123,7 +123,7 @@ func main() {
 		klog.Errorf("Error adding submariner types to the scheme: %v", err)
 	}
 
-	var cableHealthchecker *healthchecker.HealthChecker
+	var cableHealthchecker healthchecker.Interface
 	if len(submSpec.GlobalCidr) == 0 {
 		cableHealthchecker, err = healthchecker.New(&watcher.Config{RestConfig: cfg}, submSpec.Namespace, submSpec.ClusterID)
 		if err != nil {
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ import (
 	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cable"
 	"github.com/submariner-io/submariner/pkg/cableengine"
+	"github.com/submariner-io/submariner/pkg/cableengine/healthchecker"
 	"github.com/submariner-io/submariner/pkg/cableengine/syncer"
 	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	"github.com/submariner-io/submariner/pkg/controllers/datastoresyncer"
@@ -115,12 +116,20 @@ func main() {
 		klog.Fatalf("Error creating local endpoint object from %#v: %v", submSpec, err)
 	}
 
+	latencyMap := new(sync.Map)
 	cableEngine := cableengine.NewEngine(localCluster, localEndpoint)
+
+	if len(submSpec.GlobalCidr) == 0 {
+		cableHealthchecker := healthchecker.NewHealthChecker(localEndpoint.Spec.Hostname,
+			submarinerClient.SubmarinerV1().Gateways(submSpec.Namespace), latencyMap)
+
+		cableHealthchecker.Run(stopCh)
+	}
 
 	cableEngineSyncer := syncer.NewGatewaySyncer(
 		cableEngine,
 		submarinerClient.SubmarinerV1().Gateways(submSpec.Namespace),
-		VERSION)
+		VERSION, latencyMap)
 
 	cableEngineSyncer.Run(stopCh)
 
--- a/pkg/cableengine/healthchecker/healthchecker.go
+++ b/pkg/cableengine/healthchecker/healthchecker.go
@@ -2,154 +2,136 @@ package healthchecker
 
 import (
 	"sync"
-	"time"
 
-	"github.com/go-ping/ping"
-	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
-	v1typed "github.com/submariner-io/submariner/pkg/client/clientset/versioned/typed/submariner.io/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
+	"github.com/submariner-io/admiral/pkg/log"
+	"github.com/submariner-io/admiral/pkg/watcher"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog"
 )
 
-var HealthCheckInterval = 15 * time.Second
+type LatencyInfo struct {
+	ConnectionError string
+	Spec            *submarinerv1.LatencySpec
+}
+
+type Interface interface {
+	Start(stopCh <-chan struct{})
+
+	GetLatencyInfo(hostName string) *LatencyInfo
+}
 
 type HealthChecker struct {
-	localEndpointName string
-	client            v1typed.GatewayInterface
-	latencyMap        *sync.Map
+	endpointWatcher watcher.Interface
+	healthCheckers  sync.Map
+	clusterID       string
 }
 
-type LatencyInfo struct {
-	Status  v1.ConnectionStatus
-	Latency *v1.LatencySpec
-}
-
-// NewEngine creates a new Engine for the local cluster
-func NewHealthChecker(localEndpointName string, client v1typed.GatewayInterface,
-	latencyMap *sync.Map) *HealthChecker {
-	return &HealthChecker{
-		localEndpointName: localEndpointName,
-		client:            client,
-		latencyMap:        latencyMap,
-	}
-}
-
-func (h *HealthChecker) Run(stopCh <-chan struct{}) {
-	go func() {
-		wait.Until(h.checkRemoteGatewayHealth, HealthCheckInterval, stopCh)
-	}()
-
-	klog.Infof("CableEngine healthchecker started")
-}
-func (h *HealthChecker) checkRemoteGatewayHealth() {
-	existingGw, err := h.client.Get(h.localEndpointName, metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		klog.V(2).Infof("Gateway object not found: %v", err)
-		return
-	} else if err != nil {
-		klog.Errorf("Error retrieving the gateway object: %v", err)
-		return
+func New(config *watcher.Config, endpointNameSpace, clusterID string) (*HealthChecker, error) {
+	serviceExportController := HealthChecker{
+		clusterID: clusterID,
 	}
 
-	if len(existingGw.Status.Connections) == 0 {
-		klog.V(2).Info("No endpoints present in the gateway")
-		return
+	config.ResourceConfigs = []watcher.ResourceConfig{
+		{
+			Name:         "HelathChecker Endoint Controller",
+			ResourceType: &submarinerv1.Endpoint{},
+			Handler: watcher.EventHandlerFuncs{
+				OnCreateFunc: serviceExportController.endpointCreatedorUpdated,
+				OnUpdateFunc: serviceExportController.endpointCreatedorUpdated,
+				OnDeleteFunc: serviceExportController.endpointDeleted,
+			},
+			SourceNamespace: endpointNameSpace,
+		},
 	}
 
-	deleteMap := copyLatencyMap(h.latencyMap)
-	var wg sync.WaitGroup
-
-	for _, endPointInfo := range existingGw.Status.Connections {
-		wg.Add(1)
-
-		hostName := endPointInfo.Endpoint.Hostname
-		healthCheckIp := endPointInfo.Endpoint.HealthCheckIP
-
-		// Remove the entry that is found in the gateway connection status
-		delete(deleteMap, hostName)
-
-		go func() {
-			latencyInfo := h.sendPing(&wg, hostName, healthCheckIp)
-			if latencyInfo != nil {
-				h.latencyMap.Store(hostName, latencyInfo)
-			}
-		}()
-	}
-
-	wg.Wait()
-
-	for k := range deleteMap {
-		// Remove the entry that are missing in the gateway connection status from the
-		// latencymap
-		h.latencyMap.Delete(k)
-	}
-}
-
-func copyLatencyMap(latencyMap *sync.Map) map[string]*LatencyInfo {
-	newMap := make(map[string]*LatencyInfo)
-
-	latencyMap.Range(func(key, val interface{}) bool {
-		if val != nil {
-			newMap[key.(string)] = val.(*LatencyInfo)
-		}
-		return true
-	})
-
-	return newMap
-}
-
-func (h *HealthChecker) sendPing(wg *sync.WaitGroup, remoteHostName, healthCheckIp string) *LatencyInfo {
-	defer wg.Done()
-	var lastRtt time.Duration
-
-	pinger, err := ping.NewPinger(healthCheckIp)
+	endpointWatcher, err := watcher.New(config)
 	if err != nil {
-		klog.Errorf("Creating pinger for hostname %v to the ip %v failed due to %v", remoteHostName, healthCheckIp, err)
-		return nil
+		return nil, err
 	}
 
-	pinger.Count = 3
-	pinger.SetPrivileged(true)
-	pinger.Timeout = 3 * time.Second
+	serviceExportController.endpointWatcher = endpointWatcher
 
-	pinger.OnRecv = func(packet *ping.Packet) {
-		lastRtt = packet.Rtt
-	}
-	var latencyinfo *LatencyInfo
+	return &serviceExportController, nil
+}
 
-	pinger.OnFinish = func(stats *ping.Statistics) {
-		if stats.PacketLoss == 100 {
-			latencySpec := &v1.LatencySpec{
-				LastRTT:    0,
-				MinRTT:     0,
-				AverageRTT: 0,
-				MaxRTT:     0,
-				StdDevRTT:  0,
-			}
-			latencyinfo = &LatencyInfo{
-				Status:  v1.ConnectionError,
-				Latency: latencySpec,
-			}
-		} else {
-			latencySpec := &v1.LatencySpec{
-				LastRTT:    float64(lastRtt.Microseconds()) / 1000,
-				MinRTT:     float64(stats.MinRtt.Microseconds()) / 1000,
-				AverageRTT: float64(stats.AvgRtt.Microseconds()) / 1000,
-				MaxRTT:     float64(stats.MaxRtt.Microseconds()) / 1000,
-				StdDevRTT:  float64(stats.StdDevRtt.Microseconds()) / 1000,
-			}
-			latencyinfo = &LatencyInfo{
-				Status:  v1.Connected,
-				Latency: latencySpec,
-			}
+func (h *HealthChecker) GetLatencyInfo(hostName string) *LatencyInfo {
+	if obj, found := h.healthCheckers.Load(hostName); found {
+		pinger := obj.(*Pinger)
+		statistics := pinger.statistics
+
+		return &LatencyInfo{
+			ConnectionError: pinger.Status,
+			Spec: &submarinerv1.LatencySpec{
+				LastRTT:    statistics.lastRtt,
+				MinRTT:     statistics.minRtt,
+				AverageRTT: statistics.mean,
+				MaxRTT:     statistics.maxRtt,
+				StdDevRTT:  statistics.stdDev,
+			},
 		}
 	}
-	err = pinger.Run()
-	if err != nil {
-		klog.Errorf("Ping to %q with ip %q failed due to  %v", remoteHostName, healthCheckIp, err)
+
+	return nil
+}
+
+func (h *HealthChecker) Start(stopCh <-chan struct{}) error {
+	if err := h.endpointWatcher.Start(stopCh); err != nil {
+		return err
 	}
 
-	return latencyinfo
+	return nil
+}
+
+func (h *HealthChecker) endpointCreatedorUpdated(obj runtime.Object) bool {
+	klog.V(log.TRACE).Infof("Endpoint created: %v", obj)
+	endpointCreated := obj.(*submarinerv1.Endpoint)
+	if endpointCreated.Spec.ClusterID == h.clusterID {
+		klog.Infof("Skipping endpoint creation for the local cluster %q", h.clusterID)
+		return false
+	}
+
+	if endpointCreated.Spec.HealthCheckIP == "" || endpointCreated.Spec.Hostname == "" {
+		klog.Infof("HealthCheckIP or Hostname is Nil, hence returning %q: %q", endpointCreated.Spec.HealthCheckIP,
+			endpointCreated.Spec.Hostname)
+		return false
+	}
+
+	if obj, found := h.healthCheckers.Load(endpointCreated.Spec.Hostname); found {
+		klog.V(log.TRACE).Infof("HealthChecker is already running %q: %q, and hence stopping", endpointCreated.Spec.HealthCheckIP,
+			endpointCreated.Spec.Hostname)
+
+		pinger := obj.(*Pinger)
+		if pinger.healthCheckIp == endpointCreated.Spec.HealthCheckIP {
+			return false
+		}
+
+		close(pinger.stopCh)
+		pinger.stop()
+		h.healthCheckers.Delete(endpointCreated.Spec.Hostname)
+	}
+
+	klog.V(log.TRACE).Infof("Starting Pinger for Hostname: %q, with HealthCheckIP: %q",
+		endpointCreated.Spec.HealthCheckIP, endpointCreated.Spec.Hostname)
+
+	pinger := NewPinger(endpointCreated.Spec.Hostname, endpointCreated.Spec.HealthCheckIP)
+	h.healthCheckers.Store(endpointCreated.Spec.Hostname, pinger)
+	pinger.Run()
+
+	return false
+}
+
+func (h *HealthChecker) endpointDeleted(obj runtime.Object) bool {
+	endpointDeleted := obj.(*submarinerv1.Endpoint)
+	if endpointDeleted.Spec.Hostname == "" {
+		return false
+	}
+
+	if obj, found := h.healthCheckers.Load(endpointDeleted.Spec.Hostname); found {
+		pinger := obj.(*Pinger)
+		pinger.stop()
+		h.healthCheckers.Delete(endpointDeleted.Spec.Hostname)
+	}
+
+	return false
 }
--- a/pkg/cableengine/healthchecker/healthchecker.go
+++ b/pkg/cableengine/healthchecker/healthchecker.go
@@ -1,0 +1,155 @@
+package healthchecker
+
+import (
+	"sync"
+	"time"
+
+	"github.com/go-ping/ping"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	v1typed "github.com/submariner-io/submariner/pkg/client/clientset/versioned/typed/submariner.io/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+)
+
+var HealthCheckInterval = 15 * time.Second
+
+type HealthChecker struct {
+	localEndpointName string
+	client            v1typed.GatewayInterface
+	latencyMap        *sync.Map
+}
+
+type LatencyInfo struct {
+	Status  v1.ConnectionStatus
+	Latency *v1.LatencySpec
+}
+
+// NewEngine creates a new Engine for the local cluster
+func NewHealthChecker(localEndpointName string, client v1typed.GatewayInterface,
+	latencyMap *sync.Map) *HealthChecker {
+	return &HealthChecker{
+		localEndpointName: localEndpointName,
+		client:            client,
+		latencyMap:        latencyMap,
+	}
+}
+
+func (h *HealthChecker) Run(stopCh <-chan struct{}) {
+	go func() {
+		wait.Until(h.checkRemoteGatewayHealth, HealthCheckInterval, stopCh)
+	}()
+
+	klog.Infof("CableEngine healthchecker started")
+}
+func (h *HealthChecker) checkRemoteGatewayHealth() {
+	existingGw, err := h.client.Get(h.localEndpointName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		klog.V(2).Infof("Gateway object not found: %v", err)
+		return
+	} else if err != nil {
+		klog.Errorf("Error retrieving the gateway object: %v", err)
+		return
+	}
+
+	if len(existingGw.Status.Connections) == 0 {
+		klog.V(2).Info("No endpoints present in the gateway")
+		return
+	}
+
+	deleteMap := copyLatencyMap(h.latencyMap)
+	var wg sync.WaitGroup
+
+	for _, endPointInfo := range existingGw.Status.Connections {
+		wg.Add(1)
+
+		hostName := endPointInfo.Endpoint.Hostname
+		healthCheckIp := endPointInfo.Endpoint.HealthCheckIP
+
+		// Remove the entry that is found in the gateway connection status
+		delete(deleteMap, hostName)
+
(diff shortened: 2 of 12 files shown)


pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+	defer wg.Done()
+	var lastRtt time.Duration
+
+	pinger, err := ping.NewPinger(healthCheckIp)
+	if err != nil {
+		klog.Errorf("Creating pinger for hostname %v to the ip %v failed due to %v", remoteHostName, healthCheckIp, err)
+		return nil
+	}
+
+	pinger.Count = 3
```
mangelajo, line 113: If you use this to send 3 pings periodically, stdev/min/max will be not very meaningful.

Ideally we should keep the pingers running in the background all the time along with each connection, and stop them/destroy them if the connections cease to exist.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+				MaxRTT:     0,
+				StdDevRTT:  0,
+			}
+			latencyinfo = &LatencyInfo{
+				Status:  v1.ConnectionError,
+				Latency: latencySpec,
+			}
+		} else {
+			latencySpec := &v1.LatencySpec{
+				LastRTT:    float64(lastRtt.Microseconds()) / 1000,
```
mangelajo, line 137:
```suggestion
				LastRTT:    float64(lastRtt.Microseconds()) / 1000,
```

I haven't looked but do they have .Milliseconds instead?, then you don't need to divide /1000

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (h *HealthChecker) Run(stopCh <-chan struct{}) {
+	go func() {
+		wait.Until(h.checkRemoteGatewayHealth, HealthCheckInterval, stopCh)
+	}()
+
+	klog.Infof("CableEngine healthchecker started")
+}
+func (h *HealthChecker) checkRemoteGatewayHealth() {
+	existingGw, err := h.client.Get(h.localEndpointName, metav1.GetOptions{})
```
mangelajo, line 47: Instead of reading/writing from the API, I think it should be better to integrate the sync module here:

https://github.com/submariner-io/submariner/blob/master/pkg/cableengine/syncer/syncer.go

With the heatlhchecker, so we only write it once with the existing pinger status, while keeping pingers running in the background all the time.

mangelajo (changes requested):
Some comments, 

1) I believe ping should be running in the background all the time to get meaningful statistics and reliable detection of the connection going down.
2) Better integration to get the connection list from memory, or from the syncer that checks the connection status periodically instead of querying the k8s API for this.

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: The do have milli seconds , but to make the similar to the way the ping returns I just used floating point than an integer. And divided the microseconds.

aswinsuryan:
> Some comments,
> 
> 1. I believe ping should be running in the background all the time to get meaningful statistics and reliable detection of the connection going down.
> 2. Better integration to get the connection list from memory, or from the syncer that checks the connection status periodically instead of querying the k8s API for this.
I will try if this is possible with the library and  refactor the code based on  the findings.

pkg/cableengine/healthchecker/healthchecker.go
mangelajo: > > Some comments,
> > 
> > 1. I believe ping should be running in the background all the time to get meaningful statistics and reliable detection of the connection going down.
> > 2. Better integration to get the connection list from memory, or from the syncer that checks the connection status periodically instead of querying the k8s API for this.
> >    I will try if this is possible with the library and  refactor the code based on  the findings.

You sent no message :) I looked at the go-ping and probably will need some contribution to provide real-time statistics instead of just "at the end".

But I think it's easily doable. May be until then we can calculate the min/max/etc ... all by ourselves from the single packet reports.

mangelajo: > The do have milli seconds , but to make the similar to the way the ping returns I just used floating point than an integer. And divided the microseconds.

Ack, ok, I expected them to expose float64 and not int. Makes sense then.

pkg/util/util.go
```diff
@@ -114,9 +114,55 @@ func GetLocalEndpoint(clusterID, backend string, backendConfig map[string]string
...
+	if err != nil {
+		return types.SubmarinerEndpoint{}, fmt.Errorf("Error getting CNI Interface IP address: %v", err)
+	}
+
 	return endpoint, nil
 }
 
+func getCNIInterfaceIPAddress(clusterCIDRs []string) (string, error) {
+	for _, clusterCIDR := range clusterCIDRs {
+		_, clusterNetwork, err := net.ParseCIDR(clusterCIDR)
```
mangelajo, line 128: Isn't this code shared with route agent somewhere? I remember we already had that this somewhere.

@sridhargaddam do you remember?

pkg/cableengine/syncer/syncer.go
```diff
@@ -201,6 +204,19 @@ func (i *GatewaySyncer) generateGatewayObject() *v1.Gateway {
...
 
 	if connections != nil {
+		for index, connection := range *connections {
+			latencySpec, ok := i.latencyMap.Load(connection.Endpoint.Hostname)
+			if ok {
+				latencyInfo := latencySpec.(*healthchecker.LatencyInfo)
+				connection.Latency = latencyInfo.Latency
+				if connection.Status == v1.Connected && latencyInfo.Status != v1.Connected {
+					connection.Status = latencyInfo.Status
+				}
```
tpantelis, line 214: For the `ConnectionError` status, we should augment it with detail via the `StatusMessage` as we do  elsewhere. I suggest changing the `LatencyInfo` `Status v1.ConnectionStatus` field to `ConnectionError string` and here check:

```suggestion
				if connection.Status == v1.Connected && latencyInfo.ConnectionError != "" {
					connection.Status = vi.ConnectionError
					connection.StatusMessage = latencyInfo.ConnectionError
				}
```

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+
+type HealthChecker struct {
+	localEndpointName string
+	client            v1typed.GatewayInterface
+	latencyMap        *sync.Map
+}
+
+type LatencyInfo struct {
+	Status  v1.ConnectionStatus
+	Latency *v1.LatencySpec
```
tpantelis, line 26:
```suggestion
	Spec *v1.LatencySpec
```

tpantelis (changes requested):

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: Changed to continuous pinger.

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: now changed to unit64 as per the comments received for the API change.

pkg/util/util.go
aswinsuryan: Yes there is a similar code but not exactly the same. 
https://github.com/submariner-io/submariner/blob/master/pkg/routeagent/controllers/route/driver.go#L15

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// NewEngine creates a new Engine for the local cluster
+func NewHealthChecker(localEndpointName string, client v1typed.GatewayInterface,
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+	go func() {
+		wait.Until(h.checkRemoteGatewayHealth, HealthCheckInterval, stopCh)
+	}()
+
+	klog.Infof("CableEngine healthchecker started")
+}
+func (h *HealthChecker) checkRemoteGatewayHealth() {
+	existingGw, err := h.client.Get(h.localEndpointName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		klog.V(2).Infof("Gateway object not found: %v", err)
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+	"k8s.io/klog"
+)
+
+var HealthCheckInterval = 15 * time.Second
+
+type HealthChecker struct {
+	localEndpointName string
+	client            v1typed.GatewayInterface
+	latencyMap        *sync.Map
+}
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/cableengine/syncer/syncer.go
```diff
@@ -201,6 +204,19 @@ func (i *GatewaySyncer) generateGatewayObject() *v1.Gateway {
...
+
+				(*connections)[index] = connection
+			}
+		}
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/cableengine/syncer/syncer.go
aswinsuryan: Done.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+	if errors.IsNotFound(err) {
+		klog.V(2).Infof("Gateway object not found: %v", err)
+		return
+	} else if err != nil {
+		klog.Errorf("Error retrieving the gateway object: %v", err)
+		return
+	}
+
+	if len(existingGw.Status.Connections) == 0 {
+		klog.V(2).Info("No endpoints present in the gateway")
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/go-ping/ping"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	v1typed "github.com/submariner-io/submariner/pkg/client/clientset/versioned/typed/submariner.io/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+)
+
+var HealthCheckInterval = 15 * time.Second
```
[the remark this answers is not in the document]
aswinsuryan: The parameter is now removed.

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func New(config *watcher.Config, endpointNameSpace, clusterID string) (*HealthChecker, error) {
+	serviceExportController := HealthChecker{
+		clusterID: clusterID,
+	}
+
+	config.ResourceConfigs = []watcher.ResourceConfig{
+		{
+			Name:         "HelathChecker Endoint Controller",
```
mangelajo, line 37:
```suggestion
			Name:         "HealthChecker Endoint Controller",
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,138 @@
...
+		healthCheckIp:      healthCheckIP,
+		statistics: Statistics{
+			size:         size,
+			previousRtts: make([]uint64, size),
+		},
+		stopCh: make(chan struct{}),
+	}
+}
+
+func (s *Statistics) updateStatistics(rtt uint64) {
```
mangelajo, line 48: I'd move all "Statistics" to statistics.go

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if (s.index + 1) == s.size {
+		// Resetting since the incremental SD calculated have an error factor due to truncation which
+		// could be significant as count increases.
+		s.index = 1
+		s.previousRtts[0] = s.previousRtts[s.size-2]
+		s.previousRtts[1] = s.previousRtts[s.size-1]
+		s.sum = s.previousRtts[0] + s.previousRtts[1]
+		s.mean = s.sum / 2
```
mangelajo, line 84: This will make the mean limited just to a few samples. 

I think we should probably update this implementation later to do a moving average (for example, for the last N hrs.) 


@nyechiel  ^ what do you think would be most useful? 1h / 2h/ 4h/ .. 24h?

pkg/cableengine/syncer/syncer.go
```diff
@@ -201,6 +203,20 @@ func (i *GatewaySyncer) generateGatewayObject() *v1.Gateway {
 	}
 
 	if connections != nil {
+		if i.healthcheck != nil {
+			for index := range *connections {
+				connection := &(*connections)[index]
+				latencyInfo := i.healthcheck.GetLatencyInfo(connection.Endpoint.Hostname)
```
mangelajo, line 209: What if two endpoints share the same hostname on the remote side? (for example, two clusters configured equally on some cloud.. this could happen)

I believe we should probably index them by the IP address being pinged.

pkg/util/util.go
mangelajo: Ok, we will need to refactor this later, this code is also undergoing some changes, so better not to rely on it now. 

Can you add a TODO: to handle de-duplication of code/finding common parts with the route agent ?

pkg/cableengine/healthchecker/pinger.go
mangelajo: Can we add a TODO to handle this on the future?

main.go
```diff
@@ -109,26 +110,41 @@ func main() {
...
+	klog.Errorf("The local endpoint is %v", localEndpoint)
 	cableEngine := cableengine.NewEngine(localCluster, localEndpoint)
 
+	err = subv1.AddToScheme(scheme.Scheme)
+	if err != nil {
+		klog.Errorf("Error adding submariner types to the scheme: %v", err)
+	}
+
+	var cableHealthchecker *healthchecker.HealthChecker
+	if len(submSpec.GlobalCidr) == 0 {
```
mangelajo, line 128: This will break the "one slide globalnet" , another side not globalnet usecase we found for the storage cluster. 

But we can fix it by adding globalnet support as a bugfix ;)

mangelajo (changes requested):

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,138 @@
...
+type Pinger struct {
+	remoteEndpointName string
+	healthCheckIp      string
+	statistics         Statistics
+	Status             string
+	stopCh             chan struct{}
+	pinger             *ping.Pinger
+}
+
+type Statistics struct {
```
mangelajo, line 23: Can we add some unit testing for the Statistics object?

pkg/cableengine/healthchecker/pinger.go
aswinsuryan: I will add a TODO.

We can do a moving average with this algorithm. But since there an approximation when we calculate the square root, an error factor will be introduced. With a 100 samples itself, I could see a small difference with this algorithm for SD ( which computes incremental SD for each element addition)  in the least significant digits. This will grow as the packet count increases.

So we will have to reset the SD periodically.  If we make window large say 24 hours, we will have some computational overhead to recalculate SD for that period.

pkg/util/util.go
aswinsuryan: Added TODO

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/cableengine/healthchecker/pinger.go

pkg/cableengine/syncer/syncer.go
aswinsuryan: Done.

main.go
aswinsuryan: yes, we can fix it, once the datapath changes are ready.

pkg/cableengine/healthchecker/pinger.go

pkg/cableengine/healthchecker/pinger.go
aswinsuryan: Added TODO.

pkg/cableengine/healthchecker/statistics_test.go
```diff
@@ -0,0 +1,45 @@
...
+})
+
+func testStatistics() {
+	size := 10
+	statistics := &Statistics{
+		size:         uint64(size),
+		previousRtts: make([]uint64, size),
+	}
+
+	sampleSpace := [10]uint64{404351, 490406, 530333, 609556, 609650, 685106, 726265, 785707, 1044609, 1048263}
```
mangelajo, line 21: some nits sorry ':D
to make the test easier to follow / check on reading

I'd use some constants

```
sampleSpace := [10]uint64{testMinRTT, 490406, 530333, 609556, 609650, 685106, 726265, 785707, testMaxRTT, testLastRTT}

```

(you need to change the max, because it's also the last.)

pkg/cableengine/healthchecker/statistics_test.go
```diff
@@ -0,0 +1,45 @@
...
+		statistics.updateStatistics(v)
+	}
+
+	Expect(statistics.maxRtt).To(Equal(uint64(1048263)))
+	Expect(statistics.minRtt).To(Equal(uint64(404351)))
+	Expect(statistics.lastRtt).To(Equal(uint64(1048263)))
+	Expect(statistics.mean).To(Equal(uint64(693424)))
+	Expect(statistics.stdDev).To(Equal(uint64(205994)))
+
+	statistics.updateStatistics(404300)
```
mangelajo, line 32:
```suggestion
	statistics.updateStatistics(testNewMinRTT)
```

pkg/cableengine/healthchecker/statistics_test.go
```diff
@@ -0,0 +1,45 @@
...
+	statistics.updateStatistics(1048264)
```
mangelajo, line 33:
```suggestion
	statistics.updateStatistics(testNewMaxRTT)
	statistics.updateStatistics(testNewLastRTT)
```

pkg/cableengine/healthchecker/statistics_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	Expect(statistics.maxRtt).To(Equal(uint64(1048264)))
+	Expect(statistics.minRtt).To(Equal(uint64(404300)))
+	Expect(statistics.lastRtt).To(Equal(uint64(1048264)))
```
mangelajo, line 37:
```suggestion
	Expect(statistics.maxRtt).To(Equal(uint64(testNewMaxRTT)))
	Expect(statistics.minRtt).To(Equal(uint64(testNewMinRTT)))
	Expect(statistics.lastRtt).To(Equal(uint64(testNewLastRTT)))
```

pkg/cableengine/healthchecker/statistics_test.go
aswinsuryan: Done.

pkg/cableengine/healthchecker/statistics_test.go
aswinsuryan: Done.

pkg/cableengine/healthchecker/statistics_test.go
aswinsuryan: Done.

pkg/cableengine/healthchecker/statistics_test.go
aswinsuryan: Done.

mangelajo (approved):

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+
+				p.sendPing(p.pinger, p.healthCheckIp)
+			}
+		}
+	}()
+	klog.Infof("CableEngine HealthChecker started")
+}
+
+func (p *Pinger) stop() {
+	p.pinger.Stop()
```
tpantelis, line 59:
```suggestion
	close(p. stopCh)
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+
+				p.pinger, err = ping.NewPinger(p.healthCheckIp)
+				if err != nil {
+					klog.Errorf("Creating pinger for the ip %v failed due to %v", p.healthCheckIp, err)
+
+					return
+				}
+
+				p.sendPing(p.pinger, p.healthCheckIp)
+			}
```
tpantelis, line 52: This should use a timer instead of `Sleep`  so stop is immediate.

```suggestion
			select {
			case <-p.stopCh:
				return
			case <-time.After(waitTime):
				p.sendPing()
			}
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+var waitTime time.Duration = 15
+var size uint64 = 1000
+
+type Pinger struct {
+	healthCheckIp string
+	statistics    Statistics
+	Status        string
+	stopCh        chan struct{}
+	pinger        *ping.Pinger
+}
```
tpantelis, line 19:
```suggestion
type pingerInfo struct {
	healthCheckIP string
	statistics    Statistics
	failureMsg        string
	stopCh        chan struct{}
}
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}()
+	klog.Infof("CableEngine HealthChecker started")
+}
+
+func (p *Pinger) stop() {
+	p.pinger.Stop()
+}
+
+func (p *Pinger) sendPing(pinger *ping.Pinger, healthCheckIp string) {
```
tpantelis, line 62:
```suggestion
func (p *Pinger) sendPing() {
	pinger, err := ping.NewPinger(p.healthCheckIP)
	if err != nil {
		klog.Errorf("Error creating pinger for IP %q: %v", p.healthCheckIP, err)
		return
	}
	
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+		p.Status = ""
+		p.statistics.updateStatistics(uint64(packet.Rtt.Nanoseconds()))
+	}
+
+	pinger.OnFinish = func(stats *ping.Statistics) {
+		p.Status = "Failed to successfully ping the remote endpoint"
+	}
+	err := pinger.Run()
+	if err != nil {
+		klog.Errorf("Ping to ip %q failed due to  %v", healthCheckIp, err)
```
tpantelis, line 76: Shouldn't we set the `failureMsg` (aka `Status`) field here?

```suggestion
		klog.Errorf("Ping to ip %q failed: %v", p.healthCheckIP, err)
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+	pinger.RecordRtts = false
+
+	pinger.OnRecv = func(packet *ping.Packet) {
+		p.Status = ""
+		p.statistics.updateStatistics(uint64(packet.Rtt.Nanoseconds()))
+	}
+
+	pinger.OnFinish = func(stats *ping.Statistics) {
+		p.Status = "Failed to successfully ping the remote endpoint"
+	}
```
tpantelis, line 73: The docs say this is called when the Pinger exits so doesn't seem indicative of an error. It seems we can remove this.

pkg/cableengine/healthchecker/statistics.go
```diff
@@ -0,0 +1,63 @@
+package healthchecker
+
+import (
+	"math"
+)
+
+type Statistics struct {
```
tpantelis, line 7:
```suggestion
type statistics struct {
```

pkg/cableengine/healthchecker/statistics.go
```diff
@@ -0,0 +1,63 @@
...
+	stdDev       uint64
+	lastRtt      uint64
+	minRtt       uint64
+	maxRtt       uint64
+	sqrDiff      uint64
+	index        uint64
+	size         uint64
+}
+
+func (s *Statistics) updateStatistics(rtt uint64) {
```
tpantelis, line 20:
```suggestion
func (s *Statistics) update(rtt uint64) {
```

pkg/cableengine/healthchecker/statistics_test.go
```diff
@@ -0,0 +1,61 @@
...
+	Expect(statistics.minRtt).To(Equal(uint64(testNewMinRTT)))
+	Expect(statistics.lastRtt).To(Equal(uint64(testNewLastRTT)))
+	Expect(statistics.mean).To(Equal(uint64(newExpectedMean)))
+	Expect(statistics.stdDev).To(Equal(uint64(newExpectedSD)))
+}
+
+func TestSyncer(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Statistics Suite")
+}
```
tpantelis, line 61: This should be in a _healthchecker_suite_test.go_ file.

```suggestion
func TestHealthChecker(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Health Checker Suite")
}
```

pkg/cableengine/syncer/syncer.go
```diff
@@ -44,11 +45,12 @@ func init() {
 
 // NewEngine creates a new Engine for the local cluster
 func NewGatewaySyncer(engine cableengine.Engine, client v1typed.GatewayInterface,
-	version string) *GatewaySyncer {
+	version string, healthcheck *healthchecker.HealthChecker) *GatewaySyncer {
```
tpantelis, line 48:
```suggestion
	version string, healthcheck healthchecker.Interface) *GatewaySyncer {
```

pkg/cableengine/syncer/syncer_test.go
```diff
@@ -335,9 +342,22 @@ func (t *testDriver) run() {
...
+	restMapper := test.GetRESTMapperFor(&submarinerv1.Endpoint{})
+
 	client := fakeClientset.NewSimpleClientset()
 	t.gateways.GatewayInterface = client.SubmarinerV1().Gateways(namespace)
-	t.syncer = syncer.NewGatewaySyncer(t.engine, t.gateways, t.expectedGateway.Status.Version)
+	t.healthChecker, _ = healthchecker.New(&watcher.Config{
+		RestMapper: restMapper,
+		Client:     dynamicClient,
+		Scheme:     scheme,
+	}, namespace, "west")
```
tpantelis, line 359: The syncer should take a `healthchecker.Interface` so you can create a mock/fake.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,137 @@
...
+	endpointWatcher watcher.Interface
+	healthCheckers  sync.Map
+	clusterID       string
+}
+
+func New(config *watcher.Config, endpointNameSpace, clusterID string) (*HealthChecker, error) {
+	serviceExportController := HealthChecker{
+		clusterID: clusterID,
+	}
+
```
tpantelis, line 34:
```suggestion
```

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,137 @@
...
+	serviceExportController.endpointWatcher = endpointWatcher
+
+	return &serviceExportController, nil
+}
+
+func (h *HealthChecker) GetLatencyInfo(hostName string) *LatencyInfo {
+	if obj, found := h.healthCheckers.Load(hostName); found {
+		pinger := obj.(*Pinger)
+		statistics := pinger.statistics
```
tpantelis, line 61: No need to copy here - access `pinger.statistics` below.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,137 @@
...
+	if obj, found := h.healthCheckers.Load(endpointCreated.Spec.Hostname); found {
+		klog.V(log.TRACE).Infof("HealthChecker is already running %q: %q, and hence stopping", endpointCreated.Spec.HealthCheckIP,
+			endpointCreated.Spec.Hostname)
+
+		pinger := obj.(*Pinger)
+		if pinger.healthCheckIp == endpointCreated.Spec.HealthCheckIP {
+			return false
+		}
+
+		close(pinger.stopCh)
```
tpantelis, line 109:
```suggestion
```

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,137 @@
...
+		pinger.stop()
+		h.healthCheckers.Delete(endpointCreated.Spec.Hostname)
+	}
+
+	klog.V(log.TRACE).Infof("Starting Pinger for Hostname: %q, with HealthCheckIP: %q",
+		endpointCreated.Spec.HealthCheckIP, endpointCreated.Spec.Hostname)
+
+	pinger := NewPinger(endpointCreated.Spec.HealthCheckIP)
+	h.healthCheckers.Store(endpointCreated.Spec.HealthCheckIP, pinger)
```
tpantelis, line 118:
```suggestion
	h.healthCheckers.Store(endpointCreated.Spec.Hostname, pinger)
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+		healthCheckIp: healthCheckIP,
+		statistics: Statistics{
+			size:         size,
+			previousRtts: make([]uint64, size),
+		},
+		stopCh: make(chan struct{}),
+	}
+}
+
+func (p *Pinger) Run() {
```
tpantelis, line 32:
```suggestion
func (p *Pinger) start() {
```

tpantelis (changes requested):

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return nil
+}
+
+func (h *HealthChecker) endpointCreatedorUpdated(obj runtime.Object) bool {
+	klog.V(log.TRACE).Infof("Endpoint created: %v", obj)
+	endpointCreated := obj.(*submarinerv1.Endpoint)
+	if endpointCreated.Spec.ClusterID == h.clusterID {
+		klog.Infof("Skipping endpoint creation for the local cluster %q", h.clusterID)
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: Since hostname could be duplicate changed to HealtcheckIP, but I missed to change it on a delete, did that now,

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"time"
+
+	"github.com/go-ping/ping"
+	"k8s.io/klog"
+)
+
+var waitTime time.Duration = 15
+var size uint64 = 1000
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/cableengine/healthchecker/pinger.go
aswinsuryan: since we are not setting the count, it will be an endless ping. ff there are no successful packets till  the Timeout the onFinish will be called and it is a failed ping.

The timeout setting seems to be missed during the patch rework I guess I have added it back.

pkg/cableengine/syncer/syncer_test.go
aswinsuryan: Done.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -16,121 +16,116 @@ type LatencyInfo struct {
 }
 
 type Interface interface {
-	Start(stopCh <-chan struct{})
+	Start(stopCh <-chan struct{}) error
 
-	GetLatencyInfo(hostName string) *LatencyInfo
+	GetLatencyInfo(hostName string) interface{}
```
tpantelis, line 21: Why was this changed from a concrete type to `interface{}`?

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -7,72 +7,72 @@ import (
...
 
 	pinger.OnFinish = func(stats *ping.Statistics) {
-		p.Status = "Failed to successfully ping the remote endpoint"
+		p.failureMsg = "Failed to successfully ping the remote endpoint"
 	}
-	err := pinger.Run()
+	err = pinger.Run()
 	if err != nil {
-		klog.Errorf("Ping to ip %q failed due to  %v", healthCheckIp, err)
+		klog.Errorf("Ping to ip %q failed: %v", p.healthCheckIP, err)
```
tpantelis, line 76: Shouldn't we set the `failureMsg` field here?

```suggestion
		klog.Errorf("Error running ping for the remote endpoint IP %q: %v", p.healthCheckIP, err)
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -7,72 +7,72 @@ import (
...
 	pinger.OnRecv = func(packet *ping.Packet) {
-		p.Status = ""
-		p.statistics.updateStatistics(uint64(packet.Rtt.Nanoseconds()))
+		p.failureMsg = ""
+		p.statistics.update(uint64(packet.Rtt.Nanoseconds()))
 	}
 
 	pinger.OnFinish = func(stats *ping.Statistics) {
-		p.Status = "Failed to successfully ping the remote endpoint"
+		p.failureMsg = "Failed to successfully ping the remote endpoint"
```
tpantelis, line 72: Can you put your explanation into a code comment: "_Since we are setting a timeout and not a count, it will be an endless ping. ff the timeout is reached with no successful packets, onFinish will be called and it is a failed ping._".  Also add more context to the message.

```suggestion
		p.failureMsg = fmt.Sprintf("Failed to successfully ping the remote endpoint IP %q", p.healthCheckIP)
```

pkg/cableengine/healthchecker/statistics_test.go
```diff
@@ -0,0 +1,54 @@
...
+		testMinRTT     = 404351
+		testMaxRTT     = 1048263
+		testLastRTT    = 1044609
+		testNewMinRTT  = 404300
+		testNewMaxRTT  = 1048264
+		testNewLastRTT = 609555
+	)
+
+	When("Update statistics is called with a sample space", func() {
+		It("The statistics should be computes as expected", func() {
```
tpantelis, line 19:
```suggestion
	When("update is called with a sample space", func() {
		It("should correctly compute the statistics", func() {
```

pkg/cableengine/syncer/syncer_test.go
tpantelis, line 359: I don't think it makes sense to use the real health checker implementation. We should use a fake one so we can test  the interaction with it.  But we can leave this as is for now - I can follow up with another PR.

tpantelis (changes requested):

pkg/cableengine/syncer/syncer_test.go
aswinsuryan: ok , thanks . :)

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: it was an oversight , when I was fixing some compilation issues reverted it.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -16,121 +16,116 @@ type LatencyInfo struct {
...
 	}
 
 	klog.V(log.TRACE).Infof("Starting Pinger for Hostname: %q, with HealthCheckIP: %q",
 		endpointCreated.Spec.HealthCheckIP, endpointCreated.Spec.Hostname)
 
-	pinger := NewPinger(endpointCreated.Spec.HealthCheckIP)
-	h.healthCheckers.Store(endpointCreated.Spec.HealthCheckIP, pinger)
-	pinger.Run()
+	pinger := newPinger(endpointCreated.Spec.HealthCheckIP)
+	h.pingers.Store(endpointCreated.Spec.HealthCheckIP, pinger)
```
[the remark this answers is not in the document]
aswinsuryan: Using cableName now as discussed.

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,132 @@
...
+
+type LatencyInfo struct {
+	ConnectionError string
+	Spec            *submarinerv1.LatencySpec
+}
+
+type Interface interface {
+	Start(stopCh <-chan struct{}) error
+
+	GetLatencyInfo(endpoint submarinerv1.EndpointSpec) *LatencyInfo
```
tpantelis, line 21:
```suggestion
	GetLatencyInfo(endpoint *submarinerv1.EndpointSpec) *LatencyInfo
```

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	pinger.OnFinish = func(stats *ping.Statistics) {
+		// Since we are setting a timeout and not a count, it will be an endless ping.
+		// If the timeout is reached with no successful packets, onFinish will be called and it is a failed ping.
+		p.failureMsg = fmt.Sprintf("Failed to successfully ping the remote endpoint IP %q", p.healthCheckIP)
+	}
+	err = pinger.Run()
+	if err != nil {
+		klog.Errorf("Error running ping for the remote endpoint IP %q: %v", p.healthCheckIP, err)
```
tpantelis, line 79: Shouldn't we set the `failureMsg` field here?

tpantelis (changes requested):

pkg/cableengine/healthchecker/pinger.go
aswinsuryan: This error occurs when something is wrong with the pinger configuration, not the actual datapath error, should we be setting error in that  case too?

aswinsuryan: Have raised an issue to address this https://github.com/submariner-io/submariner/issues/969

tpantelis (approved):

Issue #678, opened 2020-06-24
Prometheus metrics


skitt:
**What would you like to be added**:

The various Submariner components should export metrics which can then be used with _e.g._ Prometheus. Useful information includes gateway reachability, error rates, transfer rates (bytes rx/tx, packets rx/tx etc.).

**Why is this needed**:

Prometheus metrics are commonly used to monitor K8s clusters. Providing these means we can piggy-back on the Prometheus-based ecosystem for graphs, alerts etc.

IMO this is worth bearing in mind for https://github.com/submariner-io/submariner-operator/issues/359, at least the parts thereof which will affect the information provided by cable drivers and the `Gateway` object.

**Cross cluster connection**

- [x] (p1) RX/TX bytes and packets between gateway nodes: https://github.com/submariner-io/submariner/pull/846 

- [x] (p1) Connected/not connected 	(We have the number of connections.) — number of connections and corresponding status (as labels), with connection details as labels (endpoint IPs): [#942]( https://github.com/submariner-io/submariner/pull/942
)
- [x] (p1) Connection  uptime: [#942]( https://github.com/submariner-io/submariner/pull/942
)

- [x] (p1) Connection latency: [#978 ](https://github.com/submariner-io/submariner/pull/978)

tpantelis:
bump

roytman:
Should we add a number of reconnects in last 24hs or something like this. I mean that "Connection uptime" can be not enough.

skitt:
> Should we add a number of reconnects in last 24hs or something like this. I mean that "Connection uptime" can be not enough.

Prometheus keeps the history for us; we’d see multiple reconnects as the uptime graph starting from zero several times. This gives a more global view than can be tracked in the cable driver only: it will show for example reconnects which come from the whole pod restarting, or the gateway node changing.

skitt:
When we’re tracking connection states, we need to reset the states at each iteration, otherwise Prometheus will remember old states for us. We can’t do this globally because the metrics track multiple cable drivers potentially; instead, we’ll require each cable driver to explicitly set unused statuses to 0.

skitt:
We need to check whether it’s useful to measure the active gateway’s uptime from Submariner: since the engine is recording the metrics, it can only see itself, and not other gateways. We already track something similar to this in the operator.

maayanf24:
Additional metrics definition in Epic: https://github.com/submariner-io/submariner-operator/issues/745

PR #942, opened 2020-11-12
Add connection metrics and track from Libreswan cable driver

maayanf24:
Related to https://github.com/submariner-io/submariner/issues/678


pkg/cable/metrics.go
```diff
@@ -55,32 +60,126 @@ var (
...
+		cableDriverLabel:      cableDriverName,
+		localClusterLabel:     localEndpoint.ClusterID,
+		localHostnameLabel:    localEndpoint.Hostname,
+		localEndpointIpLabel:  localEndpoint.PublicIP,
+		remoteClusterLabel:    remoteEndpoint.ClusterID,
+		remoteHostnameLabel:   remoteEndpoint.Hostname,
+		remoteEndpointIpLabel: remoteEndpoint.PublicIP,
+	}
+	txGauge.Delete(labels)
+	rxGauge.Delete(labels)
```
skitt, line 181: I can’t find the details in the docs right now, but I think that deleting results in the gauge not being exported for those labels, and that might not be what we want for rx/tx. I’ll check tomorrow and update.

pkg/cable/metrics.go
```diff
@@ -55,32 +60,126 @@ var (
...
+			cableDriverLabel,
+			localClusterLabel,
+			localHostnameLabel,
+			localEndpointIpLabel,
 			remoteClusterLabel,
 			remoteHostnameLabel,
+			remoteEndpointIpLabel,
+		},
+	)
+	connectionLatencySecondsGague = prometheus.NewGaugeVec(
```
skitt, line 99: `Gauge` ;-).

pkg/cable/metrics.go
```diff
@@ -55,32 +60,126 @@ var (
...
+		},
+		[]string{
+			cableDriverLabel,
+			localClusterLabel,
+			localHostnameLabel,
+			localEndpointIpLabel,
+			remoteClusterLabel,
+			remoteHostnameLabel,
+			remoteEndpointIpLabel,
+		},
```
skitt, line 82: Shouldn’t this have a label for the status too?

skitt (changes requested):

pkg/cable/metrics.go
maayanf24: then what will be the gauge's value?

pkg/cable/metrics.go
maayanf24: shouldn't we use same labels for rx/tx gauges as all gauges in this file?

pkg/cable/metrics.go
skitt: The number of connections per status, same as the operator.

pkg/cable/metrics.go
```diff
@@ -59,28 +63,115 @@ var (
...
+		localEndpointIpLabel:  localEndpoint.PublicIP,
+		remoteClusterLabel:    remoteEndpoint.ClusterID,
+		remoteHostnameLabel:   remoteEndpoint.Hostname,
+		remoteEndpointIpLabel: remoteEndpoint.PublicIP,
+	}
+}
+
+func getTxRxLabels(cableDriverName string, localEndpoint, remoteEndpoint *submv1.EndpointSpec) prometheus.Labels {
+	return prometheus.Labels{
 		cableDriverLabel:    cableDriverName,
```
skitt, line 132: Regarding your point about whether the tx/rx gauges should have the same labels as the other metrics, you’re right, they should; so we should use `getLabels()` everywhere, and drop `getTxRxLabels()`.

skitt (changes requested):
Looks good, thanks, just a minor tweak (as you suggested) on the tx/rx labels!

skitt (approved):

mangelajo (approved):

PR #944, opened 2020-11-14
Enable multiarch images work

mangelajo:
This allows compiling the binaries and the images with

```bash
GOARCH=arm ARCH=arm32v7 make bin/submariner-route-agent

docker build . -f package/Dockerfile.submariner --build-arg ARCH=arm32v7/ -t mangelajo/submariner-route-agent:manifest-arm32v7
```

while still not perferct, it enables starting the work,
the route-agent image has been moved to fedora:33 to allow arm32v7 images.


skitt:
See https://github.com/submariner-io/submariner/pull/964 for an alternative approach which doesn’t require prefixing the base image name, and which will ultimately allow multiarch images to be built.

mangelajo:
Love your patches, more like a finished product ;-)

Issue #858, opened 2020-10-07
Convert the existing implementation into a "kube-proxy iptables" driver which will be the default.


mangelajo:
**What would you like to be added**:

Convert the existing iptables based implementation into a driver as defined by #857 

**Why is this needed**:

As part of the base work for supporting OVN

Depends-On: #857 

See:  https://github.com/submariner-io/enhancements/pull/11
Related-Issue: https://github.com/submariner-io/submariner/issues/778

PR #946, opened 2020-11-15
Route Agent Kubeproxy iptables driver

sridhargaddam:
This PR implements the following features

1. Creates vxLAN tunnels between the worker nodes and the local active gateway node.
2. Handles gateway failover scenarios there is a transition.
3. Handles programming of necessary routes and iptable rules when a new endpoint is added.
4. Programs the FDB entries on the GatewayNode when a new node is added to the cluster.
5. Handles routing rules for HostNetwork to remoteCluster support.
6. Handles overlapping CIDRs from remote clusters
7. Discovers and annotates the nodes with cniInterfaceIP
8. Handles cleanup during gateway transitions

Fixes issue: https://github.com/submariner-io/submariner/issues/858


pkg/event/registry.go
```diff
@@ -63,6 +63,12 @@ func (er *Registry) AddHandlers(eventHandlers ...Handler) error {
 	return nil
 }
 
+func (er *Registry) StartHandlers(clusterID string, ClusterCidr, ServiceCidr []string) error {
```
mangelajo, line 66: I'm doing this in the OVN side when receiving the first local-endpoint.

May be we don't need to receive this anymore for the route agents and we can discover it from the local endpoints?

The controller already knows the local clusterID... and will send LocalEndpoint... updates

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,100 @@
...
+func NewSyncHandler() *SyncHandler {
+	return &SyncHandler{}
+}
+
+func (kp *SyncHandler) GetName() string {
+	return "kubeproxy-iptables-handler"
+}
+
+func (kp *SyncHandler) GetNetworkPlugin() string {
+	return event.AnyNetworkPlugin
```
mangelajo, line 31: In our case it will be "generic" or "Generic" whatever subctl reports on network discovery.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (kp *SyncHandler) Init() error {
+	return nil
+}
+
+func (kp *SyncHandler) TransitionToNonGateway() error {
```
mangelajo, line 38: You don't need to add the logging again, just add the logging handler during initialization of your controller., see: https://github.com/submariner-io/submariner/blob/f000f3605e05f334623c7227c9827c4e920c9439/pkg/networkplugin-syncer/main.go#L30

pkg/routeagent-driver/main.go
```diff
@@ -0,0 +1,58 @@
...
+func main() {
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	klog.Info("SGM: Starting submariner-route-agent using the event framework")
+	// set up signals so we handle the first shutdown signal gracefully
+	stopCh := signals.SetupSignalHandler()
+
+	eventHandlers := event.NewRegistry("routeagent-driver", os.Getenv("NETWORK_PLUGIN"))
+	if err := eventHandlers.AddHandlers(logger.NewHandler(), kp.NewSyncHandler()); err != nil {
```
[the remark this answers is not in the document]
sridhargaddam: Thanks for fixing it in admiral and the logger. I updated the references to the latest admiral repo in this PR (i.e., go.mod and go.sum)

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
sridhargaddam: Yes, I will change the log messages in further PRs. This was just for my debugging until I handle the corresponding events.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
sridhargaddam: Okay, I'll update it.

pkg/event/registry.go
sridhargaddam: I thought about it, but its prone to race conditions if we take that approach. Let me explain...

As you know, route-agent needs to discover the local CNI Interface on the hosts and also annotate the nodes with the cni-iface ip. As part of discoverCNIInterface API, we need the localClusterCIDR info. I plan to handle this in the NodeCreated event. So, when NodeCreated event is fired, we need the localClusterCIDR.
If we use the localEndpoint events, it might arrive later (most likely) compared to NodeEvents.
So to have a proper initialization sequence, IMHO its better to have a new API called "Start" (or something on those lines) where we can initialize the necessary things on the handler side. Hope it makes sense.

sridhargaddam (commented):
thanks for the review comments @mangelajo

pkg/event/registry.go
mangelajo: Makes sense, it's not a problem for OVN.

Can you take those details from the environment variables instead in the driver itself?


Also, can we separate the "annotation of local CNI ip" to a separate handler?,

In that way it would be reusable if we wanted to use it also... in OVN or other drivers for example.

Also code is cleaner and more isolated.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
mangelajo: Well, now that I think it can be generic or OpenshiftSDN, probably when the generic is ready we can make  a sub-class driver, which only changes the name...

In that way if we had any openshift SDN specifics we could handle them there :)

pkg/event/registry.go
sridhargaddam: Yes, the controller.go is getting the info from the environment variables and is passing it to the handlers via StartHandlers() API.

You mean an handler to "annotate the nodes with CNI IP"? something like logger handler?

mangelajo: Yes, something like that, if that makes sense.

That way it would be a reusable functionality for OVN too.

mangelajo: > Yes, the controller.go is getting the info from the environment variables and is passing it to the handlers via StartHandlers() API.
> 
> You mean an handler to "annotate the nodes with CNI IP"? something like logger handler?

I mean, instead of doing that, if you need that info, grab it from the environment variables. Could also be passed during Init, why having a separate start handlers?

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (kp *SyncHandler) Init() error {
+	var err error
+	kp.hostname, err = os.Hostname()
+	if err != nil {
+		klog.Fatalf("unable to determine hostname: %v", err)
+	}
+
+	cniIface, err := cni_driver.DiscoverCNIInterface(kp.localClusterCidr[0])
```
mangelajo, line 48: I would move this to the main of the route agent, since we will most likely leed the CNI interface details always for any driver. Other services rely on this tag being present on the node. Even the test? I can't remember what exactly but it's something we expect.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,67 @@
...
+	if err == nil {
+		// Configure CNI Specific changes
+		kp.cniIface = cniIface
+		err := cni_driver.ToggleCNISpecificConfiguration(kp.cniIface.Name)
+		if err != nil {
+			return fmt.Errorf("ToggleCNISpecificConfiguration returned error. %v", err)
+		}
+	} else {
+		klog.Errorf("DiscoverCNIInterface returned error %v", err)
+	}
```
mangelajo, line 58: I would make this block a whole function we can call.

cni_interface.Setup()

pkg/routeagent-driver/cni-driver/cni-iface.go
```diff
@@ -0,0 +1,67 @@
+package cni_driver
```
mangelajo, line 1: Since we are doing refactors, some suggestions, but take it only as opinion.

I would name this module cni_interface (instead of cni-driver)

Also I believe we always use the "_" convention instead of "-" for modules because it's problematic for go... (the import names don't translate directly)

pkg/routeagent-driver/cni-driver/cni-iface.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/submariner-io/admiral/pkg/log"
+)
+
+type CniInterface struct {
+	Name      string
+	IPAddress string
+}
+
+func DiscoverCNIInterface(clusterCIDR string) (*CniInterface, error) {
```
mangelajo, line 18: Discover  (avoid shuttering)

pkg/routeagent-driver/cni-driver/cni-iface.go
```diff
@@ -0,0 +1,67 @@
...
+					return &CniInterface{IPAddress: ipAddr.String(), Name: iface.Name}, nil
+				}
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("unable to find CNI Interface on the host which has IP from %q", clusterCIDR)
+}
+
+func ToggleCNISpecificConfiguration(iface string) error {
```
mangelajo, line 56: With the suggestion I'm doing later of a top level function Setup, for this module, this would become a private function, may be:

configureRpFilter(iface string)  ?

pkg/routeagent-driver/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+	var env specification
+	err := envconfig.Process("submariner", &env)
+	if err != nil {
+		klog.Fatalf("Error reading the environment variables: %s", err.Error())
+	}
+
+	eventHandlers := event.NewRegistry("routeagent-driver", os.Getenv("NETWORK_PLUGIN"))
+	if err := eventHandlers.AddHandlers(logger.NewHandler(),
+		kp.NewSyncHandler(env.ClusterID, env.Namespace, env.ClusterCidr, env.ServiceCidr)); err != nil {
```
mangelajo, line 45: nit: You could just pass the env object, but you may need to move the env variable specification into a common module

pkg/routeagent-driver/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+
+	"github.com/kelseyhightower/envconfig"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+
+	"github.com/submariner-io/submariner/pkg/event"
+	"github.com/submariner-io/submariner/pkg/event/controller"
+	"github.com/submariner-io/submariner/pkg/event/logger"
+	kp "github.com/submariner-io/submariner/pkg/routeagent-driver/handlers/kubeproxy-iptables"
```
mangelajo, line 14: kp -> kp_iptables (so the code later is more understandable in the kp.NewHandler)

mangelajo (commented):
Sent a bunch of comments ':D

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
sridhargaddam: As I explained above, AFAIU this is specific to kp-iptables driver. I'm not sure we require this for OVN - please let me know.

I'll add similar code in the main() method of route-agent where we discover and annotate the respective node - this piece of code seems to be common to all CNIs and can be re-used.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
sridhargaddam: I can do that, but we also need an API that returns the CNIInterface on the host.

pkg/routeagent-driver/main.go
sridhargaddam: Ok

pkg/routeagent-driver/main.go
sridhargaddam: Yes I thought about it, but wanted to have explicit arguments as it will be clear on what we are passing to the SyncHandler.

pkg/routeagent-driver/cni-driver/cni-iface.go
mangelajo: It's the file names, then when you do an import I believe golang will need a cni_drive "pkg/routeagent-driver/cni-driver" , at least on my side GoLand editor was getting crazy otherwise   

But I'm not sure now. Not important anyway

pkg/routeagent-driver/cni-driver/cni-iface.go
mangelajo: Hehe I was proposing to rename the module to cni_interface.

So it would be cni_interface.Discover()

pkg/routeagent-driver/cni-driver/cni-iface.go
mangelajo: Oh right, the rp_filter. You're right @sridhargaddam

mangelajo: At least I would use the explicit name ConfigureRpFilter if possible

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
mangelajo: Makes sense Sridhar

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,67 @@
+package kp_iptables
+
+import (
+	"fmt"
+	"os"
+
+	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/event"
+	cni_driver "github.com/submariner-io/submariner/pkg/routeagent-driver/cni-driver"
```
mangelajo, line 10: ah see, this is what I was meaning ^ we end up needing to add the cni_driver in front ':)

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+func (kp *SyncHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
+	kp.syncHandlerMutex.Lock()
+	defer kp.syncHandlerMutex.Unlock()
+	kp.localCableDriver = endpoint.Spec.Backend
+
+	// We are on nonGateway node
+	if endpoint.Spec.Hostname != kp.hostname {
+		kp.isGatewayNode = false
```
mangelajo, line 20: May be we should handle this block in TransitionToNonGateway, probably we can pass newEndpoint into the TransitionToNonGateway so this information would become available.

What do you think?

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,169 @@
...
+func (kp *SyncHandler) LocalEndpointUpdated(endpoint *submV1.Endpoint) error {
+	return nil
+}
+
+func (kp *SyncHandler) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
+	kp.syncHandlerMutex.Lock()
+	defer kp.syncHandlerMutex.Unlock()
+	kp.isGatewayNode = false
+
+	if kp.vxlanDevice != nil {
```
mangelajo, line 52: What if we receive this information out of order?, for example

1) Gateway has moved:

a) new Endpoint is created,
b) old Endpoint is removed.

I think you need to verify if the vxlanDevice you are going to delete was for this endpoint.

Even if probably the order will always be the same, chances are that if the first removal fails for some reason and it's later retried... we could end cleaning up for the wrong endpoint.

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,169 @@
...
+			return fmt.Errorf("Failed to delete the the vxlan interface on Endpoint removal: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func (kp *SyncHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	if err := kp.overlappingSubnets(endpoint.Spec.Subnets); err != nil {
+		// Skip processing the endpoint when CIDRs overlap
```
mangelajo, line 66: In this case we should probably log, and don't return error?

Otherwise it will be retried forever.

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (kp *SyncHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	kp.syncHandlerMutex.Lock()
+	defer kp.syncHandlerMutex.Unlock()
+	for _, inputCidrBlock := range endpoint.Spec.Subnets {
+		if !kp.remoteSubnets.Contains(inputCidrBlock) {
+			kp.remoteSubnets.Delete(inputCidrBlock)
+		}
+	}
+	// TODO: Handle a remote endpoint removal use-case
```
mangelajo, line 101: I'd open an issue for this.

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,169 @@
...
+	kp.updateRoutingRulesForHostNetworkSupport(endpoint.Spec.Subnets, Delete)
+	kp.updateIptableRulesForInterClusterTraffic(endpoint.Spec.Subnets, Delete)
+
+	return nil
+}
+
+func (kp *SyncHandler) overlappingSubnets(remoteSubnets []string) error {
+	// If the remoteSubnets [*] overlap with local cluster Pod/Service CIDRs we
+	// should not update the IPTable rules on the host, as it will disrupt the
+	// functionality of the local cluster. So, lets validate that subnets do not
```
mangelajo, line 117: Ideally we may want to look also for the local worker nodes network CIDR at some point, but that's an existing problem today, not related to your new code.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
mangelajo: Let's do this on a separate PR when we start having more than one.

mangelajo (changes requested):
Some comments.

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
sridhargaddam: Done

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
sridhargaddam: Good point @mangelajo

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
sridhargaddam: Ack, thanks

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
sridhargaddam: Let me think about it...

pkg/routeagent-driver/handlers/kubeproxy-iptables/constants.go
```diff
@@ -0,0 +1,59 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,184 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (kp *SyncHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	if err := kp.overlappingSubnets(endpoint.Spec.Subnets); err != nil {
+		// Skip processing the endpoint when CIDRs overlap and return nil to avoid re-queuing.
+		klog.Errorf("IsoverlappingSubnets returned error: %v", err)
```
tpantelis, line 80: It seems more context would be helpful?

```suggestion
		klog.Errorf("overlappingSubnets for new remote %#v returned error: %v", endpoint, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	kp.syncHandlerMutex.Lock()
+	defer kp.syncHandlerMutex.Unlock()
+	for _, inputCidrBlock := range endpoint.Spec.Subnets {
+		if !kp.remoteSubnets.Contains(inputCidrBlock) {
+			kp.remoteSubnets.Add(inputCidrBlock)
+		}
+	}
+
+	if err := kp.updateRoutingRulesForInterClusterSupport(endpoint.Spec.Subnets, Add); err != nil {
+		klog.Errorf("updateRoutingRulesForInterClusterSupport returned error: %+v", err)
```
tpantelis, line 93:
```suggestion
		klog.Errorf("updateRoutingRulesForInterClusterSupport for new remote %#v returned error: %+v", endpoint, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	defer kp.syncHandlerMutex.Unlock()
+	for _, inputCidrBlock := range endpoint.Spec.Subnets {
+		if !kp.remoteSubnets.Contains(inputCidrBlock) {
+			kp.remoteSubnets.Delete(inputCidrBlock)
+		}
+	}
+	// TODO: Handle a remote endpoint removal use-case
+	//         - remove related iptable rules
+	if err := kp.updateRoutingRulesForInterClusterSupport(endpoint.Spec.Subnets, Delete); err != nil {
+		klog.Errorf("updateRoutingRulesForInterClusterSupport returned error: %+v", err)
```
tpantelis, line 118:
```suggestion
		klog.Errorf("updateRoutingRulesForInterClusterSupport for removed remote %#v returned error: %+v", err, endpoint)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	// globalCIDRs allocated to the clusters.
+	for _, serviceCidr := range kp.localServiceCidr {
+		overlap, err := util.IsOverlappingCIDR(remoteSubnets, serviceCidr)
+		if err != nil {
+			// Ideally this case will never hit, as the subnets are valid CIDRs
+			klog.Warningf("unable to validate overlapping Service CIDR: %s", err)
+		}
+
+		if overlap {
+			return fmt.Errorf("Local Service CIDR %q, overlaps with remote cluster %s",
```
tpantelis, line 145:
```suggestion
			return fmt.Errorf("Local Service CIDR %q, overlaps with remote cluster subnets %s",
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	for _, podCidr := range kp.localClusterCidr {
+		overlap, err := util.IsOverlappingCIDR(remoteSubnets, podCidr)
+		if err != nil {
+			klog.Warningf("unable to validate overlapping Pod CIDR: %s", err)
+		}
+
+		if overlap {
+			return fmt.Errorf("Local Pod CIDR %q, overlaps with remote cluster %s", podCidr, remoteSubnets)
```
tpantelis, line 157:
```suggestion
			return fmt.Errorf("Local Pod CIDR %q, overlaps with remote cluster subnets %s", podCidr, remoteSubnets)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
```diff
@@ -0,0 +1,184 @@
...
+	addrs, err := kp.defaultHostIface.Addrs()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(addrs) > 0 {
+		for i := range addrs {
+			ipAddr, _, err := net.ParseCIDR(addrs[i].String())
+			if err != nil {
+				klog.Errorf("Unable to ParseCIDR : %v\n", addrs)
```
tpantelis, line 174:
```suggestion
				klog.Errorf("Unable to ParseCIDR  %v: %v", addrs, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/gw_transition.go
```diff
@@ -0,0 +1,55 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/iptables-iface.go
```diff
@@ -0,0 +1,113 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/iptables-iface.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return fmt.Errorf("error initializing iptables: %v", err)
+	}
+
+	for _, localClusterCidr := range kp.localClusterCidr {
+		ruleSpec := []string{"-s", localClusterCidr, "-d", remoteCidrBlock, "-j", "ACCEPT"}
+		klog.V(log.DEBUG).Infof("Installing iptables rule for outgoing traffic: %s", strings.Join(ruleSpec, " "))
+
+		if err = ipt.AppendUnique("nat", constants.SmPostRoutingChain, ruleSpec...); err != nil {
+			return fmt.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
```
tpantelis, line 100:
```suggestion
			return fmt.Errorf("error appending iptables rule %q: %v\n", strings.Join(ruleSpec, " "), err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/iptables-iface.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		// TODO: revisit, we only have to program rules to allow traffic from the podCidr
+		ruleSpec = []string{"-s", remoteCidrBlock, "-d", localClusterCidr, "-j", "ACCEPT"}
+		klog.V(log.DEBUG).Infof("Installing iptables rule for incoming traffic: %s", strings.Join(ruleSpec, " "))
+
+		if err = ipt.AppendUnique("nat", constants.SmPostRoutingChain, ruleSpec...); err != nil {
+			return fmt.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpec, " "), err)
```
tpantelis, line 108:
```suggestion
			return fmt.Errorf("error appending iptables rule %q: %v\n", strings.Join(ruleSpec, " "), err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,107 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,107 @@
...
+	namespace        string
+	localCableDriver string
+	localClusterCidr []string
+	localServiceCidr []string
+
+	remoteSubnets    *util.StringSet
+	remoteVTEPs      *util.StringSet
+	routeCacheGWNode *util.StringSet
+
+	syncHandlerMutex     *sync.Mutex
```
tpantelis, line 32: This does not need to be a pointer - avoids having to initialize.

```suggestion
	syncHandlerMutex     sync.Mutex
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (kp *SyncHandler) GetNetworkPlugin() string {
+	return event.AnyNetworkPlugin
+}
+
+func (kp *SyncHandler) Init() error {
+	var err error
+	kp.hostname, err = os.Hostname()
+	if err != nil {
+		klog.Fatalf("unable to determine hostname: %v", err)
```
tpantelis, line 77: The method signature returns an error so we should do so.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	kp.defaultHostIface, err = util.GetDefaultGatewayInterface()
+	if err != nil {
+		klog.Fatalf("Unable to find the default interface on host: %s", err.Error())
```
tpantelis, line 82: Same here

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,107 @@
...
+	cniIface, err := cni_interface.Discover(kp.localClusterCidr[0])
+	if err == nil {
+		// Configure CNI Specific changes
+		kp.cniIface = cniIface
+		err := cni_interface.ConfigureRpFilter(kp.cniIface.Name)
+		if err != nil {
+			return fmt.Errorf("ConfigureRpFilter returned error. %v", err)
+		}
+	} else {
+		klog.Errorf("DiscoverCNIInterface returned error %v", err)
```
tpantelis, line 94: Return the error?

```suggestion
		return errors.Wrap(err, "Error discovering the CNI interface")
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/node_handler.go
```diff
@@ -0,0 +1,71 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
...
+		LinkIndex: ifaceIndex,
+		Protocol:  4,
+		Table:     RouteAgentHostNetworkTableID,
+	}
+
+	switch operation {
+	case Add:
+		err = netlink.RouteAdd(&route)
+		if err != nil && !os.IsExist(err) {
+			return fmt.Errorf("error adding the route %s: %v", route.String(), err)
```
tpantelis, line 89:
```suggestion
			return fmt.Errorf("error adding the route %s: %v", route, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
...
+		}
+	case Delete:
+		err = netlink.RouteDel(&route)
+		if err != nil {
+			return fmt.Errorf("error deleting the route %s: %v", route.String(), err)
```
tpantelis, line 94:
```suggestion
			return fmt.Errorf("error deleting the route %s: %v", route, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
...
+			} else if curRoute.Gw.Equal(route.Gw) && curRoute.Dst.String() == route.Dst.String() {
+				klog.V(log.DEBUG).Infof("Found equivalent route, not adding")
+				found = true
+			}
+		}
+
+		if !found {
+			err = netlink.RouteAdd(&route)
+			if err != nil {
+				klog.Errorf("Error adding route %s: %v", route.String(), err)
```
tpantelis, line 198:
```suggestion
				klog.Errorf("Error adding route %s: %v", route, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
...
+				Gw:        *kp.vxlanGwIP,
+				Scope:     unix.RT_SCOPE_UNIVERSE,
+				LinkIndex: link.Attrs().Index,
+				Protocol:  4,
+			}
+
+			if operation == Add {
+				err = netlink.RouteAdd(&route)
+				if err != nil {
+					return fmt.Errorf("error adding route %s: %v", route.String(), err)
```
tpantelis, line 236:
```suggestion
					return fmt.Errorf("error adding route %s: %v", route, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
...
+				}
+			} else if operation == Delete {
+				err = netlink.RouteDel(&route)
+				if err != nil {
+					return fmt.Errorf("error deleting route %s: %v", route.String(), err)
```
tpantelis, line 241:
```suggestion
					return fmt.Errorf("error deleting route %s: %v", route, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
...
+	if kp.cniIface != nil {
+		rule := netlink.NewRule()
+		rule.Table = RouteAgentHostNetworkTableID
+		rule.Priority = RouteAgentHostNetworkTableID
+
+		switch operation {
+		case Add:
+			err := netlink.RuleAdd(rule)
+			if err != nil && !os.IsExist(err) {
+				return fmt.Errorf("failed to add ip rule %s: %v", rule.String(), err)
```
tpantelis, line 260:
```suggestion
				return fmt.Errorf("failed to add ip rule %s: %v", rule, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/routes-iface.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+		case Delete:
+			err := netlink.RuleDel(rule)
+			if err != nil && !os.IsNotExist(err) {
+				return fmt.Errorf("failed to delete ip rule %s: %v", rule.String(), err)
```
tpantelis, line 265:
```suggestion
				return fmt.Errorf("failed to delete ip rule %s: %v", rule, err)
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan.go
```diff
@@ -0,0 +1,276 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan_test.go
```diff
@@ -0,0 +1,34 @@
+package kp_iptables
```
tpantelis, line 1:
```suggestion
package kubeproxy-iptables
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"strconv"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("getVxlanVtepIPAddress", func() {
+	Describe("Unit tests for getVxlanVtepIPAddress", func() {
```
tpantelis, line 12:
```suggestion
var _ = Describe("Function getVxlanVtepIPAddress tests", func() {
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan_test.go
```diff
@@ -0,0 +1,34 @@
...
+		Context("When a valid ipaddress is provided to getVxlanVtepIPAddress", func() {
+			It("Should return the VxLAN VtepIP that can be configured", func() {
```
tpantelis, line 14:
```suggestion
		When("a valid IP address is provided", func() {
			It("should return the VxLAN VtepIP that can be configured", func() {
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan_test.go
```diff
@@ -0,0 +1,34 @@
...
+				routeController := SyncHandler{}
+				vtepIP, _ := routeController.getVxlanVtepIPAddress("192.168.100.24")
+				Expect(vtepIP.String()).Should(Equal(strconv.Itoa(VxLANVTepNetworkPrefix) + ".168.100.24"))
+			})
+		})
+
+		Context("When an invalid ipaddress is provided to getVxlanVtepIPAddress", func() {
+			It("Should return an error", func() {
```
tpantelis, line 22:
```suggestion
		When("an invalid IP address is provided", func() {
			It("should return an error", func() {
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan_test.go
```diff
@@ -0,0 +1,34 @@
...
+				routeController := SyncHandler{}
+				_, err := routeController.getVxlanVtepIPAddress("10.0.0")
+				Expect(err).ShouldNot(Equal(nil))
+			})
+		})
+	})
+})
+
+func TestRoute(t *testing.T) {
```
tpantelis, line 31:
```suggestion
func TestKubeProxyIPTables(t *testing.T) {
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan_test.go
```diff
@@ -0,0 +1,34 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Route Suite")
```
tpantelis, line 33:
```suggestion
	RunSpecs(t, "Kube Proxy IP Tables Suite")
```

tpantelis (changes requested):

pkg/routeagent-driver/cni_interface/cni_iface.go
```diff
@@ -0,0 +1,101 @@
...
+
+type CniInterface struct {
+	Name      string
+	IPAddress string
+}
+
+func Discover(clusterCIDR string) (*CniInterface, error) {
+	_, clusterNetwork, err := net.ParseCIDR(clusterCIDR)
+	if err != nil {
+		return nil, fmt.Errorf("unable to ParseCIDR %q : %v", clusterCIDR, err)
```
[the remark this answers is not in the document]
mangelajo: right, errors.Wrap also has the advantage that later on you can do...

errors.Is(err, TypeOfErrror)

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
mangelajo: I believe we should do it, let's modify those transition functions to pass down the endpoint info. Otherwise you are replicating the isGateway/Gateway code in here, while it could be done by using the transition functions.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/submariner-io/submariner/pkg/event"
+	"github.com/submariner-io/submariner/pkg/routeagent-driver/cni_interface"
+	"github.com/submariner-io/submariner/pkg/routeagent-driver/constants"
+	"github.com/submariner-io/submariner/pkg/routeagent/cleanup"
+	"github.com/submariner-io/submariner/pkg/util"
+)
+
+type SyncHandler struct {
+	event.HandlerBase
+	clusterID        string
```
mangelajo, line 22: Not used anywhere

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,107 @@
...
+	namespace        string
```
mangelajo, line 23: not used anywhere

mangelajo (changes requested):

pkg/routeagent-driver/constants/constants.go
```diff
@@ -0,0 +1,32 @@
...
+	// with the corresponding globalIP assigned to the node. Since globalnet controller does
+	// not run on all the worker-nodes and there is no well defined mechanism to get the
+	// CNIInterfaceIP for each of the nodes, we annotate the node with CNIInterfaceIPInfo as
+	// part of route-agent and this will subsequently be used in globalnet controller for
+	// supporting connectivity from HostNetwork to remoteClusters.
+	// [#] interface on the node that has an IPAddress from the clusterCIDR
+	CniInterfaceIp = "submariner.io/cniIfaceIp"
+)
+
+type Specification struct {
```
mangelajo, line 27: This may be deserves a different package, it's not a constant.

May be environment ?

pkg/routeagent-driver/handlers/kubeproxy-iptables/cleanup_handling.go
```diff
@@ -0,0 +1,38 @@
+package kp_iptables
```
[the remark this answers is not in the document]
mangelajo: I think it's better to use `kubeproxy_iptables`  the normal hyphens `-` are problematic later as you
need to do import with an alias

pkg/routeagent-driver/handlers/kubeproxy-iptables/constants.go
mangelajo: kubeproxy_iptables?

pkg/routeagent-driver/cni_interface/cni_iface.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to update rp_filter for cni_interface %q, err: %s", iface, err)
+	} else {
+		klog.V(log.DEBUG).Infof("Successfully configured rp_filter to loose mode(2) on cniInterface %q", iface)
+	}
+
+	return nil
+}
+
+func AnnotateNodeWithCNIInterfaceIP(nodeName, cniAnnotation string, clientSet kubernetes.Interface, clusterCidr []string) error {
```
mangelajo, line 72:
```suggestion
func AnnotateNodeWithCNIInterfaceIP(nodeName, clientSet kubernetes.Interface, clusterCidr []string) error {
```

pkg/routeagent-driver/cni_interface/cni_iface.go
```diff
@@ -0,0 +1,101 @@
...
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		node, err := clientSet.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("unable to get node info for node %v, err: %s", nodeName, err)
+		} else {
+			annotations := node.GetAnnotations()
+			if annotations == nil {
+				annotations = map[string]string{}
+			}
+			annotations[cniAnnotation] = cniIface.IPAddress
```
mangelajo, line 87:
```suggestion
			annotations[constants.CniInterfaceIp] = cniIface.IPAddress
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/cleanup_handling.go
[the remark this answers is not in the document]
tpantelis: ahh right - it doesn't allow hyphens in imports.

pkg/routeagent-driver/constants/constants.go
sridhargaddam: IMHO creating a package just for this env seems to be an overkill for me.
One option is to pass individual params to NewSyncHandler() API so that we can remove this structure from the constants package. 

Please let me know what you prefer?

sridhargaddam (commented):
Many thanks @tpantelis and @mangelajo for taking time and reviewing this longgg PR.

pkg/routeagent-driver/cni_interface/cni_iface.go
```diff
@@ -0,0 +1,101 @@
+package cni_interface
```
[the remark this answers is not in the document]
tpantelis: I see.  I still prefer just `cni`  so references read `cni.Interface`, `cni.Discover`, etc as opposed to `cni_interface.Interface` which seems a bit awkward.

pkg/routeagent-driver/handlers/kubeproxy-iptables/cleanup_handling.go
```diff
@@ -0,0 +1,38 @@
+package kubeproxy_iptables
```
tpantelis, line 1: The main point of my prior comment was that the package name and import path match, so both be `kubeproxy_iptables`.  Also, as Miguel mentioned, eliminating the hyphen in the import path avoids users having to alias.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,105 @@
...
+		return errors.Wrapf(err, "Unable to find the default interface on host: %s", kp.hostname)
+	}
+
+	cniIface, err := cni_interface.Discover(kp.localClusterCidr[0])
+	if err == nil {
+		// Configure CNI Specific changes
+		kp.cniIface = cniIface
+		err := cni_interface.ConfigureRpFilter(kp.cniIface.Name)
+		if err != nil {
+			return fmt.Errorf("ConfigureRpFilter returned error. %v", err)
```
tpantelis, line 87:
```suggestion
			return errors.Wrapf(err, "ConfigureRpFilter returned error")
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -0,0 +1,105 @@
...
+	} else {
+		// This is not a fatal error. Hostnetworking to remote cluster support will be broken
+		// but other use-cases can continue to work.
+		klog.Errorf("Error discovering the CNI interface %v", err)
+	}
+
+	// Create the necessary IPTable chains in the filter and nat tables.
+	err = kp.createIPTableChains()
+	if err != nil {
+		return fmt.Errorf("createIPTableChains returned error. %v", err)
```
tpantelis, line 98:
```suggestion
		return errors.Wrapf(err, "createIPTableChains returned error")
```

pkg/routeagent-driver/handlers/kubeproxy-iptables/vxlan_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"strconv"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("getVxlanVtepIPAddress", func() {
+	Describe("Function getVxlanVtepIPAddress tests", func() {
```
tpantelis, line 12:
```suggestion
var _ = Describe("Function getVxlanVtepIPAddress tests", func() {
```

tpantelis (changes requested):

sridhargaddam (commented):
Thanks @tpantelis

sridhargaddam:
@tpantelis @mangelajo While merging this PR, please don't Squash the commits. I tried to use individual commits for various use-cases - in future, it will help us to understand the changes we did for a specific use-case (for example: to support HostNetworking to RemoteCluster etc) if we have individual commits.

pkg/routeagent-driver/handlers/kubeproxy-iptables/endpoint_handler.go
mangelajo: > Yeah, I wanted to discuss with you on this.
>
> Naturally if the node was a gateway and transitions to NonGateway, it will be invoked. But what would happen for the following scenarios?


> If admin removes the label from active Gateway node, which translates to deletion of local Endpoint and there will NOT be a new active Gateway node. Will we invoke TransitionToNonGateway in such cases or not? If so, with what endpoint info as argument?

If we need the endpoint to be received we can add that logic into the controller (keeping the previous local endpoint)
 
> The Route-agent comes up (aka bootup) on a nonGateway Node, will TransitionToNonGateway() be called on that node?

Same, ... if you are missing logic that it's necessary on the controller, let's add it.

I think calling TransitionToNonGateway should only be called if we detect that the node is not a gateway at boot up (may be waiting for all endpoints to have synchronized and then sending the transition if we are not gateway..., we could have a timer, or we could improve the admiral watcher to let us know that initial sync finished)

pkg/routeagent_driver/handlers/kubeproxy_iptables/endpoint_handler.go
```diff
@@ -0,0 +1,187 @@
...
+func (kp *SyncHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
+	kp.syncHandlerMutex.Lock()
+	defer kp.syncHandlerMutex.Unlock()
+	kp.localCableDriver = endpoint.Spec.Backend
+
+	// We are on nonGateway node
+	if endpoint.Spec.Hostname != kp.hostname {
+		// If the node already has a vxLAN interface that points to an oldEndpoint
+		// (i.e., during gateway migration), delete it.
+		if kp.vxlanDevice != nil && kp.vxlanDevice.activeEndpointHostname != endpoint.Spec.Hostname {
```
mangelajo, line 22: ok to merge as long as we follow up moving this to use the TransitionToGateway/NonGw logic before release

pkg/routeagent_driver/handlers/kubeproxy_iptables/endpoint_handler.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (kp *SyncHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	kp.syncHandlerMutex.Lock()
+	defer kp.syncHandlerMutex.Unlock()
+	for _, inputCidrBlock := range endpoint.Spec.Subnets {
+		if !kp.remoteSubnets.Contains(inputCidrBlock) {
+			kp.remoteSubnets.Delete(inputCidrBlock)
+		}
+	}
+	// TODO: Handle a remote endpoint removal use-case
```
mangelajo, line 116: I'm my case (ovn) I'm reading the system state, and calculating the diff or rules to set.

mangelajo (approved):

mangelajo:
I think it's good to merge as long as we continue PRs to move all GW transition handling to use the event logic from the controller, even if we need to refine it.

sridhargaddam:
> I think it's good to merge as long as we continue PRs to move all GW transition handling to use the event logic from the controller, even if we need to refine it.

Thanks @mangelajo. Yes, lets continue to work on the GW transition handling as separate PRs.

pkg/routeagent-driver/cni_interface/cni_iface.go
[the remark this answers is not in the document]
tpantelis: Then that indicates to me that `Discover` is not the proper name for the function. IMO, `cni` and `cni.Interface ` imply the same meaning except the latter has stutter.  But if you guys feel strongly then we can leave as is.

tpantelis (approved):

PR #952, opened 2020-11-19
Open UDP/4800 ports between master and worker nodes on AWS

sridhargaddam:
Recently Submariner pods are configured as infrastructure pods
which will ignore any taints. Because of this change, they will
get scheduled even on the Master nodes. To validate this use-case
Submariner e2e test pods are also modified to tolerate any taints.

So, we should open UDP/4800 ports between the master and worker
nodes in an AWS cluster. This PR handles this change in terraform
scripts.

### DIFF
--- a/tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
+++ b/tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
@@ -13,6 +13,22 @@ data "aws_security_group" "worker_sg" {
   }
 }
 
+# Get security group associated with master nodes.
+data "aws_security_group" "master_sg" {
+  vpc_id = data.aws_vpc.env_vpc.id
+
+  filter {
+    name   = "tag:kubernetes.io/cluster/${var.cluster_id}"
+    values = ["owned"]
+  }
+
+  filter {
+    name   = "tag:Name"
+    values = ["${var.cluster_id}-master-sg"]
+  }
+}
+
+
 # Add a rule for vxlan traffic for all workers.
 resource "aws_security_group_rule" "worker_sg_vxlan_rule" {
   security_group_id        = data.aws_security_group.worker_sg.id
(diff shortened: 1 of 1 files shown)


nyechiel (approved):
LGTM, thanks @sridhargaddam

dfarrell07 (approved):

tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
```diff
@@ -23,6 +39,27 @@ resource "aws_security_group_rule" "worker_sg_vxlan_rule" {
   type                     = "ingress"
 }
 
+# Add a rule for vxlan traffic from master nodes to worker nodes.
```
mangelajo, line 42: We also need to add master node to master node, like here:

https://github.com/submariner-io/submariner/blob/master/tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf#L17

mangelajo (changes requested):

tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
mangelajo: Oh, as talked in slack, in the existing IPI topology, the gateway node won't ever be part of the master SG.
and 4800 is used for node to GW communication.

mangelajo (approved):

manosnoam (commented):
A general question - isn't this fix relevant only for E2E execution, but not for Submariner functionality ? 
Shouldn't we fix it on the E2E test flow, instead of in ocp-ipi-aws ?

sridhargaddam:
> A general question - isn't this fix relevant only for E2E execution, but not for Submariner functionality ?
> Shouldn't we fix it on the E2E test flow, instead of in ocp-ipi-aws ?

The problem was that our e2e pods are now tolerating all taints (i.e., they will even get scheduled on Master nodes when we run e2e tests). So we should open the necessary UDP/4800 ports between Master/worker nodes. We don't need any additional changes to our e2e test flow.

PR #955, opened 2020-11-19
Change the statistics type from float64 to uint64

aswinsuryan:
The code generator does not support float64 and hence updating
the API to use uint64

### DIFF
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -86,11 +86,11 @@ type GatewayStatus struct {
 // LatencySpec describes the round trip time information in miliseconds for a packet
 // between the gateway pods of two clusters.
 type LatencySpec struct {
-	LastRTT    float64 `json:"lastRTT"`
-	MinRTT     float64 `json:"minRTT"`
-	AverageRTT float64 `json:"averageRTT"`
-	MaxRTT     float64 `json:"maxRTT"`
-	StdDevRTT  float64 `json:"stddevRTT"`
+	LastRTT    *float64 `json:"lastRTT"`
+	MinRTT     *float64 `json:"minRTT"`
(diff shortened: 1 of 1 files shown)


tpantelis:
Can you provide more detail? What code generator and how did it fail?

aswinsuryan:
> Can you provide more detail? What code generator and how did it fail?

cd vendor/github.com/submariner-io/submariner && controller-gen "crd:crdVersions=v1,trivialVersions=false" paths="./..." output:crd:artifacts:config=../../../../deploy/submariner/crds
/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/submariner/pkg/apis/submariner.io/v1/types.go:89:13: unsupported type "float64"

This is the error it shows.

SteveMattar (approved):

tpantelis:
Discussed  on chat - we should just change them to uint64.

aswinsuryan:
> Discussed on chat - we should just change them to uint64.
done.

SteveMattar (approved):

tpantelis (approved):

pkg/apis/submariner.io/v1/types.go
```diff
@@ -86,11 +86,11 @@ type GatewayStatus struct {
 // LatencySpec describes the round trip time information in miliseconds for a packet
```
skitt, line 86: To preserve the granularity, shouldn’t the values use a finer unit, say microseconds or nanoseconds?

skitt (changes requested):

pkg/apis/submariner.io/v1/types.go
aswinsuryan: Ok the time period gives it in nano seconds, I will change it to nano seconds.

skitt (approved):

Issue #960, opened 2020-11-23
Avoid deleting local endpoint when Engine pod is restarted


sridhargaddam:
**What happened**:
Currently its seen that when we restart the EnginePod we get two events.
1. Deletion of localEndpoint
2. Creation of localEndpoint with the same params.

We should ideally not delete the localEndpoint when the params are exactly the same.

sridhargaddam:
The necessary changes are merged.

PR #961, opened 2020-11-23
Avoid deleting local endpoint when Engine pod is restarted

sridhargaddam:
In datastoresyncer, ensureExclusiveEndpoint API when the local endpoints
are compared, they were not matching due to small mismatch in BackendConfig
and this was causing deletion of local endpoint followed by creation of
local endpoint.

Fixes issue: https://github.com/submariner-io/submariner/issues/960

### DIFF
--- a/main.go
+++ b/main.go
@@ -108,7 +108,7 @@ func main() {
 
 	submSpec.CableDriver = strings.ToLower(submSpec.CableDriver)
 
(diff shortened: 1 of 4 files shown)


mangelajo (approved):
Thank you for handling this bugfix

tpantelis (approved):

PR #962, opened 2020-11-23
Extend the generic routeagent handler to the other drivers

mangelajo:
This will allow for all those network plugins to be recognized with this same
route agent driver, and at the same time we could make specific modification
if this was ever required.

### DIFF
--- a/pkg/networkplugin-syncer/main.go
+++ b/pkg/networkplugin-syncer/main.go
@@ -26,7 +26,13 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	registry := event.NewRegistry("networkplugin-syncer", os.Getenv("NETWORK_PLUGIN"))
+	networkPlugin := os.Getenv("SUBMARINER_NETWORKPLUGIN")
+
+	if networkPlugin == "" {
+		networkPlugin = "generic"
+	}
+
+	registry := event.NewRegistry("networkplugin-syncer", networkPlugin)
 	if err := registry.AddHandlers(logger.NewHandler(), ovn.NewSyncHandler()); err != nil {
 		klog.Fatalf("Error registering the handlers: %s", err.Error())
 	}
--- a/pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
+++ b/pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
@@ -1,26 +1,34 @@
 package kp_iptables
 
 import (
+	"fmt"
+	"os"
+
 	"k8s.io/klog"
 
-	"github.com/submariner-io/admiral/pkg/log"
-	k8sV1 "k8s.io/api/core/v1"
-
-	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/event"
+	"github.com/submariner-io/submariner/pkg/routeagent-driver/cni_interface"
+	"github.com/submariner-io/submariner/pkg/routeagent-driver/constants"
 )
 
 type SyncHandler struct {
 	event.HandlerBase
 	clusterID        string
+	namespace        string
 	localClusterCidr []string
 	localServiceCidr []string
 
-	cniIface *cniInterface
+	hostname string
+	cniIface *cni_interface.CniInterface
 }
 
-func NewSyncHandler() *SyncHandler {
-	return &SyncHandler{}
+func NewSyncHandler(env constants.Specification) *SyncHandler {
+	return &SyncHandler{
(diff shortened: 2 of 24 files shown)


pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -44,31 +44,35 @@ type SyncHandler struct {
...
-		remoteVTEPs:          util.NewStringSet(),
-		routeCacheGWNode:     util.NewStringSet(),
-		syncHandlerMutex:     &sync.Mutex{},
-		isGatewayNode:        false,
-		wasGatewayPreviously: false,
-		vxlanDevice:          nil,
-		vxlanGwIP:            nil,
-		smClientSet:          smClientSet,
-	}
+func NewSyncHandler(env constants.Specification, smClientSet clientset.Interface) event.Handler {
```
sridhargaddam, line 47: As discussed over Slack, IMHO GetNetworkPlugin -> GetNetworkPlugins() might be more simpler.
Also, xfrm specific fixes are required even in the Generic driver (as we should not delete the xfrm rules which are not programmed by route-agent).

CC: @tpantelis @skitt

mangelajo:
@sridhargaddam I will remake with that simpler solution

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
mangelajo: done

mangelajo:
Oh, @sridhargaddam @tpantelis ah, we are reading the network plugin from the wrong env var, it must be "SUBMARINER_NETWORKPLUGIN" ... changing that.

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
```diff
@@ -66,8 +66,8 @@ func (kp *SyncHandler) GetName() string {
 	return "kubeproxy-iptables-handler"
 }
 
-func (kp *SyncHandler) GetNetworkPlugin() string {
-	return event.AnyNetworkPlugin
+func (kp *SyncHandler) GetNetworkPlugins() []string {
+	return []string{"generic", "canal-flannel", "weave-net", "OpenShiftSDN"}
```
sridhargaddam, line 70: "flannel"?

sridhargaddam (approved):
LGTM, thanks @mangelajo

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
mangelajo: We need to use what our discovery mechanism provides today in the operator, which is canal-flannel

https://github.com/submariner-io/submariner-operator/blob/master/pkg/discovery/network/canal.go#L51

In an ideal world we would fetch the constants from there, but... we cannot make submariner depend on the operator.

We could define constants here and use them from the operator instead.

mangelajo:
@skitt do you have any idea of why could the update be failing?

pkg/routeagent-driver/handlers/kubeproxy-iptables/kp-iptables.go
tpantelis: We could also define such constants in admiral.

pkg/networkplugin-syncer/main.go
```diff
@@ -26,7 +26,13 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	registry := event.NewRegistry("networkplugin-syncer", os.Getenv("NETWORK_PLUGIN"))
+	networkPlugin := os.Getenv("SUBMARINER_NETWORKPLUGIN")
```
tpantelis, line 29: As a side note, we can make this `NETWORK_PLUGIN` is the operator. There was a recent discussion about `NODE_NAME` and whether to use the `SUBMARINER_`.  We should be consistent. @sridhargaddam

mangelajo: it's fine as SUBMARINER_NETWORKPLUGIN too, it's the network plugin we are instantiating in the different submariner components, 

I'd leave it as it is for now give the time restrictions we have.

sridhargaddam: W.r.t NODE_NAME, I wanted to avoid SUBMARINER as its not something related to SUBMARINER but is OS/Host information. Regarding NETWORK_PLUGIN, I'm fine with anything. AFAIK, it will not be parsed along (or at the same place) where other SUBMARINER_* env variables are parsed/used. So, using NETWORK_PLUGIN might make more sense.

mangelajo:
This patch is now part of: https://github.com/submariner-io/submariner/pull/971

PR #965, opened 2020-11-24
Switch to Fedora for all images

skitt:
... so we can build ARM images.

This needs https://github.com/submariner-io/submariner/pull/964

### DIFF
--- a/package/Dockerfile.submariner-globalnet
+++ b/package/Dockerfile.submariner-globalnet
@@ -1,11 +1,11 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM fedora:33
 ARG TARGETPLATFORM
 
 WORKDIR /var/submariner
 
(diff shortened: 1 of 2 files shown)


mangelajo (approved):

dfarrell07 (approved):

mangelajo:
If we will be doing backports I guess this would be fine to merge

tpantelis (approved):

mkolesnik:
Why isn't `ubi-minimal` enough? It also seems to support `arm64`

```
[mkolesni@localhost ~]$ docker manifest inspect --verbose registry.access.redhat.com/ubi8/ubi-minimal:8.3 | grep architecture
				"architecture": "amd64",
				"architecture": "arm64",
				"architecture": "ppc64le",
				"architecture": "s390x",
```

Or am I missing something?

PR #971, opened 2020-11-25
Ovn routeagent handler

mangelajo:
OVN route-agent handler
    
This handler manages the ip routing rules & routes on the submariner gateway,
as well as the necessary FORWARD iptables.
   
Extend the generic routeagent handler to the other drivers
     
This will allow for all those network plugins to be recognized with this same
route agent driver, and at the same time we could make specific modification
if this was ever required.
    
### DIFF
--- a/pkg/networkplugin-syncer/handlers/ovn/infrastructure.go
+++ b/pkg/networkplugin-syncer/handlers/ovn/infrastructure.go
@@ -32,14 +32,15 @@ func (ovn *SyncHandler) ensureSubmarinerInfra() error {
 }
 
 func (ovn *SyncHandler) ensureSubmarinerJoinSwitch() error {
+	klog.Infof("Ensuring %q switch", submarinerDownstreamSwitch)
 	lsCmd, err := ovn.nbdb.LSAdd(submarinerDownstreamSwitch)
 	if err == nil {
-		klog.Infof("error creating submariner switch %q", submarinerDownstreamSwitch)
+		klog.Infof("Creating submariner join switch %q", submarinerDownstreamSwitch)
 
 		err = ovn.nbdb.Execute(lsCmd)
 	}
 
-	if !errors.Is(err, goovn.ErrorExist) {
+	if err != nil && !errors.Is(err, goovn.ErrorExist) {
 		return errors.Wrapf(err, "error creating submariner switch %q", submarinerDownstreamSwitch)
 	}
 
@@ -47,6 +48,7 @@ func (ovn *SyncHandler) ensureSubmarinerJoinSwitch() error {
 }
 
 func (ovn *SyncHandler) ensureSubmarinerRouter() error {
+	klog.Infof("Ensuring %q", submarinerLogicalRouter)
 	lrCmd, err := ovn.nbdb.LRAdd(submarinerLogicalRouter, nil)
 	if err == nil {
 		klog.Infof("Creating submariner router %q", submarinerLogicalRouter)
@@ -54,7 +56,7 @@ func (ovn *SyncHandler) ensureSubmarinerRouter() error {
 		err = ovn.nbdb.Execute(lrCmd)
 	}
 
-	if !errors.Is(err, goovn.ErrorExist) {
+	if err != nil && !errors.Is(err, goovn.ErrorExist) {
 		return errors.Wrapf(err, "error creating submariner router %q", submarinerLogicalRouter)
 	}
 
@@ -65,15 +67,15 @@ func (ovn *SyncHandler) ensureSubmarinerRouter() error {
 	delOldLSP, _ := ovn.nbdb.LSPDel(submarinerDownstreamSwPort)
 	_ = ovn.nbdb.Execute(delOldLSP)
 
-	linkCmd, _ := ovn.nbdb.LinkSwitchToRouter(
+	linkCmd, err := ovn.nbdb.LinkSwitchToRouter(
 		submarinerDownstreamSwitch, submarinerDownstreamSwPort,
 		submarinerLogicalRouter, submarinerDownstreamRPort,
 		submarinerDownstreamMAC,
 		[]string{submarinerDownstreamNET}, nil,
 	)
 
-	if err != nil {
-		return errors.Wrapf(err, "error link for port %q", submarinerDownstreamRPort)
+	if err != nil && !errors.Is(err, goovn.ErrorExist) {
+		return errors.Wrapf(err, "error link for port %q, linkCmd: %#v", submarinerDownstreamRPort, linkCmd)
 	}
 
 	err = ovn.nbdb.Execute(linkCmd)
--- a/pkg/networkplugin-syncer/handlers/ovn/infrastructure.go
+++ b/pkg/networkplugin-syncer/handlers/ovn/infrastructure.go
@@ -72,6 +72,10 @@ func (ovn *SyncHandler) ensureSubmarinerRouter() error {
 		[]string{submarinerDownstreamNET}, nil,
 	)
 
+	if err != nil {
+		return errors.Wrapf(err, "error link for port %q", submarinerDownstreamRPort)
+	}
+
 	err = ovn.nbdb.Execute(linkCmd)
 	if err != nil {
 		return errors.Wrapf(err, "error creating %q port %q", submarinerLogicalRouter, submarinerDownstreamRPort)
--- a/pkg/networkplugin-syncer/main.go
+++ b/pkg/networkplugin-syncer/main.go
@@ -30,7 +30,13 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	registry := event.NewRegistry("networkplugin-syncer", os.Getenv("NETWORK_PLUGIN"))
+	networkPlugin := os.Getenv("SUBMARINER_NETWORKPLUGIN")
+
+	if networkPlugin == "" {
+		networkPlugin = "generic"
+	}
+
+	registry := event.NewRegistry("networkplugin-syncer", networkPlugin)
 	if err := registry.AddHandlers(logger.NewHandler(), ovn.NewSyncHandler(getK8sClient())); err != nil {
 		klog.Fatalf("Error registering the handlers: %s", err.Error())
 	}
--- a/pkg/routeagent_driver/handlers/kubeproxy_iptables/constants.go
+++ b/pkg/routeagent_driver/handlers/kubeproxy_iptables/constants.go
@@ -32,7 +32,7 @@ const (
 	// To support connectivity for Pods with HostNetworking on the GatewayNode, we program
 	// certain routing rules in table 150. As part of these routes, we set the source-ip of
 	// the egress traffic to the corresponding CNIInterfaceIP on that host.
-	RouteAgentHostNetworkTableID = 150
+
 )
 
 type Operation int
--- a/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
@@ -1,31 +1,46 @@
 package ovn
 
 import (
-	"fmt"
 	"net"
 	"os"
 
 	"github.com/coreos/go-iptables/iptables"
 	"github.com/pkg/errors"
 	"github.com/vishvananda/netlink"
+	"k8s.io/klog"
 
+	npSyncerOvn "github.com/submariner-io/submariner/pkg/networkplugin-syncer/handlers/ovn"
+	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	"github.com/submariner-io/submariner/pkg/util"
 )
 
 func (ovn *Handler) cleanupGatewayDataplane() error {
 	currentRemoteSubnets, err := ovn.getExistingRuleSubnets()
 	if err != nil {
-		return errors.Wrapf(err, "error reading ip rule list for IPv4 in TransitionToNonGateway")
+		return errors.Wrapf(err, "error reading ip rule list for IPv4")
 	}
 
-	err = ovn.handleSubnets(currentRemoteSubnets.Elements(), netlink.RuleDel, os.IsNotExist, "TransitionToNonGateway")
+	err = ovn.handleSubnets(currentRemoteSubnets.Elements(), netlink.RuleDel, os.IsNotExist)
 	if err != nil {
 		return errors.Wrapf(err, "error removing routing rule")
 	}
 
 	err = netlink.RouteDel(getSubmDefaultRoute())
 	if err != nil && !os.IsNotExist(err) {
-		return errors.Wrap(err, "error deleting submariner default route during TransitionToNonGateway")
+		return errors.Wrap(err, "error deleting submariner default route")
+	}
+
+	for _, handler := range ovn.cleanupHandlers {
+		// TODO: Make those handlers have a single cleanup on GatewayToNonGateway transition,
+		//       we can make them a normal event handler, instead of a cleanup handler, but
+		//       we need to make sure that TransitionToNonGateway is fired at least one for
+		//       a controller which is started once and no endpoint is received on for current
+		//       host.
+		if err := handler.NonGatewayCleanup(); err != nil {
+			return errors.Wrap(err, "NonGatewayCleanup")
+		} else if err := handler.GatewayToNonGatewayTransition(); err != nil {
+			return errors.Wrap(err, "GatewayToNonGatewayTransition")
+		}
 	}
 
 	return ovn.cleanupForwardingIptables()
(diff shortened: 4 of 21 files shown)


pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -0,0 +1,105 @@
...
+		return errors.Wrapf(err, "error reading ip rule list for IPv4 in TransitionToGateway")
+	}
+
+	endpointSubnets := ovn.getRemoteSubnets()
+
+	toAdd := currentRuleRemotes.Difference(endpointSubnets)
+
+	err = ovn.handleSubnets(toAdd, netlink.RuleAdd, os.IsExist, "TransitionToGateway")
+	if err != nil {
+		return errors.Wrapf(err, "error adding routing rule")
```
sridhargaddam, line 46: It may be helpful to print the toAdd rules here.

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	toRemove := endpointSubnets.Difference(currentRuleRemotes)
+
+	err = ovn.handleSubnets(toRemove, netlink.RuleDel, os.IsNotExist, "TransitionToGateway")
+	if err != nil {
+		return errors.Wrapf(err, "error removing routing rule")
```
sridhargaddam, line 53: same as above

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	_ = ipt.Delete("filter", "FORWARD", "-i", "ovn-k8s-gw0", "-o", ovn.cableRoutingInterface.Name, "-j", "ACCEPT")
+	_ = ipt.Delete("filter", "FORWARD", "-i", ovn.cableRoutingInterface.Name, "-o", "ovn-k8s-gw0", "-j", "ACCEPT")
+
+	return nil
+}
+
+// TODO: Use the constant defined in networkplugin-syncer handler when that's merged.
+const submarinerUpstreamIP = "169.254.33.7"
```
sridhargaddam, line 98: Just for my understanding: Is this some hardcoded IP in OVN?

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -0,0 +1,105 @@
...
+	ruleSpec := []string{"-i", "ovn-k8s-gw0", "-o", ovn.cableRoutingInterface.Name, "-j", "ACCEPT"}
+
+	if err = util.PrependUnique(ipt, "filter", "FORWARD", ruleSpec); err != nil {
+		return fmt.Errorf("unable to insert iptable rule in filter table to forward submariner traffic: %v", err)
+	}
+
+	ruleSpec = []string{"-i", ovn.cableRoutingInterface.Name, "-o", "ovn-k8s-gw0", "-j", "ACCEPT"}
+
+	if err = util.PrependUnique(ipt, "filter", "FORWARD", ruleSpec); err != nil {
+		return fmt.Errorf("unable to insert iptable rule in filter table to allow vxlan traffic: %v", err)
```
sridhargaddam, line 79: Is it vxlan traffic? or was it a typo in the log message?

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+func (ovn *Handler) cleanupForwardingIptables() error {
+	ipt, err := iptables.New()
+	if err != nil {
+		return fmt.Errorf("error initializing iptables: %v", err)
+	}
+
+	_ = ipt.Delete("filter", "FORWARD", "-i", "ovn-k8s-gw0", "-o", ovn.cableRoutingInterface.Name, "-j", "ACCEPT")
```
sridhargaddam, line 91: Can we safely ignore the errors or do we want to retry?

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -0,0 +1,155 @@
...
+func (ovn *Handler) Init() error {
+	var err error
+
+	ovn.cableRoutingInterface, err = util.GetDefaultGatewayInterface()
+	if err != nil {
+		klog.Fatalf("Unable to find the default interface on host: %s", err.Error())
+	}
+
+	// For now we get all the cleanups
+	ovn.cleanupHandlers = cableCleanup.GetCleanupHandlers()
```
sridhargaddam, line 56: Looks like we are not calling the cleanupHandlers anywhere?

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -0,0 +1,155 @@
...
+	ovn.mutex.Lock()
+	defer ovn.mutex.Unlock()
+
+	ovn.localEndpoint = endpoint
+
+	// TODO: this logic belongs to the cabledrivers instead
+	if endpoint.Spec.Backend == "wireguard" {
+		//NOTE: This assumes that LocalEndpointCreated happens before than TransitionToGatewayNode
+		if wg, err := net.InterfaceByName(wireguard.DefaultDeviceName); err == nil {
+			ovn.cableRoutingInterface = wg
```
sridhargaddam, line 71: We are overriding the interface we discovered in the Init method. Can you please explain the significance of cableRoutingInterface? Is it related to cniInterface?

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return nil
+}
+
+func (ovn *Handler) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
+	ovn.mutex.Lock()
+	defer ovn.mutex.Unlock()
+
+	if ovn.localEndpoint.Name == endpoint.Name {
+		ovn.localEndpoint = nil
```
sridhargaddam, line 94: no need to handle any cleanups?

pkg/routeagent_driver/handlers/ovn/south_rules.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"net"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/stringset"
+	"github.com/vishvananda/netlink"
+)
+
+const ovnRoutingRulesTable = 5
```
sridhargaddam, line 11: As you know, in Route-agent we are using routing table 150 and Strongswan uses routing table of 220. Can we use the same table as route-agent (i.e., 150) for OVN?

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
mangelajo: oh yes, copy & paste error ;D thanks

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
mangelajo: I didn't find a way of making a difference between "rule does not exist" or "other error", may be there is?.

The alternative was listing before, but that's expensive.

At worst case (the rules are left on the system) I believe those won't cause any specific issues. But wanted to make sure that we at least clean them up.

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
mangelajo: Yes, I have that patch now in tree, so I can use that constant.

the 169.254.33.0/24 is a range that OVNKubernetes uses for a ovs switch that exists on every node, on that switch, there is an interface from the host "ovn-k8s-gw0" 169.254.33.2 so the host has an ip on such network.

And also there is another interface from OVNKubernetes routers (one per host) 169.254.33.1.

Now, there is another interface we add in that network, which is the submariner upstream router leg 169.254.33.7

pkg/routeagent_driver/handlers/ovn/handler.go
mangelajo: oh right, thanks! that part is missing.

pkg/routeagent_driver/handlers/ovn/south_rules.go
mangelajo: makes sense, let's make that a constant somewhere in the code.

pkg/routeagent_driver/handlers/ovn/handler.go
mangelajo: It's the interface to which we forward to/from traffic for the remote clusters, it's used here:

https://github.com/submariner-io/submariner/pull/971/files#diff-f45ea0a8359161724c3b2f14cbfffa677d410ddf4a2856be29f42e1eb705b3e4R70

I will refactor this to make it clearer, when this event happens and we don't have an existing interface detected, I will make the detection. Also, only once, not multiple times like we would be doing now for wireguard.

pkg/routeagent_driver/main.go
```diff
@@ -60,9 +55,17 @@ func main() {
 		klog.Errorf("Error while annotating the node: %s", err.Error())
 	}
 
-	registry := event.NewRegistry("routeagent_driver", os.Getenv("NETWORK_PLUGIN"))
-	if err := registry.AddHandlers(logger.NewHandler(),
-		kp_iptables.NewSyncHandler(env.ClusterCidr, env.ServiceCidr, smClientset)); err != nil {
+	np := os.Getenv("SUBMARINER_NETWORKPLUGIN")
```
vthapar, line 58: Not specific to this review, but we should put these constants in a common file which can be reused in operator to avoid mismatch.

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
mangelajo: We could also free the 240/8 range for OVN, we could use that, although I think it's better not to use that range.

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
mangelajo: ok, I need to figure out the difference between "it doesn't exist", and other errors, I'll check that

pkg/routeagent_driver/main.go
mangelajo: yup, may be in admiral. Sounds like a good idea.

sridhargaddam (approved):

vthapar (approved):

pkg/networkplugin-syncer/handlers/ovn/infrastructure.go
```diff
@@ -72,6 +72,10 @@ func (ovn *SyncHandler) ensureSubmarinerRouter() error {
 		[]string{submarinerDownstreamNET}, nil,
 	)
 
+	if err != nil {
+		return errors.Wrapf(err, "error link for port %q", submarinerDownstreamRPort)
+	}
```
tpantelis, line 77: The `err` referenced here pertains to a prior operation. Is this intended be above L63? Or is this intended to check the error return from L68 (which is currently ignored)?.

pkg/networkplugin-syncer/main.go
```diff
@@ -30,7 +30,13 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	registry := event.NewRegistry("networkplugin-syncer", os.Getenv("NETWORK_PLUGIN"))
+	networkPlugin := os.Getenv("SUBMARINER_NETWORKPLUGIN")
+
+	if networkPlugin == "" {
+		networkPlugin = "generic"
```
tpantelis, line 36: What does the "generic" plugin name refer to?  Should this be a constant somewhere?

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -0,0 +1,139 @@
...
+
+	npSyncerOvn "github.com/submariner-io/submariner/pkg/networkplugin-syncer/handlers/ovn"
+	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
+	"github.com/submariner-io/submariner/pkg/util"
+)
+
+func (ovn *Handler) cleanupGatewayDataplane() error {
+	currentRemoteSubnets, err := ovn.getExistingRuleSubnets()
+	if err != nil {
+		return errors.Wrapf(err, "error reading ip rule list for IPv4")
```
tpantelis, line 20: Trying to get the error message and method name to match more...

```suggestion
	currentRemoteSubnets, err := ovn.getExistingIPv4RuleSubnets()
	if err != nil {
		return errors.Wrapf(err, "error reading IP rule subnets for IPv4")
```

pkg/routeagent_driver/handlers/ovn/iptables.go
```diff
@@ -0,0 +1 @@
+package ovn
```
tpantelis, line 1: This is an empty file.

pkg/routeagent_driver/handlers/ovn/south_rules.go
```diff
@@ -0,0 +1,67 @@
...
+
+// handleSubnets builds ip rules, and passes them to the specified netlink function
+//               for provided subnet list
+func (ovn *Handler) handleSubnets(subnets []string, ruleFunc func(rule *netlink.Rule) error,
+	ignoredErrorFunc func(error) bool) error {
+	for _, subnetToHandle := range subnets {
+		for _, localSubnet := range ovn.localEndpoint.Spec.Subnets {
+			rule, err := ovn.ruleForSouthTraffic(localSubnet, subnetToHandle)
+			if err != nil {
+				return errors.Wrapf(err, "creating rule %#v", rule)
```
tpantelis, line 21:
```suggestion
				return errors.Wrapf(err, "error creating rule %#v", rule)
```

pkg/routeagent_driver/handlers/ovn/south_rules.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+
+			err = ruleFunc(rule)
+			if err != nil && !ignoredErrorFunc(err) {
+				return errors.Wrapf(err, "handling rule %#v with dst=%q src=%q", rule, rule.Dst.String(), rule.Src.String())
```
tpantelis, line 26:
```suggestion
				return errors.Wrapf(err, "error handling rule %#v", rule, rule.Dst.String(), rule.Src.String())
```

pkg/networkplugin-syncer/handlers/ovn/infrastructure.go
mangelajo: oops!, I was trying to add error handling to the linkCmd generation, I'll fix that.

pkg/networkplugin-syncer/main.go
mangelajo: Yes, let's make constants for the network plugins on admiral on a 2nd pass, as for OVNKubernetes, etc...

mangelajo: generic is means kubeproxy-iptables based, but not identified. It will trigger the kubeproxy driver as default

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
mangelajo: Done.

pkg/routeagent_driver/handlers/ovn/iptables.go
mangelajo: Opps, deleted.

pkg/routeagent_driver/handlers/ovn/south_rules.go
mangelajo: Done.

pkg/networkplugin-syncer/handlers/ovn/infrastructure.go
mangelajo: Done.

tpantelis (approved):

pkg/routeagent_driver/handlers/kubeproxy_iptables/constants.go
```diff
@@ -32,7 +32,7 @@ const (
 	// To support connectivity for Pods with HostNetworking on the GatewayNode, we program
```
dfarrell07, line 32: Should this comment move with the variable?

mangelajo: Oh it should! good catch!

mangelajo: I will address this on separate PR if that's ok.

PR #975, opened 2020-11-26
Fix issue with HealthChecker

aswinsuryan:
The pinger is calling onFinish at frequent intervals even
when there is no packet loss. Now the logic is updated to
mark error state for the connection only if there 100% packet
loss

### DIFF
--- a/pkg/cableengine/healthchecker/pinger.go
+++ b/pkg/cableengine/healthchecker/pinger.go
@@ -11,6 +11,8 @@ import (
 var waitTime time.Duration = 15 * time.Second
 var timeout = 3 * time.Second
 
+const allpackets = 100
+
 // The RTT will be stored and will be used to calculate the statistics until
 // the size is reached. Once the size is reached the array will be reset and
 // the last elements will be added to the array for statistics.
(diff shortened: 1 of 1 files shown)


mangelajo (approved):

mangelajo (approved):
This is still creating a pinger every 15 seconds. Then pings for 3 seconds.

I think it's better if we keep the pinger alive forever, and detect ourselves that we haven't receive the ping response after X seconds.

This will mark a failing link as late as 15 + 3 sec  while we could probably detect it in 3 sec.

I think it's ok for a temporary workaround now, but not a final solution.

vthapar (approved):

aswinsuryan:
> This is still creating a pinger every 15 seconds. Then pings for 3 seconds.
> 
> I think it's better if we keep the pinger alive forever, and detect ourselves that we haven't receive the ping response after X seconds.
> 
> This will mark a failing link as late as 15 + 3 sec while we could probably detect it in 3 sec.
> 
> I think it's ok for a temporary workaround now, but not a final solution.

Yes we need fix it , I will work on it. It is expected that if set Count to -1 it should be indefinite. But  there seems to be an issue.

PR #977, opened 2020-11-30
Make Healthchecker ping continuous

aswinsuryan:
When the Timeout value is set in the pinger, it is calcuated from the
time pinger starts and not the time interval between send and receive.
So now we are now using a threshold of packet count after which the connection
will be marked as an error, if the packet loss continues to be 100%.

### DIFF
--- a/pkg/cableengine/healthchecker/pinger.go
+++ b/pkg/cableengine/healthchecker/pinger.go
@@ -11,7 +11,6 @@ import (
 var waitTime = 15 * time.Second
 
 const threshold = 5
-const allpackets = 100
 
 // The RTT will be stored and will be used to calculate the statistics until
 // the size is reached. Once the size is reached the array will be reset and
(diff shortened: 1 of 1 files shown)


pkg/cableengine/healthchecker/pinger.go
```diff
@@ -63,8 +63,16 @@ func (p *pingerInfo) sendPing() {
...
 	pinger.RecordRtts = false
-	// After 3 seconds stop waiting.
-	pinger.Timeout = timeout
+
+	pinger.OnSend = func(packet *ping.Packet) {
+		// Pinger will be stopped once the packet loss reaches threshold and
+		// if packet loss is not 100% recreates the Pinger without marking an error
+		if pinger.Statistics().PacketsSent-pinger.Statistics().PacketsRecv > threshold {
+			klog.Infof("Packet loss reached a threshold of %d and hence stopping the pinger for HealthCheckIP: %q",
+				threshold, p.healthCheckIP)
```
mangelajo, line 72: But, why would we stop it?.

I believe we should keep it up until connection is back, or the cable connection is removed.

mangelajo (changes requested):

pkg/cableengine/healthchecker/pinger.go

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -8,10 +8,9 @@ import (
 	"k8s.io/klog"
 )
 
-var waitTime time.Duration = 15 * time.Second
-var timeout = 3 * time.Second
+var waitTime = 15 * time.Second
 
-const allpackets = 100
+const threshold = 5
```
dfarrell07, line 13: Maybe a better name would be something like `maxLostPackets` or `maxLosablePackets`?

mangelajo (approved):

dfarrell07 (approved):

PR #978, opened 2020-11-30
Record connection latency metric

maayanf24:
Related issue #678

### DIFF
--- a/pkg/cableengine/syncer/syncer.go
+++ b/pkg/cableengine/syncer/syncer.go
@@ -210,12 +210,14 @@ func (i *GatewaySyncer) generateGatewayObject() *v1.Gateway {
 				latencyInfo := i.healthCheck.GetLatencyInfo(&connection.Endpoint)
 				if latencyInfo != nil {
 					connection.LatencyRTT = latencyInfo.Spec
-					lastRTT, _ := time.ParseDuration(latencyInfo.Spec.Last)
-					cable.RecordConnectionLatency(localEndpoint.Spec.CableName, &localEndpoint.Spec, &connection.Endpoint, lastRTT.Seconds())
+					if connection.Status == v1.Connected {
+						lastRTT, _ := time.ParseDuration(latencyInfo.Spec.Last)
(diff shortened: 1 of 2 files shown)


pkg/cableengine/syncer/syncer.go
```diff
@@ -209,6 +210,9 @@ func (i *GatewaySyncer) generateGatewayObject() *v1.Gateway {
 				latencyInfo := i.healthCheck.GetLatencyInfo(&connection.Endpoint)
 				if latencyInfo != nil {
 					connection.Latency = latencyInfo.Spec
+					cable.RecordConnectionLatency(localEndpoint.Spec.CableName, &localEndpoint.Spec, &connection.Endpoint,
+						float64(latencyInfo.Spec.AverageRTT/uint64(time.Second)))
```
mangelajo, line 214: In the case of prometheus, I think it'd be better to expose the LastRTT, in that way prometheus can be queried for average over specific times with it's query language.

mangelajo (commented):
AverageRTT vs LastRTT

dfarrell07 (approved):

pkg/cableengine/syncer/syncer.go
```diff
@@ -209,6 +210,9 @@ func (i *GatewaySyncer) generateGatewayObject() *v1.Gateway {
 				latencyInfo := i.healthCheck.GetLatencyInfo(&connection.Endpoint)
 				if latencyInfo != nil {
 					connection.Latency = latencyInfo.Spec
+					cable.RecordConnectionLatency(localEndpoint.Spec.CableName, &localEndpoint.Spec, &connection.Endpoint,
+						float64(latencyInfo.Spec.LastRTT/uint64(time.Second)))
```
tpantelis, line 214: Shouldn't we only record this if connected (ie `connection.Status == v1. Connected`)?

skitt:
This will need to be squashed when merged.

mangelajo (approved):

mangelajo (approved):

PR #987, opened 2020-12-02
Use health check config values

aswinsuryan:

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -14,17 +14,20 @@ type SubmarinerEndpoint struct {
 }
 
 type SubmarinerSpecification struct {
-	ClusterCidr []string
-	ColorCodes  []string
-	GlobalCidr  []string
-	ServiceCidr []string
-	Broker      string
-	CableDriver string
-	ClusterID   string
-	Namespace   string
-	Token       string
-	Debug       bool
-	NatEnabled  bool
+	ClusterCidr                   []string
+	ColorCodes                    []string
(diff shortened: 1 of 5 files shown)


pkg/types/types.go
```diff
@@ -14,17 +14,20 @@ type SubmarinerEndpoint struct {
...
+	Broker              string
+	CableDriver         string
+	ClusterID           string
+	Namespace           string
+	Token               string
+	Debug               bool
+	NatEnabled          bool
+	HealthCheckEnabled  bool
+	HealthCheckInterval string
+	HealthCheckTimeout  string
```
tpantelis, line 30: We need to make maxLosablePackets configurable

pkg/types/types.go
```diff
@@ -14,17 +14,20 @@ type SubmarinerEndpoint struct {
...
+	ServiceCidr         []string
+	Broker              string
+	CableDriver         string
+	ClusterID           string
+	Namespace           string
+	Token               string
+	Debug               bool
+	NatEnabled          bool
+	HealthCheckEnabled  bool
+	HealthCheckInterval string
```
tpantelis, line 29: How about we make this uint64 so it's parsed for us?

aswinsuryan: Done.

tpantelis (approved):

mangelajo (approved):

mangelajo (approved):

PR #989, opened 2020-12-02
Fix OVNKubernetes PMTU discovery bug

mangelajo:
This toggles an ACL generated by default in ovn-kubernetes, that allows
traffic from the k8s-host port (a virtual port that exist on each host)
so that fragment ICMP packets will also get back to pods in the gateway nodes.

Without this change, if a packet needs to be fragmented while traversing the
IPSEC tunnel the sender pod won't be properly notified, and the packet will
never be fragmented in the origin, as we use the DNF flag all around
for performance purposes.

### DIFF
--- a/pkg/networkplugin-syncer/handlers/ovn/router_ovn_cluster_north.go
+++ b/pkg/networkplugin-syncer/handlers/ovn/router_ovn_cluster_north.go
@@ -1,6 +1,9 @@
 package ovn
 
 import (
+	"fmt"
+	"strings"
+
 	goovn "github.com/ebay/go-ovn"
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

pkg/networkplugin-syncer/handlers/ovn/router_ovn_cluster_north.go
```diff
@@ -45,6 +48,80 @@ func (ovn *SyncHandler) connectOvnClusterRouterToSubm() error {
...
+
+	klog.V(log.DEBUG).Infof("ovn-kubernetes host mgmt ACL wasn't present as allow-related in switch %q", chassisSwitch)
+
+	return nil
+}
+
+func (ovn *SyncHandler) getMgmtPortIP(chassisSwitch string) (string, error) {
+	// NOTE: tried to grab k8s- constant from ovn-kubernetes but that pulls a lot of go mod dependencies
+	// which conflict with ours
+	port, err := ovn.nbdb.LSPGet("k8s-" + chassisSwitch)
```
vthapar, line 108: Make it a local constant and in a comment add link to file in ovn-kubernetes where it is originally defined. Useful for future reference and in case the constant changes in ovn-kubernetes.

pkg/networkplugin-syncer/handlers/ovn/router_ovn_cluster_north.go
```diff
@@ -45,6 +48,80 @@ func (ovn *SyncHandler) connectOvnClusterRouterToSubm() error {
...
+		if err != nil {
+			return errors.Wrapf(err, "error building ACLDel cmd for %#v", acl)
+		}
+
+		err = ovn.nbdb.Execute(delCmd)
+		if err != nil {
+			return errors.Wrapf(err, "error executing ACLDel for allow-related in switch %q", chassisSwitch)
+		}
+
+		addCmd, err := ovn.nbdb.ACLAdd(chassisSwitch, acl.Direction, acl.Match, "allow",
```
vthapar, line 86: Replacing `allow-related` with `allow` will have security impact. We're allowing many more pkt types than just ones needed for PMTU. Okay for now, but should revisit this later and come up with a more secure rule.

vthapar (approved):
Other than one nitpick, approving tentatively to unblock release. We should revisit this to come up with a better fix if possible.

Issue #999, opened 2020-12-03
Submariner Engine metrics are not exposed properly


nyechiel:
**What happened**:

Looking at the OpenShift Metrics UI, I am not able to query the ones exposed from the Engine (e.g `gateway_rx_bytes`, `gateway_tx_bytes`, and so on). `submariner_gateways` and `submariner_connections` work as expected.

**What you expected to happen**:

The metrics to be discoverable.

**How to reproduce it (as minimally and precisely as possible)**:

1. Deploy Submariner across two OCP clusters
2. Enable user workload monitoring in the cluster-monitoring-config configmap 

**Anything else we need to know?**:

Looking at the metrics services within the submariner-operator namespace, it seems like `submariner-engine-metrics` is exposed on 8080/TCP and not 8383/TCP,8686/TCP like `submariner-operator-metric`:

```
$ kubectl get svc -n submariner-operator
NAME                            TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)             AGE
submariner-engine-metrics       ClusterIP   172.30.253.50    <none>        8080/TCP            33m
submariner-lighthouse-coredns   ClusterIP   172.30.35.242    <none>        53/UDP              33m
submariner-operator-metrics     ClusterIP   172.30.174.198   <none>        8383/TCP,8686/TCP   33m
```

**Environment**: 

 OCP 4.6.1, subctl 0.8.0-rc0

sridhargaddam:
As discussed on the slack, we should capture notes saying that GatewayNode should not be running any process that listens on the same port (tcp/8080) that's used for metrics.

skitt:
> As discussed on the slack, we should capture notes saying that GatewayNode should not be running any process that listens on the same port (tcp/8080) that's used for metrics.

I’ve added this to the prerequisites in https://github.com/submariner-io/submariner-website/pull/392

nyechiel:
Thanks @skitt. In addition to the the prerequisites, I think that we should also add this info to the OCP-IPI tool (https://github.com/submariner-io/submariner/tree/master/tools/openshift/ocp-ipi-aws) and OpenShift quickstarts (https://github.com/submariner-io/submariner-website/blob/master/src/resources/shared/openshift/download_prep_for_subm.md).

skitt:
> I think that we should also add this info to the OCP-IPI tool (https://github.com/submariner-io/submariner/tree/master/tools/openshift/ocp-ipi-aws)

[...]

PR #1000, opened 2020-12-04
Give access to the gateway metrics port on AWS

skitt:
Since the gateway runs on the host network, we need to give access to
its metrics port (8080) if we want Prometheus to be able to scrape its
metrics.

Fixes: #999

### DIFF
--- a/tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
+++ b/tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
@@ -61,6 +61,28 @@ resource "aws_security_group_rule" "worker_to_master_sg_vxlan_rule" {
 
 
 
+# Add a rule for metrics traffic for all workers.
+resource "aws_security_group_rule" "worker_sg_metrics_rule" {
+  security_group_id        = data.aws_security_group.worker_sg.id
+  source_security_group_id = data.aws_security_group.worker_sg.id
+  from_port                = 8080
+  protocol                 = "tcp"
+  to_port                  = 8080
+  type                     = "ingress"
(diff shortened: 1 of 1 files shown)


tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
```diff
@@ -61,6 +61,38 @@ resource "aws_security_group_rule" "worker_to_master_sg_vxlan_rule" {
...
+  security_group_id        = data.aws_security_group.worker_sg.id
+  source_security_group_id = data.aws_security_group.master_sg.id
+  from_port                = 8080
+  protocol                 = "tcp"
+  to_port                  = 8080
+  type                     = "ingress"
+}
+
+# Add a rule for metrics traffic from worker nodes to master nodes.
+resource "aws_security_group_rule" "worker_to_master_sg_metrics_rule" {
```
sridhargaddam, line 85: This is not required as we do not schedule Submariner-Engine pod on the Master nodes. The gateway-machine that we prepare as part of prep_for_subm.sh script associates the node with the worker_sg.id, so we just have to allow traffic from worker_sg.id and master_sg.id

sridhargaddam (changes requested):

tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/ec2-resources.tf
skitt: Done, thanks.

sridhargaddam (approved):
LGTM, thanks @skitt

tpantelis (approved):

PR #1009, opened 2020-12-08
Use the K8s SIG iptables wrapper

skitt:
... instead of our own. This ensures that our container images use the
correct iptables implementation without placing any particular
requirements on the host. It also removes the requirement for mounting
the host file system in the container (which appears to be restricted
in OpenShift 4.7 and later).

This requires installing both variants of iptables in the container;
we can do that on Fedora, but not on UBI8, so this is limited to the
gateway image for the time being.

See https://github.com/kubernetes-sigs/iptables-wrappers for details.

### DIFF
--- a/package/submariner.sh
+++ b/package/submariner.sh
@@ -18,34 +18,4 @@ mkdir -p /etc/ipsec
 
 sysctl -w net.ipv4.conf.all.send_redirects=0
 
-function find_iptables_on_host() {
-    chroot /host test -x /usr/sbin/$1 && { echo "/usr/sbin"; return; }
-    chroot /host test -x /sbin/$1 && { echo "/sbin"; return; }
-    echo "unknown"
-}
-
-
-# If host is mounted on /host and host has its own iptables version
-# use that one instead via the shipped iptables wrapper, we do this
-# to avoid configuring iptables and nftables on the host which
-# could lead to functional failures because some hosts use an iptables
-# and iptables-save which program nftables under the hood.
-# Since we're using UBI8, we only have nftables in the container so we
-# can't use the Kubernetes approach (see
-# https://github.com/kubernetes/kubernetes/pull/82966 and
-# https://github.com/kubernetes/website/pull/16271 for details).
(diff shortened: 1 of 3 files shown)


mangelajo:
@nyechiel this looks more like a change for 0.9, can we move it to that board instead?

nyechiel:
> @nyechiel this looks more like a change for 0.9, can we move it to that board instead?

Done

skitt:
This is still relevant.

package/submariner.sh
```diff
@@ -18,34 +18,6 @@ mkdir -p /etc/ipsec
...
-    echo "$f is present on the host at ${location}/$f"
-	sed "s!@@PATH@@!${location}!" /usr/sbin/iptables-wrapper.in > /usr/sbin/$f
-  else
-    echo "WARNING: not using iptables wrapper because iptables was not detected on the"
-    echo "host at the following paths [/usr/sbin, /sbin]."
-    echo "Either the host file system isn't mounted or the host does not have iptables"
-    echo "installed. The pod will use the image installed iptables version."
-  fi
-done
+export PATH=$PATH:/usr/libexec/strongswan
```
mangelajo: We don't need that right ^ no more strongswan

mangelajo (changes requested):
the path for strongswan isn't necessary anymore

package/submariner.sh
skitt: Aargh yes I meant to delete that

mangelajo (approved):

sridhargaddam (approved):

nyechiel:
@skitt @mangelajo this seems critical for 0.9, but should we also consider backport to 0.8?

skitt:
> @skitt @mangelajo this seems critical for 0.9, but should we also consider backport to 0.8?

Once we figure out whether it actually is required for OCP 4.7, yes. If it _is_ required, we’d also need to apply this to the other `iptables`-using images and backport that as well.

PR #1012, opened 2020-12-08
Drop the libreswan-git images

skitt:
These are difficult to maintain, let's stop suggesting we want to.


SteveMattar (approved):
Thanks @skitt

dfarrell07 (approved):

tpantelis (approved):

Issue #993, opened 2020-12-03
OVN: host-network to remote pod/service does not work


mangelajo:
**What happened**:

When deploying with ovn-kubernetes, and running subctl verify, hostNetwork based tests will fail.

**What you expected to happen**:

All tests to pass.

**How to reproduce it (as minimally and precisely as possible)**:

1) Deploy at least one cluster with OVN
2) Install submariner
3) subctl verify ovn-cluster another-cluster

**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`): 0.8.0-rc0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

mangelajo:
Something like this needs to be done in every worker/gw node, considering that 10.132.0.0 and 172.31.0.0/16 are remote cluster CIDRs. also 10.128.2.1 is the local router for that worker node, which can be found by listing ip routes. Every worker node has it's own router ip.
```
# ip r list
default via 10.0.128.1 dev br-ex proto dhcp metric 800 
10.0.128.0/19 dev br-ex proto kernel scope link src 10.0.133.75 metric 800 
10.128.0.0/14 via 10.128.2.1 dev ovn-k8s-mp0 
10.128.2.0/23 dev ovn-k8s-mp0 proto kernel scope link src 10.128.2.2 
169.254.0.0/20 dev ovn-k8s-gw0 proto kernel scope link src 169.254.0.1 
172.30.0.0/16 via 10.128.2.1 dev ovn-k8s-mp0 
# ip r add 10.132.0.0/14 via 10.128.2.1   
# ip r add 172.31.0.0/16 via 10.128.2.1  
```

mangelajo:
Thanks @sridhargaddam for giving a hand on this

nyechiel:
Verified on 0.8.0-rc1

PR #1020, opened 2020-12-10
OVN: Host-network to remote pod/service support

sridhargaddam:
This PR implements Host network to remote pod/service support
in OVN route-agent driver.

Fixes issue: https://github.com/submariner-io/submariner/issues/993
Co-Authored-By: Miguel Angel Ajo <majopela@redhat.com>

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/host_networking.go
+++ b/pkg/routeagent_driver/handlers/ovn/host_networking.go
@@ -1,0 +1,128 @@
+package ovn
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"syscall"
+
+	"github.com/submariner-io/admiral/pkg/log"
+	"k8s.io/klog"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/stringset"
+	"github.com/vishvananda/netlink"
+
+	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
+)
+
+const (
+	OVNK8sMgmntIntfName = "ovn-k8s-mp0"
+)
+
+func (ovn *Handler) updateHostNetworkDataplane() error {
+	currentRuleRemotes, err := ovn.getExistingIPv4HostNetworkRoutes()
+	if err != nil {
+		return errors.Wrapf(err, "error reading ip rule list for IPv4")
+	}
+
+	endpointSubnets := ovn.getRemoteSubnets()
+
+	toAdd := currentRuleRemotes.Difference(endpointSubnets)
+
+	err = ovn.programRulesForRemoteSubnets(toAdd, netlink.RuleAdd, os.IsExist)
+	if err != nil {
+		return errors.Wrap(err, "error adding routing rule")
+	}
+
+	toRemove := endpointSubnets.Difference(currentRuleRemotes)
+
+	err = ovn.programRulesForRemoteSubnets(toRemove, netlink.RuleDel, os.IsNotExist)
+	if err != nil {
+		return errors.Wrapf(err, "error removing routing rule")
+	}
+
+	nextHop, err := ovn.getNextHopOnK8sMgmtIntf()
+	if err != nil {
+		return errors.Wrapf(err, "getNextHopOnK8sMgmtIntf returned error")
+	}
+
+	route := &netlink.Route{
+		Gw:    *nextHop,
+		Table: constants.RouteAgentHostNetworkTableID,
+	}
+
+	err = netlink.RouteAdd(route)
+	if err != nil && !os.IsExist(err) {
+		return errors.Wrap(err, "error adding submariner default")
+	}
+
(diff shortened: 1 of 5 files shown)


pkg/routeagent_driver/handlers/ovn/host_networking.go
```diff
@@ -0,0 +1,126 @@
...
+	toRemove := endpointSubnets.Difference(currentRuleRemotes)
+
+	err = ovn.programRulesForRemoteSubnets(toRemove, netlink.RuleDel, os.IsNotExist)
+	if err != nil {
+		return errors.Wrapf(err, "error removing routing rule")
+	}
+
+	nextHop, err := ovn.getNextHopOnK8sMgmtIntf()
+	if err != nil {
+		return errors.Wrapf(err, "error removing routing rule")
```
tpantelis, line 47:
```suggestion
		return errors.Wrapf(err, "getNextHopOnK8sMgmtIntf  returned error")
```

pkg/routeagent_driver/handlers/ovn/host_networking.go
```diff
@@ -0,0 +1,126 @@
...
+	currentRouteList, err := netlink.RouteList(link, syscall.AF_INET)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error retrieving routes on the link %s", OVNK8sMgmntIntfName)
+	}
+
+	for i := range currentRouteList {
+		klog.V(log.DEBUG).Infof("Processing route %v", currentRouteList[i])
+
+		if currentRouteList[i].Dst == nil || currentRouteList[i].Gw == nil {
+			klog.V(log.DEBUG).Infof("Found nil gw or dst")
```
mangelajo, line 113:
```suggestion
			klog.V(log.DEBUG).Infof("Found nil gw or dst")
```

Let's remove this bit once we verify it's working fine.

pkg/routeagent_driver/handlers/ovn/host_networking.go
```diff
@@ -0,0 +1,126 @@
...
+
+		for _, subnet := range ovn.localEndpoint.Spec.Subnets {
+			_, localCIDR, _ := net.ParseCIDR(subnet)
+			if currentRouteList[i].Dst == localCIDR && currentRouteList[i].Gw != nil {
+				return &currentRouteList[i].Gw, nil
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("could not find the necessary route")
```
mangelajo, line 125:
```suggestion
	return nil, fmt.Errorf("could not find the route to %q via %q", localCIDR, OVNK8sMgmntIntfName)
```

pkg/routeagent_driver/handlers/ovn/south_rules.go
```diff
@@ -31,21 +31,29 @@ func (ovn *Handler) handleSubnets(subnets []string, ruleFunc func(rule *netlink.
...
+
+		rule.Src = srcCIDR
 	}
 
-	rule := netlink.NewRule()
-	rule.Dst = dstCIDR
-	rule.Src = srcCIDR
-	rule.Table = constants.RouteAgentHostNetworkTableID
+	rule.Table = tableId
+	rule.Priority = tableId
```
mangelajo, line 56:
```suggestion
	rule.Priority = tableId
```

May be we should just use a fixed priority for now, or accept a parameter for it?, setting the prioity as the table ID is confusing :)

mangelajo (changes requested):
@sridhargaddam it looks great, good job. A few nit comments.

sridhargaddam (commented):
thanks for reviewing @tpantelis @mangelajo

sridhargaddam:
Verified the PR on an OVN build and all the HostNetworking tests pass

```
•SSSSSSSSSSSSSS
Ran 10 of 30 Specs in 140.283 seconds
SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 20 Skipped
```

tpantelis (approved):

pkg/routeagent_driver/handlers/ovn/south_rules.go
mangelajo: Ok, I understand now, but it's confusing. Can we document via comments at least?

I would be more explicit and separate the input of priority & table, even if they will be the same value
in the end.
```
 programRule(dest, src string, tableId, priority int) 
```

Can we do something like this later as a follow up?

mangelajo (approved):
See comments here: 

https://github.com/submariner-io/submariner/pull/1020/files#r542223756

If we could do a follow up PR, it'd make the code/intent clearer.

pkg/routeagent_driver/handlers/ovn/south_rules.go
sridhargaddam: Sure, we can do that.

Issue #1022, opened 2020-12-11
OVN: The ovn loadbalancers (ClusterIP) seem to choke on jumboframes / icmp requesting frag to the service


mangelajo:
**What happened**:

Our E2E tests fail since we test jumboframe fragmentation. (by chance... not by design)

https://mail.openvswitch.org/pipermail/ovs-discuss/2020-December/050834.html

iptables -I FORWARD -o ovn-k8s-gw0 -p tcp --tcp-flags SYN,RST SYN -j TCPMSS  --set-mss 1398   

**What you expected to happen**:

Service tests to work

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

Something that seems to workaround this, although it's not perfect, is to find the PMTU to the destination cluster....
```
sh-4.4# tracepath 10.133.2.2
 1?: [LOCALHOST]                      pmtu 8938
 1?: [LOCALHOST]                      pmtu 1438
 1:  ip-10-133-2-2.us-east-2.compute.internal              0.440ms reached
 1:  ip-10-133-2-2.us-east-2.compute.internal              0.410ms reached
     Resume: pmtu 1438 hops 1 back 1 
```

And then clamp TCP/MSS with pmtu-40 (40 = IP + TCP overhead)
```
iptables -I FORWARD -o ovn-k8s-gw0 -p tcp --tcp-flags SYN,RST SYN -j TCPMSS  --set-mss 1398
```

This won't help with UDP, but will help until we can figure out a solution in ovn. (if it's really in ovn.)

This was tried without success:
```
iptables -I FORWARD -p tcp --tcp-flags SYN,RST SYN -j TCPMSS  --clamp-mss-to-pmtu
```

**Environment**:
- Submariner version (use `subctl version`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

sridhargaddam:
@mangelajo can you try if the following solution works with OVN?

https://github.com/submariner-io/submariner/issues/995#issuecomment-738010826

mangelajo:
> @mangelajo can you try if the following solution works with OVN?
> 
> [#995 (comment)](https://github.com/submariner-io/submariner/issues/995#issuecomment-738010826)

I tried, it doesn't work in this case with OVN.

[...]

PR #1025, opened 2020-12-14
Insert MSS clamping for traffic arriving services

mangelajo:
OVN load balancing does not propagate fragment ICMPs down to the
 services, so when fragmentation becomes necessary the TCP stack
 of the service pod won't be able to detect this situation.
    
 Until this is fixed in OVN this workaround (to be enhanced later)
 will fix TCP.
    
 Fixes-Issue: #1022
    
### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
@@ -3,6 +3,7 @@ package ovn
 import (
 	"net"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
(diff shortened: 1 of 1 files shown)


skitt (approved):

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -83,16 +84,31 @@ func (ovn *Handler) updateGatewayDataplane() error {
...
 
-	return [][]string{
+	rules := [][]string{
 		{"-i", ovnK8sGatewayInterface, "-o", ovn.cableRoutingInterface.Name, "-j", "ACCEPT"},
-		{"-i", ovn.cableRoutingInterface.Name, "-o", ovnK8sGatewayInterface, "-j", "ACCEPT"},
-	}, nil
+		{"-i", ovn.cableRoutingInterface.Name, "-o", ovnK8sGatewayInterface, "-j", "ACCEPT"}}
+
+	// NOTE: This is a workaround for submariner issue https://github.com/submariner-io/submariner/issues/1022
+	//		  1398 is 1500
```
sridhargaddam, line 105: "1398 is 1500" -> Did you miss conveying something here?

mangelajo:
one note after questions in slack:

   My plan is to implement PMTU discovery in 0.8.1 so we can use it here. Also... we need to go down to the OVN bug itself and get it fixed.

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
mangelajo: oops

mangelajo: Done.

tpantelis (approved):

sridhargaddam (approved):

Issue #1031, opened 2020-12-15
Duplicated connections on upgrade test (re-enable-upgrade-test)


mangelajo:
**What happened**:

The upgrade tests on the submariner repo are failing , for example: 
https://github.com/submariner-io/submariner/pull/1030/checks?check_run_id=1556110859

This change to shipyard probably unveils it:
https://github.com/submariner-io/shipyard/pull/380

After being applied, we wee the upgrade jobs timeout sometimes, and also:

```
[submariner]$ [cluster3] grep connected
cluster1-worker                 cluster1                172.17.0.8      libreswan           100.1.0.0/16, 10.1.0.0/16               connected       
cluster2-worker                 cluster2                172.17.0.5      libreswan           100.2.0.0/16, 10.2.0.0/16               connected       
cluster2-worker                 cluster2                172.17.0.5      libreswan           100.2.0.0/16, 10.2.0.0/16               connected       
cluster3-worker                 cluster3                172.17.0.11     libreswan           100.3.0.0/16, 10.3.0.0/16               connected       
cluster1-worker                 cluster1                172.17.0.8      libreswan           100.1.0.0/16, 10.1.0.0/16               connected       
cluster3-worker                 cluster3                172.17.0.11     libreswan           100.3.0.0/16, 10.3.0.0/16               connected
```

**What you expected to happen**:

Upgrade test passing

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

tpantelis:
Is this still an issue? The upgrade test is passing now.

skitt:
> Is this still an issue? The upgrade test is passing now.

We worked around this by waiting for the CR deletion to be complete before creating the new one; ideally we should be capable of handling simultaneous additions and removals of CRs.

PR #1035, opened 2020-12-16
Increase upgrade test timeout 30->45 mins

dfarrell07:
The upgrade tests have recently started being frequently canceled by
their 30min job timeouts.

github.com/submariner-io/submariner/actions?query=workflow%3AUpgrade

"The job running on runner GitHub Actions 2 has exceeded the maximum
execution time of 30 minutes."

It seems they are hitting a failure described by #1031 and then being
canceled before they have time to run their post mortem.

### DIFF
--- a/.github/workflows/upgrade-e2e.yml
+++ b/.github/workflows/upgrade-e2e.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   upgrade-e2e:
     name: Latest Release to Latest Version
-    timeout-minutes: 30
+    timeout-minutes: 45
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false


mangelajo (approved):
I'm not sure if the issue is the timeout itself, or it would timeout equally because of : https://github.com/submariner-io/submariner/issues/1031

Let's let @skitt have a look at this before bumping the timeout

mangelajo:
@dfarrell07 a normal working pass is 19 min, that's 10min headroom which I suspect only happens when something is really going wrong.

dfarrell07:
> @dfarrell07 a normal working pass is 19 min, that's 10min headroom which I suspect only happens when something is really going wrong.

The history here shows the old passing jobs at 24 mins.

https://github.com/submariner-io/submariner/actions?query=workflow%3AUpgrade

nyechiel (approved):

Issue #1036, opened 2020-12-16
Upgrade Job: Missing delete endpoint notification for remote clusters


sridhargaddam:
**What happened**:
Latest release to Latest version tests are failing 
https://github.com/submariner-io/submariner/pull/1010/checks?check_run_id=1562617791

The e2e test fails on the [redundancy] job. As part of e2e redundancy tests, we perform a gateway-failover on "cluster2", wait for a new Engine Pod to be created on the newly labelled node and ensure that all connections are established. The last step fails - i.e., connection to cluster1 stays in "connecting..." stage itself.

Looking at the logs on cluster1 enginepod, we see the following error. Also, there is no delete event for cluster2 old endpoint in the logs of cluster1 enginePod. This needs to be investigated and fixed.

cluster1 EnginePod logs: https://pastebin.com/raw/UU0j1cHt

```
2020-12-16T09:26:24.5140018Z [cluster1] E1216 09:20:56.988121       1 tunnel.go:59] error installing cable for Endpoint types.SubmarinerEndpoint{Spec:v1.EndpointSpec{ClusterID:"cluster2", CableName:"submariner-cable-cluster2-172-17-0-10", HealthCheckIP:"10.2.0.1", Hostname:"cluster2-worker2", Subnets:[]string{"100.2.0.0/16", "10.2.0.0/16"}, PrivateIP:"172.17.0.10", PublicIP:"", NATEnabled:false, Backend:"libreswan", BackendConfig:map[string]string(nil)}}, found a pre-existing cable "submariner-cable-cluster2-172-17-0-5" that belongs to this cluster cluster2
```

**What you expected to happen**:
The redundancy test should pass and "cluster1" enginePod should be getting the delete notification for old Endpoint.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

tpantelis:
[...]

PR #1037, opened 2020-12-16
Establish connection to remoteEndpoints upon notification

sridhargaddam:
Instead of waiting for a delete notification for an endpoint,
if add/update notification is received, delete the old connection
and restablish with the new parameters.

Related to: https://github.com/submariner-io/submariner/issues/1036

### DIFF
--- a/pkg/cableengine/cableengine.go
+++ b/pkg/cableengine/cableengine.go
@@ -1,16 +1,16 @@
 package cableengine
 
 import (
-	"fmt"
 	"reflect"
 	"sync"
 
 	"github.com/submariner-io/admiral/pkg/log"
+	"k8s.io/klog"
+
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cable"
 	"github.com/submariner-io/submariner/pkg/types"
 	"github.com/submariner-io/submariner/pkg/util"
-	"k8s.io/klog"
 
 	// Add supported drivers
 	_ "github.com/submariner-io/submariner/pkg/cable/libreswan"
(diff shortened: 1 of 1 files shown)


sridhargaddam:
@tpantelis @mangelajo @skitt I marked this as Draft, but would like to get your feedback on this change.

tpantelis (dismissed):

pkg/cableengine/cableengine.go
```diff
@@ -138,7 +138,13 @@ func (i *engine) InstallCable(endpoint types.SubmarinerEndpoint) error {
...
 
 		if util.GetClusterIDFromCableName(active) == endpoint.Spec.ClusterID {
-			return fmt.Errorf("found a pre-existing cable %q that belongs to this cluster %s", active, endpoint.Spec.ClusterID)
+			klog.V(log.DEBUG).Infof("Found a pre-existing cable %q that belongs to this cluster %s", active, endpoint.Spec.ClusterID)
+			preExistingEndpoint := endpoint
+			preExistingEndpoint.Spec.CableName = active
+			err = i.driver.DisconnectFromEndpoint(preExistingEndpoint)
+			if err != nil {
+				return err
+			}
```
tpantelis, line 147: Shouldn't we obtain the actual previous endpoint info:

```suggestion
			activeEndpointInfo := i.getEndpointInfo(active, connections)
			if activeEndpointInfo != nil {
				err = i.driver.DisconnectFromEndpoint(types.SubmarinerEndpoint{*activeEndpointInfo})
				if err != nil {
					return err
				}
			}
```

tpantelis: Also I think the `ClusterID` check should be an else to the preceding `if`. Otherwise if the `CableName` was the same but the `PublicIP` changed then it would fall thru to this code b/c the `ClusterID` would be the same. I don't think this is the intention. Previously it would've returned an error - with these changes it would try to disconnect again.  I think we should structure it like this:

```
for _, active := range activeConnections {
    ...
    activeEndpointInfo := i.getEndpointInfo(active, connections)
    if activeEndpointInfo == nil {
        klog.Warningf("No Connection found for active Endpoint cable name %q", active)
        continue
    }

    if active == endpoint.Spec.CableName {
        ...
    } else if util.GetClusterIDFromCableName(active) == endpoint.Spec.ClusterID {
        ...
    }
}
```

tpantelis (changes requested):

pkg/cableengine/cableengine.go
sridhargaddam: Thats a good point @tpantelis, thanks.

sridhargaddam (commented):
thanks for reviewing @tpantelis

tpantelis (approved):

skitt (approved):

PR #1042, opened 2020-12-17
Limit the number of times a service is requeued

sridhargaddam:
Currently, Globalnet works only with kubeproxy iptables driver
and re-uses the IPTable Service Chains created by kubeproxy for
supporting Globalnet ingress use-case for Services.
If there is any issue with kubeproxy or if the cluster uses
kubeproxy with IPVS mode, the IPTable Service Chain will not
be created on the Gateway node and Globalnet continues to
requeue the service waiting for the Chain to be created.
In this PR, we requeue max of 20 times before giving up.

### DIFF
--- a/pkg/globalnet/controllers/ipam/constants.go
+++ b/pkg/globalnet/controllers/ipam/constants.go
@@ -22,6 +22,8 @@ const (
 	kubeProxyServiceChainPrefix = "KUBE-SVC-"
 	kubeProxyServiceChainName   = "KUBE-SERVICES"
 
+	maxServiceRequeues = 20
+
 	AddRules    = true
 	DeleteRules = false
 
--- a/pkg/globalnet/controllers/ipam/ipam.go
+++ b/pkg/globalnet/controllers/ipam/ipam.go
@@ -174,6 +174,15 @@ func (i *Controller) processNextObject(objWorkqueue workqueue.RateLimitingInterf
 					objWorkqueue.Forget(obj)
 					return nil
 				case Requeue:
+					// If the kubeproxy chain for the service is missing on the node (might happen if there is
+					// some issue with kubeproxy pod itself or if using non-iptables based kubeproxy), instead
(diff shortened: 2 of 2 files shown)


pkg/globalnet/controllers/ipam/constants.go
```diff
@@ -22,6 +22,8 @@ const (
 	kubeProxyServiceChainPrefix = "KUBE-SVC-"
 	kubeProxyServiceChainName   = "KUBE-SERVICES"
 
+	GLOBALNET_SVC_MAX_REQUEUES = 20
```
tpantelis, line 25:
```suggestion
	maxServiceRequeues = 20
```

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -174,6 +174,15 @@ func (i *Controller) processNextObject(objWorkqueue workqueue.RateLimitingInterf
 					objWorkqueue.Forget(obj)
 					return nil
 				case Requeue:
+					// If the kubeproxy chain for the service is missing on the node (might happen if there is
+					// some issue with kubeproxy pod itself or if using non-iptables based kubeproxy), instead
+					// of re-queuing forever, we place a hard-limit of GLOBALNET_SVC_MAX_REQUEUES.
+					if objWorkqueue.NumRequeues(obj) >= GLOBALNET_SVC_MAX_REQUEUES {
+						objWorkqueue.Forget(obj)
+						klog.V(log.DEBUG).Infof("Service %s requeued max(%q) allowed iterations, ignoring it",
```
tpantelis, line 182: ?

```suggestion
						klog.Warningf("Service %s requeued max(%q) allowed iterations, ignoring it",
```

tpantelis (changes requested):

sridhargaddam (commented):
Thanks for review comments @tpantelis

tpantelis (approved):

skitt (approved):

PR #1045, opened 2020-12-18
Don't try to remove local cable

tpantelis:
`RemoveCable` should check if the EP is local like `InstallCable` does.


mangelajo:
Is this related to an issue? shouldn't the cable engine be able to remove an old endpoint on his own cluster?

tpantelis:
> Is this related to an issue? shouldn't the cable engine be able to remove an old endpoint on his own cluster?

No reported issue just something I noticed when reviewing logs. We don't install cables for the local cluster - it checks for that in `InstallCable` -  so we should do the same in `RemoveCable` rather than calling the driver for an EP connection we know doesn't exist.  This scenario didn't occur prior to the admiral conversion.

pkg/cableengine/cableengine.go
```diff
@@ -168,6 +168,11 @@ func (i *engine) getEndpointInfo(cableName string, connections *[]v1.Connection)
 }
 
 func (i *engine) RemoveCable(endpoint types.SubmarinerEndpoint) error {
+	if endpoint.Spec.ClusterID == i.localCluster.ID {
+		klog.V(log.DEBUG).Infof("Not removing cable for local cluster")
```
dfarrell07, line 172:
```suggestion
		klog.V(log.DEBUG).Infof("Cables are not added/removed for local cluster, skipping removal")
```

To incorporate your explanation?

skitt (changes requested):
See Daniel’s suggestion, I think it will keep the logs’ meaning clearer in several months’ time when we’ve all forgotten this PR...

skitt (approved):

mangelajo (approved):

Issue #996, opened 2020-12-03
E2e redundancy test failing in KIND Globalnet env with Wireguard driver


sridhargaddam:
**What happened**:
In KIND Setup with Globalnet enabled, when e2e tests are run, one of the redundancy tests is failing randomly

[redundancy] Gateway fail-over tests when a new node is labeled as a gateway node and the label on the existing gateway node is removed 
  should start a submariner engine on the new gateway node and be able to connect from another cluster
  /go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/submariner/test/e2e/redundancy/gateway_failover.go:37
STEP: Creating namespace objects with basename "gateway-redundancy"
STEP: Generated namespace "e2e-tests-gateway-redundancy-l9rhp" in cluster "cluster1" to execute the tests in
STEP: Creating namespace "e2e-tests-gateway-redundancy-l9rhp" in cluster "cluster2"
STEP: Found gateway node "cluster1-worker" on "cluster1"
STEP: Found non-gateway node "cluster1-worker2" on "cluster1"
STEP: Found submariner engine pod "submariner-gateway-frbsg" on "cluster1"
STEP: Deleting namespace "e2e-tests-gateway-redundancy-l9rhp" on cluster "cluster1"
STEP: Deleting namespace "e2e-tests-gateway-redundancy-l9rhp" on cluster "cluster2"

• Failure [240.147 seconds]
[redundancy] Gateway fail-over tests
/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/submariner/test/e2e/redundancy/gateway_failover.go:18
  when a new node is labeled as a gateway node and the label on the existing gateway node is removed
  /go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/submariner/test/e2e/redundancy/gateway_failover.go:36
    should start a submariner engine on the new gateway node and be able to connect from another cluster [It]
    /go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/submariner/test/e2e/redundancy/gateway_failover.go:37

    Failed to find the submariner endpoint for cluster1. No endpoint found
    Unexpected error:                                                                                                                                                                         
        <*errors.errorString | 0xc0002e6200>: {                                                                                                                                               
            s: "timed out waiting for the condition",                                                                                                                                         
        }                                                                                                                                                                                     
        timed out waiting for the condition                                                                                                                                                   
[...]

PR #1046, opened 2020-12-18
Resetting the pinger after connection error

aswinsuryan:
The healthchecker use the same pinger even if the ping fails.
This is leading into stale conntrack entries, which causes pakcet
loss even if the connection is established. So in the event of
a failure a new pinger will be created.

Fixes: #996

### DIFF
--- a/pkg/cableengine/healthchecker/pinger.go
+++ b/pkg/cableengine/healthchecker/pinger.go
@@ -111,19 +111,11 @@ func (p *pingerInfo) doPing() error {
 		if pinger.PacketsSent-pinger.PacketsRecv > int(p.maxPacketLossCount) {
 			p.Lock()
 			defer p.Unlock()
-			// The counter will be reset to once to ensure that it is continuous packet loss
-			// before marking the connection as down.
-			if pinger.PacketsSent <= int(p.maxPacketLossCount)+1 {
-				p.failureMsg = fmt.Sprintf("Failed to successfully ping the remote endpoint IP %q", p.ip)
-				p.connectionStatus = ConnectionError
 
-				pinger.Stop()
+			p.connectionStatus = ConnectionError
+			p.failureMsg = fmt.Sprintf("Failed to successfully ping the remote endpoint IP %q", p.ip)
 
-				return
-			}
-
-			pinger.PacketsSent = 0
-			pinger.PacketsRecv = 0
+			pinger.Stop()
 		}
 	}
 
@@ -134,6 +126,9 @@ func (p *pingerInfo) doPing() error {
 		p.connectionStatus = Connected
 		p.failureMsg = ""
 		p.statistics.update(uint64(packet.Rtt.Nanoseconds()))
+
+		pinger.PacketsSent = 0
+		pinger.PacketsRecv = 0
 	}
 
 	err = pinger.Run()
--- a/pkg/cableengine/healthchecker/pinger.go
+++ b/pkg/cableengine/healthchecker/pinger.go
@@ -111,9 +111,17 @@ func (p *pingerInfo) doPing() error {
 		if pinger.PacketsSent-pinger.PacketsRecv > int(p.maxPacketLossCount) {
 			p.Lock()
 			defer p.Unlock()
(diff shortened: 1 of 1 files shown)


pkg/cableengine/healthchecker/pinger.go
```diff
@@ -111,9 +111,17 @@ func (p *pingerInfo) doPing() error {
 		if pinger.PacketsSent-pinger.PacketsRecv > int(p.maxPacketLossCount) {
 			p.Lock()
 			defer p.Unlock()
+			// The counter will be reset to once to ensure that it is continuous packet loss
+			// before marking the connection as down.
+			if pinger.PacketsSent <= int(p.maxPacketLossCount)+1 {
+				p.failureMsg = fmt.Sprintf("Failed to successfully ping the remote endpoint IP %q", p.ip)
+				p.connectionStatus = ConnectionError
+
+				pinger.Stop()
```
mangelajo, line 120: But we stop the pinger forever? :?

I'm not understand the patch :)

mangelajo: I mean, when is it started again?

aswinsuryan: We restart the pinger in infinite [1] for loop and will be stopped only if a stop signal, which happens when the endpoint is deleted.

[1] https://github.com/submariner-io/submariner/blob/master/pkg/cableengine/healthchecker/pinger.go#L67

mangelajo: I don't understand how's it started again, I'm missing something,

mangelajo: ahh, it's pinger.stop , which you created in L89

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -111,9 +111,17 @@ func (p *pingerInfo) doPing() error {
 		if pinger.PacketsSent-pinger.PacketsRecv > int(p.maxPacketLossCount) {
 			p.Lock()
 			defer p.Unlock()
+			// The counter will be reset to once to ensure that it is continuous packet loss
+			// before marking the connection as down.
```
tpantelis, line 115: I'm not clear on this comment. Which counter?

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -111,9 +111,17 @@ func (p *pingerInfo) doPing() error {
...
+			if pinger.PacketsSent <= int(p.maxPacketLossCount)+1 {
```
tpantelis, line 116: So basically it seems we want to mark it down if we we've sent out `maxPacketLossCount + 1` packets with no replies.  I think it would be clearer to check:

```suggestion
			if pinger.PacketsRecv == 0  {
```

Either way once we trip the threshold it will take another `maxPacketLossCount + 1` packets lost before we mark it as down (unless the very first `maxPacketLossCount + 1` packets happened failed). So it will take `(maxPacketLossCount + 1)*2` total packets lost before we mark it as down. 

As an alternative, we could reset `PacketsSent` and `PacketsRecv` when we get a reply in `OnRecv`.  Then here we would do:

```
if pinger.PacketsSent-pinger.PacketsRecv > int(p.maxPacketLossCount) {
    p.Lock()
    defer p.Unlock()

    p.failureMsg = fmt.Sprintf("Failed to successfully ping the remote endpoint IP %q", p.ip)
    p.connectionStatus = ConnectionError

    pinger.Stop()
}
```

tpantelis (changes requested):

mangelajo:
Can we rename the commit / PR explaining what we are doing exactly? (resetting the pinger after connection error..)

pkg/cableengine/healthchecker/pinger.go

aswinsuryan:
> Can we rename the commit / PR explaining what we are doing exactly? (resetting the pinger after connection error..)

Done.

tpantelis (approved):

skitt (approved):

PR #1050, opened 2020-12-18
Change GetConnections to return non-pointer

tpantelis:
Change return type from `*[]v1.Connection` to `[]v1.Connection`. This simplifies the code. I don't see any compelling reason to return a pointer - the `CableEngine` code doesn't handle nil and all drivers return non-nil anyway. If there are no connections then a driver can return a nil or empty array.


dfarrell07 (approved):
I wonder if we had the golangci-lint performance tests enabled, if they would suggest using a pointer.

https://github.com/submariner-io/submariner/blob/master/.golangci.yml#L7

skitt (approved):
We’ll need to do some proper performance analysis if performance turns out to be a problem, we can revisit this then if it turns out to be significant.

tpantelis:
A slice value is a header containing a pointer to the backing array.  It is small so passing by value is negligible. Slices are commonly passed by value so it needs to be performant, ie we'd see performance issues in many other areas if it were not.

PR #1052, opened 2020-12-22
Fix cable driver in connection latency metric

maayanf24:


dfarrell07 (approved):

nyechiel (approved):

skitt (approved):
For future reference, it would be good to have a little more explanation of the change in the commit message ;-).

Issue #1063, opened 2021-01-04
ovn-nbctl in fedora:33 image fails talking to ovn-kubernetes


mangelajo:
**What happened**:

```
I0104 16:45:58.900913       1 nbctl.go:57] ovn-nbctl [--db=tcp:172.17.0.10:6641 lr-policy-add ovn_cluster_router 20000 ip4.dst == 100.2.0.0/16 reroute 169.254.254.1]
E0104 16:45:58.913724       1 nbctl.go:64] error running ovn-nbctl exit status 1, output:
2021-01-04T16:45:58Z|00002|ovsdb_idl|WARN|Load_Balancer table in OVN_Northbound database lacks options column (database needs upgrade?)
2021-01-04T16:45:58Z|00003|ovsdb_idl|WARN|Logical_Router_Policy table in OVN_Northbound database lacks nexthops column (database needs upgrade?)
2021-01-04T16:45:58Z|00004|ovsdb_idl|WARN|transaction error: {"details":"No column nexthops in table Logical_Router_Policy.","error":"unknown column","syntax":"{\"action\":\"reroute\",\"match\":\"ip4.dst == 100.2.0.0/16\",\"nexthops\":\"169.254.254.1\",\"priority\":20000}"}
ovn-nbctl: transaction error: {"details":"No column nexthops in table Logical_Router_Policy.","error":"unknown column","syntax":"{\"action\":\"reroute\",\"match\":\"ip4.dst == 100.2.0.0/16\",\"nexthops\":\"169.254.254.1\",\"priority\":20000}"}
E0104 16:45:58.913791       1 endpoint_created.go:22] Error handling created endpoint: RemoteEndpointCreated failed: "ovn-sync-handler" returned error: error adding the "100.2.0.0/16" routing rule to router "ovn_cluster_router": exit status 1
I0104 16:45:58.913813       1 resource_syncer.go:352] Syncer "Endpoint watcher for networkplugin-syncer registry": transform function returned nil - not syncing - requeue: true
I0104 16:45:58.913842       1 queue.go:88] Endpoint watcher for networkplugin-syncer registry: enqueued "submariner-operator/cluster2-submariner-cable-cluster2-172-17-0-6" for retry - # of times re-queued: 17
```

**What you expected to happen**:

For lr-policy-add in network plugin syncer to work.

**How to reproduce it (as minimally and precisely as possible)**:

[...]

PR #1065, opened 2021-01-05
Switch the submariner-networkplugin-syncer to fedora:32

mangelajo:
We use fedora to get ovn-nbctl, but fedora:33 now ships an ovn-nbctl
version which is incompatible with previous versions of the database.

The version in fedora:32 should be fine, and this ovn-nbctl functionality
should be migrated into go-ovn during 0.9.0.

Fixes-Issue: #1063

### DIFF
--- a/package/Dockerfile.submariner-networkplugin-syncer
+++ b/package/Dockerfile.submariner-networkplugin-syncer
@@ -1,4 +1,4 @@
-FROM fedora:33 
+FROM fedora:32
 
 WORKDIR /var/submariner
 


package/Dockerfile.submariner-networkplugin-syncer
```diff
@@ -1,4 +1,4 @@
-FROM fedora:33 
+FROM fedora:32
```
skitt, line 1: Please add a comment in the Dockerfile to point to the issue, and perhaps the upstream reference (so we know where to look if things break before `go-ovn` is ready).

skitt (changes requested):
Please add some explanation in the code itself.

mangelajo:
> Please add some explanation in the code itself.

Makes sense, but I'm closing the issue, it doesn't work.

PR #1078, opened 2021-01-12
Add IPAM unit tests

tpantelis:


sridhargaddam:
Thank you @tpantelis for working on this. As this is a big change (of course its mostly the unit tests, but its also modifying the flow to some extent), let's target this for 0.9 release (instead of 0.8.1). Your thoughts?

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -102,10 +101,7 @@ func NewController(spec *SubmarinerIpamControllerSpecification, config *Informer
 	return ipamController, nil
 }
 
-func (i *Controller) Run(stopCh <-chan struct{}) error {
-	defer utilruntime.HandleCrash()
-	defer cleanup.ClearGlobalnetChains()
```
sridhargaddam, line 107: When the active gateway migrates to a new node, the ipamController will be stopped and we want the cleanup to happen. Can you please point me where the cleanup is handled in this change?

pkg/globalnet/main.go
```diff
@@ -91,24 +96,18 @@ func main() {
...
+	if err = gatewayMonitor.Start(stopCh); err != nil {
+		klog.Fatalf("Error running gatewayMonitor: %s", err.Error())
+	}
 
-		if err = gatewayMonitor.Run(stopCh); err != nil {
-			klog.Fatalf("Error running gatewayMonitor: %s", err.Error())
-		}
-	}()
+	<-stopCh
+	gatewayMonitor.Stop()
```
sridhargaddam, line 109: Okay, along with the cleanup when the process itself dies, we also want the cleanup (of certain iptable chains) even when there is a Gateway Migration.

tpantelis:
> Thank you @tpantelis for working on this. As this is a big change (of course its mostly the unit tests, but its also modifying the flow to some extent), let's target this for 0.9 release (instead of 0.8.1). Your thoughts?

Yes - I intended this for 0.9.0.

pkg/globalnet/controllers/ipam/ipam.go
tpantelis: `cleanup.ClearGlobalnetChains` is now called by `GatewayMonitor.stopIpamController` and the chain cleanup is verified by the first test spec on L59 in _gatewaymonitor_test.go_. I made this function synchronous b/c it really didn't need to be async and keep a goroutine blocked until stopped, but, more importantly, it was problematic for the unit tests, ie it caused intermittent failures due to an issue in the fake clients (they're a bit flaky with concurrency on startup).

dfarrell07 (approved):

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -372,15 +366,15 @@ func (i *Controller) handleUpdateNode(old, newObj interface{}) {
 
 	oldCniIfaceIpOnNode := old.(*k8sv1.Node).GetAnnotations()[constants.CniInterfaceIP]
 	newCniIfaceIpOnNode := newObj.(*k8sv1.Node).GetAnnotations()[constants.CniInterfaceIP]
-	if oldCniIfaceIpOnNode == "" && newCniIfaceIpOnNode == "" {
-		klog.V(log.DEBUG).Infof("In handleUpdateNode, node %q is not yet annotated with cniIfaceIP, enqueing", newObj.(*k8sv1.Node).Name)
+	if oldCniIfaceIpOnNode == "" && newCniIfaceIpOnNode != "" {
```
sridhargaddam, line 369: Thanks for fixing.

pkg/globalnet/controllers/ipam/ipam.go
sridhargaddam: Cool, thanks @tpantelis

sridhargaddam (approved):

Issue #825, opened 2020-09-22
Add 1 HA Failure Scenario to E2E testing (POC)


pinikomarov:
Purpose of this issue : 
To automate - add to e2e (GO) a single HA failure test from the scenarios here : 
https://docs.google.com/spreadsheets/d/11MV7CZmt13D8D4WYJTaddzQxSrXFRrn3UG0LiQKFVwY/edit#gid=0

More specifically :
scenario no. 1 of : setup : KIND Setup: Vanilla Submariner + Libreswan Driver + without Lighthouse

Sridhar's gave me the manual steps which are as follows : 

> 
> Okay, I did the following manually.
> 1. Deploy clusters using KIND. In submariner repo, execute the command "make deploy using=libreswan"
> 2. The above command creates three clusters - cluster1 (broker), cluster2, and cluster3.
> 3. I used cluster2 as West cluster and cluster3 as East cluster (sample topology diagram in the [*])
> 
> The above steps are common for all the scenarios.
> Now let's take the scenario "Submariner Engine Pod restarts (i.e., same node and no failover) on the West cluster."
> 
> For this, I deploy a few fedora pods on cluster2 and cluster3.
> kubectl run fedora-pod -i --tty --image fedora -- /bin/bash
> 
> Before triggering any failure, I ensure that from the fedora pod on cluster2 I'm able to ping to the fedora pod on cluster3 (install ping utility if its missing).
> 
> I continue to run ping command as shown below to understand the impact of datapath loss.
> i.e., on fedora pod on Cluster2, run the following command
> ping -D <ipAddressOfNetshootPodOnCluster3> -i 0.1
> 
> Now identify the submariner-gateway pod on cluster2 (say it's named as submariner-gateway-8pj66) and delete it (to simulate restart scenario).
> kubectl delete pod submariner-gateway-8pj66 -n submariner-operator
> 
> Wait for the submariner-gateway pod to restart and enter Running state.
> Look at the output of "ping -D ...." in your fedora pod and measure the amount of time where ping did not get any response.
> As you can see, this was a manual test.
> 
>

pinikomarov:
Hi @tpantelis can you help me on this effort ? first time writing e2e tests here so some things are new to me..
test : 
"Submariner Engine Pod restarts (i.e., same node and no failover) on the West cluster.)"

As I see it a code plan might be : 
1.  f.AwaitGatewayFullyConnected - implemented in shipyard
[...]

PR #1061, opened 2021-01-03
Add E2E route agent redundancy tests

pinikomarov:

PR #1084, opened 2021-01-13
Add E2E route agent restart tests

pinikomarov:
Closes https://github.com/submariner-io/submariner/issues/825

### DIFF
--- a/test/e2e/redundancy/route_agent_restart.go
+++ b/test/e2e/redundancy/route_agent_restart.go
@@ -1,0 +1,80 @@
+/*
+© 2021 Red Hat, Inc. and others
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package redundancy
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	"github.com/submariner-io/shipyard/test/e2e/framework"
+	"github.com/submariner-io/shipyard/test/e2e/tcp"
+
+	subFramework "github.com/submariner-io/submariner/test/e2e/framework"
+)
+
+var _ = Describe("[redundancy] Route Agent restart tests", func() {
+	f := subFramework.NewFramework("route-agent-restart")
+
+	When("a route agent pod running on a gateway node is restarted", func() {
+		It("should start a new route agent pod and be able to connect from another cluster", func() {
+			testRouteAgentRestart(f, true)
+		})
+	})
+
+	When("a route agent pod running on a non-gateway node is restarted", func() {
+		It("should start a new route agent pod and be able to connect from another cluster", func() {
+			testRouteAgentRestart(f, false)
+		})
+	})
+})
+
+func testRouteAgentRestart(f *subFramework.Framework, onGateway bool) {
+	clusterAName := framework.TestContext.ClusterIDs[framework.ClusterA]
+	clusterBName := framework.TestContext.ClusterIDs[framework.ClusterB]
+
+	nodes := f.FindNodesByGatewayLabel(framework.ClusterA, onGateway)
+	By(fmt.Sprintf("Found node %q on %q", nodes[0].Name, clusterAName))
+	node := nodes[0]
+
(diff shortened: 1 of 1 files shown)


pinikomarov:
@tpantelis Hi , I corrected what you commented about before.
Two caveats though :

test/e2e/redundancy/route_agent_failover.go
```diff
@@ -0,0 +1,76 @@
...
+	Expect(gatewayNodes).To(HaveLen(1), fmt.Sprintf("Expected only one gateway node on %q", clusterAName))
+	By(fmt.Sprintf("Found gateway on node %q on %q", gatewayNodes[0].Name, clusterAName))
+
+	routeAgentPod := f.AwaitRouteAgentPod(framework.ClusterA)
+	By(fmt.Sprintf("Found route agent pod %q on %q", routeAgentPod.Name, clusterAName))
+
+	By(fmt.Sprintf("Deleting route agent pod %q", routeAgentPod.Name))
+	f.DeletePod(framework.ClusterA, routeAgentPod.Name, framework.TestContext.SubmarinerNamespace)
+
+	newRouteAgentPod := f.AwaitRouteAgentPod(framework.ClusterA)
```
pinikomarov: 1) Here , there is no need for another method:
https://github.com/submariner-io/submariner/pull/1061#discussion_r551628916
Suggested change
        NewRouteAgentPod := f.AwaitRouteAgentPod(framework.ClusterA) ->
        newRouteAgentPod := f.AwaitNewRouteAgentPod(framework.ClusterA)

since I can resuse the f.AwaitRouteAgentPod and put it in a newRouteAgentPod variable.

pinikomarov:
@tpantelis and another concern , this patch won't pass gates untill the :  
https://github.com/submariner-io/shipyard/pull/405/files
PR is merged since this one uses shipyard PR's AwaitRouteAgentPod function.
Am I correct? 
Thanks

tpantelis:
> @tpantelis and another concern , this patch won't pass gates untill the :
> https://github.com/submariner-io/shipyard/pull/405/files
> PR is merged since this one uses shipyard PR's AwaitRouteAgentPod function.
> Am I correct?
> Thanks

This can't be merged until the shipyard PR is merged. That said, I think it would be fine to put that function in submariner since the route agent is specific to submariner, ie I don't see it would be generally useful outside of submariner.

test/e2e/redundancy/route_agent_failover.go
tpantelis: Please see `AwaitNewSubmarinerEnginePod` and how it is used in the other test.

mangelajo: where is AwaitRouteAgentPod defined?

Does it wait for a route agent on any node?, or only the Gateway node?

If it's only in the gateway node, please rename to AwaitRouteAgentPodOnGatewayNode, or make
a function that receives the worker node name: AwaitRouteAgentPodOn(nodeName), as deployments will have as many route agents as nodes.

mangelajo (changes requested):
Some early comments, thanks for working on this.

test/e2e/framework/framework.go
```diff
@@ -247,3 +251,37 @@ func GetDeletionChannel(informer cache.SharedIndexInformer) chan string {
 
 	return deletionChannel
 }
+
+func (f *Framework) AwaitRouteAgentPodOnNode(cluster framework.ClusterIndex, nodeName string, prevPodUID types.UID) *corev1.Pod {
```
sridhargaddam: Thanks for working on it @pinikomarov. 

Suggestion: We already have few APIs for GatewayPod which does similar functionality. Can we create wrappers to those APIs 

For example in Shipyard we can add the following API and use it from Submariner.

// AwaitSubmarinerRouteAgentPod finds the Route Agent pod in a given cluster, waiting if necessary for a period of time
// for the pod to materialize.
func (f *Framework) AwaitSubmarinerRouteAgentPod(cluster ClusterIndex) *v1.Pod {
	return &f.AwaitPodsByAppLabel(cluster, RouteAgent, TestContext.SubmarinerNamespace, 1).Items[0]
}

and another wrapper to check that its a new Pod (similar to what we have in the API AwaitNewSubmarinerEnginePod, file: gateway_failover.go)?

@tpantelis @mangelajo your thoughts?

tpantelis: Actually I'm finishing up this PR. Originally @pinikomarov pushed a PR to add such a method to shipyard - we can do that but I thought it wasn't generally useful outside of submariner since route agent is specific to submariner.

sridhargaddam: As Submariner Engine/Gateway falls in the same category, IMO it's better to keep the APIs together (either in Shipyard or in Submariner repo)

tpantelis: https://github.com/submariner-io/shipyard/pull/440

sridhargaddam (approved):

mangelajo (approved):
Looks good to me

PR #1085, opened 2021-01-14
Configure the base branch for Markdown linting

skitt:
When checking only changed files, the checker needs to know what the
reference branch is; this sets it to release-0.8 for the 0.8 branch.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -62,6 +62,7 @@ jobs:
         with:
           config-file: ".markdownlinkcheck.json"
           check-modified-files-only: "yes"
+          base-branch: "release-0.8"
 
   markdownlint:
     name: Markdown


.github/workflows/linting.yml
```diff
@@ -62,6 +62,7 @@ jobs:
         with:
           config-file: ".markdownlinkcheck.json"
           check-modified-files-only: "yes"
+          base-branch: "release-0.8"
```
mangelajo, line 65: I wonder if we can grab this dynamically from the github context.

mangelajo: https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions

May be we can capture this env var and then use it here?, at least as a patch for the master branch, this one seems fine now.

mangelajo (approved):

.github/workflows/linting.yml
skitt: Ah yes, `$GITHUB_BASE_REF`

skitt: https://github.com/submariner-io/submariner/pull/1086

skitt:
Closed in favour of #1086.

PR #1086, opened 2021-01-14
Use the PR base branch as reference when linting

skitt:
To determine the changed files, the Markdown linter needs to know the
base branch; instead of using the default everywhere, use the base
branch set in the PR.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -72,6 +72,7 @@ jobs:
         with:
           config-file: ".markdownlinkcheck.json"
           check-modified-files-only: "yes"
+          base-branch: ${{ github.base_ref }}
 
   markdownlint:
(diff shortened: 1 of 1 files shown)


.github/workflows/linting.yml
```diff
@@ -72,6 +72,7 @@ jobs:
         with:
           config-file: ".markdownlinkcheck.json"
           check-modified-files-only: "yes"
+	  base-branch: "${GITHUB_BASE_REF}"
```
mkolesnik, line 75: This doesnt seem right

skitt: Yup, none of it was right.

mkolesnik (approved):

mangelajo (approved):

Issue #1091, opened 2021-01-14
Globalnet fails to annotate service without port name


mkimuram:
**What happened**: If service without port name is created, globalnet fails to annotate
service, as a result, the service can't be accessed from another cluster.

**What you expected to happen**: The service is annotated properly and be accessible from another cluster.

**How to reproduce it (as minimally and precisely as possible)**:
Create service without port name in an environment where submariner w/ globalnet is enabled.
```
$ kubectl create ns ns1
$ kubectl run my-pod1 -n ns1
--image=registry.access.redhat.com/ubi7/ubi:latest --command -- python
-m SimpleHTTPServer 80
$ kubectl expose pod my-pod1 -n ns1 --name=my-service1 --port=80
```
Then, check that the service isn't annotated with `submariner.io/globalIP`.
```
$ kubectl get svc -n ns1 my-service1 -o jsonpath='{.metadata.annotations}'
```

**Anything else we need to know?**:

iptables rule for the service is as follows: 
```
# iptables -t nat -nvL --line-numbers | grep my-service1
1        0     0 KUBE-MARK-MASQ  all  --  *      *       10.44.0.12
      0.0.0.0/0            /* ns1/my-service1 */
2        0     0 DNAT       tcp  --  *      *       0.0.0.0/0
0.0.0.0/0            /* ns1/my-service1 */ tcp to:10.44.0.12:80
19       0     0 KUBE-MARK-MASQ  tcp  --  *      *      !10.44.0.0/16
      10.45.55.222         /* ns1/my-service1 cluster IP */ tcp dpt:80
20       0     0 KUBE-SVC-YNXEIGB6TEIVWVFA  tcp  --  *      *
0.0.0.0/0            10.45.55.222         /* ns1/my-service1 cluster IP
*/ tcp dpt:80
1        0     0 KUBE-SEP-ELHPCOKDLRTOJBKE  all  --  *      *
0.0.0.0/0            0.0.0.0/0            /* ns1/my-service1 */
```

`KUBE-SVC-YNXEIGB6TEIVWVFA` chain was created for the service `ns1/my-service1`.
[...]

PR #1092, opened 2021-01-14
Handle empty port name in kubeProxyClusterIpServiceChainName

tpantelis:
Fixes https://github.com/submariner-io/submariner/issues/1091.

### DIFF
--- a/pkg/globalnet/controllers/ipam/globalnet_ingress.go
+++ b/pkg/globalnet/controllers/ipam/globalnet_ingress.go
@@ -48,16 +48,36 @@ func (i *Controller) updateIngressRulesForService(globalIP, chainName string, ad
 	return nil
 }
 
-func (i *Controller) kubeProxyClusterIpServiceChainName(service *k8sv1.Service) string {
+func (i *Controller) kubeProxyClusterIPServiceChainName(service *k8sv1.Service) (string, bool, error) {
 	// CNIs that use kube-proxy with iptables for loadbalancing create an iptables chain for each service
 	// and incoming traffic to the clusterIP Service is directed into the respective chain.
 	// Reference: https://bit.ly/2OPhlwk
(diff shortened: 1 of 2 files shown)


mangelajo (approved):

pkg/globalnet/controllers/ipam/globalnet_ingress.go
```diff
@@ -48,16 +48,35 @@ func (i *Controller) updateIngressRulesForService(globalIP, chainName string, ad
...
-	serviceName := service.GetNamespace() + "/" + service.GetName() + ":" + service.Spec.Ports[0].Name
-	protocol := strings.ToLower(string(service.Spec.Ports[0].Protocol))
-	hash := sha256.Sum256([]byte(serviceName + protocol))
-	encoded := base32.StdEncoding.EncodeToString(hash[:])
+	prefix := service.GetNamespace() + "/" + service.GetName()
+	serviceNames := []string{prefix + ":" + service.Spec.Ports[0].Name}
 
-	return kubeProxyServiceChainPrefix + encoded[:16]
+	if service.Spec.Ports[0].Name == "" {
+		// In newer k8s versions, they omit the ":" if the port name is empty so we need to handle both formats.
```
sridhargaddam, line 59:
```suggestion
		// In newer k8s versions (v1.19+), they omit the ":" if the port name is empty so we need to handle both formats.
		// Reference: https://github.com/kubernetes/kubernetes/commit/341022f8d1361992f5c462d31cd0159a7c2e5a9f#
```

sridhargaddam (commented):
Thanks for pushing the fix @tpantelis 

Can you please include the link to the commit which introduced the change as commented.

pkg/globalnet/controllers/ipam/globalnet_ingress.go
tpantelis: Done

skitt (approved):

sridhargaddam (approved):

PR #1107, opened 2021-01-26
Cleanup additional space and stop removing swap

mangelajo:
There are 30GB+ of files in tools that we don't need or use
for building submariner, we mostly work inside of docker.
Those files are removed now.

Swap isn't disabled anymore, now GitHub uses a separate disk
for swap /dev/sdb instead of the main disk.


skitt (approved):

dfarrell07:
Do we still want this, or are we going with something based on https://github.com/submariner-io/shipyard/pull/422?

mangelajo:
> Do we still want this, or are we going with something based on [submariner-io/shipyard#422](https://github.com/submariner-io/shipyard/pull/422)?

I would merge for now, that will need more work (we also want to pull e2e in general into there).

So we need to merge now so we can unblock ovn-kind testing, (and backport )

vthapar (approved):

PR #1111, opened 2021-01-27
Add tests for Datastoresyncer Node handling

tpantelis:


pkg/controllers/datastoresyncer/datastoresyncer_suite_test.go
```diff
@@ -46,6 +48,7 @@ const (
 	otherClusterID  = "west"
 	localNamespace  = "submariner"
 	brokerNamespace = "submariner-broker"
+	nodeName        = "raiders"
```
dfarrell07, line 51: Raiders fan? :football: :wink:

dfarrell07 (approved):

pkg/controllers/datastoresyncer/datastoresyncer_suite_test.go
tpantelis: yup - after '03 I used to be embarrassed to say that but they've been more respectable lately.

skitt (approved):

PR #1121, opened 2021-01-29
Enable goconst Go linters

Jaanki:
Enable goconst Go linter

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,6 +28,7 @@ linters:
     # - funlen
     # - gochecknoglobals
     # - gochecknoinits
+    - goconst
     - gocritic
     - gocyclo
     - gofmt
--- a/pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
+++ b/pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
@@ -50,6 +50,8 @@ const (
 	localCIDR       = "169.254.1.0/24"
 	remoteCIDR      = "169.254.2.0/24"
 	nodeName        = "raiders"
+	oldIP           = "169.254.1.10"
+	newIP           = "169.254.1.20"
 )
 
 var _ = Describe("Endpoint monitoring", func() {
@@ -119,14 +121,14 @@ var _ = Describe("Service monitoring", func() {
 
 	When("a Service's global IP is updated", func() {
 		BeforeEach(func() {
-			service.Annotations[ipam.SubmarinerIpamGlobalIP] = "169.254.1.10"
+			service.Annotations[ipam.SubmarinerIpamGlobalIP] = oldIP
 		})
 
 		It("should update the appropriate IP tables rule", func() {
 			t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain,
 				ContainSubstring(service.Annotations[ipam.SubmarinerIpamGlobalIP]))
 
-			service.Annotations[ipam.SubmarinerIpamGlobalIP] = "169.254.1.20"
+			service.Annotations[ipam.SubmarinerIpamGlobalIP] = newIP
 			_, err := t.k8sClient.CoreV1().Services(namespace).Update(service)
 			Expect(err).To(Succeed())
 			t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain,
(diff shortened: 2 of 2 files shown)


pkg/cable/strongswan/getconnections.go
```diff
@@ -7,6 +7,9 @@ import (
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 )
 
+const ConnectingState = "CONNECTING"
+const EstablishedState = "ESTABLISHED"
+
```
skitt: The strongSwan code has been removed ;-).

skitt (changes requested):
Please rebase your PR on top of the main branch and resolve the conflicts (the strongSwan cable driver has been removed).

pkg/cable/strongswan/getconnections.go
Jaanki: Oops. Done

skitt (approved):

sridhargaddam (approved):

tpantelis (approved):

.golangci.yml
```diff
@@ -41,6 +43,7 @@ linters:
     - maligned
     - misspell
     - nakedret
+    - scopelint
```
dfarrell07: This is deprecated in favor of exportloopref, which we already have enabled.

https://github.com/kyoh86/scopelint#obsoleted

.golangci.yml
```diff
@@ -29,7 +29,9 @@ linters:
     # - gochecknoinits
     - gocritic
     - gocyclo
+    - goconst
     - gofmt
+    - goheader
```
dfarrell07: We have https://github.com/submariner-io/submariner/blob/master/.github/workflows/linting.yml#L53-L61 to check for legal headers. Maybe we should do a comparison of the two and decide which we want to enable?

.golangci.yml
```diff
@@ -29,7 +29,9 @@ linters:
     # - gochecknoinits
     - gocritic
     - gocyclo
+    - goconst
```
dfarrell07: I'm wondering if this check is a subset of other checks, since it didn't flag anything new. Does it flag anything in any of the other Go-heavy repos?

dfarrell07 (dismissed):
Thanks @Jaanki! Some comments inline.

.golangci.yml
Jaanki: It did send 2 warnings for strongswan code which is now removed. See @skitt's comment above. So this check was not covered before.

dfarrell07: Ah, okay that's good! Agreed that means it's good to enable. :heavy_check_mark:

.golangci.yml
```diff
@@ -11,6 +11,24 @@ linters-settings:
       - unnamedResult
   gocyclo:
     min-complexity: 20
+  goheader:
+    values:
+      regexp:
+        copyright-header: Red Hat, Inc. and others.
```
mangelajo: Will this mean that only "Red Hat, Inc. and others." could be written?

I'm not sure I understand. We could have other contributors

.golangci.yml
mangelajo: Yeah, may be the other is fine already?

.golangci.yml
skitt: Good point, we shouldn’t require this specifically (the existing header linter doesn’t check for this, it looks for the license only).

Jaanki: Its a regex so its not a hard-coded line. There are files that do not include this line and yet linter does not complain. For ex. https://github.com/submariner-io/submariner/blob/devel/pkg/client/listers/submariner.io/v1/endpoint.go

.golangci.yml
Jaanki: yes. I was just thinking to keep all go linters together. However, goheaders is acting wierd and throwing an error which I can't figure out.

`pkg/routeagent_driver/cleanup/cleanup.go:13:31: Missed string: (goheader)
                              ^

pkg/routeagent_driver/constants/constants.go:13:31: Missed string: (goheader)
                              ^

pkg/routeagent_driver/environment/env.go:13:31: Missed string: (goheader)
                              ^

make: *** [/opt/shipyard/Makefile.inc:65: golangci-lint] Error 1
[submariner]$ make golangci-lint
time="2021-02-09T10:40:59Z" level=fatal msg="exit status 2"
make: *** [golangci-lint] Error 1
Makefile.dapper:14: recipe for target 'golangci-lint' failed
Error: Process completed with exit code 2.`

Reverting to using GitHub workflows for license checks.

mangelajo (approved):

skitt (approved):

Issue #1106, opened 2021-01-25
ip route loop when the service/pod CIDR is shared with the worker node CIDRs


mangelajo:
**What happened**:

As it can be seen in the following example, the di-aks-poc-2 cluster is trying to connect to the remote gateway on 10.43.0.4, but
that overlaps with the pod CIDR 10.43.0.0/23 .
```
Showing information for cluster "di-aks-poc-2":
GATEWAY                         CLUSTER                 REMOTE IP       CABLE DRIVER        SUBNETS                                 STATUS          
aks-agentpool-12291320-vmss0000 di-aks-poc-1            10.43.0.4       libreswan           10.0.0.0/18, 10.43.0.0/23               connecting      
```

Submariner-route agent creates routing rules to target the remote pods from the host-network of the gateway (and all other nodes). But that will conflict with the remote IP of the gateway. So the IPSEC, or any other connection can't be established, it's not possible anymore.

After we flush table 150 on both gateways:

```
[majopela@bluehat BrandonWitter]$ subctl show connections --kubecontext di-aks-poc-1-admin

Showing information for cluster "di-aks-poc-1":
GATEWAY                         CLUSTER                 REMOTE IP       CABLE DRIVER        SUBNETS                                 STATUS          
aks-agentpool-35329906-vmss0000 di-aks-poc-2            10.43.2.4       libreswan           10.1.0.0/18, 10.43.2.0/23               connected     
```

**What you expected to happen**:

connection working, and traffic possible between the gateways.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

Workaround:

```
[majopela@bluehat BrandonWitter]$ CONTEXT="di-aks-poc-2-admin" ./snetshoot-hostnet.sh aks-agentpool-35329906-vmss000000
If you don't see a command prompt, try pressing enter.
bash-5.0# ip route flush table 150
bash-5.0# exit
exit
Session ended, resume using 'kubectl attach netshoot-hostmount-bbdfe80c-385c-4ffc-8f83-f3a75ab1af07 -c netshoot-hostmount -i -t' command when the pod is running
pod "netshoot-hostmount-bbdfe80c-385c-4ffc-8f83-f3a75ab1af07" deleted
[majopela@bluehat BrandonWitter]$ CONTEXT="di-aks-poc-1-admin" ./snetshoot-hostnet.sh 
please provide a node name as the first argument

[...]

PR #1122, opened 2021-01-29
Setup a via parameter in route for remote CIDRS overlapping to gw

mangelajo:
Setup a via parameter in route for remote CIDRS overlapping to gw

Fixes-Issue: #1106


sridhargaddam (approved):

skitt (approved):

mangelajo:
Thanks, @skitt @sridhargaddam , moved it to draft.

I was speaking to @sridhargaddam , and while this fixes the issue (just tested it and made sure it works), if this is encountered it seems that connectivity for those CIDRs is already available, so submariner is probably not necessary.

We will add a warning in such case.

It could be the case that the CIDR connectivity is only available on the gateway, that would be more unlikely. We can always remove such warning if it's reported in the future.

mangelajo:
@skitt @sridhargaddam I was adding something like this:

```go
klog.Warning("the IP address %q of the remote endpoint connecting to %q seem to overlap, this indicates "+
					"that you are using routed pod networks, at least between the gateways. This means that probably "+
					"the routing and tunneling that submariner provides is not necessary.", gwIP, inputCidrBlock)
```

But may be there are use cases, where:
1) The routed network between the clusters is considered to be insecure, and encryption is expected
2) There are more participating clusters which aren't routed.. ?

sridhargaddam:
> @skitt @sridhargaddam I was adding something like this:
> 
> ```go
> klog.Warning("the IP address %q of the remote endpoint connecting to %q seem to overlap, this indicates "+
> 					"that you are using routed pod networks, at least between the gateways. This means that probably "+
> 					"the routing and tunneling that submariner provides is not necessary.", gwIP, inputCidrBlock)
> ```
> 
> But may be there are use cases, where:
> 
> 1. The routed network between the clusters is considered to be insecure, and encryption is expected
> 2. There are more participating clusters which aren't routed.. ?

IMHO, it's hard to visualize all possible cases. So we can add this Warning as a starting point and if we come across any issues we can revisit accordingly.

PR #1123, opened 2021-02-01
Enale below linters

Jaanki:
1. prealloc
2. dogsled


.golangci.yml
```diff
@@ -21,6 +21,7 @@ linters:
     - bodyclose
     - deadcode
     - depguard
+    - dogsled
```
dfarrell07, line 24: I think this is a subset of other, already enabled, linters. I think remember having to fix all the "too many blank identifiers" while enabling another linter.

.golangci.yml
```diff
@@ -41,6 +42,7 @@ linters:
     - maligned
     - misspell
     - nakedret
+    - prealloc
```
dfarrell07, line 45: I see this one doesn't flag anything. I wonder if it flags anything in any of our other Go-heavy repos? If not, I wonder if it's worth the runtime (or if it's already covered by other linting).

dfarrell07 (changes requested):
I'd suggest a commit something like "Enable x, y Go linters", see: https://submariner.io/development/code-review/

.golangci.yml
mangelajo: May be it doesn't flag now but could flag in the future?

mangelajo (approved):

.golangci.yml
Jaanki: That must be blank lines warning by wsl which was excluded https://github.com/submariner-io/submariner/blob/devel/.golangci.yml#L57

dogsled is for too may blank identifiers in assigments/declarations https://github.com/alexkohler/dogsled

tpantelis (approved):

.golangci.yml
dfarrell07: Okay, well I guess I don't want to argue for fewer linters. :wink:

.golangci.yml
```diff
@@ -31,7 +32,7 @@ linters:
     - gocyclo
     - gofmt
     - goimports
-    # - golint
+    #- golint
```
dfarrell07: Do we need this change?

dfarrell07 (commented):
Can we also revise the commit message per my previous feedback?

Jaanki:
> Can we also revise the commit message per my previous feedback?

@dfarrell07 Done :).

dfarrell07 (approved):

PR #1124, opened 2021-02-02
ServiceExport support for globalnet

### LINKED ISSUE #720: Aligning Globalnet to LightHouse ServiceExport objects
Currently, Globalnet annotates every service in a Cluster. This enhancement request is to modify globalnet implementation to annotate only exported services and NOT every service. This will make Globalnet implementation aligned with latest Lighthouse code as well.

vthapar:
Fixes #720 


vthapar:
Will add E2E in subsequent commit, for easier reviews.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -98,6 +106,14 @@ func NewController(spec *SubmarinerIpamControllerSpecification, config *Informer
 		},
 		DeleteFunc: ipamController.handleRemovedNode,
 	}, handlerResync)
+	config.SvcExInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			ipamController.enqueueObject(obj, ipamController.svcExWorkqueue)
+		},
+		UpdateFunc: func(old, newObj interface{}) {
```
sridhargaddam: nil?

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -469,6 +493,67 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
...
+	if globalIp, ok := annotations[submarinerIpamGlobalIp]; ok {
+		if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+			utilruntime.HandleError(err)
+			return
+		}
+
+		err = i.syncServiceRules(svc, globalIp, DeleteRules)
+		if err != nil {
+			klog.Errorf("Error while cleaning up Service %q ingress rules. %v", key, err)
+			logAndRequeue(key, i.svcExWorkqueue)
```
sridhargaddam: As this is a delete event, if we requeue the event, will we have access to the ServiceExport Object when the event is retriggered? Can you please check the behavior?

pkg/globalnet/controllers/ipam/globalnet.go
```diff
@@ -127,6 +128,14 @@ func (i *Controller) isServiceSupported(service *k8sv1.Service) bool {
 		return false
 	}
 
+	_, err := i.svcExGetter(service.Namespace, service.Name)
+	if errors.IsNotFound(err) {
+		return false
+	} else if err != nil {
+		klog.Errorf("Failed to get ServiceExport for %#v due to %v", service, err)
```
sridhargaddam: As we have listeners for Service objects as well (which may not be exported), lets not print it as an "Errorf", it can be an "Infof" with a message that "ServiceExport is missing for Service, so skipping it"

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -469,6 +493,67 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
...
+
+		i.pool.Release(key)
+		klog.V(log.DEBUG).Infof("Released ip %s for service %s", globalIp, key)
+
+		delete(annotations, submarinerIpamGlobalIp)
+		svc.SetAnnotations(annotations)
+
+		_, err = i.kubeClientSet.CoreV1().Services(svc.Namespace).Update(svc)
+		if err != nil {
+			logAndRequeue(key, i.svcExWorkqueue)
```
sridhargaddam: Instead of logAndRequeue, shall we use retry.RetryOnConflict?

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -607,6 +696,26 @@ func (i *Controller) podUpdater(obj runtime.Object, key string) error {
...
+
+	unstructObj := obj.(*unstructured.Unstructured).UnstructuredContent()
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructObj, &svcEx)
+	if err != nil {
+		klog.Errorf("Failed to extract ServiceExport from unstructured Object %#v", unstructObj)
+	}
+
+	svc, err := i.serviceGetter(svcEx.Namespace, svcEx.Name)
+	if err != nil && !errors.IsNotFound(err) {
+		klog.Errorf("Failed to get service %s/%s", svcEx.Namespace, svcEx.Name)
```
sridhargaddam: I think this should be an Infof

sridhargaddam (changes requested):
Nice work @vthapar

pkg/globalnet/controllers/ipam/globalnet.go
vthapar: If service not exported it will be handled by IsNotFound. This error means some other API error.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: It is difficult to reproduce this. That is why added an Error log. If we hit issues due to this, we'll know from logs and can revisit this. Am still thinking of a way to handle the scenario even if we don't have object anymore.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: Same as above. Error is something other than NotFound.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: Didn't know we can use nil, will try that.

pkg/globalnet/controllers/ipam/globalnet.go
tpantelis: I think in this case we want to re-queue so I suggest moving this to `evaluateService` and return `Ignore`/`Requeue` appropriately.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -55,14 +57,20 @@ func NewController(spec *SubmarinerIpamControllerSpecification, config *Informer
 		return nil, err
 	}
 
+	config.ServiceInformer.Lister()
```
tpantelis: What is the purpose of this call?

pkg/globalnet/controllers/ipam/ipam.go
tpantelis: You can omit it altogether - `ResourceEventHandlerFuncs` provides a stub.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -607,6 +682,26 @@ func (i *Controller) podUpdater(obj runtime.Object, key string) error {
 	return nil
 }
 
+func (i *Controller) svcExUpdater(obj runtime.Object, key string) error {
+	var svcEx mcsv1a1.ServiceExport
+
+	unstructObj := obj.(*unstructured.Unstructured).UnstructuredContent()
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructObj, &svcEx)
+	if err != nil {
+		klog.Errorf("Failed to extract ServiceExport from unstructured Object %#v", unstructObj)
```
tpantelis: This is likely an unrecoverable error so no reason to re-queue. However we don't really need to convert it. You only need to access `Namespace`  and `Name` which are easily available on `Unstructured`.

```suggestion
		klog.Errorf("Failed to extract ServiceExport from unstructured Object %#v", unstructObj)
		return nil
```

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -469,6 +493,53 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
...
+			svc.SetAnnotations(annotations)
+
+			_, err = i.kubeClientSet.CoreV1().Services(svc.Namespace).Update(svc)
+			return err
+		}
+		return nil
+	})
+
+	if retryErr != nil {
+		logAndRequeue(key, i.svcExWorkqueue)
```
tpantelis: This is a delete event so re-queueing would have no effect.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -469,6 +493,53 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
 	}
 }
 
+func (i *Controller) handleRemovedSvcEx(obj interface{}) {
+	var key, name, namespace string
+	var err error
+
```
tpantelis, line 488: Need to check for tombstone here.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -469,6 +493,53 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
...
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+
+	if namespace, name, err = cache.SplitMetaNamespaceKey(key); err != nil {
```
tpantelis, line 494: `obj` is an `Unstructured` so you can obtain namespace and name directly.

tpantelis (changes requested):

pkg/globalnet/controllers/ipam/ipam.go
vthapar: Missed cleanup.

pkg/globalnet/controllers/ipam/globalnet.go
vthapar: > I think in this case we want to re-queue so I suggest moving this to `evaluateService` and return `Ignore`/`Requeue` appropriately.

This is called from `handleRemovedService` too. If we move it to `evaluateService` we will lose this check in `handleRemovedService`

pkg/globalnet/controllers/ipam/ipam.go
vthapar: Yeah, I have done a similar change. We don't even need Unstructured. We can get name+namespace from key itself.

pkg/globalnet/controllers/ipam/ipam.go
vthapar: I did this way so that we don't even need to check for tombstone. We re-queue with the key and if service or annotation already cleaned up we stop requeuing.

pkg/globalnet/controllers/ipam/globalnet.go
tpantelis: yeah and `handleUpdateService`. For `handleUpdateService`, I don't think we want to do this check prior to queueing. We want such checks to be quick.  Plus it checks if the `GlobalIP`  changed which wouldn't be the case if not previously exported. For `handleRemovedService`, `isServiceSupported` is only called to log a message which I don't think is necessary.

vthapar: Right, `processNextObject` already calls `evaluateService` before calling `handleUpdateService` so calling `isServiceSupported` again is kinda redundant. Moved this to `evaluateService` and agree it works better.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
```diff
@@ -176,6 +184,52 @@ var _ = Describe("Service monitoring", func() {
...
+
+	BeforeEach(func() {
+		service = newService("nginx")
+	})
+
+	JustBeforeEach(func() {
+		t.createEndpoint(newEndpointSpec(clusterID, t.hostName, localCIDR))
+	})
+
+	When("a Service without a global IP is Exported", func() {
```
tpantelis: This is the same test case as L121 above.  It's also technically the same as L209.

pkg/globalnet/controllers/ipam/ipam.go
tpantelis: Missed cleanup 😄

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -463,6 +484,53 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
 	}
 }
 
+func (i *Controller) handleRemovedSvcEx(obj interface{}) {
+	var key, name, namespace string
+	var err error
+
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
```
tpantelis:
```suggestion
	if key, err = cache.DeletionHandlingMetaNamespaceKeyFunc(obj); err != nil {
```

pkg/globalnet/controllers/ipam/ipam.go
tpantelis: bump

pkg/globalnet/main.go
```diff
@@ -73,6 +76,16 @@ func main() {
...
 	}
 
+	dynClientSet, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		klog.Fatalf("Error building dynamic clientset: %s", err.Error())
+	}
+
+	err = mcsv1a1.AddToScheme(scheme.Scheme)
+	if err != nil {
+		klog.Exitf("Error adding Multicluster v1alpha1 to the scheme: %v", err)
```
tpantelis: ?

```suggestion
		klog. Fatalf("Error adding Multicluster v1alpha1 to the scheme: %v", err)
```

tpantelis (changes requested):

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
```diff
@@ -176,6 +184,52 @@ var _ = Describe("Service monitoring", func() {
...
+	When("a Service Export is created before service", func() {
+		It("should assign a global IP and add an appropriate IP tables rule after service creation", func() {
+			t.createSvcExport(service)
+			t.ipt.AwaitNoRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIPSubstr))
+
+			t.createService(service)
+			globalIP := t.awaitServiceGlobalIP(service.Name)
+			t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))
+		})
+	})
```
tpantelis, line 220: We can combine these test cases with "_Service monitoring_" above since `Service` and `ServiceExport` are closely tied. Also we need a test for `ServiceExport` delete.

```
When("a Service without a global IP is created", func() {
    JustBeforeEach(func() {
        t.createEndpoint(newEndpointSpec(clusterID, t.hostName, localCIDR))
    })

    When("the Service was already exported", func() {
        It("should assign a global IP and add an appropriate IP tables rule", func() {
            t.createSvcExport(service)
            t.ipt.AwaitNoRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIPSubstr))

            t.createService(service)
            ...
        })
    })

    When("the Service is subsequently exported", func() {
        It("should eventually assign a global IP and add an appropriate IP tables rule", func() {
            service = t.createService(service)
            t.awaitNoServiceGlobalIP(service)
            t.ipt.AwaitNoRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIPSubstr))

            t.createSvcExport(service)
            ...
        })
    })
})

Context("", func() {
    JustBeforeEach(func() {
        t.createSvcExport(service)
        service = t.createService(service)
        t.createEndpoint(newEndpointSpec(clusterID, t.hostName, localCIDR))
    })

    When("a Service with a global IP is unexported", func() {
        It("should unassign its global IP and remove the appropriate IP tables rule", func() {
            globalIP := t.awaitServiceGlobalIP(service.Name, localCIDR) 
            t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))

            // delete ServiceExport
	    t.ipt.AwaitNoRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))
            t.awaitNoServiceGlobalIP(service)
        })
    })

    When("a Service's global IP is updated", func() { ... the rest of the tests
})
```

tpantelis (changes requested):

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
sridhargaddam: We can modify the test-case to create ServiceExport first and later create the necessary service.

sridhargaddam: Oh, the test-case is already present in L221. So, I agree with @tpantelis  this is kind of redundant.

pkg/globalnet/controllers/ipam/globalnet.go
```diff
@@ -135,6 +136,14 @@ func (i *Controller) evaluateService(service *k8sv1.Service) Operation {
 		return Ignore
 	}
 
+	_, err := i.svcExGetter(service.Namespace, service.Name)
+	if errors.IsNotFound(err) {
+		return Ignore
+	} else if err != nil {
+		klog.Errorf("Failed to get ServiceExport for %#v due to %v", service, err)
+		return Requeue
```
sridhargaddam, line 144: I think we should simply return Ignore here if ServiceExport object is missing. For example, say the user created a regular ClusterIPService and is not interested in exporting it, so we will be requeuing it again and again. Am I missing something?

sridhargaddam (changes requested):

pkg/globalnet/controllers/ipam/globalnet.go
tpantelis: We return `Ignore` for `IsNotFound` above.

sridhargaddam: Okay, that makes sense. I missed the IsNotFound() check.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
vthapar: Yeah, was planning to merge them once I figured out test cases and got them working. Find it easier to write separate and just run that test case when writing new ones. Thanks for easy to copy-paste refactor :D

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -54,14 +55,20 @@ func NewController(spec *SubmarinerIpamControllerSpecification, config *Informer
 		return nil, err
 	}
 
+	config.ServiceInformer.Lister()
```
tpantelis: Please remove this line

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
```diff
@@ -104,74 +111,128 @@ var _ = Describe("Service monitoring", func() {
...
+			t.createEndpoint(newEndpointSpec(clusterID, t.hostName, localCIDR))
 		})
 
-		It("should not assign a global IP", func() {
-			t.awaitNoServiceGlobalIP(service)
+		When("a Service without a global IP is created", func() {
+			It("should assign a global IP and add an appropriate IP tables rule", func() {
+				t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain,
+					ContainSubstring(t.awaitServiceGlobalIP(service.Name)))
+			})
```
tpantelis: This is redundant with the spec on L119.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
```diff
@@ -104,74 +111,128 @@ var _ = Describe("Service monitoring", func() {
...
 
-			service.Annotations[ipam.SubmarinerIpamGlobalIp] = "169.254.1.20"
-			_, err := t.k8sClient.CoreV1().Services(namespace).Update(service)
-			Expect(err).To(Succeed())
-			t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain,
-				ContainSubstring(service.Annotations[ipam.SubmarinerIpamGlobalIp]))
+				t.createService(service)
+				globalIP := t.awaitServiceGlobalIP(service.Name)
+				t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))
+			})
```
tpantelis: This is the same as the spec on L119.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
```diff
@@ -104,74 +111,128 @@ var _ = Describe("Service monitoring", func() {
...
+				service = t.createService(service)
+				t.createSvcExport(service)
+				globalIP := t.awaitServiceGlobalIP(service.Name)
+				t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))
 
-			Expect(t.k8sClient.CoreV1().Services(namespace).Delete(service.Name, nil)).To(Succeed())
-			t.ipt.AwaitNoRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))
+				Expect(t.dynClient.Resource(*t.svcExGvr).Namespace(service.Namespace).Delete(service.Name, nil)).To(Succeed())
+				t.ipt.AwaitNoRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIPSubstr))
+			})
```
tpantelis: I suggested to put this in the next parent spec below after L171 since it relies on the service already assigned, like the update case. Saves some redundant code that is done in `JustBeforeEach` on L167:

```
    When("a Service with a global IP is unexported", func() {
        It("should unassign its global IP and remove the appropriate IP tables rule", func() {
            globalIP := t.awaitServiceGlobalIP(service.Name, localCIDR) 
            t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))

            Expect(t.dynClient.Resource(*t.svcExGvr).Namespace(service.Namespace).Delete(service.Name, nil)).To(Succeed())
	    t.ipt.AwaitNoRule("nat", constants.SmGlobalnetIngressChain, ContainSubstring(globalIP))
            t.awaitNoServiceGlobalIP(service)
        })
    })


```

tpantelis (changes requested):

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
vthapar: > This is the same as the spec on L119.

No, the order is different. L119 creates Service > ServiceExport. This is ServiceExport > Service. This is to make sure things work if you create in either order.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
vthapar: Done.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
tpantelis: L119 does do ServiceExport > Service.  L130 does Service > ServiceExport.

tpantelis (changes requested):

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
```diff
@@ -104,74 +111,126 @@ var _ = Describe("Service monitoring", func() {
...
+				Expect(err).To(Succeed())
+				t.ipt.AwaitRule("nat", constants.SmGlobalnetIngressChain,
+					ContainSubstring(service.Annotations[ipam.SubmarinerIpamGlobalIp]))
+			})
 		})
 
-		It("should not assign a global IP", func() {
-			t.awaitNoServiceGlobalIP(service)
+		When("the Service Export is deleted", func() {
+			It("should eventually remove global IP and the IP tables rule", func() {
```
tpantelis: I know this is a nit but with my suggested wording below, it implies that the service was previously exported.

```suggestion
		When("a Service with a global IP is unexported", func() {
                     It("should unassign its global IP and remove the appropriate IP tables rule", func() {
```

tpantelis (changes requested):

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
vthapar: L130 has an extra test that IP Tables rule is not added if service is created without ServiceExport. That test is not present in L142 coz we already created everything in JustBeforeEach()

EDIT: nvm, @sridhargaddam clarified it. L119 is duplicate of L142.

sridhargaddam: @tpantelis is talking about test-case in L142 (not L130). I see that they are repeated :)

sridhargaddam (changes requested):
Also, there is a linter error at 
pkg/globalnet/controllers/ipam/gatewaymonitor_test.go:186: 

please take a look.

vthapar:
@tpantelis @sridhargaddam New linting patch got merged and rebase was messy. Do take another look to make sure I didn't end up missing anything.

go.mod
```diff
@@ -6,27 +6,24 @@ require (
...
-	github.com/submariner-io/shipyard v0.8.1-0.20210204102811-7d5759c94fae
+	github.com/submariner-io/shipyard v0.8.1-0.20210205083436-029e7183dbea
 	github.com/vishvananda/netlink v1.1.0
-	go.uber.org/zap v1.15.0 // indirect
 	golang.org/x/sys v0.0.0-20210112080510-489259a85091
 	golang.zx2c4.com/wireguard/wgctrl v0.0.0-20200324154536-ceff61240acf
-	k8s.io/api v0.17.0
-	k8s.io/apimachinery v0.17.0
-	k8s.io/client-go v0.17.0
+	k8s.io/api v0.18.4
```
mangelajo, line 21: I wonder why does it change here (and in lighthouse)... if it's all pinned (L32) :?

mangelajo (approved):
I thought the plan was doing this as part of globalnetv2, I won't block it since the work is now done :-) but I'm confused :)

vthapar:
> I thought the plan was doing this as part of globalnetv2, I won't block it since the work is now done :-) but I'm confused :)

Probably coz it is independent enough of rest of v2 changes :) I wanted this early in 0.9.0 cycle so it gets more time for testing.

go.mod
skitt: The MCS API dependency bumps these transitively, but the overrides win ultimately.

sridhargaddam:
> > I thought the plan was doing this as part of globalnetv2, I won't block it since the work is now done :-) but I'm confused :)
> 
> Probably coz it is independent enough of rest of v2 changes :) I wanted this early in 0.9.0 cycle so it gets more time for testing.

True. This is kind of independent change and might require some modifications as we make other Globalnet v2.0, but we can get this in as it works fine with the current codebase.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
tpantelis: Correction: this one is redundant with the spec on L130 - creates Service then ServiceExport..

tpantelis:
This looks good from my end - just one more comment re: redundant test.

pkg/globalnet/controllers/ipam/gatewaymonitor_test.go
vthapar: Done

vthapar:
> This looks good from my end - just one more comment re: redundant test.

Done. Please do squash before merge.

sridhargaddam (approved):

tpantelis (approved):

PR #1125, opened 2021-02-02
Enable stylecheck linter

Jaanki:
Correct the variable and package names to be
compatible with stylecheck recommendations.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,7 +32,6 @@ linters:
     - gocyclo
     - gofmt
     - goimports
-    # - golint
     - gosec
     - gosimple
     - govet
@@ -45,7 +44,7 @@ linters:
     - prealloc
     - staticcheck
     - structcheck
-    # - stylecheck
+    - stylecheck # instead of golint
     # - testpackage
     - typecheck
     - unconvert
@@ -72,3 +71,14 @@ issues:
     - linters:
         - wsl
       text: "only one cuddle assignment allowed before range statement"
+
+    # Allow dot imports for `fake` directory and e2e tests
+    - linters:
+        - stylecheck
+      text: "ST1001: "
+      path: /fake
+
+    - linters:
+        - stylecheck
+      text: "ST1001: "
+      path: test/e2e
--- a/main.go
+++ b/main.go
@@ -88,7 +88,7 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	httpServer := startHttpServer()
+	httpServer := startHTTPServer()
 
 	var submSpec types.SubmarinerSpecification
 	err := envconfig.Process("submariner", &submSpec)
@@ -236,7 +236,7 @@ func submarinerClusterFrom(submSpec *types.SubmarinerSpecification) types.Submar
 	}
 }
 
-func startHttpServer() *http.Server {
+func startHTTPServer() *http.Server {
 	srv := &http.Server{Addr: ":8080"}
 
 	http.Handle("/metrics", promhttp.Handler())
@@ -256,7 +256,7 @@ func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder res
 
 	err := envconfig.Process(leadershipConfigEnvPrefix, &gwLeadershipConfig)
 	if err != nil {
-		return fmt.Errorf("Error processing environment config for %s: %v", leadershipConfigEnvPrefix, err)
+		return fmt.Errorf("error processing environment config for %s: %v", leadershipConfigEnvPrefix, err)
 	}
 
 	// Use default values when GatewayLeadership environment variables are not configured
@@ -276,7 +276,7 @@ func startLeaderElection(leaderElectionClient kubernetes.Interface, recorder res
 
 	id, err := os.Hostname()
 	if err != nil {
-		return fmt.Errorf("Error getting hostname: %v", err)
+		return fmt.Errorf("error getting hostname: %v", err)
 	}
 
 	kubeconfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
--- a/pkg/cable/cleanup/xfrm_cleanup_handler.go
+++ b/pkg/cable/cleanup/xfrm_cleanup_handler.go
@@ -40,7 +40,7 @@ func (xc *XFRMCleanupHandler) NonGatewayCleanup() error {
 	currentXfrmPolicyList, err := netlink.XfrmPolicyList(syscall.AF_INET)
 
 	if err != nil {
-		return fmt.Errorf("Error retrieving current xfrm policies: %v", err)
+		return fmt.Errorf("error retrieving current xfrm policies: %v", err)
 	}
 
 	if len(currentXfrmPolicyList) > 0 {
@@ -51,7 +51,7 @@ func (xc *XFRMCleanupHandler) NonGatewayCleanup() error {
 		klog.V(log.DEBUG).Infof("Deleting XFRM policy %s", currentXfrmPolicyList[i])
 
 		if err = netlink.XfrmPolicyDel(&currentXfrmPolicyList[i]); err != nil {
-			return fmt.Errorf("Error Deleting XFRM policy %s: %v", currentXfrmPolicyList[i], err)
+			return fmt.Errorf("error Deleting XFRM policy %s: %v", currentXfrmPolicyList[i], err)
 		}
 	}
 
--- a/pkg/cableengine/syncer/syncer.go
+++ b/pkg/cableengine/syncer/syncer.go
@@ -70,56 +70,56 @@ func NewGatewaySyncer(engine cableengine.Engine, client v1typed.GatewayInterface
 	}
 }
 
-func (s *GatewaySyncer) Run(stopCh <-chan struct{}) {
+func (gs *GatewaySyncer) Run(stopCh <-chan struct{}) {
 	go func() {
-		wait.Until(s.syncGatewayStatus, GatewayUpdateInterval, stopCh)
-		s.CleanupGatewayEntry()
+		wait.Until(gs.syncGatewayStatus, GatewayUpdateInterval, stopCh)
+		gs.CleanupGatewayEntry()
 	}()
 
 	klog.Info("CableEngine syncer started")
 }
 
-func (i *GatewaySyncer) syncGatewayStatus() {
-	i.Lock()
-	defer i.Unlock()
+func (gs *GatewaySyncer) syncGatewayStatus() {
+	gs.Lock()
+	defer gs.Unlock()
 
-	i.syncGatewayStatusSafe()
+	gs.syncGatewayStatusSafe()
 }
 
-func (i *GatewaySyncer) SetGatewayStatusError(err error) {
-	i.Lock()
-	defer i.Unlock()
+func (gs *GatewaySyncer) SetGatewayStatusError(err error) {
+	gs.Lock()
+	defer gs.Unlock()
 
-	i.statusError = err
-	i.syncGatewayStatusSafe()
+	gs.statusError = err
+	gs.syncGatewayStatusSafe()
 }
 
-func (i *GatewaySyncer) syncGatewayStatusSafe() {
+func (gs *GatewaySyncer) syncGatewayStatusSafe() {
 	klog.V(log.TRACE).Info("Running Gateway status sync")
 	gatewaySyncIterations.Inc()
 
-	gatewayObj := i.generateGatewayObject()
+	gatewayObj := gs.generateGatewayObject()
 
-	existingGw, err := i.getLastSyncedGateway(gatewayObj.Name)
+	existingGw, err := gs.getLastSyncedGateway(gatewayObj.Name)
 
 	if errors.IsNotFound(err) {
 		klog.V(log.TRACE).Infof("Gateway does not exist - creating: %+v", gatewayObj)
-		_, err = i.client.Create(gatewayObj)
+		_, err = gs.client.Create(gatewayObj)
 		if err != nil {
-			utilruntime.HandleError(fmt.Errorf("Error creating Gateway object %+v: %s", gatewayObj, err))
+			utilruntime.HandleError(fmt.Errorf("error creating Gateway object %+v: %s", gatewayObj, err))
 			return
 		}
 	} else if err != nil {
-		utilruntime.HandleError(fmt.Errorf("Error getting existing Gateway: %s", err))
+		utilruntime.HandleError(fmt.Errorf("error getting existing Gateway: %s", err))
 		return
 	} else if !reflect.DeepEqual(gatewayObj.Status, existingGw.Status) {
 		klog.V(log.TRACE).Infof("Gateway already exists - updating %+v", gatewayObj)
 		existingGw.Status = gatewayObj.Status
 		existingGw.Annotations = gatewayObj.Annotations
 
-		_, err := i.client.Update(existingGw)
+		_, err := gs.client.Update(existingGw)
 		if err != nil {
-			utilruntime.HandleError(fmt.Errorf("Error updating Gateway object %+v: %s", gatewayObj, err))
+			utilruntime.HandleError(fmt.Errorf("error updating Gateway object %+v: %s", gatewayObj, err))
 			return
 		}
 	} else {
@@ -127,15 +127,15 @@ func (i *GatewaySyncer) syncGatewayStatusSafe() {
 	}
 
 	if gatewayObj.Status.HAStatus == v1.HAStatusActive {
-		err := i.cleanupStaleGatewayEntries(gatewayObj.Name)
+		err := gs.cleanupStaleGatewayEntries(gatewayObj.Name)
 		if err != nil {
-			utilruntime.HandleError(fmt.Errorf("Error cleaning up stale gateway entries: %s", err))
+			utilruntime.HandleError(fmt.Errorf("error cleaning up stale gateway entries: %s", err))
 		}
 	}
 }
 
-func (i *GatewaySyncer) cleanupStaleGatewayEntries(localGatewayName string) error {
-	gateways, err := i.client.List(metav1.ListOptions{})
+func (gs *GatewaySyncer) cleanupStaleGatewayEntries(localGatewayName string) error {
+	gateways, err := gs.client.List(metav1.ListOptions{})
 	if err != nil {
 		return err
 	}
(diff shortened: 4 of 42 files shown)


.golangci.yml
```diff
@@ -70,3 +69,12 @@ issues:
     - linters:
         - wsl
       text: "only one cuddle assignment allowed before range statement"
+
+    # These are stylecheck warnings that can be suppressed. This is temporary and will be deleted in further commits.
```
dfarrell07: This is a good strategy for breaking up very large changes like this, fixing issues in different commits per-issue-type.

In general, I found it useful to:

1. First commit to enable the comitter but add ignores for all failing rules
2. Loop over each rule, enabling it and fixing the issues it flags in a separate commit.

main.go
```diff
@@ -88,7 +88,7 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	httpServer := startHttpServer()
+	httpServer := starthttpServer()
```
tpantelis: hmm... this seems odd to me that the linter would want lower case. According to [guidelines](https://github.com/golang/go/wiki/CodeReviewComments#initialisms), protocols et al should be capitalized, so `startHTTPServer`.

pkg/globalnet/controllers/ipam/constants.go
```diff
@@ -21,7 +21,7 @@ type Operation string
 
 const (
 	handlerResync          = time.Hour * 24
-	submarinerIpamGlobalIp = "submariner.io/globalIp"
+	submarinerIpamGlobalIP = "submariner.io/globalIP"
```
tpantelis: We can't change the label name w/o changing it elsewhere and causing backwards compatibility and migration issues.  We should leave "_globalIp_" as is.

pkg/cableengine/syncer/syncer.go
```diff
@@ -70,41 +70,41 @@ func NewGatewaySyncer(engine cableengine.Engine, client v1typed.GatewayInterface
 	}
 }
 
-func (s *GatewaySyncer) Run(stopCh <-chan struct{}) {
+func (gws *GatewaySyncer) Run(stopCh <-chan struct{}) {
```
tpantelis: Usually 1 or 2 letters suffices as outlined [here](https://github.com/golang/go/wiki/CodeReviewComments#receiver-names).

```suggestion
func (gs *GatewaySyncer) Run(stopCh <-chan struct{}) {
```

pkg/globalnet/main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
 
-	httpServer := startHttpServer()
+	httpServer := starthttpServer()
```
tpantelis:
```suggestion
	httpServer := startHTTPServer()
```

pkg/routeagent_driver/cniinterface/cni_iface.go
```diff
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package cni_interface
+package cniinterface
```
tpantelis: Since _interface_ is already part of CNI:

```suggestion
package cni
```

pkg/routeagent_driver/constants/constants.go
```diff
@@ -36,10 +36,10 @@ const (
 	// part of route-agent and this will subsequently be used in globalnet controller for
 	// supporting connectivity from HostNetwork to remoteClusters.
 	// [#] interface on the node that has an IPAddress from the clusterCIDR
-	CniInterfaceIP = "submariner.io/cniIfaceIp"
+	CniInterfaceIP = "submariner.io/cniIfaceIP"
 
 	// TODO (revisit): Unify this constant across different components
-	SmGlobalIP = "submariner.io/globalIp"
+	SmGlobalIP = "submariner.io/globalIP"
```
tpantelis: As above we shouldn't change label names.

pkg/routeagent_driver/handlers/kubeproxyiptables/cleanup_handling.go
```diff
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package kubeproxy_iptables
+package kubeproxyiptables
```
tpantelis: I think the '_' makes it more readable.  In lieu of that, how about we just make it _kubeproxy_?

pkg/util/clusterfiles/cluster_files.go
```diff
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package cluster_files
+package clusterfiles
```
tpantelis, line 16: How about shorten it to just _files_?

tpantelis (changes requested):

pkg/globalnet/controllers/ipam/constants.go
mangelajo: Correct, the submariner.io/globalIp part needs to stay as it 's (the constant name can change, but not the content).

main.go
mangelajo: Correct

pkg/routeagent_driver/handlers/kubeproxyiptables/cleanup_handling.go
mangelajo: agreed about readability but I don't see a good match that works with the "kubeproxyiptables" directory name? can they be different, if they can be different ... may be

pkg/util/clusterfiles/cluster_files.go
mangelajo: I think also clusterfiles would be ok, not as long as "kubeproxyiptables" ':D this starts to sound like german

mangelajo (changes requested):
I agree with the first two or three by tom

pkg/routeagent_driver/handlers/kubeproxyiptables/cleanup_handling.go
Jaanki: directory name and package name can be different but its a convention that they should be the same (not necessarily though) so I changed the directory name.

Jaanki: How about kpiptables?

tpantelis: @mangelajo How about just shorten to _kubeproxy_? But if the _iptables_ part is important, ie there could be another kubeproxy implementation,  then how about a subdirectory _kubeproxy/iptables_?

Re: '_' in general in package names, I'm OK with it although we should still strive for brevity to avoid the need - if others agree, perhaps we can disable this check/failure?

.golangci.yml
```diff
@@ -70,3 +69,20 @@ issues:
...
+    # Allow dot imports for `fake` directory
+    - linters:
+        - stylecheck
+      text: "ST1001: "
+      path: pkg/cable/fake
+
+    - linters:
+        - stylecheck
+      text: "ST1001: "
+      path: pkg/cableengine/fake
```
tpantelis: We can handle all _fake_ dirs with one rule. Also need to exclude _test/e2e_.

```suggestion
    # Allow dot imports for gingko and gomega in test-related directories
    - linters:
        - stylecheck
      text: "ST1001: "
      path: /fake

    - linters:
        - stylecheck
      text: "ST1001: "
      path: test/e2e
```

.golangci.yml
```diff
@@ -70,3 +69,20 @@ issues:
     - linters:
         - wsl
       text: "only one cuddle assignment allowed before range statement"
+
+    # Ignore new line warning at L30. New line is added to reduce line length according to lll
+    - linters:
+        - stylecheck
+      text: "ST1005: "
+      path: pkg/networkplugin-syncer/handlers/ovn/handler.go
```
tpantelis: The error seems like a false-positive bug with the linter. We should ignore this one line in code via:

`//nolint:stylecheck This seems like a false-positive bug with the linter - the actual string has no newline`

tpantelis (changes requested):

.golangci.yml
Jaanki: ahh yes...thats a good alternative. Forgot this one

pkg/networkplugin-syncer/handlers/ovn/handler.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/networkplugin-syncer/handlers/ovn/nbctl"
 )
 
+//nolint:stylecheck // Limit line length as per lll
```
tpantelis: The "_ST1005: error strings should not end with punctuation or a newline_" error is actually due to the ending period and not line-length. So we just need to remove the period 😄

tpantelis (changes requested):

pkg/routeagent_driver/handlers/kubeproxyiptables/cleanup_handling.go
mangelajo: Yeah, that makes sense we could have kubeproxy now, and if we ever support the other implementation use folders.

.golangci.yml
```diff
@@ -70,3 +69,14 @@ issues:
...
+    # Allow dot imports for `fake` directory
+    - linters:
+        - stylecheck
+      text: "ST1001: "
+      path: /fake
+
+    - linters:
+        - stylecheck
+      text: "ST1001: "
+      path: test/e2e
```
mangelajo: This needs a return at the end of the line.

mangelajo:
No more comments on my side, just what @tpantelis said & agreeing with "kubeproxy" instead of kubeproxyiptables, but that's a nit :-P :D

mangelajo (changes requested):

pkg/networkplugin-syncer/handlers/ovn/handler.go
```diff
@@ -27,7 +27,8 @@ import (
 	"github.com/submariner-io/submariner/pkg/networkplugin-syncer/handlers/ovn/nbctl"
 )
 
-var WaitingForLocalEndpoint = errors.New("Waiting for the local endpoint details before we can " +
+//nolint:stylecheck // Limit line length as per lll
+var ErrWaitingForLocalEndpoint = errors.New("Waiting for the local endpoint details before we can " +
 	"setup any remote endpoint related information, this will be retried.")
```
skitt:
```suggestion
var ErrWaitingForLocalEndpoint = errors.New("waiting for the local endpoint details before we can " +
	"setup any remote endpoint related information, this will be retried")
```

pkg/networkplugin-syncer/handlers/ovn/nbctl/nbctl.go
```diff
@@ -103,9 +103,9 @@ func (n *NbCtl) GetGatewayChassis(lrp, chassis string) (string, error) {
 	return output, err
 }
 
-func (n *NbCtl) LrSetLoadBalancers(lrId string, lbIDs []string) error {
+func (n *NbCtl) LrSetLoadBalancers(lrID string, lbIDs []string) error {
```
skitt, line 106: Although the linter doesn’t flag this, `Lr` should be `LR` throughout since it’s an acronym. (But we can ignore this for now.)

pkg/networkplugin-syncer/handlers/ovn/services.go
```diff
@@ -29,23 +29,23 @@ func (ovn *SyncHandler) ensureServiceLoadBalancersFrom(logicalSwitch string) err
 		return err
 	}
 
-	submarinerRouterId, err := ovn.getSubmarinerRouterId()
+	submarinerRouterID, err := ovn.getSubmarinerRouterID()
```
skitt, line 32: “Id” is an abbreviation, not an acronym, but this is a bug in the linter...

skitt (changes requested):

pkg/networkplugin-syncer/handlers/ovn/services.go
skitt: And the rest of the world disagrees with me, so “ID” it is.

pkg/networkplugin-syncer/handlers/ovn/nbctl/nbctl.go
Jaanki: Ignoring for now.

.golangci.yml
dfarrell07: Can we have this comment explain both rules?

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -364,19 +364,19 @@ func (i *Controller) handleUpdateNode(old, newObj interface{}) {
 		return
 	}
 
-	oldCniIfaceIpOnNode := old.(*k8sv1.Node).GetAnnotations()[constants.CniInterfaceIP]
-	newCniIfaceIpOnNode := newObj.(*k8sv1.Node).GetAnnotations()[constants.CniInterfaceIP]
-	if oldCniIfaceIpOnNode == "" && newCniIfaceIpOnNode != "" {
-		klog.V(log.DEBUG).Infof("In handleUpdateNode, node %q is now annotated with cniIfaceIP, enqueing", newObj.(*k8sv1.Node).Name)
+	oldCniIfaceIPOnNode := old.(*k8sv1.Node).GetAnnotations()[constants.CniInterfaceIP]
```
dfarrell07, line 367: I guess the linter doesn't flag "Cni"? This is a bit of a strange var name, it expands to include "interface interface". I guess this is out of scope for this PR.

pkg/globalnet/controllers/ipam/ipam.go
```diff
@@ -442,31 +442,31 @@ func (i *Controller) handleRemovedPod(obj interface{}) {
 	}
 
 	if !i.excludeNamespaces[pod.Namespace] {
-		globalIp := pod.Annotations[SubmarinerIpamGlobalIp]
-		if globalIp != "" && pod.Status.PodIP != "" {
+		globalIP := pod.Annotations[SubmarinerIpamGlobalIP]
```
dfarrell07, line 445: I guess the linter doesn't flag "Ipam"

pkg/routeagent_driver/cni/cni_iface.go
```diff
@@ -79,7 +79,7 @@ func ConfigureRpFilter(iface string) error {
 	// #nosec G306
 	err := ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+iface+"/rp_filter", []byte("2"), 0644)
 	if err != nil {
-		return fmt.Errorf("unable to update rp_filter for cni_interface %q, err: %s", iface, err)
+		return fmt.Errorf("unable to update rp_filter for cni interface %q, err: %s", iface, err)
```
dfarrell07, line 82: CNI?

pkg/globalnet/controllers/ipam/ipam.go
Jaanki: I have only changed the names that the linter flagged.

pkg/globalnet/controllers/ipam/ipam.go
Jaanki: Ditto

pkg/routeagent_driver/cni/cni_iface.go
Jaanki: This applies to all `cni`s throughtout the codebase and its out of the scope of this patch.

pkg/globalnet/controllers/ipam/ipam.go
tpantelis: CNI isn't a well-known acronym outside k8s, like ID and IP, so the linter doesn't know about it.

skitt: We can add our own configuration with other acronyms we want to watch for...

pkg/routeagent_driver/cni/cni_iface.go
Jaanki: This applies to all `CNI Interfaces` throughtout the code and is out of the scope of this PR.

pkg/routeagent_driver/constants/constants.go
```diff
@@ -36,7 +36,7 @@ const (
 	// part of route-agent and this will subsequently be used in globalnet controller for
 	// supporting connectivity from HostNetwork to remoteClusters.
 	// [#] interface on the node that has an IPAddress from the clusterCIDR
-	CniInterfaceIP = "submariner.io/cniIfaceIp"
+	CniInterfaceIP = "submariner.io/cniIfaceIP"
```
tpantelis: Changing the label name may have other effects and would be outside the scope of this PR.

```suggestion
	CniInterfaceIP = "submariner.io/cniIfaceIp"
```

tpantelis (changes requested):

pkg/routeagent_driver/cni/cni_iface.go
dfarrell07: Kinda, I mean we are adding this comment and introducing a new "cni" that should be "CNI". As @skitt suggested, probably best to raise an issue to track adding additional config for custom acronyms we want to watch for.

tpantelis: It's referred to in several different ways in messages in this file, ie _CNI Interface_, _cni_interface_, _cniInterface_, _cniIface_. We should be consistent - IMO _CNI_ as @dfarrell07 suggest. However I do agree it's a bit out of scope for this PR - this PR is already quite large and refactors quite a bit of files so it would be nice to get this merged ASAP. I have work to do in several of the files but have been holding off to avoid significant conflict resolution.

tpantelis: @Jaanki Can  you address this in another PR?

tpantelis (approved):

dfarrell07 (approved):

mangelajo (approved):

pkg/cable/cleanup/xfrm_cleanup_handler.go
```diff
@@ -51,7 +51,7 @@ func (xc *XFRMCleanupHandler) NonGatewayCleanup() error {
 		klog.V(log.DEBUG).Infof("Deleting XFRM policy %s", currentXfrmPolicyList[i])
 
 		if err = netlink.XfrmPolicyDel(&currentXfrmPolicyList[i]); err != nil {
-			return fmt.Errorf("Error Deleting XFRM policy %s: %v", currentXfrmPolicyList[i], err)
+			return fmt.Errorf("error Deleting XFRM policy %s: %v", currentXfrmPolicyList[i], err)
```
mangelajo, line 54:
```suggestion
			return fmt.Errorf("error deleting XFRM policy %s: %v", currentXfrmPolicyList[i], err)
```

skitt (approved):

PR #1147, opened 2021-02-16
Enable PLPMTUD on the active gateway node when using Globalnet

### LINKED ISSUE #995: AWS/OCP 4.5.13: Submariner + Globalnet: Pod with HostNetworking on GW Node to remoteService failing
**What happened**:
In a Submariner Globalnet deployment, when e2e tests are executed on AWS/OCP Clusters, one of them is consistently failing.
Basically when a Pod with HostNetworking is trying to connect to a remoteService its failing.

```
STEP: Creating a listener pod in cluster "cluster-f", which will wait for a handshake over TCP
STEP: Pointing a service ClusterIP to the listener pod in cluster "cluster-f"
Dec  3 15:27:05.661: INFO: Will send traffic to IP: 169.254.32.200
STEP: Creating a connector pod in cluster "cluster-e", which will attempt the specific UUID handshake over TCP
STEP: Waiting for the connector pod "tcp-check-podxfwlq" to exit, returning what connector sent
Dec  3 15:27:11.621: INFO: Pod "tcp-check-podxfwlq" output:
169.254.32.200 (169.254.32.200:1234) open
[dataplane] listener says 26d4329b-2467-4dd8-84f1-6c2102d6f7f1

STEP: Waiting for the listener pod "tcp-check-listener6k5bp" to exit, returning what listener sent
Dec  3 15:31:12.234: INFO: Pod "tcp-check-listener6k5bp" output:
listening on 0.0.0.0:1234 ...
connect to 10.129.0.52:1234 from 169.254.0.135:34739 (169.254.0.135:34739)
```
Interestingly, if you look at the output of the listener pod, we can see that the connector pod was indeed able to reach the listener, but it was unable to send the UUID string (that is sent as part of e2e tests).

e2e tests should pass consistently.

**Environment**:
- Submariner version (use `subctl version`): v0.8.0-rc0
- Kubernetes version (use `kubectl version`): v1.18.3+47c0e71
- Cloud provider or hardware configuration: AWS/OCP 4.5.13
- OS (e.g: `cat /etc/os-release`): Alpine Linux
- Kernel (e.g. `uname -a`): 4.18.0-193.23.1.el8_2.x86_64
- Install tools:
- Network plugin and version (if this is a network-related bug): OpenshiftSDN
- Others:

sridhargaddam:
On some platforms like AWS when using Globalnet, it was seen
that Path MTU discovery was not happening properly. Because of
this, one of the e2e test is failing when the sourcePod is on
Gateway Node with HostNetworking enabled. This PR enables TCP
Packetization-Layer Path MTU discovery when an ICMP black hole
is detected by configuring the appropriate proc entry.

Also, we update the base mss value to RFC4821 recommended value
of 1024. This change is done only on the active Gateway node of
the cluster.

Fixes issue: https://github.com/submariner-io/submariner/issues/995

### DIFF
--- a/pkg/globalnet/controllers/ipam/gatewaymonitor.go
+++ b/pkg/globalnet/controllers/ipam/gatewaymonitor.go
@@ -17,6 +17,7 @@ package ipam
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"time"
 
@@ -88,6 +89,8 @@ func NewGatewayMonitor(spec *SubmarinerIpamControllerSpecification,
 		},
 		DeleteFunc: gatewayMonitor.handleRemovedEndpoint,
 	}, handlerResync)
+
+	gatewayMonitor.ReadTcpMTUConfigurationOnNode()
 
 	return gatewayMonitor, nil
 }
@@ -201,6 +204,10 @@ func (gm *GatewayMonitor) processNextEndpoint() bool {
 		if endpoint.Spec.Hostname == hostname {
 			klog.V(log.DEBUG).Infof("We are now on GatewayNode %s", endpoint.Spec.PrivateIP)
 
+			if gm.mtuConfig.successfullyRead {
+				ConfigureTcpMTUProbeValue([]byte("1"), []byte("1024"))
+			}
+
 			gm.syncMutex.Lock()
 			if !gm.isGatewayNode {
 				gm.isGatewayNode = true
@@ -209,6 +216,10 @@ func (gm *GatewayMonitor) processNextEndpoint() bool {
 			gm.syncMutex.Unlock()
 		} else {
 			klog.V(log.DEBUG).Infof("We are on non-gatewayNode. GatewayNode ip is %s", endpoint.Spec.PrivateIP)
+			if gm.mtuConfig.successfullyRead {
+				ConfigureTcpMTUProbeValue(gm.mtuConfig.default_mtu_probing, gm.mtuConfig.default_tcp_base_mss)
+			}
+
 			gm.syncMutex.Lock()
 			if gm.isGatewayNode {
 				gm.stopIpamController()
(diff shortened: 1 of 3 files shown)


pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -209,6 +216,10 @@ func (gm *GatewayMonitor) processNextEndpoint() bool {
 			gm.syncMutex.Unlock()
 		} else {
 			klog.V(log.DEBUG).Infof("We are on non-gatewayNode. GatewayNode ip is %s", endpoint.Spec.PrivateIP)
+			if gm.mtuConfig.successfullyRead {
```
mangelajo: Is this to restore?, What if the pod is restarted when the settings are configured already?

Shouldn't we remove also on pod stop?

BTW, I'm not sure if it's worth bothering with the removal of this.

sridhargaddam: Yes @mangelajo this is to restore the old values when the Gateway migrates to a new node. 
TBH from what I understand there does not seem to be any harm even if we leave the modified configuration, but I just thought of getting some feedback and included this change.

So, are you suggesting that we can skip it?

pkg/globalnet/controllers/ipam/types.go
```diff
@@ -68,6 +68,12 @@ type Controller struct {
 	ipt iptables.Interface
 }
 
+type MTUConfig struct {
+	successfullyRead          bool
+	default_mtu_probing       []byte
```
tpantelis:
```suggestion
	defaultMTUProbing       []byte
```

pkg/globalnet/controllers/ipam/types.go
```diff
@@ -68,6 +68,12 @@ type Controller struct {
...
+	default_tcp_base_mss      []byte
```
tpantelis:
```suggestion
	defaultTCPBaseMss      []byte
```

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -326,3 +337,41 @@ func (gm *GatewayMonitor) stopIpamController() {
 		cleanup.ClearGlobalnetChains(gm.ipt)
 	}
 }
+
+func ConfigureTcpMTUProbeValue(mtuProbe, mssValue []byte) {
+	// If we are unable to update the values, just log a warning, Globalnet functionality works fine except
+	// for one use-case where Pod with HostNetworking on Gateway node has mtu issues connecting to remoteServices.
+	if err := ioutil.WriteFile(tcpMtuProbingProcEntry, mtuProbe, 0644); err == nil {
```
tpantelis, line 341: For the unit tests, I think it would be better to abstract this functionality behind an interface.  I propose to make `MTUConfig` an interface with the `ReadTcpMTUConfigurationOnNode` and `ConfigureTcpMTUProbeValue` methods and make it configurable on `GatewayMonitor`.  If this sounds reasonable, I could do it in a follow-up if you want.

sridhargaddam: Sure @tpantelis. Thank you :)

sridhargaddam (commented):
Thanks for reviewing @mangelajo @tpantelis

pkg/globalnet/controllers/ipam/gatewaymonitor.go
mangelajo: I would simplify the code by avoiding the cleanup as we do for similar other parameters.

tpantelis (approved):

mangelajo (approved):

Issue #1157, opened 2021-02-24
Create GlobalnetEgressIP and associated objects


sridhargaddam:
**What happened**:
As part of [Globalnet Enhancement](https://github.com/submariner-io/enhancements/pull/22/files), we now have a new CRD, GlobalnetEgressIP for requesting a Egress globalIP on Pods. This issue is to track the necessary CRD changes including the creation of Clientsets, informers and listers.

tpantelis:
bump

sridhargaddam:
@tpantelis do you want to take a look at this? Thank you

tpantelis:
> @tpantelis do you want to take a look at this? Thank you

I updated #1172.

PR #1158, opened 2021-02-24
Create GlobalnetEgressIP CRD types and associated items

sridhargaddam:
This PR adds the necessary types and includes the clientsets,
informers and listers using codegenerator.

Fixes issue: https://github.com/submariner-io/submariner/issues/1157


tpantelis:
Why is the new CRD in `v1alpha1` and not `v1`?

sridhargaddam:
> Why is the new CRD in `v1alpha1` and not `v1`?

This will be the first implementation of the CRD and there is a potential for changes in the future based on the feedback we get from users. So, I wanted to use v1alpha1. Do you see any problem with this @tpantelis ?

tpantelis:
> > Why is the new CRD in `v1alpha1` and not `v1`?
> 
> This will be the first implementation of the CRD and there is a potential for changes in the future based on the feedback we get from users. So, I wanted to use v1alpha1. Do you see any problem with this @tpantelis ?

It's not a problem - just wondering what the version progression would be. Would GA release graduate to v1 such that we won't have to worry about supporting/migrating v1alpha1?

sridhargaddam:
> > > Why is the new CRD in `v1alpha1` and not `v1`?
> > 
> > 
> > This will be the first implementation of the CRD and there is a potential for changes in the future based on the feedback we get from users. So, I wanted to use v1alpha1. Do you see any problem with this @tpantelis ?
> 
> It's not a problem - just wondering what the version progression would be. Would GA release graduate to v1 such that we won't have to worry about supporting/migrating v1alpha1?

Thats a good point @tpantelis . I'd like to hear from @mangelajo, @nyechiel @skitt @mkolesnik if we should go with v1 itself or use v1alpha1.

skitt:
> > > > Why is the new CRD in `v1alpha1` and not `v1`?
> > > 
> > > 
> > > This will be the first implementation of the CRD and there is a potential for changes in the future based on the feedback we get from users. So, I wanted to use v1alpha1. Do you see any problem with this @tpantelis ?
> > 
> > 
> > It's not a problem - just wondering what the version progression would be. Would GA release graduate to v1 such that we won't have to worry about supporting/migrating v1alpha1?
> 
> Thats a good point @tpantelis . I'd like to hear from @mangelajo, @nyechiel @skitt @mkolesnik if we should go with v1 itself or use v1alpha1.

IMO, as long as we’re not reasonably convinced we won’t need to change the CRD, we should start with v1alpha1.

The GA point in general is a good one, we’ll need to address that sooner rather than later (across all CRDs).

pkg/apis/submariner.io/v1alpha1/types.go
```diff
@@ -0,0 +1,112 @@
...
+	// +optional
+	NumGlobalIPs int `json:"numGlobalIPs,omitempty"`
+
+	// Selects the Namespaces to which this GlobalnetEgressIP object applies.
+	// If an empty namespaceSelector: {} is configured, it selects all namespaces.
+	// If an empty namespaceSelector: {} is configured along with empty PodSelector, it applies
+	// to whole Cluster.
+	// On the other hand, if you omit specifying namespaceSelector, it does not select any
+	// namespaces and selects only Pods from the namespace where the GlobalnetEgressIP is
+	// deployed to.
```
dfarrell07, line 61:
```suggestion
	// deployed.
```

pkg/apis/submariner.io/v1alpha1/types.go
```diff
@@ -0,0 +1,112 @@
...
+	// User can explicitly request a specific GlobalIP and this should be a valid IPaddress
+	// from globalnetCIDR allocated to the Cluster. If the requested globalIP falls outside
+	// of the globalnetCIDR or is already allocated (either fully or partially), then Status
+	// field will be set to Error with appropriate message.
+	// If this field is not specified, a single globalIP will be allocated.
+	// This field cannot be changed through updates.
+	// +optional
+	GlobalIP string `json:"globalIP,omitempty"`
+
+	// The number of globalIP's requested. Globalnet Controller will allocate the requested
```
dfarrell07, line 47:
```suggestion
	// The number of globalIPs requested. Globalnet Controller will allocate the requested
```

skitt:
> > > > > Why is the new CRD in `v1alpha1` and not `v1`?
> > > > 
> > > > 
> > > > This will be the first implementation of the CRD and there is a potential for changes in the future based on the feedback we get from users. So, I wanted to use v1alpha1. Do you see any problem with this @tpantelis ?
> > > 
> > > 
> > > It's not a problem - just wondering what the version progression would be. Would GA release graduate to v1 such that we won't have to worry about supporting/migrating v1alpha1?
> > 
> > 
> > Thats a good point @tpantelis . I'd like to hear from @mangelajo, @nyechiel @skitt @mkolesnik if we should go with v1 itself or use v1alpha1.
> 
> IMO, as long as we’re not reasonably convinced we won’t need to change the CRD, we should start with v1alpha1.
> 
> The GA point in general is a good one, we’ll need to address that sooner rather than later (across all CRDs).

I’d missed the fact that all the existing Submariner CRDs are already v1; so let’s stick to that now that we’ve set the precedent.

pkg/apis/submariner.io/v1alpha1/types.go
```diff
@@ -0,0 +1,112 @@
...
+	// to whole Cluster.
+	// On the other hand, if you omit specifying namespaceSelector, it does not select any
+	// namespaces and selects only Pods from the namespace where the GlobalnetEgressIP is
+	// deployed to.
+	// +optional
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+
+	// Selects the pods whose label match the definition. This is an optional field and
+	// in case it is not set, it results in all the Pods selected from the NamespaceSelector.
+	// In case it is set, its intersected with NamespaceSelector, and only Pods that match the
```
dfarrell07, line 67:
```suggestion
	// In case it is set, it's intersected with NamespaceSelector, and only Pods that match the
```

pkg/apis/submariner.io/v1alpha1/types.go
```diff
@@ -0,0 +1,112 @@
...
+	// PodSelector and present in the NamespaceSelector will have the GlobalIPs as EgressIPs.
+	// Either namespaceSelector or podSelector have to be specified. If both are omitted,
+	// its considered as an error.
```
dfarrell07, line 70:
```suggestion
	// it's considered as an error.
```

dfarrell07 (changes requested):

sridhargaddam:
v1alpha1 modified lot of existing files and it was kind of mess to revert those changes manually.
I pushed the following PR that uses v1 - https://github.com/submariner-io/submariner/pull/1172

Please review the above PR, thanks.

PR #1159, opened 2021-02-25
Enable PLPMTUD on the active gateway node when using Globalnet

sridhargaddam:
On some platforms like AWS when using Globalnet, it was seen
that Path MTU discovery was not happening properly. Because of
this, one of the e2e test is failing when the sourcePod is on
Gateway Node with HostNetworking enabled. This PR enables TCP
Packetization-Layer Path MTU discovery when an ICMP black hole
is detected by configuring the appropriate proc entry.

Also, we update the base mss value to RFC4821 recommended value
of 1024. This change is done only on the active Gateway node of
the cluster.

Fixes issue: #995
Signed-Off-by: Sridhar Gaddam sgaddam@redhat.com

### DIFF
--- a/pkg/globalnet/controllers/ipam/constants.go
+++ b/pkg/globalnet/controllers/ipam/constants.go
@@ -22,6 +22,9 @@ const (
 	kubeProxyServiceChainPrefix = "KUBE-SVC-"
 	kubeProxyServiceChainName   = "KUBE-SERVICES"
 
+	tcpMtuProbingProcEntry = "/proc/sys/net/ipv4/tcp_mtu_probing"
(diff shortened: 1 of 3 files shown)


sridhargaddam:
@manosnoam @SteveMattar This backport will fix the following issue
```
Submariner E2E suite.[dataplane-globalnet] Basic TCP connectivity tests across overlapping clusters without discovery when a pod with HostNetworking connects via TCP to the globalIP of a remote service when the pod is on a gateway and the remote service is not on a gateway should have sent the expected data from the pod to the other pod
```

mangelajo (approved):

tpantelis (approved):

PR #1169, opened 2021-03-02
Add kubeproxy unit tests

### LINKED ISSUE #380: Add unit tests for route agent controller

tpantelis:
See individual commits.

### DIFF
--- a/pkg/routeagent_driver/cleanup/fake/cleanup.go
+++ b/pkg/routeagent_driver/cleanup/fake/cleanup.go
@@ -1,0 +1,64 @@
+/*
+© 2021 Red Hat, Inc. and others
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package fake
+
+import (
+	"time"
+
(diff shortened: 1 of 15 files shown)


pkg/routeagent_driver/cleanup/fake/cleanup.go
```diff
@@ -0,0 +1,64 @@
...
+	return nil
+}
+
+func (h *Handler) NonGatewayCleanup() error {
+	h.nonGatewayCleanup <- true
+	return nil
+}
+
+func (h *Handler) AwaitGatewayToNonGatewayTransition() {
+	Eventually(h.gatewayToNonGatewayTransition, 5).Should(Receive())
```
mangelajo, line 51: The gomega docs aren't clear because they only talk about float64 and not int, so I went down again into checking that int = seconds and not milliseconds: 

https://github.com/onsi/gomega/blob/3e036b855ae3e332843aff25edf11a0a37033156/gomega_dsl.go#L434

So 5 == seconds. (checked!) %)

mangelajo (approved):
It looks great Tom, It would be amazing if we were able to contribute those fake wrappers to the netlink library itself "upstream"ier %)

pkg/routeagent_driver/cleanup/fake/cleanup.go
tpantelis: yeah the behavior is documented:

"_Both the timeout and polling interval are configurable as optional arguments: The first optional argument is the timeout The second optional argument is the polling interval
Both intervals can either be specified as time.Duration, parsable duration strings or as floats/integers. **In the last case they are interpreted as seconds**._"

tpantelis:
> It looks great Tom, It would be amazing if we were able to contribute those fake wrappers to the netlink library itself "upstream"ier %)

Thanks for reviewing.  Yeah I could look into contributing it to netlink.

dfarrell07 (approved):
Wow this looks great @tpantelis, thanks!

PR #1171, opened 2021-03-02
Vxlan cable driver

### LINKED ISSUE #1148: Add CI For un-encrypted tunnels
As a part of  https://github.com/submariner-io/submariner/issues/674 we are adding support for new Cable driver using vxlan . We need to have a dedicated CI job for this. Since there proposal to have a different kind of tunnels we could also have CI job in which one tunnel in un-encrypted.

**Why is this needed**:
To automate testing of  https://github.com/submariner-io/submariner/issues/674

- [x] Add `vxlan` option to shipyard Makefile.inc https://github.com/submariner-io/shipyard/pull/553
- [x] Add 'vxlan' to e2e test matrix https://github.com/submariner-io/submariner/pull/1171

### LINKED ISSUE #1149: Un-Encrypted vxlan tunnel between clusters
Un-Encrypted vxlan tunnel between clusters
**Why is this needed**:
All current cable drivers involve encrypting content, which is great for privacy but involves some overhead. On private connections, it would be useful to enable unencrypted connections, using VxLAN.

aswinsuryan:
Fixes: #1149

### DIFF
--- a/pkg/cable/vxlan/vxlan.go
+++ b/pkg/cable/vxlan/vxlan.go
@@ -1,0 +1,537 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package vxlan
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/log"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cable"
+	"github.com/submariner-io/submariner/pkg/natdiscovery"
+	netlinkAPI "github.com/submariner-io/submariner/pkg/netlink"
+	"github.com/submariner-io/submariner/pkg/types"
+	"github.com/submariner-io/submariner/pkg/util"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+	"k8s.io/klog"
+)
+
+const (
+	VxlanIface             = "vxlan-tunnel"
+	VxlanOverhead          = 50
+	VxlanVTepNetworkPrefix = 241
+	CableDriverName        = "vxlan"
+	TableID                = 100
+	defaultNATTPort        = 4500
+)
+
+type Operation int
+
+const (
+	Add Operation = iota
+	Delete
+	Flush
+)
+
+type vxlan struct {
+	localEndpoint types.SubmarinerEndpoint
+	connections   []v1.Connection
+	mutex         sync.Mutex
+	vxlanIface    *vxlanIface
+	netLink       netlinkAPI.Interface
+}
+
+type vxlanIface struct {
+	activeEndpointHostname string
+	link                   *netlink.Vxlan
+	vtepIP                 net.IP
+}
+
+type vxlanAttributes struct {
+	name     string
+	vxlanID  int
+	group    net.IP
+	srcAddr  net.IP
+	vtepPort int
+	mtu      int
+}
+
+func init() {
+	cable.AddDriver(CableDriverName, NewDriver)
+}
+
+func NewDriver(localEndpoint types.SubmarinerEndpoint, localCluster types.SubmarinerCluster) (cable.Driver, error) {
+	var err error
+
+	v := vxlan{
+		localEndpoint: localEndpoint,
+		netLink:       netlinkAPI.New(),
+	}
+
+	port, err := localEndpoint.Spec.GetBackendPort(v1.UDPPortConfig, defaultNATTPort)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the UDP port configuration: %v", err)
+	}
+
+	if err = v.createVxlanInterface(localEndpoint.Spec.Hostname, int(port)); err != nil {
+		return nil, fmt.Errorf("failed to setup Vxlan link: %v", err)
+	}
+
+	return &v, nil
+}
+
+func (v *vxlan) createVxlanInterface(activeEndPoint string, port int) error {
+	ipAddr := v.localEndpoint.Spec.PrivateIP
+
+	vtepIP, err := v.getVxlanVtepIPAddress(ipAddr)
+	if err != nil {
+		return fmt.Errorf("failed to derive the vxlan vtepIP for %s, %v", ipAddr, err)
+	}
+
+	defaultHostIface, err := util.GetDefaultGatewayInterface()
+
+	if err != nil {
+		return errors.Wrapf(err, "Unable to find the default interface on host: %s",
+			v.localEndpoint.Spec.Hostname)
+	}
+
+	// Derive the MTU based on the default outgoing interface
+	vxlanMtu := defaultHostIface.MTU - VxlanOverhead
+
+	attrs := &vxlanAttributes{
+		name:     VxlanIface,
+		vxlanID:  1000,
+		group:    nil,
+		srcAddr:  nil,
+		vtepPort: port,
+		mtu:      vxlanMtu,
+	}
+
+	v.vxlanIface, err = newVxlanIface(attrs, activeEndPoint)
+	if err != nil {
+		return fmt.Errorf("failed to create vxlan interface on Gateway Node: %v", err)
+	}
+
+	v.vxlanIface.vtepIP = vtepIP
+
+	err = v.addIPRule()
+	if err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to add ip rule: %v", err)
+	}
+
+	err = v.netLink.EnableLooseModeReversePathFilter(VxlanIface)
+	if err != nil {
+		return fmt.Errorf("unable to update vxlan rp_filter proc entry, err: %s", err)
+	}
+
+	klog.V(log.DEBUG).Infof("Successfully configured rp_filter to loose mode(2) on %s", VxlanIface)
+
+	err = v.vxlanIface.configureIPAddress(vtepIP, net.CIDRMask(8, 32))
+	if err != nil {
+		return fmt.Errorf("failed to configure vxlan interface ipaddress on the Gateway Node %v", err)
+	}
+
+	return nil
+}
+
+func newVxlanIface(attrs *vxlanAttributes, activeEndPoint string) (*vxlanIface, error) {
+	iface := &netlink.Vxlan{
+		LinkAttrs: netlink.LinkAttrs{
+			Name:  attrs.name,
+			MTU:   attrs.mtu,
+			Flags: net.FlagUp,
+		},
+		VxlanId: attrs.vxlanID,
+		SrcAddr: attrs.srcAddr,
+		Group:   attrs.group,
+		Port:    attrs.vtepPort,
+	}
+
+	vxlanIface := &vxlanIface{
+		link:                   iface,
+		activeEndpointHostname: activeEndPoint,
+	}
+
+	if err := createvxlanIface(vxlanIface); err != nil {
+		return nil, err
+	}
+
+	return vxlanIface, nil
+}
+
+func createvxlanIface(iface *vxlanIface) error {
+	err := netlink.LinkAdd(iface.link)
+	if err == syscall.EEXIST {
+		klog.Errorf("Got error: %v, %v", err, iface.link)
+		// Get the properties of existing vxlan interface
+		existing, err := netlink.LinkByName(iface.link.Name)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve link info: %v", err)
+		}
+
+		if isVxlanConfigTheSame(iface.link, existing) {
+			klog.V(log.DEBUG).Infof("VxLAN interface already exists with same configuration.")
+
+			iface.link = existing.(*netlink.Vxlan)
+
+			return nil
+		}
+
+		// Config does not match, delete the existing interface and re-create it.
+		if err = netlink.LinkDel(existing); err != nil {
+			return fmt.Errorf("failed to delete the existing vxlan interface: %v", err)
+		}
+
+		if err = netlink.LinkAdd(iface.link); err != nil {
+			return fmt.Errorf("failed to re-create the the vxlan interface: %v", err)
+		}
+	} else if err != nil {
+		return fmt.Errorf("failed to create the the vxlan interface: %v", err)
+	}
+
+	return nil
+}
+
+func isVxlanConfigTheSame(newLink, currentLink netlink.Link) bool {
+	required := newLink.(*netlink.Vxlan)
+	existing := currentLink.(*netlink.Vxlan)
+
+	if required.VxlanId != existing.VxlanId {
+		klog.Warningf("VxlanId of existing interface (%d) does not match with required VxlanId (%d)", existing.VxlanId, required.VxlanId)
+		return false
+	}
+
+	if len(required.Group) > 0 && len(existing.Group) > 0 && !required.Group.Equal(existing.Group) {
+		klog.Warningf("Vxlan Group (%v) of existing interface does not match with required Group (%v)", existing.Group, required.Group)
+		return false
+	}
+
+	if len(required.SrcAddr) > 0 && len(existing.SrcAddr) > 0 && !required.SrcAddr.Equal(existing.SrcAddr) {
+		klog.Warningf("Vxlan SrcAddr (%v) of existing interface does not match with required SrcAddr (%v)", existing.SrcAddr, required.SrcAddr)
+		return false
+	}
+
+	if required.Port > 0 && existing.Port > 0 && required.Port != existing.Port {
+		klog.Warningf("Vxlan Port (%d) of existing interface does not match with required Port (%d)", existing.Port, required.Port)
+		return false
+	}
+
+	return true
+}
+
+func (v *vxlan) getVxlanVtepIPAddress(ipAddr string) (net.IP, error) {
+	ipSlice := strings.Split(ipAddr, ".")
+	if len(ipSlice) < 4 {
+		return nil, fmt.Errorf("invalid ipAddr [%s]", ipAddr)
+	}
+
+	ipSlice[0] = strconv.Itoa(VxlanVTepNetworkPrefix)
+	vxlanIP := net.ParseIP(strings.Join(ipSlice, "."))
+
+	return vxlanIP, nil
+}
+
+func (v *vxlan) addIPRule() error {
+	if v.vxlanIface != nil {
+		rule := netlink.NewRule()
+		rule.Table = TableID
+		rule.Priority = TableID
+
+		err := netlink.RuleAdd(rule)
+		if err != nil && !os.IsExist(err) {
+			return fmt.Errorf("failed to add ip rule %s: %v", rule, err)
+		}
+	}
+
+	return nil
+}
+
+func (v *vxlan) ConnectToEndpoint(endpointInfo *natdiscovery.NATEndpointInfo) (string, error) {
+	remoteEndpoint := endpointInfo.Endpoint
+	if v.localEndpoint.Spec.ClusterID == remoteEndpoint.Spec.ClusterID {
+		klog.V(log.DEBUG).Infof("Will not connect to self")
+		return "", nil
+	}
+
+	remoteIP := net.ParseIP(endpointInfo.UseIP)
+	if remoteIP == nil {
+		return "", fmt.Errorf("failed to parse remote IP %s", endpointInfo.UseIP)
+	}
+
+	allowedIPs := parseSubnets(remoteEndpoint.Spec.Subnets)
+
+	klog.V(log.DEBUG).Infof("Connecting cluster %s endpoint %s",
+		remoteEndpoint.Spec.ClusterID, remoteIP)
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+
+	cable.RecordConnection(CableDriverName, &v.localEndpoint.Spec, &remoteEndpoint.Spec, string(v1.Connected), true)
+
+	remoteVtepIP, err := v.getVxlanVtepIPAddress(endpointInfo.UseIP)
+
+	if err != nil {
+		return endpointInfo.UseIP, fmt.Errorf("failed to derive the vxlan vtepIP for %s, %v", endpointInfo.UseIP, err)
+	}
+
+	err = v.vxlanIface.AddFDB(remoteIP, "00:00:00:00:00:00")
+
+	if err != nil {
+		return endpointInfo.UseIP, fmt.Errorf("failed to add remoteIP %q to the forwarding database", remoteIP)
+	}
+
+	err = v.vxlanIface.AddRoute(allowedIPs, remoteVtepIP, v.vxlanIface.vtepIP)
+
+	if err != nil {
+		return endpointInfo.UseIP, fmt.Errorf("failed to add route for the CIDR %q with remoteVtepIP %q and vxlanInterfaceIP %q",
+			allowedIPs, remoteVtepIP, v.vxlanIface.vtepIP)
+	}
+
+	v.connections = append(v.connections, v1.Connection{Endpoint: remoteEndpoint.Spec, Status: v1.Connected,
+		UsingIP: endpointInfo.UseIP, UsingNAT: endpointInfo.UseNAT})
+
+	klog.V(log.DEBUG).Infof("Done adding endpoint for cluster %s", remoteEndpoint.Spec.ClusterID)
+
+	return endpointInfo.UseIP, nil
+}
+
+func (v *vxlan) DisconnectFromEndpoint(remoteEndpoint types.SubmarinerEndpoint) error {
+	klog.V(log.DEBUG).Infof("Removing endpoint %#v", remoteEndpoint)
+
+	if v.localEndpoint.Spec.ClusterID == remoteEndpoint.Spec.ClusterID {
+		klog.V(log.DEBUG).Infof("Will not disconnect self")
+		return nil
+	}
+
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+
+	// parse remote addresses and allowed IPs
+	ip := submarinerEndpointIP(&remoteEndpoint)
+
+	remoteIP := net.ParseIP(ip)
+
+	if remoteIP == nil {
+		return fmt.Errorf("failed to parse remote IP %s", ip)
+	}
+
+	allowedIPs := parseSubnets(remoteEndpoint.Spec.Subnets)
+
+	err := v.vxlanIface.DelFDB(remoteIP, "00:00:00:00:00:00")
+
+	if err != nil {
+		return fmt.Errorf("failed to delete remoteIP %q from the forwarding database", remoteIP)
+	}
+
+	err = v.vxlanIface.DelRoute(allowedIPs)
+
+	if err != nil {
+		return fmt.Errorf("failed to remove route for the CIDR %q",
+			allowedIPs)
+	}
+
+	v.connections = removeConnectionForEndpoint(v.connections, remoteEndpoint)
+	cable.RecordDisconnected(CableDriverName, &v.localEndpoint.Spec, &remoteEndpoint.Spec)
+
+	klog.V(log.DEBUG).Infof("Done removing endpoint for cluster %s", remoteEndpoint.Spec.ClusterID)
+
+	return nil
+}
+
+func removeConnectionForEndpoint(connections []v1.Connection, endpoint types.SubmarinerEndpoint) []v1.Connection {
+	for j := range connections {
+		if connections[j].Endpoint.CableName == endpoint.Spec.CableName {
+			copy(connections[j:], connections[j+1:])
+			return connections[:len(connections)-1]
+		}
+	}
+
+	return connections
+}
+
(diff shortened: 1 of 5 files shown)


pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -64,6 +66,13 @@ func (kp *SyncHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
 		}
 	}
 
+	// For now we get all the cleanups
+	if kp.localCableDriver == vxlan.CableDriverName {
+		kp.installCleanupHandlers(cableCleanup.GetVXLANCleanupHandlers())
+	} else {
+		kp.installCleanupHandlers(cableCleanup.GetCleanupHandlers())
+	}
```
tpantelis: I think this should be abstracted in `cableCleanup.GetCleanupHandlers ` (pass in `kp.localCableDriver`).

I'm wondering how this is going to work when we have multiple cable drivers, ie we may have `vxlan` and `libreswan` drivers running concurrent.  Can the cleanup handlers co-exist?

pkg/routeagent_driver/handlers/kubeproxy/gw_transition.go
```diff
@@ -31,7 +32,10 @@ func (kp *SyncHandler) TransitionToNonGateway() error {
 
 	kp.cleanVxSubmarinerRoutes()
 	// If the active Gateway transitions to a new node, we flush the HostNetwork routing table.
-	kp.updateRoutingRulesForHostNetworkSupport(nil, Flush)
+	if kp.localCableDriver != vxlan.VxLANIface {
```
tpantelis:
```suggestion
	if kp.localCableDriver != vxlan.CableDriverName {
```

pkg/routeagent_driver/handlers/kubeproxy/gw_transition.go
```diff
@@ -66,7 +70,9 @@ func (kp *SyncHandler) TransitionToGateway() error {
 	}
 
 	// Add routes to the new endpoint on the GatewayNode.
-	kp.updateRoutingRulesForHostNetworkSupport(kp.remoteSubnets.Elements(), Add)
+	if kp.localCableDriver != vxlan.VxLANIface {
```
tpantelis:
```suggestion
	if kp.localCableDriver != vxlan.CableDriverName {
```

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -138,7 +151,10 @@ func (kp *SyncHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 		return err
 	}
 
-	kp.updateRoutingRulesForHostNetworkSupport(endpoint.Spec.Subnets, Delete)
+	if kp.localCableDriver != vxlan.CableDriverName {
+		kp.updateRoutingRulesForHostNetworkSupport(endpoint.Spec.Subnets, Delete)
+	}
```
tpantelis: Again what happens with multiple cable drivers....  also would be nice if this code does not need knowledge of a specific driver. Perhaps this could be part of the cable driver impl?

aswinsuryan: This is the part of the route-agent, probably we need to add a cable-driver interface there too.

tpantelis: yeah we have the `cleanup.Handler` interface. In this case we need `RemoteEndpointCreated`, `RemoteEndpointRemoved` etc - it seems we could extend the `event.Handler` interface to the cable drivers. In fact `cleanup.Handler` is really just a subset of `event.Handler`. Both `GatewayToNonGatewayTransition` and `NonGatewayCleanup` are called from `TransitionToNonGateway` so both can be consolidated to `TransitionToNonGateway`. I propose we eliminate `cleanup.Handler` and use `event.Handler` for the cable driver interactions. The cable driver handlers can be added to the event `Registry` same as the `SyncHandler`.

tpantelis: Actually let's back up a bit. Why isn't `updateRoutingRulesForHostNetworkSupport` done for the vxlan driver? Is it b/c it isn't needed but doesn't hurt anything if it is done or it's problematic for the `vxlan` driver? If the latter then how will we handle this when both the `vxlan` and `libreswan` drivers are present?

tpantelis:
bump

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
aswinsuryan: Yes it is done since the route inserted by the vxlan driver ( which can handle hostnetworks too) will conflict with the one added by updateRoutingRulesForHostNetworkSupport. We will only have either vxlan or libreswan between a pair of endpoints so either one of the rule should be present.

pkg/routeagent_driver/handlers/kubeproxy/kp_iptables.go
```diff
@@ -114,8 +113,5 @@ func (kp *SyncHandler) Init() error {
 		return errors.Wrapf(err, "createIPTableChains returned error")
 	}
 
-	// For now we get all the cleanups
-	kp.installCleanupHandlers(cableCleanup.GetCleanupHandlers())
```
sridhargaddam: Why are we removing this from Init and handling it as part of LocalEndpointCreated?

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
sridhargaddam: IMO the cleanup handlers can query the current state and depending on the configuration perform the cleanup operation or skip it. For example, we do the same for Globalnet cleanup - it checks if Globalnet is enabled and only if enabled, cleanup is done, otherwise skipped.

sridhargaddam: Also, can you please explain why we moved this from Init to here?

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
sridhargaddam: Will it be possible to remove any hostNetworkSupport specific routes from vxlan driver and let them be handled in this API, just like how it works for other drivers (libreswan, wireguard etc)?

tpantelis:
bump

aswinsuryan:
bump

.github/workflows/e2e.yml
```diff
@@ -9,6 +9,22 @@ jobs:
...
+        ovn: ['', 'ovn']
+        k8s_version: ['1.20.2']
+        exclude:
+          - ovn: 'ovn'
+            globalnet: 'globalnet'
+          - ovn: 'ovn'
+            cable_driver: 'wireguard'
+        include:
+          # This is the oldest K8s version we try to support
+          - k8s_version: 1.17.17
```
mangelajo: Please note that your changes should now go in: 
https://github.com/submariner-io/submariner/blob/devel/.github/workflows/e2e-full.yml

As those tests are only activated when the ready-to-test label is assigned due to enough reviews.
```suggestion
```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,533 @@
...
+	"github.com/submariner-io/submariner/pkg/types"
+	"github.com/submariner-io/submariner/pkg/util"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+	"k8s.io/klog"
+)
+
+const (
+	VxLANIface             = "vxlan-tunnel"
+	VxLANPort              = 4800
```
mangelajo: 4800/UDP will conflict with the port for vx-submariner.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,533 @@
...
+	"github.com/submariner-io/submariner/pkg/natdiscovery"
+	"github.com/submariner-io/submariner/pkg/types"
+	"github.com/submariner-io/submariner/pkg/util"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+	"k8s.io/klog"
+)
+
+const (
+	VxLANIface             = "vxlan-tunnel"
```
mangelajo: can we call it `submariner`  as for wireguard?

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,533 @@
...
+
+	// Derive the MTU based on the default outgoing interface
+	vxlanMtu := defaultHostIface.MTU - VxLANOverhead
+
+	attrs := &vxLanAttributes{
+		name:     VxLANIface,
+		vxlanID:  1000,
+		group:    nil,
+		srcAddr:  nil,
+		vtepPort: VxLANPort,
```
mangelajo: Please look at how wireguard or libreswan handle the UDP port to be used, as it's a configurable parameter.

We should use the UDP encapsulation port configured for this gateway.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,533 @@
...
+
+	return nil
+}
+
+func isVxlanConfigTheSame(newLink, currentLink netlink.Link) bool {
+	required := newLink.(*netlink.Vxlan)
+	existing := currentLink.(*netlink.Vxlan)
+
+	if required.VxlanId != existing.VxlanId {
+		klog.Errorf("VxlanId of existing interface (%d) does not match with required VxlanId (%d)", existing.VxlanId, required.VxlanId)
```
mangelajo:
```suggestion
		klog.Warningf("VxlanId of existing interface (%d) does not match with required VxlanId (%d)", existing.VxlanId, required.VxlanId)
```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,533 @@
...
+		return false
+	}
+
+	if len(required.Group) > 0 && len(existing.Group) > 0 && !required.Group.Equal(existing.Group) {
+		klog.Errorf("Vxlan Group (%v) of existing interface does not match with required Group (%v)", existing.Group, required.Group)
```
mangelajo: ditto.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,533 @@
...
+		return false
+	}
+
+	if len(required.SrcAddr) > 0 && len(existing.SrcAddr) > 0 && !required.SrcAddr.Equal(existing.SrcAddr) {
+		klog.Errorf("Vxlan SrcAddr (%v) of existing interface does not match with required SrcAddr (%v)", existing.SrcAddr, required.SrcAddr)
```
mangelajo: ditto.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,533 @@
...
+		return false
+	}
+
+	if required.Port > 0 && existing.Port > 0 && required.Port != existing.Port {
+		klog.V(log.DEBUG).Infof("Vxlan Port (%d) of existing interface does not match with required Port (%d)", existing.Port, required.Port)
```
mangelajo: ditto.

mangelajo (changes requested):

pkg/routeagent_driver/cabledriver/vxlan_tunnel_cleanup_handler.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/submariner-io/submariner/pkg/cable/vxlan"
+	"github.com/vishvananda/netlink"
+	"k8s.io/klog"
+)
+
+type vxlanCleanup struct {
+	event.HandlerBase
+}
+
+func NewvxlanCleanup() event.Handler {
```
mangelajo:
```suggestion
func NewVXLANCleanup() event.Handler {
```
?

pkg/routeagent_driver/cabledriver/vxlan_tunnel_cleanup_handler.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (h *vxlanCleanup) NonGatewayCleanup() error {
+	klog.V(log.DEBUG).Infof("Cleaning up the routes")
+
+	link, err := netlink.LinkByName(vxlan.VxLANIface)
+
+	if err != nil {
+		klog.Errorf("Unable to cleanup routes, error retrieving the link %s: %v", vxlan.VxLANIface, err)
+		return nil
+	}
```
mangelajo: This should not print anything, even an error, because otherwise for other cable drivers we will get errors all the time. Just silently exit

pkg/routeagent_driver/cabledriver/vxlan_tunnel_cleanup_handler.go
```diff
@@ -0,0 +1,83 @@
...
+
+	currentRouteList, err := netlink.RouteList(link, syscall.AF_INET)
+
+	if err != nil {
+		klog.Errorf("Unable to cleanup routes, error retrieving routes on the link %s: %v", vxlan.VxLANIface, err)
```
mangelajo: This should not exit at this time. Probably print the warning and not execute L65-L71, but then we really need to try delete the link interface in L73.

mangelajo (changes requested):

pkg/cable/vxlan/vxlan.go
aswinsuryan: I missed this  change, in the spec, it was agreed to use the common UDP port for cable drivers, changed it know

pkg/cable/vxlan/vxlan.go
aswinsuryan: Done.

pkg/cable/vxlan/vxlan.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
aswinsuryan: The code is refactored and cleanup handlers have been removed. So marking these comments as resolved.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+
+	// Enable loose mode (rp_filter=2) reverse path filtering on the vxlan interface.
+	// We won't ever create rp_filter, and its permissions are 644
+	// #nosec G306
+	err = ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+VxLANIface+"/rp_filter", []byte("2"), 0600)
+	if err != nil {
+		return fmt.Errorf("unable to update vxlan rp_filter proc entry, err: %s", err)
+	} else {
+		klog.V(log.DEBUG).Infof("Successfully configured rp_filter to loose mode(2) on %s", VxLANIface)
+	}
```
tpantelis:
```suggestion
	} 
		
	klog.V(log.DEBUG).Infof("Successfully configured rp_filter to loose mode(2) on %s", VxLANIface)

```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+
+type Operation int
+
+const (
+	Add Operation = iota
+	Delete
+	Flush
+)
+
+type vxLan struct {
```
tpantelis: You use a mix of `vxlan`, `vxLan` and `vxLAN` throughout this file - can we pick one for consistency? `vxlan`?

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+	localEndpoint types.SubmarinerEndpoint
+	connections   []v1.Connection
+	mutex         sync.Mutex
+	vxLanIface    *vxLanIface
+	NATTPort      int `default:"4500"`
```
tpantelis:
```suggestion
	nattPort      int `default:"4500"`
```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+	}
+
+	if err := createVxLanIface(vxLANIface); err != nil {
+		return nil, err
+	}
+
+	return vxLANIface, nil
+}
+
+func createVxLanIface(iface *vxLanIface) error {
```
tpantelis: It looks like this and maybe other code in this file was copied from _pkg/routeagent_driver/handlers/kubeproxy/vxlan.go_. Perhaps we should create a _pkg/vxlan_ package to reuse code between the two? (in a follow-up PR)

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+		State:        netlink.NUD_PERMANENT | netlink.NUD_NOARP,
+		HardwareAddr: macAddr,
+	}
+
+	err = netlink.NeighAppend(neigh)
+	if err != nil {
+		return fmt.Errorf("unable to add the bridge fdb entry %v, err: %s", neigh, err)
+	} else {
+		klog.V(log.DEBUG).Infof("Successfully added the bridge fdb entry %v", neigh)
+	}
```
tpantelis:
```suggestion
	} 
		
	klog.V(log.DEBUG).Infof("Successfully added the bridge fdb entry %v", neigh)
```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+
+	return nil
+}
+
+func (iface *vxLanIface) DelFDB(ipAddress net.IP, hwAddr string) error {
```
tpantelis: It's recommended to keep receiver var names to 1 or 2 chars (I thought we had linting that caught this). 
 
```suggestion
func (v *vxLanIface) DelFDB(ipAddress net.IP, hwAddr string) error {
```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+
+		if err == syscall.EEXIST {
+			err = netlink.RouteReplace(route)
+		}
+
+		if err != nil {
+			return fmt.Errorf("unable to add the route entry %v, err: %s", route, err)
+		} else {
+			klog.V(log.DEBUG).Infof("Successfully added the route entry %v and gw ip %v", route, gwIP)
+		}
```
tpantelis:
```suggestion
		} 
			
		klog.V(log.DEBUG).Infof("Successfully added the route entry %v and gw ip %v", route, gwIP)
```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+			Flags:     netlink.NTF_SELF,
+			Priority:  100,
+			Table:     TableID,
+		}
+		err := netlink.RouteDel(route)
+		if err != nil {
+			return fmt.Errorf("unable to add the route entry %v, err: %s", route, err)
+		} else {
+			klog.V(log.DEBUG).Infof("Successfully deleted the route entry %v", route)
+		}
```
tpantelis:
```suggestion
		} 
			
		klog.V(log.DEBUG).Infof("Successfully deleted the route entry %v", route)
```

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+	if v.localEndpoint.Spec.ClusterID == remoteEndpoint.Spec.ClusterID {
+		klog.V(log.DEBUG).Infof("Will not disconnect self")
+		return nil
+	}
+
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+
+	// parse remote addresses and allowed IPs
+	ip := submarinerEndpointIP(&remoteEndpoint)
```
tpantelis: hmm... `ConnectToEndpoint` uses `NATEndpointInfo.UseIP` but this is extracting the remote IP from the EP which I believe may not be the same.  Is that a potential issue?

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+
+func (v *vxLan) GetConnections() ([]v1.Connection, error) {
+	return v.connections, nil
+}
+
+func (v *vxLan) GetActiveConnections() ([]v1.Connection, error) {
+	return v.connections, nil
+}
+
+func (iface *vxLanIface) AddFDB(ipAddress net.IP, hwAddr string) error {
```
tpantelis: This looks to be duplicated from _pkg/routeagent_driver/handlers/kubeproxy/vxlan.go_ as well. Perhaps we can make `vxLanIface` reusable?

tpantelis (changes requested):

pkg/cable/vxlan/vxlan.go
aswinsuryan: I have raised this issue for it, will work on m3. https://github.com/submariner-io/submariner/issues/1393

pkg/cable/vxlan/vxlan.go
aswinsuryan: rasied https://github.com/submariner-io/submariner/issues/1393

pkg/cable/vxlan/vxlan.go
aswinsuryan: They should be in sync  I think @mangelajo any thoughts?

tpantelis: There must have been a good reason why @mangelajo changed `ConnectToEndpoint` to take the `NATEndpointInfo` 😄 I think the private IP can be selected with NAT enabled.

pkg/cable/vxlan/vxlan.go
aswinsuryan: Changed the name back to vxlan-tunnel since cleanup logic which is no-ops now deleted the line created by wireguard. It may be a challenge when we support cable driver per connection.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+}
+
+func NewDriver(localEndpoint types.SubmarinerEndpoint, localCluster types.SubmarinerCluster) (cable.Driver, error) {
+	var err error
+
+	v := vxlan{
+		localEndpoint: localEndpoint,
+	}
+
+	port, err := localEndpoint.Spec.GetBackendPort(v1.UDPPortConfig, int32(v.nattPort))
```
tpantelis: `v.nattPort` will be 0 here - it seems you want a default value of 4500 since you've defined a field tag as such. But there must be some code to parse the tag and set the field but I don't see that here.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+		return fmt.Errorf("failed to add ip rule: %v", err)
+	}
+
+	// Enable loose mode (rp_filter=2) reverse path filtering on the vxlan interface.
+	// We won't ever create rp_filter, and its permissions are 644
+	// #nosec G306
+	err = ioutil.WriteFile("/proc/sys/net/ipv4/conf/"+VxlanIface+"/rp_filter", []byte("2"), 0600)
+	if err != nil {
+		return fmt.Errorf("unable to update vxlan rp_filter proc entry, err: %s", err)
+	}
```
tpantelis: Please use the `netlink.EnableLooseModeReversePathFilter` for this.

tpantelis (changes requested):

pkg/cable/vxlan/vxlan.go
aswinsuryan: Made the logic simpler using a constant.

pkg/cable/vxlan/vxlan.go
aswinsuryan: done.

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,540 @@
...
+	v.vxlanIface.vtepIP = vtepIP
+
+	err = v.addIPRule()
+	if err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to add ip rule: %v", err)
+	}
+
+	// Enable loose mode (rp_filter=2) reverse path filtering on the vxlan interface.
+	// We won't ever create rp_filter, and its permissions are 644
+	// #nosec G306
```
tpantelis: nit - the #nosec comment no longer applies here - in fact I think all the comments here can be removed

tpantelis (changes requested):
Just one more nit - after that I'm good. @aswinsuryan There's quite a few commits with this PR - there some specific to review comments but I'm not sure if you want to squash all of them. Let me know if you want them all squashed on merge, otherwise please squash the ones you want when you push a final change.

tpantelis:
> Just one more nit - after that I'm good. @aswinsuryan There's quite a few commits with this PR - there some specific to review comments but I'm not sure if you want to squash all of them. Let me know if you want them all squashed on merge, otherwise please squash the ones you want when you push again.

pkg/cable/vxlan/vxlan.go
aswinsuryan: Done.

pkg/cable/vxlan/vxlan.go
aswinsuryan: > Just one more nit - after that I'm good. @aswinsuryan There's quite a few commits with this PR - there some specific to review comments but I'm not sure if you want to squash all of them. Let me know if you want them all squashed on merge, otherwise please squash the ones you want when you push a final change.

I squashed those, most of them were not relevant as the cleanup driver was refactored.

tpantelis (dismissed):

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,537 @@
...
+	netlinkAPI "github.com/submariner-io/submariner/pkg/netlink"
+	"github.com/submariner-io/submariner/pkg/types"
+	"github.com/submariner-io/submariner/pkg/util"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+	"k8s.io/klog"
+)
+
+const (
+	VxlanIface             = "vxlan-tunnel"
```
mangelajo, line 43: can we just call this one "submariner" as we do with wireguard?

Then it's just a standard in documentation and we don't need to handle special cases.
```suggestion
	VxlanIface             = "submariner"
```

mangelajo: ovn + vxlan testing is failing, you also need to add an entry here:

https://github.com/submariner-io/submariner/blob/devel/pkg/routeagent_driver/handlers/ovn/handler.go#L80

May be we can just standardize it to look for "submariner" interface regardless of the cable driver (ipsec has no interface)

aswinsuryan: Making it submariner cause issues when we support different cable drivers per link. Also when I tried changing it "submariner" , the vxlan-cleanup-driver in route agent , which does not check for the cable driver type , was deleting wireguard interface.

pkg/cable/vxlan/vxlan.go
mangelajo: We need the UseIP.

Although VXLAN will not survive NAT most likely very well, unless under very specific conditions.

Also I see that for connect you use the UseIP, and in disconnect, the other. That could create errors during disconnection, right?

Is there a way you can figure out the IP you used previously on the FDB?

I suggest may be:

1) Listing the FDB entries
2) Comparing to the Public & Private IPs of the endpoint. 

That will find you the right one.

pkg/cable/vxlan/vxlan.go
aswinsuryan: > can we just call this one "submariner" as we do with wireguard?
> 
> 
> 
> Then it's just a standard in documentation and we don't need to handle special cases.
> 
> ```suggestion
> 
> 	VxlanIface             = "submariner"
> 
> ```

Making it submariner cause issues when we support different cable drivers per link. Also when I tried changing it "submariner" , the vxlan-cleanup-driver in route agent , which does not check for the cable driver type , was deleting wireguard interface.

mangelajo (changes requested):

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,537 @@
...
+	VxlanOverhead          = 50
+	VxlanVTepNetworkPrefix = 241
+	CableDriverName        = "vxlan"
+	TableID                = 100
+	defaultNATTPort        = 4500
```
mangelajo: we can probably call this 
```suggestion
	defaultPort        = 4500
```

since it's not really about nat traversal in this case ':)

pkg/cable/vxlan/vxlan.go
mangelajo: Ohh, ok, what you're saying makes sense.

pkg/cable/vxlan/vxlan.go
aswinsuryan: Done

pkg/cable/vxlan/vxlan.go
aswinsuryan: we have a copy of the connections, now using that.

tpantelis (approved):

pkg/cable/vxlan/vxlan.go
```diff
@@ -0,0 +1,538 @@
...
+	var ip string
+
+	for _, connection := range v.connections {
+		if connection.Endpoint.CableName == remoteEndpoint.Spec.CableName {
+			ip = connection.UsingIP
+		}
+	}
+
+	if ip == "" {
+		return fmt.Errorf("failed to get remote IP %s", remoteEndpoint.Spec.CableName)
```
mangelajo: If we return an error, this will keep retrying, but the IP will not be found in the future (may be
if the endpoint is re-added, but at that time we really don't want to remove it anymore)

In this case I believe we should log the error and return nil.
```suggestion
		klog.Errorf("failed to get remote IP %s", remoteEndpoint.Spec.CableName)
		return nil
```

mangelajo (changes requested):
Just a tiny last thing.

pkg/cable/vxlan/vxlan.go
tpantelis: yeah I thought the same thing but callers don't specifically re-queue. It's fine either way although the caller does provide more context in the message. So if we want to log here then we could say "_Cannot disconnect remote endpoint %q - no prior connection entry found_".

aswinsuryan: Done.

tpantelis (approved):

mangelajo (approved):

PR #1172, opened 2021-03-02
Add GlobalEgressIP CRD and associated generated code

### LINKED ISSUE #1157: Create GlobalnetEgressIP and associated objects

sridhargaddam:
This PR adds the necessary types and includes the clientsets, informers and listers using codegenerator.

Fixes #1157


sridhargaddam:
@dfarrell07 included your review comments in this PR.

dfarrell07 (approved):

sridhargaddam:
Failure in unit-tests is not related to this PR.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type GlobalnetEgressIP struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the specification of desired behavior of GlobalnetEgressIP object
+	Spec EgressIPSpec `json:"spec"`
```
```suggestion
	Spec GlobalnetEgressIPSpec `json:"spec"`
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+
+	// Observed status of GlobalnetEgressIP. Its a read-only field.
+	// +optional
+	Status EgressIPStatus `json:"status,omitempty"`
```
```suggestion
	Status GlobalnetEgressIPStatus `json:"status,omitempty"`
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+}
+
+type EgressIPSpec struct {
+	// User can explicitly request a specific GlobalIP and this should be a valid IPaddress
+	// from globalnetCIDR allocated to the Cluster. If the requested globalIP falls outside
+	// of the globalnetCIDR or is already allocated (either fully or partially), then Status
+	// field will be set to Error with appropriate message.
+	// If this field is not specified, a single globalIP will be allocated.
+	// This field cannot be changed through updates.
```
```suggestion
	// The specific GlobalIP to request. This should be a valid IP address
	// from the Globalnet CIDR assigned to the cluster. If the requested GlobalIP falls outside
	// of the Globalnet CIDR or is already allocated (either fully or partially), then the Status
	// field will be set to "Error" with an appropriate Message set.
	// If not specified, GlobalIP(s) will be auto-allocated according to the NumGlobalIPs field.
	// This field cannot be changed through updates.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// +optional
+	GlobalIP string `json:"globalIP,omitempty"`
+
+	// The number of globalIPs requested. Globalnet Controller will allocate the requested
+	// number of contiguous GlobalIPs for this GlobalnetEgressIP object.
+	// User can specify this field while omitting GlobalIP. In such cases, Globalnet will
+	// auto-allocate the requested number of contiguous GlobalIPs.
+	// If unspecified, NumGlobalIPs defaults to 1.
```
tpantelis: I assume this is mutually exclusive with `GlobalIP` and that `GlobalIP` takes precedence. I also assume the IPs are allocated from the cluster's Globalnet CIDR.

```suggestion
	// The requested number of contiguous GlobalIPs to allocate from the Globalnet CIDR assigned to the cluster.
	// This field is ignored if the GlobalIP field is set. Defaults to 1.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// +optional
+	NumGlobalIPs int `json:"numGlobalIPs,omitempty"`
+
+	// Selects the Namespaces to which this GlobalnetEgressIP object applies.
+	// If an empty namespaceSelector: {} is configured, it selects all namespaces.
+	// If an empty namespaceSelector: {} is configured along with empty PodSelector, it applies
+	// to whole Cluster.
+	// On the other hand, if you omit specifying namespaceSelector, it does not select any
+	// namespaces and selects only Pods from the namespace where the GlobalnetEgressIP is
+	// deployed.
```
```suggestion
	// Selects the specific namespaces to which this GlobalnetEgressIP applies. Pods from matching namespaces 
	// will have GlobalIPs assigned.
	// If the selector is empty, all namespaces are selected.
	// If not specified, the namespace in which the GlobalnetEgressIP is deployed is used.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// +optional
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
```
tpantelis: To distinguish between unspecified/not set and and empty selector, this should be a pointer.

```suggestion
	NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// Selects the pods whose label match the definition. This is an optional field and
+	// in case it is not set, it results in all the Pods selected from the NamespaceSelector.
+	// In case it is set, it's intersected with NamespaceSelector, and only Pods that match the
+	// PodSelector and present in the NamespaceSelector will have the GlobalIPs as EgressIPs.
+	// Either namespaceSelector or podSelector have to be specified. If both are omitted,
+	// it's considered as an error.
+	// When a Pod or Namespace matches multiple GlobalnetEgressIP objects, there is no guarantee
+	// which of the globalIP address that are assigned to GlobalnetEgressIP will be used.
+	// +optional
+	PodSelector metav1.LabelSelector `json:"podSelector,omitempty"`
```
```suggestion
	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// On the other hand, if you omit specifying namespaceSelector, it does not select any
+	// namespaces and selects only Pods from the namespace where the GlobalnetEgressIP is
+	// deployed.
+	// +optional
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+
+	// Selects the pods whose label match the definition. This is an optional field and
+	// in case it is not set, it results in all the Pods selected from the NamespaceSelector.
+	// In case it is set, it's intersected with NamespaceSelector, and only Pods that match the
+	// PodSelector and present in the NamespaceSelector will have the GlobalIPs as EgressIPs.
```
```suggestion
	// Selects specific pods in the namespaces matched by the NamespaceSelector. 
	// If the selector is empty or not specified, all pods are selected.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// Either namespaceSelector or podSelector have to be specified. If both are omitted,
+	// it's considered as an error.
```
tpantelis: This is not actually true - according to the comments above, if the `NamespaceSelector` is omitted then it uses the namespace of the `GlobalnetEgressIP` and, if the `PodSelector` is omitted, all pods from the `NamespaceSelector` are selected. So both omitted is allowed.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// When a Pod or Namespace matches multiple GlobalnetEgressIP objects, there is no guarantee
+	// which of the globalIP address that are assigned to GlobalnetEgressIP will be used.
```
```suggestion
	// If a pod matches multiple GlobalnetEgressIP objects, there is no guarantee from which GlobalnetEgressIP  its GlobalIP will be assigned.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// +optional
+	PodSelector metav1.LabelSelector `json:"podSelector,omitempty"`
+}
+
+// GlobalnetEgressIPStatus identifies the status of GlobalnetEgressIP request.
+type GlobalnetEgressIPStatus string
+
+const (
+	// GlobalnetEgressIPSuccess means that Globalnet was able to successfully
+	// allocate the GlobalIP as requested in GlobalnetEgressIP object.
```
```suggestion
	// allocate the requested GlobalIP(s).
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	GlobalnetEgressIPSuccess GlobalnetEgressIPStatus = "Success"
+
+	// GlobalnetEgressIPError means that Globalnet was unable to allocate the
+	// requested GlobalIP.
+	GlobalnetEgressIPError GlobalnetEgressIPStatus = "Error"
+)
+
+type EgressIPStatus struct {
+	// Status is one of {"Success", "Error"}
```
```suggestion
	// Status describes the current state of the GlobalIP allocation.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	Status GlobalnetEgressIPStatus `json:"status,omitempty"`
+
+	// +optional
```
tpantelis, line 217:
```suggestion
        // Message that describes the details of the Status. 
	// +optional
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -156,3 +156,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	Message *string `json:"message,omitempty"`
+
+	// The list of GlobalIPs assigned via this GlobalnetEgressIP object.
```
```suggestion
	// The list of allocated GlobalIPs.
```

tpantelis (changes requested):

sridhargaddam:
bump

tpantelis:
bump

tpantelis:
bump

tpantelis:
bump

pkg/apis/submariner.io/v1/types.go
```diff
@@ -175,3 +175,79 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+type GlobalnetEgressIPConditionType string
+
+const (
+	GlobalnetEgressIPAllocated GlobalnetEgressIPConditionType = "Allocated"
+)
+
+type GlobalnetEgressIPStatus struct {
+	// Condition contains details for the current condition of this GlobalnetEgressIP.
+	// +optional
+	//Condition metav1.Condition `json:"condition,omitempty"`
```
tpantelis: `metav1.Condition` will be available once we bump to K8s 0.19.* (https://github.com/submariner-io/shipyard/pull/539).

pkg/apis/submariner.io/v1/types.go
```diff
@@ -175,3 +175,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// Spec is the specification of desired behavior.
+	Spec ClusterGlobalEgressIPSpec `json:"spec"`
+
+	// The most recently observed status. Read-only.
+	// +optional
+	Status GlobalEgressIPStatus `json:"status,omitempty"`
+}
+
+type ClusterGlobalEgressIPSpec struct {
+	// Specifies how GlobalIPs should be allocated.
```
sridhargaddam, line 259: @tpantelis can you please update the CRDs to the latest ones in the https://github.com/submariner-io/enhancements/pull/22/files

pkg/apis/submariner.io/v1/types.go
```diff
@@ -175,3 +175,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// The requested number of contiguous GlobalIPs to allocate from the Globalnet CIDR assigned to the cluster.
+	// If not specified, defaults to 1.
+	// +optional
+	NumGlobalIPs *int `json:"numGlobalIPs,omitempty"`
+}
+
+type GlobalEgressIPSpec struct {
+	// Specifies how GlobalIPs should be allocated.
+	// +optional
+	Allocation GlobalIPAllocationSpec `json:"allocation,omitempty"`
```
sridhargaddam: @tpantelis why not directly use `NumGlobalIPs` here instead of additional `GlobalIPAllocationSpec`?

pkg/apis/submariner.io/v1/types.go
```diff
@@ -175,3 +175,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type GlobalEgressIP struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the specification of the desired behavior.
+	Spec GlobalEgressIPSpec `json:"spec"`
```
sridhargaddam, line 188: Please add a note that captures the following
"The number of globalIP's requested for selected pods or all the Pods in the namespace the CR is applied"

pkg/apis/submariner.io/v1/types.go
```diff
@@ -175,3 +175,97 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ClusterGlobalEgressIP struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the specification of desired behavior.
```
sridhargaddam, line 247: Please add a note that captures the following
"The number of globalIP's requested at the cluster-level"

pkg/apis/submariner.io/v1/types.go
tpantelis: For reuse - it avoids defining `NumGlobalIPs` and its semantics twice and also if we add `GlobalIP` or something else. If you think it's overkill, we can remove it.

tpantelis (approved):

pkg/apis/submariner.io/v1/types.go
```diff
@@ -175,3 +175,100 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+	// GlobalIP will be assigned.
+	// +optional
+	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
+}
+
+type GlobalEgressIPConditionType string
+
+const (
+	GlobalEgressIPAllocated GlobalEgressIPConditionType = "Allocated"
+	GlobalEgressIPConflict  GlobalEgressIPConditionType = "Conflict"
```
sridhargaddam, line 213: Since user is not allowed to explicitly specify the `GlobalIP`, IMHO we do not require "Conflict". But we will require "Error" to specify that Globalnet was unable to allocate GlobalIP for certain reason.

pkg/apis/submariner.io/v1/types.go
sridhargaddam: Even if we add `GlobalIP` or something else in the future, it is also possible that we might add it only to one of them (f.e, we might allow admin to specify GlobalIP as part of `ClusterGlobalEgressIP` but not for regular users in `GlobalEgressIP` - just a thought). So, maybe we can remove the Allocation for now.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -175,3 +175,100 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a
...
+
+type GlobalEgressIPStatus struct {
+	// +optional
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// The list of allocated GlobalIPs.
```
sridhargaddam, line 224: Please include an optional message field. Globalnet can populate this with appropriate message when there is some Error.

    // +optional
    Message *string `json:"message,omitempty"`

sridhargaddam:
Thanks @tpantelis for updating the PR :)

pkg/apis/submariner.io/v1/types.go
tpantelis: `metav1.Condition` also has a `ConditionStatus` field defined as:

```
ConditionTrue    ConditionStatus = "True"
ConditionFalse   ConditionStatus = "False"
ConditionUnknown ConditionStatus = "Unknown"
```

Type defines a condition and ConditionStatus tells its status wrt that condition. So an error would have type "Allocated"  with ConditionStatus "False" and Reason/Message set appropriately.  

"Conflict" could be used if ClusterGlobalEgressIP is a singleton but the user created another one. Not sure if we'll need this type but doesn't hurt to define it.

pkg/apis/submariner.io/v1/types.go
tpantelis: `metav1.Condition` has a Message field.

```
type Condition struct {
	// type of condition in CamelCase or in foo.example.com/CamelCase.
	// ---
	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
	// useful (see .node.status.conditions), the ability to deconflict is important.
	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
	// +kubebuilder:validation:MaxLength=316
	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
	// status of the condition, one of True, False, Unknown.
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Enum=True;False;Unknown
	Status ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status"`
	// observedGeneration represents the .metadata.generation that the condition was set based upon.
	// For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
	// with respect to the current state of the instance.
	// +optional
	// +kubebuilder:validation:Minimum=0
	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,3,opt,name=observedGeneration"`
	// lastTransitionTime is the last time the condition transitioned from one status to another.
	// This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable.
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Type=string
	// +kubebuilder:validation:Format=date-time
	LastTransitionTime Time `json:"lastTransitionTime" protobuf:"bytes,4,opt,name=lastTransitionTime"`
	// reason contains a programmatic identifier indicating the reason for the condition's last transition.
	// Producers of specific condition types may define expected values and meanings for this field,
	// and whether the values are considered a guaranteed API.
	// The value should be a CamelCase string.
	// This field may not be empty.
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MaxLength=1024
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
	Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
	// message is a human readable message indicating details about the transition.
	// This may be an empty string.
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MaxLength=32768
	Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
}
```

pkg/apis/submariner.io/v1/types.go
tpantelis: ok

pkg/apis/submariner.io/v1/types.go
sridhargaddam: Okay cool. Looks good then.

pkg/apis/submariner.io/v1/types.go
sridhargaddam: Thanks @tpantelis. So we can use it instead of an additional message field.

pkg/apis/submariner.io/v1/types.go
tpantelis: We could use type "Valid", as `ServiceExport` does, instead of "Allocated".

sridhargaddam (commented):
There are some linting failures, otherwise LGTM thanks @tpantelis.

pkg/apis/submariner.io/v1/types.go
sridhargaddam: Normally we use the term, globalIP allocation. Personally, I'd prefer "Allocated" over "Valid"

PR #1181, opened 2021-03-03
Fix nil pointer panic in svcExUpdater

tpantelis:
... if the Service isn't found. This was seen in a unit test run.


nyechiel (approved):

skitt (approved):
Ignore that, I missed the case where `err` is non-`nil` but isn’t “not found”.

PR #1184, opened 2021-03-05
Use the K8s iptables wrapper everywhere

skitt:
This will allow us to drop the host mount requirement on all pods.
An iptables-equipped base image is used to reduce the time spent
building the images.

This requires using Fedora; UBI images don’t carry both iptables
variants. We’ll need to resolve this at some point, but for now
dropping the host mount seems more useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
-IMAGES ?= submariner submariner-gateway submariner-route-agent submariner-globalnet submariner-networkplugin-syncer
+IMAGES ?= submariner-iptables-base submariner submariner-gateway submariner-route-agent submariner-globalnet submariner-networkplugin-syncer
 images: build
 
 include $(SHIPYARD_DIR)/Makefile.inc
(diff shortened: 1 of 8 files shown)


mangelajo (commented):
This will try to publish the new image to quay. 

We can always improve shipyard to not publish some of the images, but I think it would be simpler to not use a base image.

skitt:
> This will try to publish the new image to quay.
> 
> We can always improve shipyard to not publish some of the images, but I think it would be simpler to not use a base image.

I did this because the main objection to a previous version of this patch was that, by relying on Fedora everywhere, the build ended up taking too long.

skitt:
> > This will try to publish the new image to quay.
> > We can always improve shipyard to not publish some of the images, but I think it would be simpler to not use a base image.
> 
> I did this because the main objection to a previous version of this patch was that, by relying on Fedora everywhere, the build ended up taking too long.

The alternative is https://github.com/submariner-io/submariner/pull/1186. The resulting images are around 200MiB in #1186, double that here...

tpantelis (approved):

skitt:
Dropped in favour of #1186.

mangelajo:
Sorry clicked wrong button on project manager.

Issue #1175, opened 2021-03-03
Define a protocol for NAT discovery


mangelajo:
Given:
* Local IP, Local Public IP, Local Port
* Remote IP, Remote Public IP, Remote Port
* Local ID, Remote ID

The goals of the protocol would be:

1) **Select between Remote IP or Remote Public IP** for connectivity
2) Ensure remote ID & Local ID matches,
3) Identify if our IP is detected as "Local IP", "Local Public IP", or neither

2 & 3 are for tracing or warning about possible issues in the Datapath (for example, incorrectly mapped ports).

Issue #1174, opened 2021-03-03
Epic: Proper NAT support for all combinations


mangelajo:
**What would you like to be added**:

### Automatic NAT/NO-NAT discovery between two endpoints (fix https://github.com/submariner-io/submariner/issues/300 )
* [x] (#1175) Define protocol (see goals of the protocol in the issue), select a port number
* [x] (#1176) Implement protocol pkg + testing
* [x] (#1177) Implement in gateways

### Individual port assignments to gateways
 * [x] (#1178) (Expose the cabledriver gateway ports on the Endpoint information
 * [x] (#1179) Allow IPSec port selection with labels see #1146 code for reference
 
### Additional work
* [x] (#1224) Support client/server model for IPSec
* [x] Support enforcement of UDP encapsulation for environments where NAT is not detected, but ESP/AH is not supported.

**Why is this needed**:

With a mix of on-prem & on-cloud deployments the current global nat/no-nat doesn't work. Also from the point of view of HA, such NAT support would also require individual port numbers for each gateway node. See https://github.com/submariner-io/submariner/issues/300

PR #1185, opened 2021-03-08
Define a nat-discovery protocol for selecting public or private IP

mangelajo:
This protocol is defined for the discovery 1:1 of the IP to be used. between private or public, and expected to be used
over an UDP port.

I have been thinking about the usability/security consequences. One important fact is that the implementation should only rely on the queries made, never on the queries received. So a gateway cannot be tricked into changing the IP being used (DoS)

We could use mTLS over TCP for this discovery if we wanted using the same proto definition, but I was talking about this with @nyechiel  and may be at this point it's more reasonable to keep it simple, unless we believe we'd be sharing anything that should not be on plain text.

I suggest the cluster / endpoint ids to, may be, be hashed.

Related-Issue: https://github.com/submariner-io/submariner/issues/1174
Closes-Issue: https://github.com/submariner-io/submariner/issues/1175

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -5,6 +5,14 @@ ENV DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD GITHUB_REF CLUSTERS_ARGS DE
     DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner DAPPER_DOCKER_SOCKET=true
 ENV DAPPER_OUTPUT=${DAPPER_SOURCE}/output
 
+ENV PROTOC_VERSION=3.15.5
+
+# alternative: dnf install -y --nodocs --setopt=install_weak_deps=False golang-google-protobuf protobuf-compiler && dnf -y clean all
+RUN curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v3.15.5/protoc-${PROTOC_VERSION}-linux-x86_64.zip && \
+    unzip protoc-${PROTOC_VERSION}-linux-x86_64.zip -d /usr/local && \
+    rm -f protoc-${PROTOC_VERSION}-linux-x86_64.zip && \
+    GO111MODULE=off go get google.golang.org/protobuf/cmd/protoc-gen-go
+
 WORKDIR ${DAPPER_SOURCE}
 
 ENTRYPOINT ["/opt/shipyard/scripts/entry"]
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,10 @@ deploy: images
 reload-images: build images
 	./scripts/$@ --restart $(restart)
 
-bin/%/submariner-gateway: vendor/modules.txt main.go $(shell find pkg -not \( -path 'pkg/globalnet*' -o -path 'pkg/routeagent*' \))
+%.pb.go: %.proto
+	protoc --go_out=/go/src $<
+
+bin/%/submariner-gateway: vendor/modules.txt main.go $(shell find pkg -not \( -path 'pkg/globalnet*' -o -path 'pkg/routeagent*' \)) pkg/natdiscovery/proto/natdiscovery.pb.go
 	GOARCH=$(call dockertogoarch,$(patsubst bin/linux/%/,%,$(dir $@))) ${SCRIPTS_DIR}/compile.sh $@ main.go $(BUILD_ARGS)
 
 bin/%/submariner-route-agent: vendor/modules.txt $(shell find pkg/routeagent_driver)
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,10 @@ dockertogoarch = $(patsubst arm/v7,arm,$(1))
 
 deploy: images
 
+golangci-lint: pkg/natdiscovery/proto/natdiscovery.pb.go
+
+unit: pkg/natdiscovery/proto/natdiscovery.pb.go
+
 reload-images: build images
 	./scripts/$@ --restart $(restart)
 
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -70,14 +70,15 @@ type EndpointSpec struct {
 	ClusterID string `json:"cluster_id"`
 	CableName string `json:"cable_name"`
 	// +optional
-	HealthCheckIP string            `json:"healthCheckIP,omitempty"`
-	Hostname      string            `json:"hostname"`
-	Subnets       []string          `json:"subnets"`
-	PrivateIP     string            `json:"private_ip"`
-	PublicIP      string            `json:"public_ip"`
-	NATEnabled    bool              `json:"nat_enabled"`
-	Backend       string            `json:"backend"`
-	BackendConfig map[string]string `json:"backend_config,omitempty"`
+	HealthCheckIP    string            `json:"healthCheckIP,omitempty"`
+	Hostname         string            `json:"hostname"`
+	Subnets          []string          `json:"subnets"`
+	PrivateIP        string            `json:"private_ip"`
+	PublicIP         string            `json:"public_ip"`
+	NATEnabled       bool              `json:"nat_enabled"`
+	Backend          string            `json:"backend"`
+	BackendConfig    map[string]string `json:"backend_config,omitempty"`
+	NATDiscoveryPort *int32            `json:"natDiscoveryPort,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/natdiscovery/listener.go
+++ b/pkg/natdiscovery/listener.go
@@ -1,0 +1,73 @@
+/*
+© 2021 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package natdiscovery
+
+import (
+	"net"
+	"strconv"
+	"time"
+
+	"github.com/pkg/errors"
+	"google.golang.org/protobuf/proto"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+
+	natproto "github.com/submariner-io/submariner/pkg/natdiscovery/proto"
+)
+
+func (nd *natDiscovery) runListener(stopCh <-chan struct{}) error {
+	serverAddress, err := net.ResolveUDPAddr("udp4", ":"+strconv.Itoa(natproto.DefaultPort))
+	if err != nil {
+		return errors.Wrap(err, "Error resolving UDP address")
+	}
+
+	serverConnection, err := net.ListenUDP("udp", serverAddress)
+	if err != nil {
+		return errors.Wrapf(err, "Error listening on udp port %d", natproto.DefaultPort)
+	}
+
+	// Instead of storing the server connection I save the reference to the WriteToUDP
+	// of our server connection instance, in a way that we can use this for unit testing
+	// later too.
+	nd.serverUDPWrite = serverConnection.WriteToUDP
+
+	go wait.Until(func() {
+		buf := make([]byte, 2048)
+		if _, addr, err := serverConnection.ReadFromUDP(buf); err != nil {
+			klog.Errorf("Error receiving from udp: %s", err)
+		} else if err := nd.parseAndHandleMessageFromAddress(buf, addr); err != nil {
+			klog.Errorf("Error handling message from address %#v: %s", addr, err)
+		}
+	}, time.Second, stopCh)
+
+	return nil
+}
+
+func (nd *natDiscovery) parseAndHandleMessageFromAddress(buf []byte, addr *net.UDPAddr) error {
+	msg := natproto.SubmarinerNatDiscoveryMessage{}
+	if err := proto.Unmarshal(buf, &msg); err != nil {
+		return errors.Wrapf(err, "Error unmarshaling message received on UDP port %d", natproto.DefaultPort)
+	}
+
+	if request := msg.GetRequest(); request != nil {
+		return nd.handleRequestFromAddress(request, addr)
+	} else if response := msg.GetResponse(); response != nil {
+		return nd.handleResponseFromAddress(response, addr)
+	}
+
+	return errors.Errorf("Message without response or request received from %#v", addr)
+}
--- a/pkg/natdiscovery/natdiscovery.go
+++ b/pkg/natdiscovery/natdiscovery.go
@@ -1,0 +1,1 @@
+package natdiscovery
--- a/pkg/natdiscovery/natdiscovery.go
+++ b/pkg/natdiscovery/natdiscovery.go
@@ -1,1 +1,156 @@
+/*
+© 2021 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package natdiscovery
+
+import (
+	"crypto/rand"
+	"math/big"
+	"net"
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+type Interface interface {
+	Run(stopCh <-chan struct{}) error
+	AddEndpoint(endpoint *types.SubmarinerEndpoint)
+	RemoveEndpoint(endpointName string)
+	SetReadyChannel(readyChannel chan *NATEndpointInfo)
+}
+
+type udpWriteFunction func(b []byte, addr *net.UDPAddr) (int, error)
+type findSrcIPFunction func(destinationIP string) (string, error)
+
+type natDiscovery struct {
+	sync.Mutex
+	localEndpoint   *types.SubmarinerEndpoint
+	remoteEndpoints map[string]*remoteEndpointNAT
+	requestCounter  uint64
+	serverUDPWrite  udpWriteFunction
+	findSrcIP       findSrcIPFunction
+	serverPort      int32
+	readyChannel    chan *NATEndpointInfo
+}
+
+func (nd *natDiscovery) SetReadyChannel(readyChannel chan *NATEndpointInfo) {
+	nd.readyChannel = readyChannel
+}
+
+func New(localEndpoint *types.SubmarinerEndpoint) (Interface, error) {
+	return newNatDiscovery(localEndpoint)
+}
+
+func newNatDiscovery(localEndpoint *types.SubmarinerEndpoint) (*natDiscovery, error) {
+	port, err := extractNATDiscoveryPort(localEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	requestCounter, err := randomRequestCounter()
+	if err != nil {
+		return nil, err
+	}
+
+	return &natDiscovery{
+		localEndpoint:   localEndpoint,
+		serverPort:      port,
+		remoteEndpoints: map[string]*remoteEndpointNAT{},
+		findSrcIP:       findPreferredSourceIP,
+		requestCounter:  requestCounter,
+	}, nil
+}
+
+func randomRequestCounter() (uint64, error) {
+	n, err := rand.Int(rand.Reader, big.NewInt(1000))
+	if err != nil {
+		return 0, errors.Wrapf(err, "generating random request counter")
+	}
+
+	return n.Uint64(), nil
+}
+
+var errorNoNatDiscoveryPort = errors.New("nat discovery port missing in endpoint")
+
+func extractNATDiscoveryPort(endpoint *types.SubmarinerEndpoint) (int32, error) {
+	if endpoint.Spec.NATDiscoveryPort == nil {
+		return 0, errorNoNatDiscoveryPort
+	}
+
+	return *endpoint.Spec.NATDiscoveryPort, nil
+}
+
+func (nd *natDiscovery) Run(stopCh <-chan struct{}) error {
+	err := nd.runListener(stopCh)
+	if err != nil {
+		return err
+	}
+
+	go wait.Until(func() {
+		nd.checkEndpointList()
+	}, time.Second, stopCh)
+
+	return nil
+}
+
+func (nd *natDiscovery) AddEndpoint(endpoint *types.SubmarinerEndpoint) {
+	nd.Lock()
+	defer nd.Unlock()
+
+	if ep, exists := nd.remoteEndpoints[endpoint.Spec.CableName]; exists {
+		if reflect.DeepEqual(ep.endpoint.Spec, endpoint.Spec) {
+			return
+		} else {
+			delete(nd.remoteEndpoints, endpoint.Spec.CableName)
+		}
+	}
+
+	remoteNAT := newRemoteEndpointNAT(endpoint)
+
+	// support a remote cluster endpoint which still hasn't implemented this protocol
+	if _, err := extractNATDiscoveryPort(endpoint); err == errorNoNatDiscoveryPort {
+		remoteNAT.useLegacyNATSettings()
+	}
+
+	nd.remoteEndpoints[endpoint.Spec.CableName] = remoteNAT
+}
+
+func (nd *natDiscovery) RemoveEndpoint(endpointName string) {
+	nd.Lock()
+	defer nd.Unlock()
+	delete(nd.remoteEndpoints, endpointName)
+}
+
+func (nd *natDiscovery) checkEndpointList() {
+	nd.Lock()
+	defer nd.Unlock()
+
+	for _, endpointNAT := range nd.remoteEndpoints {
+		if endpointNAT.shouldCheck() {
+			if endpointNAT.hasTimedOut() {
+				endpointNAT.useLegacyNATSettings()
+			} else if err := nd.sendCheckRequest(endpointNAT); err != nil {
+				klog.Errorf("Error sending check request to %q: %s", endpointNAT.endpoint.Spec.CableName, err)
+			}
+		}
+	}
+}
--- a/pkg/natdiscovery/natdiscovery_suite_test.go
+++ b/pkg/natdiscovery/natdiscovery_suite_test.go
@@ -1,0 +1,1 @@
+package natdiscovery
--- a/pkg/natdiscovery/natdiscovery_suite_test.go
+++ b/pkg/natdiscovery/natdiscovery_suite_test.go
@@ -1,1 +1,131 @@
+/*
+© 2021 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
 package natdiscovery
+
+import (
+	"net"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"google.golang.org/protobuf/proto"
+	"k8s.io/klog"
+
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	natproto "github.com/submariner-io/submariner/pkg/natdiscovery/proto"
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+func init() {
+	klog.InitFlags(nil)
+}
+
+func TestNATDiscovery(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "NAT Discovery Suite")
+}
+
+const (
+	testLocalEndpointName = "cluster-a-ep-1"
+	testLocalClusterID    = "cluster-a"
+	testLocalPublicIP     = "10.1.1.1"
+	testLocalPrivateIP    = "2.2.2.2"
+
+	testRemoteEndpointName        = "cluster-b-ep-1"
+	testRemoteUnknownEndpointName = "cluster-b-ep-unknown"
+	testRemoteClusterID           = "cluster-b"
+	testRemotePublicIP            = "10.3.3.3"
+	testRemotePrivateIP           = "4.4.4.4"
+)
+
+var (
+	testLocalNATPort  int32 = 1234
+	testRemoteNATPort int32 = 4321
+)
+
+func parseProtocolRequest(buf []byte) *natproto.SubmarinerNatDiscoveryRequest {
+	msg := natproto.SubmarinerNatDiscoveryMessage{}
+	if err := proto.Unmarshal(buf, &msg); err != nil {
+		klog.Errorf("error unmarshaling message received on UDP port %d: %s", natproto.DefaultPort, err)
+	}
+
+	request := msg.GetRequest()
+	Expect(request).NotTo(BeNil())
+
+	return request
+}
+
+func parseProtocolResponse(buf []byte) *natproto.SubmarinerNatDiscoveryResponse {
+	msg := natproto.SubmarinerNatDiscoveryMessage{}
+	if err := proto.Unmarshal(buf, &msg); err != nil {
+		klog.Errorf("error unmarshaling message received on UDP port %d: %s", natproto.DefaultPort, err)
+	}
+
+	response := msg.GetResponse()
+	Expect(response).NotTo(BeNil())
+
+	return response
+}
+
+func createTestListener(endpoint *types.SubmarinerEndpoint) (*natDiscovery, chan []byte) {
+	listener, err := newNatDiscovery(endpoint)
+	readyChannel := make(chan *NATEndpointInfo, 100)
+	listener.SetReadyChannel(readyChannel)
+	Expect(err).NotTo(HaveOccurred())
+
+	udpSentChannel := make(chan []byte, 10)
+	listener.serverUDPWrite = func(b []byte, addr *net.UDPAddr) (int, error) {
+		udpSentChannel <- b
+		return len(b), nil
+	}
+
+	return listener, udpSentChannel
+}
+func createTestLocalEndpoint() types.SubmarinerEndpoint {
+	return types.SubmarinerEndpoint{
+		Spec: submarinerv1.EndpointSpec{
+			CableName:        testLocalEndpointName,
+			ClusterID:        testLocalClusterID,
+			PublicIP:         testLocalPublicIP,
+			PrivateIP:        testLocalPrivateIP,
+			NATDiscoveryPort: &testLocalNATPort,
+		},
+	}
+}
+
+func createTestRemoteEndpoint() types.SubmarinerEndpoint {
+	return types.SubmarinerEndpoint{
+		Spec: submarinerv1.EndpointSpec{
+			CableName:        testRemoteEndpointName,
+			ClusterID:        testRemoteClusterID,
+			PublicIP:         testRemotePublicIP,
+			PrivateIP:        testRemotePrivateIP,
+			NATDiscoveryPort: &testRemoteNATPort,
+		},
+	}
+}
+
+func createTestUnknownRemoteEndpoint() types.SubmarinerEndpoint {
+	return types.SubmarinerEndpoint{
+		Spec: submarinerv1.EndpointSpec{
+			CableName:        testRemoteUnknownEndpointName,
+			ClusterID:        testRemoteClusterID,
+			PublicIP:         testRemotePublicIP,
+			PrivateIP:        testRemotePrivateIP,
+			NATDiscoveryPort: &testRemoteNATPort,
+		},
+	}
+}
--- a/pkg/natdiscovery/proto/consts.go
+++ b/pkg/natdiscovery/proto/consts.go
@@ -16,4 +16,5 @@ limitations under the License.
 
 package proto
 
-const Port = 4499
+const DefaultPort = 4490
+const Version = 1
--- a/pkg/natdiscovery/proto/consts.go
+++ b/pkg/natdiscovery/proto/consts.go
@@ -1,0 +1,19 @@
+/*
+© 2021 Red Hat, Inc. and others
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proto
+
+const Port = 4499
--- a/pkg/natdiscovery/proto/natdiscovery.proto
+++ b/pkg/natdiscovery/proto/natdiscovery.proto
@@ -44,9 +44,8 @@ enum ResponseType {
   SRC_MODIFIED = 1;
   UNKNOWN_DST_CLUSTER = 2;
   UNKNOWN_DST_ENDPOINT = 3;
-  UNKNOWN_SRC_CLUSTER = 4;
-  UNKNOWN_SRC_ENDPOINT = 5;
-  MALFORMED = 6;
+  UNKNOWN_SRC_ENDPOINT = 4;
+  MALFORMED = 5;
 }
 
 message SubmarinerNatDiscoveryResponse {
--- a/pkg/natdiscovery/proto/natdiscovery.proto
+++ b/pkg/natdiscovery/proto/natdiscovery.proto
@@ -1,0 +1,73 @@
+/*
+© 2021 Red Hat, Inc. and others
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+
+syntax = "proto3";
+option go_package = "github.com/submariner-io/submariner/pkg/natdiscovery/proto";
+
+message SubmarinerNatDiscoveryMessage {
+  int32 version = 1;
+
+  oneof message {
+    SubmarinerNatDiscoveryRequest request = 2;
+    SubmarinerNatDiscoveryResponse response = 3;
+  }
+}
+
+message SubmarinerNatDiscoveryRequest {
+  uint64 request_number = 1;
+
+  EndpointDetails sender = 2;
+  EndpointDetails receiver = 3;
+
+  // The following information would allow the receiver to identify
+  // and log if any form of NAT traversal is happening on the path
+  IPPortPair using_src = 4;
+  IPPortPair using_dst = 5;
+}
+
+enum ResponseType {
+  OK = 0;
+  SRC_MODIFIED = 1;
+  UNKNOWN_DST_CLUSTER = 2;
+  UNKNOWN_DST_ENDPOINT = 3;
+  UNKNOWN_SRC_CLUSTER = 4;
+  UNKNOWN_SRC_ENDPOINT = 5;
+  MALFORMED = 6;
+}
+
+message SubmarinerNatDiscoveryResponse {
+  uint64 request_number = 1;
+
+  ResponseType response = 2;
+  EndpointDetails sender = 3;
+  EndpointDetails receiver = 4;
+
+  // The received SRC IP / SRC port is reported, which will be useful for
+  // diagnosing corner cases
+  IPPortPair received_src = 5;
+}
+
+message IPPortPair {
+  string IP = 1;
+  uint32 port = 2;
+}
+
+message EndpointDetails {
+  // should we hash this for privacy? a hash can be checked against a known list, but can't be decoded
+  string cluster_id = 1;
+  string endpoint_id = 2;
+}
--- a/pkg/natdiscovery/remote_endpoint.go
+++ b/pkg/natdiscovery/remote_endpoint.go
@@ -1,0 +1,153 @@
+/*
+© 2021 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package natdiscovery
+
+import (
+	"time"
+
+	"github.com/pkg/errors"
+
+	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+type endpointState int
+
+const (
+	testingPrivateAndPublicIPs endpointState = iota
+	waitingForResponse
+	selectedPublicIP
+	selectedPrivateIP
+	recheckTime                    = 2 * time.Second
+	totalTimeout                   = 60 * time.Second
+	publicToPrivateFailoverTimeout = 1 * time.Second
+)
+
+type remoteEndpointNAT struct {
+	endpoint               *types.SubmarinerEndpoint
+	state                  endpointState
+	lastCheck              time.Time
+	lastTransition         time.Time
+	started                time.Time
+	useNAT                 bool
+	useIP                  string
+	lastPublicIPRequestID  uint64
+	lastPrivateIPRequestID uint64
+}
+
+type NATEndpointInfo struct {
+	Endpoint subv1.EndpointSpec
+	UseNAT   bool
+	UseIP    string
+}
+
+func (rn *remoteEndpointNAT) toNATEndpointInfo() *NATEndpointInfo {
+	return &NATEndpointInfo{
+		Endpoint: rn.endpoint.Spec,
+		UseNAT:   rn.useNAT,
+		UseIP:    rn.useIP,
+	}
+}
+
+func newRemoteEndpointNAT(endpoint *types.SubmarinerEndpoint) *remoteEndpointNAT {
+	return &remoteEndpointNAT{
+		endpoint:       endpoint,
+		state:          testingPrivateAndPublicIPs,
+		started:        time.Now(),
+		lastTransition: time.Now(),
+	}
+}
+
+func (rn *remoteEndpointNAT) transitionToState(newState endpointState) {
+	rn.lastTransition = time.Now()
+	rn.state = newState
+}
+
+func (rn *remoteEndpointNAT) sinceLastTransition() time.Duration {
+	return time.Since(rn.lastTransition)
+}
+
+func (rn *remoteEndpointNAT) hasTimedOut() bool {
+	return time.Since(rn.started) > totalTimeout
+}
+
+func (rn *remoteEndpointNAT) useLegacyNATSettings() {
+	rn.useNAT = rn.endpoint.Spec.NATEnabled
+	if rn.endpoint.Spec.NATEnabled {
+		rn.useIP = rn.endpoint.Spec.PublicIP
+		rn.transitionToState(selectedPublicIP)
+	} else {
+		rn.useIP = rn.endpoint.Spec.PrivateIP
+		rn.transitionToState(selectedPrivateIP)
+	}
+}
+
+func (rn *remoteEndpointNAT) shouldCheck() bool {
+	switch rn.state {
+	case testingPrivateAndPublicIPs:
+		return true
+	case waitingForResponse:
+		return time.Since(rn.lastCheck) > recheckTime
+	default:
+		return false
+	}
+}
+
+func (rn *remoteEndpointNAT) checkSent() {
+	if rn.state == testingPrivateAndPublicIPs {
+		rn.state = waitingForResponse
+	}
+
+	rn.lastCheck = time.Now()
+}
+
+func (rn *remoteEndpointNAT) transitionToPublicIP(remoteEndpointID string, useNAT bool) error {
+	if rn.state == waitingForResponse {
+		rn.useIP = rn.endpoint.Spec.PublicIP
+		rn.useNAT = useNAT
+		rn.transitionToState(selectedPublicIP)
+	} else {
+		return errors.Errorf("received unexpected transition to public IP from endpoint %q", remoteEndpointID)
+	}
+
+	return nil
+}
+
+func (rn *remoteEndpointNAT) transitionToPrivateIP(remoteEndpointID string, useNAT bool) error {
+	switch rn.state {
+	case waitingForResponse:
+		rn.useIP = rn.endpoint.Spec.PrivateIP
+		rn.useNAT = useNAT
+		rn.transitionToState(selectedPrivateIP)
+
+	case selectedPublicIP:
+		// If a PublicIP was selected, we still allow some time for the privateIP response to arrive, and we always
+		// prefer PrivateIP with no NAT connection, as it will be more likely to work, and more efficient
+		if rn.sinceLastTransition() > publicToPrivateFailoverTimeout {
+			return errors.Errorf("response on private address received too late after response on public address for endpoint %q",
+				remoteEndpointID)
+		}
+
+		rn.useIP = rn.endpoint.Spec.PrivateIP
+		rn.useNAT = useNAT
+		rn.transitionToState(selectedPrivateIP)
+
+	default:
+		return errors.Errorf("received unexpected transition to private IP from endpoint %q", remoteEndpointID)
+	}
+
+	return nil
+}
--- a/pkg/natdiscovery/remote_endpoint_test.go
+++ b/pkg/natdiscovery/remote_endpoint_test.go
@@ -1,0 +1,188 @@
+/*
+© 2021 Red Hat, Inc. and others
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package natdiscovery
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("remoteEndpointNAT", func() {
+
+	var rnat *remoteEndpointNAT
+
+	BeforeEach(func() {
+		remoteEndpoint := createTestRemoteEndpoint()
+		rnat = newRemoteEndpointNAT(&remoteEndpoint)
+	})
+
+	When("first created", func() {
+		It("shouldCheck should return true", func() {
+			Expect(rnat.shouldCheck()).To(BeTrue())
+		})
+	})
+
+	When("right after a check has been sent", func() {
+		It("shouldCheck should return false", func() {
+			rnat.checkSent()
+			Expect(rnat.shouldCheck()).To(BeFalse())
+		})
+	})
+
+	When("the total timeout has elapsed", func() {
+		It("should report as timed out", func() {
+			rnat.started = time.Now().Add(-totalTimeout)
+			Expect(rnat.hasTimedOut()).To(BeTrue())
+		})
+	})
+
+	When("using legacy settings", func() {
+		Context("and NAT is not enabled", func() {
+			It("should select the private IP", func() {
+				rnat.endpoint.Spec.NATEnabled = false
+				rnat.useLegacyNATSettings()
+				Expect(rnat.state).To(Equal(selectedPrivateIP))
+				Expect(rnat.useIP).To(Equal(rnat.endpoint.Spec.PrivateIP))
+			})
+		})
+		Context("and NAT is enabled", func() {
+			It("should select the public IP", func() {
+				rnat.endpoint.Spec.NATEnabled = true
+				rnat.useLegacyNATSettings()
+				Expect(rnat.state).To(Equal(selectedPublicIP))
+				Expect(rnat.useIP).To(Equal(rnat.endpoint.Spec.PublicIP))
+			})
+		})
+	})
+
+	When("the public IP is selected but no check was sent", func() {
+		It("it should fail", func() {
+			err := rnat.transitionToPublicIP(testRemoteEndpointName, false)
+			Expect(err).To(HaveOccurred())
+			Expect(rnat.useIP).To(Equal(""))
+		})
+	})
+
+	When("the private IP is selected but no check was sent", func() {
+		It("it should fail", func() {
+			err := rnat.transitionToPrivateIP(testRemoteEndpointName, false)
+			Expect(err).To(HaveOccurred())
+			Expect(rnat.useIP).To(Equal(""))
+		})
+	})
+
+	When("the private IP is selected", func() {
+		var useNAT bool
+
+		JustBeforeEach(func() {
+			rnat.checkSent()
+			err := rnat.transitionToPrivateIP(testRemoteEndpointName, useNAT)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(rnat.state).To(Equal(selectedPrivateIP))
+		})
+
+		It("should use the private IP", func() {
+			Expect(rnat.useIP).To(Equal(rnat.endpoint.Spec.PrivateIP))
+		})
+
+		Context("with NAT discovered", func() {
+			BeforeEach(func() {
+				useNAT = true
+			})
+
+			It("should set useNAT to true", func() {
+				Expect(rnat.useNAT).To(BeTrue())
+			})
+		})
+
+		Context("with no NAT discovered", func() {
+			BeforeEach(func() {
+				useNAT = false
+			})
+
+			It("should set useNAT to false", func() {
+				Expect(rnat.useNAT).To(BeFalse())
+			})
+		})
+	})
+
+	When("the public IP is selected", func() {
+		var useNAT bool
+
+		JustBeforeEach(func() {
+			rnat.checkSent()
+			err := rnat.transitionToPublicIP(testRemoteEndpointName, useNAT)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(rnat.state).To(Equal(selectedPublicIP))
+		})
+
+		It("should use the public IP", func() {
+			Expect(rnat.useIP).To(Equal(rnat.endpoint.Spec.PublicIP))
+		})
+
+		Context("with NAT discovered", func() {
+			BeforeEach(func() {
+				useNAT = true
+			})
+
+			It("should set useNAT to true", func() {
+				Expect(rnat.useNAT).To(BeTrue())
+			})
+		})
+
+		Context("with no NAT discovered", func() {
+			BeforeEach(func() {
+				useNAT = false
+			})
+
+			It("should set useNAT to false", func() {
+				Expect(rnat.useNAT).To(BeFalse())
+			})
+		})
+	})
+
+	When("private IP is selected right after publicIP ", func() {
+		Context("and the grace period has not elapsed", func() {
+			It("should use the private IP", func() {
+				rnat.checkSent()
+				err := rnat.transitionToPublicIP(testRemoteEndpointName, true)
+				Expect(err).NotTo(HaveOccurred())
+				err = rnat.transitionToPrivateIP(testRemoteEndpointName, false)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(rnat.state).To(Equal(selectedPrivateIP))
+				Expect(rnat.useIP).To(Equal(rnat.endpoint.Spec.PrivateIP))
+				Expect(rnat.useNAT).To(BeFalse())
+			})
+		})
+
+		Context("and the grace period has elapsed", func() {
+			It("should still use the public IP", func() {
+				rnat.checkSent()
+				err := rnat.transitionToPublicIP(testRemoteEndpointName, true)
+				Expect(err).NotTo(HaveOccurred())
+				rnat.lastTransition = rnat.lastTransition.Add(-publicToPrivateFailoverTimeout)
+				err = rnat.transitionToPrivateIP(testRemoteEndpointName, false)
+				Expect(err).To(HaveOccurred())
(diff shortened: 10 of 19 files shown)


mangelajo:
Converting to draft by now until I fully handle the protocol implementation.

pkg/natdiscovery/listener_test.go
```diff
@@ -0,0 +1,104 @@
+/*
```
mangelajo: to remove from this patch.

mangelajo:
@skitt although this is marked as draft, the first commit of the series is reviewable. I'd thank your feedback on it, specially the makefile changes.

Dockerfile.dapper
```diff
@@ -5,6 +5,14 @@ ENV DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD GITHUB_REF CLUSTERS_ARGS DE
     DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner DAPPER_DOCKER_SOCKET=true
 ENV DAPPER_OUTPUT=${DAPPER_SOURCE}/output
 
+ENV PROTOC_VERSION=3.15.5
+
+# alternative: dnf install -y --nodocs --setopt=install_weak_deps=False golang-google-protobuf protobuf-compiler && dnf -y clean all
```
skitt, line 10: Just a side note: downloading and installing manually takes less than 20s, going through dnf outside Shipyard takes over a minute (because the repo information needs to be downloaded).

Dockerfile.dapper
```diff
@@ -5,6 +5,14 @@ ENV DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD GITHUB_REF CLUSTERS_ARGS DE
...
+RUN curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v3.15.5/protoc-${PROTOC_VERSION}-linux-x86_64.zip && \
+    unzip protoc-${PROTOC_VERSION}-linux-x86_64.zip -d /usr/local && \
+    rm protoc-${PROTOC_VERSION}-linux-x86_64.zip && \
```
skitt: rm -f protoc-${PROTOC_VERSION}-linux-x86_64.zip && \

This avoids having to confirm the removal manually when running interactively.

Makefile
```diff
@@ -50,7 +50,10 @@ deploy: images
 reload-images: build images
 	./scripts/$@ --restart $(restart)
 
-bin/%/submariner-gateway: vendor/modules.txt main.go $(shell find pkg -not \( -path 'pkg/globalnet*' -o -path 'pkg/routeagent*' \))
+pkg/natdiscovery/proto/natdiscovery.pb.go: pkg/natdiscovery/proto/natdiscovery.proto
```
skitt: For some odd reason I can’t make an actual suggestion, but this would be better as

    %.pb.go: %.proto

skitt (dismissed):

mangelajo:
Thank you for the comments @skitt still not applied in this push, I will handle them on monday

mangelajo:
Although I want to add more testing on monday it's ready for review, please see there's two patches on the series. Sorry the second got painstakingly long :/ 

@tpantelis @sridhargaddam  I'd thank review, sorry for the size.

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,231 @@
...
+		localEndpoint:   localEndpoint,
+		serverPort:      port,
+		remoteEndpoints: map[string]*RemoteEndpointNAT{},
+		findSrcIP:       findPreferredSourceIP,
+		requestCounter:  requestCounter,
+	}, nil
+}
+
+func randomRequestCounter() (uint64, error) {
+	n, err := rand.Int(rand.Reader, big.NewInt(1000))
```
mangelajo, line 84: a simple random.Uint64 didn't satisfy the go sec linter :)

Dockerfile.dapper
mangelajo: In my case it was more than 3 minutes, may be a bad network day..

Dockerfile.dapper
mangelajo: Done.

Makefile
```diff
@@ -47,10 +47,15 @@ dockertogoarch = $(patsubst arm/v7,arm,$(1))
 
 deploy: images
 
+lint: pkg/natdiscovery/proto/natdiscovery.pb.go
```
mangelajo: for some reason this is not triggering in lint, and failing the check. :?

mangelajo: ok, changing this to: 
`golangci-lint: pkg/natdiscovery/proto/natdiscovery.pb.go`

it seems to work

mangelajo:
>  Created branch: z_pr1185/mangelajo/nat-discovery-proto
welcome back dude! ;)

mangelajo:
Sorry, moving to draft, I uploaded changes that I didn't want to.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -78,8 +78,13 @@ type EndpointSpec struct {
 	NATEnabled    bool              `json:"nat_enabled"`
 	Backend       string            `json:"backend"`
 	BackendConfig map[string]string `json:"backend_config,omitempty"`
+	PortConfig    map[string]string `json:"natConfig,omitempty"`
```
tpantelis: To match the field name. Unless the field should be `portConfig`? (doesn't seem so)

```suggestion
	NatConfig    map[string]string `json:"natConfig,omitempty"`
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -78,8 +78,13 @@ type EndpointSpec struct {
...
 }
 
+const (
+	NATDiscoveryPortConfig = "nat-discovery"
```
tpantelis: ?

```suggestion
	NATDiscoveryPortConfig = "nat-discovery-port"
```

pkg/natdiscovery/listener.go
```diff
@@ -0,0 +1,73 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+
+	natproto "github.com/submariner-io/submariner/pkg/natdiscovery/proto"
+)
+
+func (nd *natDiscovery) runListener(stopCh <-chan struct{}) error {
+	serverAddress, err := net.ResolveUDPAddr("udp", ":"+strconv.Itoa(natproto.DefaultPort))
+	if err != nil {
+		return errors.Wrap(err, "Error resolving UDP address")
```
tpantelis, line 35: The linter didn't catch this.

```suggestion
		return errors.Wrap(err, "error resolving UDP address")
```

pkg/natdiscovery/listener.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	serverConnection, err := net.ListenUDP("udp", serverAddress)
+	if err != nil {
+		return errors.Wrapf(err, "listening on udp port %d", natproto.DefaultPort)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "error listening on udp port %d", natproto.DefaultPort)
```

pkg/natdiscovery/listener.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Instead of storing the server connection I save the reference to the WriteToUDP
+	// of our server connection instance, in a way that we can use this for unit testing
+	// later too.
+	nd.serverUDPWrite = serverConnection.WriteToUDP
+
+	go wait.Until(func() {
+		buf := make([]byte, 2048)
+		if _, addr, err := serverConnection.ReadFromUDP(buf); err != nil {
+			klog.Errorf("error receiving from udp: %s", err)
```
tpantelis:
```suggestion
			klog.Errorf("Error receiving from udp: %s", err)
```

pkg/natdiscovery/listener.go
```diff
@@ -0,0 +1,73 @@
...
+		} else if err := nd.parseAndHandleMessageFromAddress(buf, addr); err != nil {
+			klog.Errorf("error handling message from address %#v: %s", addr, err)
```
tpantelis:
```suggestion
			klog.Errorf("Error handling message from address %#v: %s", addr, err)
```

pkg/natdiscovery/listener.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}, time.Second, stopCh)
+
+	return nil
+}
+
+func (nd *natDiscovery) parseAndHandleMessageFromAddress(buf []byte, addr *net.UDPAddr) error {
+	msg := natproto.SubmarinerNatDiscoveryMessage{}
+	if err := proto.Unmarshal(buf, &msg); err != nil {
+		klog.Errorf("error unmarshaling message received on UDP port %d: %s", natproto.DefaultPort, err)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "error unmarshaling message received on UDP port %d: %s", natproto.DefaultPort)

File: pkg/natdiscovery/natdiscovery.go [hunk 14]
```diff
@@ -0,0 +1,188 @@
...
+	nd.Lock()
+	defer nd.Unlock()
+
+	for _, endpointNAT := range nd.remoteEndpoints {
+		if endpointNAT.hasTimedOut() {
+			endpointNAT.useLegacyNATSettings()
+		} else if endpointNAT.shouldCheck() {
+			err := nd.sendCheckRequest(endpointNAT)
+			if err != nil {
+				klog.Errorf("error sending check request to %q: %s", endpointNAT.Endpoint.Spec.CableName, err)
```
tpantelis:
```suggestion
				klog.Errorf("Error sending check request to %q: %s", endpointNAT.Endpoint.Spec.CableName, err)
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,138 @@
...
+	waitingForResponse
+	selectedPublicIP
+	selectedPrivateIP
+	recheckTime                    = 2 * time.Second
+	totalTimeout                   = 60 * time.Second
+	publicToPrivateFailoverTimeout = 1 * time.Second
+)
+
+type remoteEndpointNAT struct {
+	Endpoint               *types.SubmarinerEndpoint
```
tpantelis:
```suggestion
	endpoint               *types.SubmarinerEndpoint
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,138 @@
...
+	state                  endpointState
+	lastCheck              time.Time
+	lastTransition         time.Time
+	started                time.Time
+	UseNAT                 bool
+	UseIP                  string
```
tpantelis:
```suggestion
	useNAT                 bool
	useIP                  string
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,138 @@
...
+func newRemoteEndpointNAT(endpoint *types.SubmarinerEndpoint) *remoteEndpointNAT {
+	return &remoteEndpointNAT{
+		Endpoint:       endpoint,
+		state:          testingPrivateAndPublicIPs,
+		started:        time.Now(),
+		lastTransition: time.Now(),
+	}
+}
+
+func (rnat *remoteEndpointNAT) transitionToState(newState endpointState) {
```
tpantelis: To keep receiver names to 1 or 2 chars.

```suggestion
func (r *remoteEndpointNAT) transitionToState(newState endpointState) {
```

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,188 @@
...
+
+	remoteNat, ok := nd.remoteEndpoints[endpointName]
+	if ok {
+		return remoteNat
+	}
+
+	return nil
+}
+
+func (nd *natDiscovery) GetEndpointNAT(endpointName string) (bool, error) {
```
tpantelis:
```suggestion
func (nd *natDiscovery) UseNATForEndpoint(endpointName string) (bool, error) {
```

I haven't seen the usage of this API but it seems the usage would be:

```
if nd.UseNATForEndpoint(...) {
    nd.GetEndpointIP(...)
}
```

If so perhaps we should combine these methods, eg

`func (nd *natDiscovery) GetEndpointNATInfo(endpointName string) (useNat bool, ip string, bool found) `...

We could also add `error` although not necessary in this case.

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,138 @@
...
+		if rnat.sinceLastTransition() > publicToPrivateFailoverTimeout {
+			return errors.Errorf("too late response on private address received after response on public address for endpoint %q",
+				remoteEndpointID)
+		}
+
+		rnat.UseIP = rnat.Endpoint.Spec.PrivateIP
+		rnat.transitionToState(selectedPrivateIP)
+	case selectedPrivateIP:
+	default:
+		return errors.Errorf("received unexpected transition to public IP from endpoint %q", remoteEndpointID)
```
tpantelis:
```suggestion
		return errors.Errorf("received unexpected transition to private IP from endpoint %q", remoteEndpointID)
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,138 @@
...
+		// If a PublicIP was selected, we still allow some time for the privateIP response to arrive, and we always
+		// prefer PrivateIP with no NAT connection, as it will be more likely to work, and more efficient
+		if rnat.sinceLastTransition() > publicToPrivateFailoverTimeout {
+			return errors.Errorf("too late response on private address received after response on public address for endpoint %q",
+				remoteEndpointID)
+		}
+
+		rnat.UseIP = rnat.Endpoint.Spec.PrivateIP
+		rnat.transitionToState(selectedPrivateIP)
+	case selectedPrivateIP:
```
tpantelis: not needed

```suggestion
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,138 @@
...
+	switch rnat.state {
+	case waitingForResponse:
+		rnat.UseIP = rnat.Endpoint.Spec.PrivateIP
+		rnat.transitionToState(selectedPrivateIP)
+
+	case selectedPublicIP:
+		// If a PublicIP was selected, we still allow some time for the privateIP response to arrive, and we always
+		// prefer PrivateIP with no NAT connection, as it will be more likely to work, and more efficient
+		if rnat.sinceLastTransition() > publicToPrivateFailoverTimeout {
+			return errors.Errorf("too late response on private address received after response on public address for endpoint %q",
```
tpantelis:
```suggestion
			return errors.Errorf("response on private address received too late after response on public address for endpoint %q",
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+package natdiscovery
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("natdiscovery listener", func() {
```
tpantelis:
```suggestion
var _ = Describe("remoteEndpointNAT", func() {
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	var rnat *remoteEndpointNAT
+
+	BeforeEach(func() {
+		remoteEndpoint := createTestRemoteEndpoint()
+		rnat = newRemoteEndpointNAT(&remoteEndpoint)
+	})
+
+	When("created", func() {
+		It("should report as shouldCheck", func() {
```
tpantelis:
```suggestion
	When("first created", func() {
		It("shouldCheck should return true", func() {
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+			Expect(rnat.shouldCheck()).To(BeTrue())
+		})
+	})
+
+	When("created, and right after have checked", func() {
+		It("should not report as shouldCheck", func() {
```
tpantelis:
```suggestion
	When("right after a check has been sent", func() {
		It("shouldCheck should return false", func() {
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+			rnat.checkSent()
+			Expect(rnat.shouldCheck()).To(BeFalse())
+		})
+	})
+
+	When("more than totalTimeout seconds passed", func() {
```
tpantelis:
```suggestion
	When("the total timeout has elapsed", func() {
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+			Expect(rnat.UseIP).To(Equal(rnat.Endpoint.Spec.PrivateIP))
+		})
+
+		It("should select the public IP for NATEnabled is true", func() {
+			rnat.Endpoint.Spec.NATEnabled = true
+			rnat.useLegacyNATSettings()
+			Expect(rnat.state).To(Equal(selectedPublicIP))
+			Expect(rnat.UseIP).To(Equal(rnat.Endpoint.Spec.PublicIP))
+		})
+	})
```
tpantelis:
```suggestion
	When("using legacy settings", func() {
		Context("and NAT is not enabled", func() {
			It("should select the private IP", func() {
				...
			})
		})

		Context("and NAT is enabled", func() {
			It("should select the public IP", func() {
				...
			})
		})
	})
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	When("an public IP is selected, but no check was sent", func() {
```
tpantelis:
```suggestion
	When("the public IP is selected but no check was sent", func() {
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+		It("it should fail", func() {
+			err := rnat.transitionToPublicIP(testRemoteEndpointName, false)
+			Expect(err).To(HaveOccurred())
+			Expect(rnat.UseIP).To(Equal(""))
+		})
+	})
+
+	When("an private IP is selected, but no check was sent", func() {
```
tpantelis:
```suggestion
	When("the private IP is selected but no check was sent", func() {
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+			err := rnat.transitionToPublicIP(testRemoteEndpointName, true)
+			Expect(err).NotTo(HaveOccurred())
+			rnat.lastTransition = rnat.lastTransition.Add(-publicToPrivateFailoverTimeout)
+			err = rnat.transitionToPrivateIP(testRemoteEndpointName, false)
+			Expect(err).To(HaveOccurred())
+			Expect(rnat.state).To(Equal(selectedPublicIP))
+			Expect(rnat.UseIP).To(Equal(rnat.Endpoint.Spec.PublicIP))
+			Expect(rnat.UseNAT).To(BeTrue())
+		})
+	})
```
tpantelis:
```suggestion
	When("the private IP is selected after the public IP", func() {
		Context("and the grace period has not elapsed", func() {
			It("should use the private IP", func() {
				...
			})
		})

		Context("and the grace period has elapsed", func() {
			It("should still use the public IP", func() {
				...
			})
		})
	})
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+	When("private IP is selected", func() {
+		It("should use the private IP", func() {
+			rnat.checkSent()
+			err := rnat.transitionToPrivateIP(testRemoteEndpointName, false)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(rnat.state).To(Equal(selectedPrivateIP))
+			Expect(rnat.UseIP).To(Equal(rnat.Endpoint.Spec.PrivateIP))
+			Expect(rnat.UseNAT).To(BeFalse())
+		})
+	})
```
tpantelis: The tests on L134 below are similar and can be combined:

```suggestion
	When("the private IP is selected", func() {
		var useNAT bool

		JustBeforeEach(func() {
			rnat.checkSent()
			err := rnat.transitionToPrivateIP(testRemoteEndpointName, useNAT)
			Expect(err).NotTo(HaveOccurred())
			Expect(rnat.state).To(Equal(selectedPrivateIP))
		})

		It("should use the private IP", func() {
			Expect(rnat.UseIP).To(Equal(rnat.Endpoint.Spec.PrivateIP))
		})

		Context("with NAT enabled", func() {
			BeforeEach(func() {
				useNAT = true
			})

			It("should set useNAT to true", func() {
				Expect(rnat.UseNAT).To(BeTrue())
			})
		})

		Context("with NAT disabled", func() {
			BeforeEach(func() {
				useNAT = false
			})

			It("should set useNAT to false", func() {
				Expect(rnat.UseNAT).To(BeFalse())
			})
		})
	})
```

pkg/natdiscovery/remote_endpoint_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	When("an private IP is selected, but no check was sent", func() {
+		It("it should fail", func() {
+			err := rnat.transitionToPrivateIP(testRemoteEndpointName, false)
+			Expect(err).To(HaveOccurred())
+			Expect(rnat.UseIP).To(Equal(""))
+		})
+	})
+
+	When("public IP is selected", func() {
```
tpantelis: Similar to the private IP tests below, tests on L149 below can be combined.

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+		Receiver: req.Sender,
+		ReceivedSrc: &proto.IPPortPair{
+			Port: uint32(addr.Port),
+			IP:   addr.IP.String(),
+		},
+	}
+
+	if req.Receiver == nil || req.Sender == nil || req.UsingDst == nil || req.UsingSrc == nil {
+		klog.Warningf("received NAT discovery packet %#v from %#v which seems to be malformed ", req, addr)
```
tpantelis:
```suggestion
		klog.Warningf("Received NAT discovery packet %#v from %#v which seems to be malformed ", req, addr)
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+
+		response.Response = proto.ResponseType_MALFORMED
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if req.Receiver.GetClusterId() != nd.localEndpoint.Spec.ClusterID {
+		klog.Warningf("received NAT discovery packet for cluster %q, but we are cluster %q", req.Receiver.GetClusterId(),
```
tpantelis:
```suggestion
		klog.Warningf("Received NAT discovery packet for cluster %q, but we are cluster %q", req.Receiver.GetClusterId(),
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+			nd.localEndpoint.Spec.ClusterID)
+
+		response.Response = proto.ResponseType_UNKNOWN_DST_CLUSTER
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if req.Receiver.GetEndpointId() != nd.localEndpoint.Spec.CableName {
+		klog.Warningf("received NAT discovery packet for endpoint %q, but we are endpoint %q "+
```
tpantelis:
```suggestion
		klog.Warningf("Received NAT discovery packet for endpoint %q, but we are endpoint %q "+
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+
+		response.Response = proto.ResponseType_UNKNOWN_DST_ENDPOINT
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if !nd.isKnownCluster(req.Sender.GetClusterId()) {
+		// This can be ok for some time, because one cluster could discover an endpoint earlier on the broker than the
+		// the other endpoint has discovered him
+		klog.Infof("received a NAT discovery packet from an cluster %q we don't know about yet", req.Receiver.GetClusterId())
```
tpantelis:
```suggestion
		klog.Infof("Received a NAT discovery packet from a cluster %q we don't know about yet", req.Receiver.GetClusterId())
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+
+		response.Response = proto.ResponseType_UNKNOWN_SRC_CLUSTER
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if !nd.isKnownEndpoint(req.Sender.GetEndpointId()) {
+		// This can be ok for some time, because one cluster could discover an endpoint earlier on the broker than the
+		// the other endpoint has discovered him
+		klog.Infof("received a NAT discovery packet from an endpoint %q we don't know about yet", req.Receiver.GetEndpointId())
```
tpantelis:
```suggestion
		klog.V(log.DEBUG).Infof("Received a NAT discovery packet from an endpoint %q we don't know about yet", req.Receiver.GetEndpointId())
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+		response.Response = proto.ResponseType_UNKNOWN_SRC_ENDPOINT
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if req.UsingSrc.GetIP() != "" && req.UsingSrc.GetIP() != addr.IP.String() {
+		klog.V(log.DEBUG).Infof("received NAT packet from endpoint %q, cluster %q, where NAT traversal has been detected, "+
+			"source IP changed",
+			req.Sender.GetEndpointId(), req.Sender.GetClusterId())
+		klog.V(log.DEBUG).Infof("original src IP was %q, received src IP is %q", req.UsingSrc.IP, addr.IP.String())
```
tpantelis:
```suggestion
		klog.V(log.DEBUG).Infof("Received NAT packet from endpoint %q, cluster %q, where NAT traversal has been detected, "+
			"source IP changed",
			req.Sender.GetEndpointId(), req.Sender.GetClusterId())
		klog.V(log.DEBUG).Infof("Original src IP was %q, received src IP is %q", req.UsingSrc.IP, addr.IP.String())
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+		response.Response = proto.ResponseType_SRC_MODIFIED
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if int(req.UsingSrc.Port) != addr.Port {
+		klog.V(log.DEBUG).Infof("received NAT packet from endpoint %q, cluster %q, where NAT traversal on the source has been detected, "+
+			"src port changed",
+			req.Sender.GetEndpointId(), req.Sender.GetClusterId())
+		klog.V(log.DEBUG).Infof("original src IP was %q, received src IP is %q", req.UsingSrc.IP, addr.IP.String())
```
tpantelis:
```suggestion
		klog.V(log.DEBUG).Infof("Received NAT packet from endpoint %q, cluster %q, where NAT traversal on the source has been detected, "+
			"src port changed",
			req.Sender.GetEndpointId(), req.Sender.GetClusterId())
		klog.V(log.DEBUG).Infof("Original src IP was %q, received src IP is %q", req.UsingSrc.IP, addr.IP.String())
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+		klog.Warningf("received NAT discovery packet for endpoint %q, but we are endpoint %q "+
+			"if the port for NAT discovery has been mapped somewhere an error may exist", req.Receiver.GetEndpointId(),
+			nd.localEndpoint.Spec.CableName)
+
+		response.Response = proto.ResponseType_UNKNOWN_DST_ENDPOINT
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if !nd.isKnownCluster(req.Sender.GetClusterId()) {
```
tpantelis: Is this necessary since we verify the `EndpointId` below? If so then `isKnownCluster` and `isKnownEndpoint` should be consistent, ie `isKnownCluster` should also lookup the in `remoteEndpoints` by `EndpointId` instead of searching the map. In fact you could just combine the two:

```
ep := nd.getRemoteEndpointNAT(endpointID)
return ep != nil && ep.Endpoint.Spec.ClusterID == clusterID
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+	for ep := range nd.remoteEndpoints {
+		if nd.remoteEndpoints[ep].Endpoint.Spec.ClusterID == clusterID {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (nd *natDiscovery) isKnownEndpoint(endpointID string) bool {
```
tpantelis: This is a duplicate of `getRemoteEndpointNAT`.

tpantelis (changes requested):
I did a partial review - will go thru the rest tomorrow.

pkg/natdiscovery/proto/natdiscovery.proto
```diff
@@ -0,0 +1,73 @@
...
+
+message IPPortPair {
+  string IP = 1;
+  uint32 port = 2;
+}
+
+message EndpointDetails {
+  // should we hash this for privacy? a hash can be checked against a known list, but can't be decoded
+  string cluster_id = 1;
+  string endpoint_id = 2;
```
tpantelis, line 71:
```suggestion
  string clusterID = 1;
  string endpointID = 2;
```

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (nd *natDiscovery) checkEndpointList() {
+	nd.Lock()
+	defer nd.Unlock()
+
+	for _, endpointNAT := range nd.remoteEndpoints {
+		if endpointNAT.hasTimedOut() {
+			endpointNAT.useLegacyNATSettings()
+		} else if endpointNAT.shouldCheck() {
```
tpantelis: Once it transitions to public IP or private IP via a valid response, it should stay in that state. However `hasTimedOut` will eventually return true once the `totalTimeout` has elapsed and thus we'll revert to `useLegacyNATSettings`. We should restructure this so that doesn't happen. I think the simplest is:

```
if endpointNAT.shouldCheck() {
    if endpointNAT.hasTimedOut() {
        endpointNAT.useLegacyNATSettings()
        continue
    }

    err := nd.sendCheckRequest(endpointNAT)
    ...
}
```

Another option is to modify `hasTimedOut` to check the state but `shouldCheck` already does that.  `hasTimedOut` may be slightly delayed by the `recheckTime` but I think that's OK.

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (nd *natDiscovery) getRemoteEndpointNAT(endpointName string) *remoteEndpointNAT {
+	nd.Lock()
+	defer nd.Unlock()
+
+	remoteNat, ok := nd.remoteEndpoints[endpointName]
+	if ok {
+		return remoteNat
+	}
+
```
tpantelis: So access to the map is protected by the lock but the `remoteEndpointNAT` instance is unprotected once returned which could be problematic.  So everywhere we access the `remoteEndpoints` map and the mutable fields in `remoteEndpointNAT` needs to be protected.

pkg/natdiscovery/natdiscovery.go
```diff
@@ -1 +1,188 @@
...
+		remoteEndpoints: map[string]*remoteEndpointNAT{},
+		findSrcIP:       findPreferredSourceIP,
+		requestCounter:  requestCounter,
+	}, nil
+}
+
+func randomRequestCounter() (uint64, error) {
+	n, err := rand.Int(rand.Reader, big.NewInt(1000))
+	if err != nil {
+		return 0, errors.Wrapf(err, "generatimg random request counter")
```
sridhargaddam: nit typo: "error generating random request counter"

pkg/natdiscovery/natdiscovery.go
```diff
@@ -1 +1,188 @@
...
+	}
+
+	return n.Uint64(), nil
+}
+
+var errorNoNatDiscoveryPort = errors.New("no nat discovery port in endpoint")
```
sridhargaddam: nit: "nat discovery port missing in endpoint"

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	rnat.lastCheck = time.Now()
+}
+
+func (rnat *remoteEndpointNAT) transitionToPublicIP(remoteEndpointID string, useNAT bool) error {
+	if rnat.state == waitingForResponse {
+		rnat.UseIP = rnat.Endpoint.Spec.PublicIP
+		rnat.transitionToState(selectedPublicIP)
+		rnat.UseNAT = useNAT
```
sridhargaddam: Just for info: Since we are using publicIP here, will useNAT be false always?

pkg/natdiscovery/remote_endpoint.go
sridhargaddam: I think @mangelajo wants default to be executed only when rnat.state is NOT `selectedPrivateIP` so it might be needed.

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+		// the other endpoint has discovered him
+		klog.Infof("received a NAT discovery packet from an endpoint %q we don't know about yet", req.Receiver.GetEndpointId())
+
+		response.Response = proto.ResponseType_UNKNOWN_SRC_ENDPOINT
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if req.UsingSrc.GetIP() != "" && req.UsingSrc.GetIP() != addr.IP.String() {
+		klog.V(log.DEBUG).Infof("received NAT packet from endpoint %q, cluster %q, where NAT traversal has been detected, "+
```
sridhargaddam: Instead of saying "NAT traversal", I think we can just say "NAT"

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+			req.Sender.GetEndpointId(), req.Sender.GetClusterId())
+		klog.V(log.DEBUG).Infof("original src IP was %q, received src IP is %q", req.UsingSrc.IP, addr.IP.String())
+
+		response.Response = proto.ResponseType_SRC_MODIFIED
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if int(req.UsingSrc.Port) != addr.Port {
+		klog.V(log.DEBUG).Infof("received NAT packet from endpoint %q, cluster %q, where NAT traversal on the source has been detected, "+
```
sridhargaddam: same

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+	message := proto.SubmarinerNatDiscoveryMessage{Message: &msgResponse}
+	buf, err := proto2.Marshal(&message)
+	if err != nil {
+		return errors.Wrapf(err, "error marshaling response %#v", response)
+	}
+
+	if length, err := nd.serverUDPWrite(buf, addr); err != nil {
+		return errors.Wrapf(err, "error sending response packet %#v", response)
+	} else if length != len(buf) {
+		return errors.Errorf("the sent UDP packet was smaller than requested, sent=%d, expected=%d", length, len(buf))
```
sridhargaddam, line 129: I'm yet to take a close look at the code. A quick question in the meantime - are we retrying when only partial data is sent?

pkg/natdiscovery/request_send.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if remoteNAT.Endpoint.Spec.PublicIP != "" {
+		reqID, errPublic = nd.sendCheckRequestToTargetIP(remoteNAT, remoteNAT.Endpoint.Spec.PublicIP)
+		if errPublic == nil {
+			remoteNAT.lastPublicIPRequestID = reqID
+		}
+	}
+
+	if errPrivate != nil && errPublic != nil {
+		return errors.Errorf("error happened trying to discover both public & private IPs of endpoint %q, [%s, %s]",
```
sridhargaddam: "error while trying to discover both public ....."

pkg/natdiscovery/request_send.go
```diff
@@ -0,0 +1,152 @@
...
+
+	remoteNAT.checkSent()
+
+	return request.RequestNumber, nil
+}
+
+func findPreferredSourceIP(destinationIP string) (string, error) {
+	var ip net.IP
+	if ip = net.ParseIP(destinationIP); ip == nil {
+		return "", errors.Errorf("Error parsing destination IP %q while trying to figure out preferred source IP", destinationIP)
```
sridhargaddam: "Error ...." -> "error ...."

pkg/natdiscovery/request_send.go
```diff
@@ -0,0 +1,152 @@
...
+
+	addr := net.UDPAddr{
+		IP:   net.ParseIP(targetIP),
+		Port: targetPort,
+	}
+
+	if length, err := nd.serverUDPWrite(buf, &addr); err != nil {
+		return request.RequestNumber, errors.Wrapf(err, "error sending request packet %#v", request)
+	} else if length != len(buf) {
+		return request.RequestNumber, errors.Errorf("the sent UDP packet was smaller than requested, sent=%d, expected=%d", length,
```
sridhargaddam, line 126: As above: are we retrying when only partial data is sent?

sridhargaddam (commented):
Thanks for the PR @mangelajo 

Some minor comments based on quick pass. I'll take a close look in parallel.

pkg/apis/submariner.io/v1/types.go
mangelajo: It could also be, Since it was in a "PortConfig" dictionary it seemed like repeating.

I want to keep the "PortConfig" as portconfig ( I need to fix the json name as you pointed out it's wrong)

The idea is that PortConfig (or simply "Ports?") would hold the port numbers also for IPSEC or wireguard, etc which now it's fixed per clusterset and doesn't really work. It needs to be flexible per-endpoint for all the on-prem nat stuff to work ok.

pkg/natdiscovery/listener.go
mangelajo: I need to check if that would be ok, I wanted it to be a soft error, like "we received something strange from the net" 

Probably it's ok to return error. (checking)

pkg/natdiscovery/natdiscovery.go
mangelajo: Opps, that's a good one.... thanks tom!


From publicIP to privateIP we allow a grace time where we can still transition to privateIP (since we test both private & public in parallel) you could have reachability to both, but we will always prefer the privateIP.
After that grace time we stop accepting responses from privateIP, because it could be a forged packet to try to cheat the gateway to switch IPs (and break connectivity)

pkg/natdiscovery/natdiscovery.go
mangelajo: You're right, I'm removing this API. In the end I will be sending the ready endpoints via a channel.

Since it's a little bit asynchronous:

1) InstallCable will happen because we know of a new endpoint, at that moment we pass the endpoint to the natDiscovery controller,
2) Once the destination IP is discovered (or changed, public->private), it's passed to cableEngine again
3) Cable engine does the final installation.

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,188 @@
...
+func (nd *natDiscovery) GetEndpointNAT(endpointName string) (bool, error) {
+	endpointNAT := nd.getRemoteEndpointNAT(endpointName)
+	if endpointNAT == nil {
+		return false, UnknownEndpoint
+	}
+
+	return endpointNAT.UseNAT, nil
+}
+
+func (nd *natDiscovery) GetEndpointIP(endpointName string) (string, error) {
```
mangelajo: This one can also be killed.

pkg/natdiscovery/remote_endpoint.go
mangelajo: I will need to be public. It will be accessed outside of this module in the end. But yeah being private you're right, it should have been private "endpoint"

pkg/natdiscovery/remote_endpoint.go
mangelajo: same.

May be I will make them private and add a "GetEndpoint" GetUseNAT, GetIP" or something like that, with locking.

pkg/natdiscovery/remote_endpoint.go
mangelajo: useNAT will be dynamic, if really the source IP or source Port changed, Imagine of the case where you have a public IP on the interface, and no nat happens.

We haven't seen it, but we would make it compatible this way.

pkg/natdiscovery/remote_endpoint.go
mangelajo: correct, we need it for what @sridhargaddam said.

It looks weird, because in C that meant that selectedPrivateIP: == default: but in go it's just no action for selectedPrivateIP.

And it's a case which could happen (for example if both public & private respond) I can add at least a warning message to make it clearer.

pkg/natdiscovery/remote_endpoint_test.go
mangelajo: oops

mangelajo:
@sridhargaddam @tpantelis thanks for all the great feedback, I have work to do...

pkg/natdiscovery/proto/natdiscovery.proto
mangelajo: It's the proto buf standard of coding.

This is translated later into go I think it follows that format more or less, but not sure. I'll check

pkg/natdiscovery/request_handle.go
mangelajo: It will only be retried when we receive another request packet. The sender will retry.

pkg/natdiscovery/listener.go
mangelajo: Done

pkg/natdiscovery/listener.go
mangelajo: Done

pkg/natdiscovery/listener.go
mangelajo: oh, on the top level function we were just printing . Done

pkg/natdiscovery/natdiscovery.go
mangelajo: Done thanks

pkg/natdiscovery/natdiscovery.go
mangelajo: Done, Thanks

pkg/natdiscovery/natdiscovery.go
mangelajo: Done

pkg/natdiscovery/natdiscovery.go
mangelajo: I removed both, and moving into this patch the "channel" implementation.

pkg/natdiscovery/natdiscovery.go
mangelajo: Done.

pkg/natdiscovery/proto/natdiscovery.proto
mangelajo: https://developers.google.com/protocol-buffers/docs/style  see the "Message and field names", I think it's better if we adhere to that standard even if we will get "Id" instead of "ID" in go ':)

pkg/natdiscovery/remote_endpoint.go
mangelajo: Done.

pkg/natdiscovery/remote_endpoint.go
mangelajo: Done.

pkg/natdiscovery/remote_endpoint.go
mangelajo: I removed it, it's better to print an error in that case.

pkg/natdiscovery/remote_endpoint.go
mangelajo: Done.

pkg/natdiscovery/remote_endpoint_test.go
mangelajo: Done.

pkg/natdiscovery/remote_endpoint_test.go
mangelajo: Done.

pkg/natdiscovery/remote_endpoint_test.go
mangelajo: Done.

pkg/natdiscovery/remote_endpoint.go
tpantelis: yeah they'd need to be protected b y a lock.

pkg/natdiscovery/natdiscovery.go
tpantelis: ok - makes sense

pkg/natdiscovery/remote_endpoint.go
tpantelis: ah right - I was thinking fall-through

pkg/natdiscovery/proto/natdiscovery.proto
tpantelis: ok with or without '_', `id` -> `ID`.

pkg/natdiscovery/request_handle.go
mangelajo: It could be a known cluster, but an unknown endpoint.

For example during a gateway transition, it could happen that the new endpoint (gw) had registered to the broker, but the other side didn't get the update yet.

It would receive the packet. And return a different error.

But as you say,  it's not strictly necessary, we don't gain much but extra code. I'll remove it , also from the protocol.

pkg/natdiscovery/request_handle.go
mangelajo: Done.

pkg/natdiscovery/request_handle.go
mangelajo: Yeah makes sense.

pkg/natdiscovery/request_handle.go
mangelajo: Done

pkg/natdiscovery/request_handle.go
mangelajo: Done.

pkg/natdiscovery/request_handle.go
mangelajo: Done.

pkg/natdiscovery/request_handle.go
mangelajo: Done.

pkg/natdiscovery/proto/consts.go
```diff
@@ -0,0 +1,20 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proto
+
+const DefaultPort = 4490
```
sridhargaddam, line 19: I was just checking if port 4490 is used by some application and found some notes that say "Known UNAUTHORIZED USE: ports 4490 and 4491" Any idea what it means?

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if req.Receiver == nil || req.Sender == nil || req.UsingDst == nil || req.UsingSrc == nil {
+		klog.Warningf("received NAT discovery packet %#v from %#v which seems to be malformed ", req, addr)
+
+		response.Response = proto.ResponseType_MALFORMED
+
+		return nd.sendResponseToAddress(&response, addr)
+	}
+
+	if req.Receiver.GetClusterId() != nd.localEndpoint.Spec.ClusterID {
```
sridhargaddam, line 52: These are safety checks right? I mean, ideally we will not see these messages unless there is some bug in the code that framed the message and sent it.

pkg/natdiscovery/request_send.go
mangelajo: I forgot about this one, will handle on next iteration

pkg/natdiscovery/listener.go
```diff
@@ -0,0 +1,73 @@
...
+	natproto "github.com/submariner-io/submariner/pkg/natdiscovery/proto"
+)
+
+func (nd *natDiscovery) runListener(stopCh <-chan struct{}) error {
+	serverAddress, err := net.ResolveUDPAddr("udp", ":"+strconv.Itoa(natproto.DefaultPort))
+	if err != nil {
+		return errors.Wrap(err, "Error resolving UDP address")
+	}
+
+	serverConnection, err := net.ListenUDP("udp", serverAddress)
```
sridhargaddam, line 38: I think we will be listening on both IPv4 and IPv6 addresses. Should we listen just on IPv4?

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,133 @@
...
+func (nd *natDiscovery) sendResponseToAddress(response *proto.SubmarinerNatDiscoveryResponse, addr *net.UDPAddr) error {
+	msgResponse := proto.SubmarinerNatDiscoveryMessage_Response{Response: response}
+	message := proto.SubmarinerNatDiscoveryMessage{Message: &msgResponse}
+	buf, err := proto2.Marshal(&message)
+	if err != nil {
+		return errors.Wrapf(err, "Error marshaling response %#v", response)
+	}
+
+	if length, err := nd.serverUDPWrite(buf, addr); err != nil {
+		return errors.Wrapf(err, "Error sending response packet %#v", response)
```
sridhargaddam:
```suggestion
		return errors.Wrapf(err, "error sending response packet %#v", response)
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return ok && ep.Endpoint().Spec.ClusterID == clusterID
+}
+
+func (nd *natDiscovery) sendResponseToAddress(response *proto.SubmarinerNatDiscoveryResponse, addr *net.UDPAddr) error {
+	msgResponse := proto.SubmarinerNatDiscoveryMessage_Response{Response: response}
+	message := proto.SubmarinerNatDiscoveryMessage{Message: &msgResponse}
+	buf, err := proto2.Marshal(&message)
+	if err != nil {
+		return errors.Wrapf(err, "Error marshaling response %#v", response)
```
sridhargaddam:
```suggestion
		return errors.Wrapf(err, "error marshaling response %#v", response)
```

pkg/natdiscovery/request_handle.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if length, err := nd.serverUDPWrite(buf, addr); err != nil {
+		return errors.Wrapf(err, "Error sending response packet %#v", response)
+	} else if length != len(buf) {
+		return errors.Errorf("The sent UDP packet was smaller than requested, sent=%d, expected=%d", length, len(buf))
```
sridhargaddam:
```suggestion
		return errors.Errorf("the sent UDP packet was smaller than requested, sent=%d, expected=%d", length, len(buf))
```

pkg/natdiscovery/response_handle.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/pkg/errors"
+	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/natdiscovery/proto"
+)
+
+func (nd *natDiscovery) handleResponseFromAddress(req *proto.SubmarinerNatDiscoveryResponse, addr *net.UDPAddr) error {
+	if req.GetSender() == nil || req.GetReceiver() == nil || req.GetReceivedSrc() == nil {
+		return errors.Errorf("received malformed response")
```
sridhargaddam: Shall we include the req in the error message to aid in debugging?

pkg/natdiscovery/response_handle.go
```diff
@@ -0,0 +1,83 @@
...
+		if name, ok = proto.ResponseType_name[int32(req.Response)]; !ok {
+			name = fmt.Sprintf("%d", req.Response)
+		}
+
+		return errors.Errorf("remote endpoint %q responded with %q : %#v", req.Sender.EndpointId, name, req)
+	}
+
+	remoteNat, ok := nd.remoteEndpoints[req.GetSender().EndpointId]
+	if !ok {
+		return errors.Errorf("received response from unknown endpoint")
```
sridhargaddam: Shall we include endpoint name in the message (req.GetSender().EndpointId)?

sridhargaddam (changes requested):

pkg/apis/submariner.io/v1/types.go
```diff
@@ -78,8 +78,13 @@ type EndpointSpec struct {
 	NATEnabled    bool              `json:"nat_enabled"`
 	Backend       string            `json:"backend"`
 	BackendConfig map[string]string `json:"backend_config,omitempty"`
+	Ports         map[string]string `json:"ports,omitempty"`
```
tpantelis: As discussed on slack:

```suggestion
	NATDiscoveryPort         *int32 `json:"natDiscoveryPort,omitempty"`
```

pkg/natdiscovery/proto/consts.go
skitt: It probably means that programs have been seen using ports 4490 and 4491, but they aren’t registered with the IANA.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -78,8 +78,13 @@ type EndpointSpec struct {
...
 }
 
+const (
+	NATDiscoveryPort = "nat-discovery"
+)
```
```suggestion
```

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,176 @@
...
+	nd.Lock()
+	defer nd.Unlock()
+
+	remoteNat, ok := nd.remoteEndpoints[endpointName]
+	if ok {
+		return remoteNat
+	}
+
+	return nil
+}
```
tpantelis: unused now

```suggestion
```

pkg/natdiscovery/natdiscovery_suite_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+func init() {
+	klog.InitFlags(nil)
+}
+
+func TestNATDiscoveryProtocol(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Submariner NATdiscovery protocol test suite")
```
tpantelis:
```suggestion
func TestNATDiscovery(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "NAT Discovery Suite")
```

pkg/natdiscovery/request_send.go
```diff
@@ -0,0 +1,152 @@
...
+
+	targetPort, err := extractNATDiscoveryPort(remoteNAT.endpoint)
+
+	if err != nil {
+		return 0, err
+	}
+
+	sourceIP, err := nd.findSrcIP(targetIP)
+	if err != nil {
+		klog.Warningf("unable to determine source IP while preparing NAT discovery request to endpoint %q: %s",
```
sridhargaddam, line 78: Ideally, this case will not hit since we will have a default route on the host. But in the rare case if we are unable to get sourceIP, should we return here?

pkg/natdiscovery/proto/natdiscovery.proto
```diff
@@ -0,0 +1,72 @@
...
```
tpantelis, line 71:
```suggestion
  string cluster_ID = 1;
  string endpoint_ID = 2;
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,175 @@
...
+	waitingForResponse
+	selectedPublicIP
+	selectedPrivateIP
+	recheckTime                    = 2 * time.Second
+	totalTimeout                   = 60 * time.Second
+	publicToPrivateFailoverTimeout = 1 * time.Second
+)
+
+type RemoteEndpointNAT struct {
+	sync.Mutex
```
tpantelis: I think this lock is too granular and complicates the code and may leave it vulnerable to deadlocks. We can leave the locking on the `remoteEndpoints` map since we need to protect both anyway.

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,176 @@
...
+		nd.checkEndpointList()
+	}, time.Second, stopCh)
+
+	return nil
+}
+
+func (nd *natDiscovery) AddEndpoint(endpoint *types.SubmarinerEndpoint) {
+	remoteNAT := newRemoteEndpointNAT(endpoint)
+
+	// support a remote cluster endpoint which still hasn't implemented this protocol
```
sridhargaddam, line 129: If AddEndpoint is called periodically for the same Endpoint, we will be validating NAT-discovery repeatedly right?

pkg/apis/submariner.io/v1/types.go
mangelajo: @sridhargaddam tom was suggesting that we use NATDiscoveryPort as a field here *int instead.

And leave the backend ports in "BackendConfig"

does that sound ok for you?

pkg/natdiscovery/listener.go
mangelajo: We will need to support ipv6 or mixed environments at some point, so why restrict it or make it specific? :)

Even it if will not help with nat for ipv6, it can help for "reachability" check

pkg/natdiscovery/request_send.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+
+	if errPrivate != nil && errPublic != nil {
+		return errors.Errorf("error happened trying to discover both public & private IPs of endpoint %q, [%s, %s]",
+			remoteNAT.endpoint.Spec.CableName, errPublic, errPrivate)
+	}
+
+	if errPrivate != nil {
+		return errors.Wrapf(errPrivate, "error happened trying to NAT-discover private IP of endpoint %q",
```
tpantelis:
```suggestion
		return errors.Wrapf(errPrivate, "error while trying to NAT-discover private IP of endpoint %q",
```

pkg/natdiscovery/request_send.go
```diff
@@ -0,0 +1,152 @@
...
+			remoteNAT.endpoint.Spec.CableName)
+	}
+
+	if errPublic != nil {
+		return errors.Wrapf(errPublic, "error happened trying to NAT-discover public IP of endpoint %q",
```
tpantelis:
```suggestion
		return errors.Wrapf(errPublic, "error while trying to NAT-discover public IP of endpoint %q",
```

pkg/natdiscovery/request_send.go
```diff
@@ -0,0 +1,152 @@
...
+			remoteNAT.endpoint.Spec.CableName)
+	}
+
+	return nil
+}
+
+func (nd *natDiscovery) sendCheckRequestToTargetIP(remoteNAT *RemoteEndpointNAT, targetIP string) (uint64, error) {
+	nd.requestCounter++
```
tpantelis: We can move this after L82.

pkg/natdiscovery/request_handle_test.go
```diff
@@ -0,0 +1,194 @@
...
+	When("receiving a request from an known cluster", func() {
+		It("it should respond OK", func() {
+			remoteListener.AddEndpoint(&localEndpoint)
+			localListener.AddEndpoint(&remoteEndpoint)
+			response := requestResponseFromRemoteToLocal(&remoteUDPAddr)
+			Expect(response[0].Response).To(Equal(natproto.ResponseType_OK))
+		})
+	})
+
+	When("receiving a request from an known cluster, but port is modified", func() {
```
sridhargaddam:
```suggestion
	When("receiving a request from an known cluster, but IP is modified", func() {
```

pkg/natdiscovery/natdiscovery.go
mangelajo: Oh we need to check that case, good point.

pkg/natdiscovery/proto/consts.go
mangelajo: It's not a fixed port so we can always configure them dynamically. Should not be a problem really.

pkg/natdiscovery/natdiscovery.go
mangelajo: I forgot that one, will do on next pass

pkg/natdiscovery/response_handle.go
```diff
@@ -0,0 +1,83 @@
...
+		var name string
+
+		if name, ok = proto.ResponseType_name[int32(req.Response)]; !ok {
+			name = fmt.Sprintf("%d", req.Response)
+		}
+
+		return errors.Errorf("remote endpoint %q responded with %q : %#v", req.Sender.EndpointId, name, req)
+	}
+
+	remoteNat, ok := nd.remoteEndpoints[req.GetSender().EndpointId]
```
tpantelis, line 46:
```suggestion
       nd.Lock()
       defer nd.Unlock()
	
       remoteNat, ok := nd.remoteEndpoints[req.GetSender().EndpointId]
```

pkg/natdiscovery/request_handle.go
mangelajo: It could be if for example there is port mapping being done in a load balancer, or router, and they made a mistake between clusters when mapping the ports.

tpantelis (changes requested):

pkg/natdiscovery/request_handle.go
mangelajo: ok so @tpantelis explained to me, that in wrap we should use low-case, and on the base Errors, upcase.

pkg/apis/submariner.io/v1/types.go
sridhargaddam: If I understood it correctly, any custom IPsec/tunnel ports will be configured in BackendConfig. If so, yes, its fine for me.

mangelajo: :+1 thanks for the ack

pkg/natdiscovery/listener.go
sridhargaddam: Currently while creating the localEndpoints, we don't use IPv6 addresses (even if the underlying host has dual-stack configuration). So, listening on IPv6 UDP is not really required since we will not be connecting to the IPv6 addresses in the current implementation.

pkg/natdiscovery/request_handle.go
sridhargaddam: Ack. I'm fine either way. I just remember seeing shipyard/e2e/framework/util.go using lower case for Errors.

pkg/natdiscovery/natdiscovery.go
```diff
@@ -0,0 +1,176 @@
...
+
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/types"
+)
+
+type Interface interface {
+	Run(stopCh <-chan struct{}, natResultCh chan RemoteEndpointNAT) error
+	AddEndpoint(endpoint *types.SubmarinerEndpoint)
+	RemoveEndpoint(endpointName string)
+	SetReadyChannel(readyChannel chan *RemoteEndpointNAT)
```
tpantelis: As discussed on call, let's keep `RemoteEndpointNAT` private and expose a smaller struct, eg

```
type NATEndpointInfo struct {
    Endpoint  subv1.EndpointSpec
    UseNAT   bool
    UseIP       string
}
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,175 @@
...
+	rn.Lock()
+	defer rn.Unlock()
+	return rn.useIP
+}
+
+func (rn *RemoteEndpointNAT) UseNAT() bool {
+	rn.Lock()
+	defer rn.Unlock()
+	return rn.useNAT
+}
```
tpantelis: not needed

```suggestion
```

tpantelis (changes requested):

pkg/natdiscovery/natdiscovery.go
mangelajo: done. Now we will only replace it if the endpoint has changed (public ip changed or nat settings changed, for example)

pkg/natdiscovery/request_handle.go
tpantelis: > Ack. I'm fine either way. I just remember seeing shipyard/e2e/framework/util.go using lower case for Errors.

yeah it's recommended [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings). We have a linter that will at least flag it with `fmt.Errorf` but it doesn't seem to recognize "_github.com/pkg/errors_".  When logging a message via klog then you capitalize.

pkg/natdiscovery/natdiscovery.go
mangelajo: byebye dead code :)

pkg/natdiscovery/natdiscovery_suite_test.go
mangelajo: Done

pkg/natdiscovery/proto/natdiscovery.proto
mangelajo: Changing this results in:
```
type EndpointDetails struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// should we hash this for privacy? a hash can be checked against a known list, but can't be decoded
	Cluster_ID  string `protobuf:"bytes,1,opt,name=cluster_ID,json=clusterID,proto3" json:"cluster_ID,omitempty"`
	Endpoint_ID string `protobuf:"bytes,2,opt,name=endpoint_ID,json=endpointID,proto3" json:"endpoint_ID,omitempty"`
}
```

I think it's better to stick to cluster_id which generates ClusterId:
```
type EndpointDetails struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// should we hash this for privacy? a hash can be checked against a known list, but can't be decoded
	ClusterId  string `protobuf:"bytes,1,opt,name=cluster_id,json=clusterId,proto3" json:"cluster_id,omitempty"`
	EndpointId string `protobuf:"bytes,2,opt,name=endpoint_id,json=endpointId,proto3" json:"endpoint_id,omitempty"`
}
```

pkg/natdiscovery/remote_endpoint.go
mangelajo: ack.

pkg/natdiscovery/request_handle_test.go
mangelajo: Done

pkg/natdiscovery/request_send.go
mangelajo: Done.

pkg/natdiscovery/request_send.go
mangelajo: It happens in my laptop ':D

Apparently because I have the same route & IP on my wifi & ethernet :-)

It's a weird case I guess.

But In that case we still want to send the packet at least and try?

pkg/natdiscovery/request_send.go
mangelajo: Done.

pkg/natdiscovery/request_send.go
mangelajo: It will be retried by the controller on the upper level after recheckTime (2 seconds)

pkg/natdiscovery/request_send.go
mangelajo: Done.

pkg/natdiscovery/request_handle.go
mangelajo: done

pkg/natdiscovery/request_handle.go
mangelajo: done

pkg/natdiscovery/request_handle.go
mangelajo: done thanks

pkg/natdiscovery/remote_endpoint.go
mangelajo: lock removed.

pkg/natdiscovery/natdiscovery.go
mangelajo: done.

pkg/natdiscovery/response_handle.go
mangelajo: :+1:

mangelajo: Done.

pkg/natdiscovery/response_handle.go
mangelajo: oops, right.

Done, but being explicit about the unlock.

pkg/natdiscovery/response_handle.go
mangelajo: Ack. Done.

pkg/natdiscovery/request_send.go
mangelajo: done.

pkg/natdiscovery/request_send.go
mangelajo: done.

mangelajo (commented):
updating the patch in a sec. linting..

mangelajo:
I have rebased on devel and squashed the Handle review comment patches.

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,166 @@
...
+	lastPrivateIPRequestID uint64
+}
+
+type NATEndpointInfo struct {
+	Endpoint subv1.EndpointSpec
+	UseNAT   bool
+	UseIP    string
+}
+
+func (rn *remoteEndpointNAT) GetNATEndpointInfo() *NATEndpointInfo {
```
tpantelis: nit but this doesn't need to be exported:

```suggestion
func (rn *remoteEndpointNAT) toNATEndpointInfo() *NATEndpointInfo {
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -0,0 +1,166 @@
...
+	return rn.endpoint
+}
+
+func (rn *remoteEndpointNAT) UseIP() string {
+	return rn.useIP
+}
+
+func (rn *remoteEndpointNAT) UseNAT() bool {
+	return rn.useNAT
+}
```
tpantelis: We don't actually need these methods since we don't export it and callers can access the fields directly anyway.
 
```suggestion
```

pkg/natdiscovery/response_handle.go
```diff
@@ -0,0 +1,88 @@
...
+		if name, ok = proto.ResponseType_name[int32(req.Response)]; !ok {
+			name = fmt.Sprintf("%d", req.Response)
+		}
+
+		return errors.Errorf("remote endpoint %q responded with %q : %#v", req.Sender.EndpointId, name, req)
+	}
+
+	nd.Lock()
+	remoteNat, ok := nd.remoteEndpoints[req.GetSender().EndpointId]
+	nd.Unlock()
```
tpantelis: We can't unlock here b/c we also need to protect access to the `remoteNat` instance below, especially since it mutates state. We might as well defer.

tpantelis (changes requested):

pkg/natdiscovery/remote_endpoint.go
mangelajo: oh right :)

pkg/natdiscovery/remote_endpoint.go
mangelajo: correct.

pkg/natdiscovery/response_handle.go
mangelajo: oh right.

pkg/natdiscovery/remote_endpoint.go
mangelajo: done.

pkg/natdiscovery/remote_endpoint.go
mangelajo: done

tpantelis (approved):

pkg/natdiscovery/request_send.go
sridhargaddam: When there is an error, will it also return a valid sourceIP? 

If not, we will endup with an empty sourceIP in request.UsingSrc right?

sridhargaddam (commented):
One small clarification, otherwise it looks good.

pkg/natdiscovery/request_send.go
mangelajo: Yes that is supported 

> One small clarification, otherwise it looks good.

Yes that's supported (thinking of this case), but we won't be able to detect modified src ip, only modified src port.

https://github.com/submariner-io/submariner/pull/1185/files#diff-2ce27aa4f46793624b7693021b9cdc8ccd90aa88ebf7a2428c4c902e4e844a1fR82

sridhargaddam (approved):
Thanks for addressing the review comments @mangelajo

PR #1186, opened 2021-03-08
Use the K8s iptables wrapper everywhere

skitt:
This will allow us to drop the host mount requirement on all pods.

This requires using Fedora; UBI images don’t carry both iptables
variants. We’ll need to resolve this at some point, but for now
dropping the host mount seems more useful.


dfarrell07 (approved):

mangelajo (approved):

dfarrell07 (approved):

mangelajo (changes requested):
Before we go ahead, Let's talk to @SteveMattar how will we make this work in downstream. In downstream won't we depend on having the ability to install also nft/legacy iptables?

SteveMattar:
> Before we go ahead, Let's talk to @SteveMattar how will we make this work in downstream. In downstream won't we depend on having the ability to install also nft/legacy iptables?

is `iptables-nft` = `nftables` ?

skitt:
> > Before we go ahead, Let's talk to @SteveMattar how will we make this work in downstream. In downstream won't we depend on having the ability to install also nft/legacy iptables?
> 
> is `iptables-nft` = `nftables` ?

It’s `nftables`-backed `iptables`: it provides a backwards-compatible `iptables` command, using `nftables` in the kernel (via `libnftnl`).

RHEL 8 only has `nftables`-backed `iptables`, _i.e._ installing `iptables` gives the same as `iptables-nft` in Fedora, and Fedora’s legacy `iptables` isn’t available at all. This matters if we need to support running containers on RHEL 7 and 8: in that case, we’d need one set of container images based on RHEL 7, and another based on RHEL 8...

SteveMattar:
Thanks for the explaination @skitt . Do you think I need it even though we know openshift 4 runs on coreos and rhel8 ?

skitt:
> Thanks for the explaination @skitt . Do you think I need it even though we know openshift 4 runs on coreos and rhel8 ?

If we know it’s all RHEL 8, then no, we don’t need this, we can install RHEL 8 `iptables` directly.

SteveMattar:
@skitt does the provided `iptables-wrapper-installer.sh` script support both packages?

skitt:
> @skitt does the provided `iptables-wrapper-installer.sh` script support both packages?

It requires both packages. For RHEL 8 we’ll have to _only_ install `iptables`, without the wrapper script.

mangelajo:
> > @skitt does the provided `iptables-wrapper-installer.sh` script support both packages?
> 
> It requires both packages. For RHEL 8 we’ll have to _only_ install `iptables`, without the wrapper script.

ok, that makes sense to me, it's the part of the plan that wasn't clear. In that case it would work because the iptables/kernel will always match. 

Did I get it right @skitt ? if so, please feel free to merge.

mangelajo (approved):

skitt:
> > > @skitt does the provided `iptables-wrapper-installer.sh` script support both packages?
> > 
> > 
> > It requires both packages. For RHEL 8 we’ll have to _only_ install `iptables`, without the wrapper script.
> 
> ok, that makes sense to me, it's the part of the plan that wasn't clear. In that case it would work because the iptables/kernel will always match.
> 
> Did I get it right @skitt ? if so, please feel free to merge.

Yes, for Red Hat supported scenarios we’d only be using `nftables` now, and installing the `iptables` package from RHEL 8 in our downstream images, without any wrapper, will do the right thing and work as-is with the existing code.

dfarrell07:
The timed out `E2E (operator, globalnet, libreswan)` job actually had a (slow) failure I think?

https://github.com/submariner-io/submariner/pull/1186/checks?check_run_id=2123186424#step:3:2512

https://pastebin.com/RaaN1A96

Rerunning...

PR #1188, opened 2021-03-08
Revert "Use the K8s SIG iptables wrapper (#1009)"

mangelajo:
This reverts commit 83a140a755df973e6f8feec8d88ed5fa836022e9.


skitt:
This might not work simply because the operator no longer sets up the host mount...

mangelajo:
> This might not work simply because the operator no longer sets up the host mount...

That image never got to build because of the other E2E errors. So I was testing if this was the culprit, but it wasnt :-)

PR #1191, opened 2021-03-09
Import Makefile.dapper downloaded from Shipyard

mkolesnik:
Instead of redefining the dapper file here, download it from Shiyard


Makefile.dapper
```diff
@@ -1,19 +1,8 @@
...
+# This Makefile imports the rules to run in a Dapper-based build environment.
+# The rules are being imported via download from Shipyard where the base file can be found
 
-.dapper:
-	@echo Downloading dapper
-	@curl -sL https://releases.rancher.com/dapper/latest/dapper-`uname -s`-`uname -m` > .dapper.tmp
-	@@chmod +x .dapper.tmp
-	@./.dapper.tmp -v
-	@mv .dapper.tmp .dapper
+BASE_BRANCH := devel
```
skitt:
```suggestion
BASE_BRANCH ?= devel
```

Makefile.dapper
```diff
@@ -1,19 +1,8 @@
...
+_ := $(shell test -f Makefile.shipyard || curl -L https://raw.githubusercontent.com/submariner-io/shipyard/$(BASE_BRANCH)/Makefile.dapper > Makefile.shipyard)
```
skitt: How do we ensure that local builds don’t use a stale file?

skitt (changes requested):

Makefile.dapper
mkolesnik: Well the intention was to minimize local redownload each time, however I see were also fetching dapper each time.. So.. Maybe it's unnecessary to optimize right now?

mkolesnik: But generally what I thought of is to have `make clean` which also removes these cached files, theoretically :)

mkolesnik: > Well the intention was to minimize local redownload each time, however I see were also fetching dapper each time.. So.. Maybe it's unnecessary to optimize right now?

Actually just realized I missed the point of `make` itself caching it locally, so its the same there I guess..

Makefile.dapper
mkolesnik: Why would we want to allow it to be set from outside?

mkolesnik:
@skitt OK this is a much more "Makefile-ish" approach, let me know what you think

Makefile
```diff
@@ -84,9 +86,15 @@ else
 
 # Not running in Dapper
 
+Makefile.dapper:
+	@echo Downloading $@
+	@curl -sfL -o $@ https://raw.githubusercontent.com/submariner-io/shipyard/$(BASE_BRANCH)/$@
```
skitt:
```suggestion
	@curl -sfL -O https://raw.githubusercontent.com/submariner-io/shipyard/$(BASE_BRANCH)/$@
```

skitt (changes requested):
Much better, thanks! A minor tweak if I may...

Makefile.dapper
skitt: Right, `make clean` would be nice to have! (In another PR, not necessarily by you.)

Makefile
mkolesnik: Thanks!
Had to do it manually to avoid "Commit suggestion" commit since this PR has 2 distinct commits (though very simple ones indeed)

skitt (approved):

dfarrell07 (approved):

mangelajo (approved):

PR #1196, opened 2021-03-16
Convert IPAM to use admiral resource syncer

### LINKED ISSUE #375: Migrate controllers to use admiral APIs

tpantelis:
Fixes https://github.com/submariner-io/submariner/issues/375

### DIFF
--- a/pkg/globalnet/controllers/ipam/gatewaymonitor.go
+++ b/pkg/globalnet/controllers/ipam/gatewaymonitor.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
+	"github.com/submariner-io/admiral/pkg/stringset"
 	"github.com/submariner-io/admiral/pkg/syncer"
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
 	admUtil "github.com/submariner-io/admiral/pkg/util"
@@ -42,6 +43,7 @@ func NewGatewayMonitor(spec *SubmarinerIPAMControllerSpecification, config watch
 		clusterID:     spec.ClusterID,
 		ipamSpec:      spec,
 		isGatewayNode: false,
+		remoteSubnets: stringset.NewSynchronized(),
 	}
 
 	iptableHandler, err := iptables.New()
@@ -151,6 +153,7 @@ func (gm *GatewayMonitor) handleCreatedOrUpdatedEndpoint(obj runtime.Object, num
 		}
 
 		for _, remoteSubnet := range endpoint.Spec.Subnets {
+			gm.remoteSubnets.Add(remoteSubnet)
 			MarkRemoteClusterTraffic(gm.ipt, remoteSubnet, AddRules)
 		}
 
@@ -160,6 +163,10 @@ func (gm *GatewayMonitor) handleCreatedOrUpdatedEndpoint(obj runtime.Object, num
 	hostname, err := os.Hostname()
 	if err != nil {
 		klog.Fatalf("Unable to determine hostname: %v", err)
+	}
+
+	for _, remoteSubnet := range gm.remoteSubnets.Elements() {
+		MarkRemoteClusterTraffic(gm.ipt, remoteSubnet, AddRules)
 	}
 
 	// If the endpoint hostname matches with our hostname, it implies we are on gateway node
@@ -208,6 +215,7 @@ func (gm *GatewayMonitor) handleRemovedEndpoint(obj runtime.Object, numRequeues 
 	} else if endpoint.Spec.ClusterID != gm.clusterID {
 		// Endpoint associated with remote cluster is removed, delete the associated flows.
 		for _, remoteSubnet := range endpoint.Spec.Subnets {
+			gm.remoteSubnets.Remove(remoteSubnet)
 			MarkRemoteClusterTraffic(gm.ipt, remoteSubnet, DeleteRules)
 		}
 	}
--- a/pkg/globalnet/controllers/ipam/gatewaymonitor.go
+++ b/pkg/globalnet/controllers/ipam/gatewaymonitor.go
@@ -16,55 +16,33 @@ limitations under the License.
 package ipam
 
 import (
-	"context"
 	"fmt"
 	"os"
-	"time"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
-	"k8s.io/client-go/dynamic"
-
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/dynamic/dynamicinformer"
-
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/syncer/broker"
+	admUtil "github.com/submariner-io/admiral/pkg/util"
+	"github.com/submariner-io/admiral/pkg/watcher"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cidr"
 	"github.com/submariner-io/submariner/pkg/globalnet/cleanup"
 	"github.com/submariner-io/submariner/pkg/iptables"
 	"github.com/submariner-io/submariner/pkg/netlink"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/workqueue"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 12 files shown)


mangelajo:
Hi @tpantelis although I see the value of using common code in terms of maintenance I'm stressed out by the risks of the 0.9 release, I have preliminary assigned this one to 0.10, but feel free to bring it up in team meeting for rescheduling.

On the other hand IPAM/globalnet is something we have discarded for full support until we make globalnet v2 in 0.10, so may be it's simply not that risky

skitt (approved):

skitt:
> Hi @tpantelis although I see the value of using common code in terms of maintenance I'm stressed out by the risks of the 0.9 release, I have preliminary assigned this one to 0.10, but feel free to bring it up in team meeting for rescheduling.
> 
> On the other hand IPAM/globalnet is something we have discarded for full support until we make globalnet v2 in 0.10, so may be it's simply not that risky

It seems to me that this would simplify support by only requiring us to fix bugs in one set of code, and ensuring that any bug fixes are applied everywhere.

sridhargaddam:
This PR is very intrusive and needs some dedicated time to ensure that there are no regressions. I will take a look when I get some bandwidth.

mangelajo:
Based on @sridhargaddam  commends and mine, I think it's better if we wait for 0.10.

I don't want to introduce errors on globalnet at this point, where we should be using our full bandwidth on the 0.9 features we planned.

mangelajo (changes requested):
I vote we wait for 0.10

There are no plans to touch the globalnet code, so the code should not age badly.

tpantelis:
I didn't mean to create a ruckus - sorry 😄 This is fine for 0.10. This is something that's been on my list and I had started before the priority shift. Note that we've seen intermittent test failures due to the structure of the prior code - converting to the admiral APIs fixes that.  The main reason I wanted to at least get the patch done.

tpantelis:
> This PR is very intrusive and needs some dedicated time to ensure that there are no regressions. I will take a look when I get some bandwidth.

Understood. Regression was the reason I created the unit tests prior to refactoring. The pertinent code paths are covered by the tests.

mangelajo:
> I didn't mean to create a ruckus - sorry  This is fine for 0.10. This is something that's been on my list and I had started before the priority shift. Note that we've seen intermittent test failures due to the structure of the prior code - converting to the admiral APIs fixes that. The main reason I wanted to at least get the patch done.

Thank you, it makes a lot of sense ':D

tpantelis:
bump

sridhargaddam:
bump

tpantelis:
bump

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -16,63 +16,74 @@ limitations under the License.
...
+	if config.Scheme == nil {
+		config.Scheme = scheme.Scheme
+	}
+
+	config.ResourceConfigs = []watcher.ResourceConfig{
+		{
+			Name:         "IPAM GatewayMonitor",
+			ResourceType: &v1.Endpoint{},
+			Handler: watcher.EventHandlerFuncs{
+				OnCreateFunc: gatewayMonitor.handleCreatedOrUpdatedEndpoint,
```
mangelajo, line 77: What do you think about moving right away to the https://github.com/submariner-io/submariner/blob/devel/pkg/event/handler.go interface instead.

The event handler also provides updates on the nodes. As well as transitions to gateways or non-gateways.

@sridhargaddam  @tpantelis ^ ?

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -124,197 +127,104 @@ func (gm *GatewayMonitor) Stop() {
...
 	if err != nil {
-		utilruntime.HandleError(err)
-		return true
+		klog.Fatalf("Unable to determine hostname: %v", err)
 	}
 
-	return true
-}
+	// If the endpoint hostname matches with our hostname, it implies we are on gateway node
+	if endpoint.Spec.Hostname == hostname {
```
mangelajo, line 173: This detection is also provided by the event interface. (Transition to gateway and transition to non gateway)

mangelajo (commented):
I made one question.

mangelajo:
as @tpantelis commented wireguard+globalnet failed, but we need to determine if it's really related to the patch:

https://github.com/submariner-io/submariner/pull/1196/checks?check_run_id=2556449601

The error here was on gateway failover.

I will rerun

For the flake finder we see ~15 days ago some failures with libreswan + globalnet or wireguard+globalnet

https://github.com/submariner-io/submariner/actions/workflows/flake_finder.yml

In that case it was also redundency related, but in a different point:
```
[Fail] [redundancy] Route Agent restart tests when a route agent pod running on a gateway node is restarted [It] should start a new route agent pod and be able to connect from another cluster 
```

pkg/globalnet/controllers/ipam/gatewaymonitor.go
tpantelis: yeah we can look into using that interface but I'd rather do that in another PR. We need to move this forward to proceed with GN 2.0 enhancements.

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -124,197 +127,104 @@ func (gm *GatewayMonitor) Stop() {
...
-func (gm *GatewayMonitor) handleRemovedEndpoint(obj interface{}) {
-	var object *v1.Endpoint
-	var ok bool
-	if object, ok = obj.(*v1.Endpoint); !ok {
-		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
-		if !ok {
-			klog.Errorf("Could not convert object %v to an Endpoint", obj)
-			return
+		gm.syncMutex.Lock()
+		if !gm.isGatewayNode {
```
sridhargaddam, line 179: This part of code gets simplified if we move to Event handler interface, but that can be done in a separate PR.

pkg/globalnet/controllers/ipam/gatewaymonitor.go
mangelajo: @tpantelis that sounds good this PR already implements a good deal of improvements and testing. It should be easier to move from this one if we finally see that as reasonable.

mangelajo:
It failed again with wireguard on failover: https://github.com/submariner-io/submariner/pull/1196/checks?check_run_id=2564983655#step:3:2912

sridhargaddam:
The code looks good to me, but we have to investigate why it's failing with Wireguard driver alone.

sridhargaddam (approved):

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -124,197 +127,104 @@ func (gm *GatewayMonitor) Stop() {
...
+			klog.Warningf("unable to validate overlapping Service CIDR: %s", err)
 		}
 
-		for _, endpoint := range endpoints.Items {
-			if endpoint.Spec.ClusterID != gm.clusterID {
-				for _, remoteSubnet := range endpoint.Spec.Subnets {
-					MarkRemoteClusterTraffic(gm.ipt, remoteSubnet, AddRules)
-				}
-			}
-		}
```
sridhargaddam, line 195: @tpantelis I was debugging the issue with Wireguard driver and found the problem. The following code is accidentally removed in the new implementation.

I was wondering why it was working with Libreswan Cabledriver and noticed that there is a difference of behavior between the two. In Libreswan driver, when the Gateway pod is restarted, it was not deleting the local Endpoint but with Wireguard driver, it was deleting the local endpoint. Anyways, IMHO I feel we should fix the code in Globalnet.

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -124,197 +127,104 @@ func (gm *GatewayMonitor) Stop() {
...
 		}
 
-		gm.endpointWorkqueue.Forget(obj)
-
-		return nil
-	}()
+		return false
+	}
 
+	hostname, err := os.Hostname()
```
sridhargaddam, line 163: Here is the missing code : https://github.com/submariner-io/submariner/blob/7c57fa2974dab3c57b9e870ded0e7c61ea2649f2/pkg/globalnet/controllers/ipam/gatewaymonitor.go#L183

sridhargaddam (changes requested):

pkg/globalnet/controllers/ipam/gatewaymonitor.go
tpantelis: Thanks for looking into this. I remember this code now - it didn't seem necessary since we do the same thing above on L173 when we get notified of the remote Endpoint. So I'm not clear on why we would also do this for a local Endpoint, ie it seems redundant.

tpantelis: Fixed with https://github.com/submariner-io/submariner/pull/1196/commits/63cc50ef87f7a04d3b8e3ab47c7fc7be7ed8cd66.

sridhargaddam: Thanks for fixing it @tpantelis

sridhargaddam (approved):

sridhargaddam:
@mangelajo can you please re-review this PR. Its now passing CI as well.

PR #1199, opened 2021-03-16
Abstract GatewayMonitor ConfigureTCPMTUProbe for unit tests

tpantelis:
This causes a warning in unit tests due to file permissions but it would be ideal not to even try to update system files. So this function was abstracted behind the netlink interface.

This is not important for 0.9.0 so I added it to the 0.10.0 board.


mangelajo (approved):

sridhargaddam (approved):
Thank you @tpantelis

skitt (approved):

PR #1205, opened 2021-03-19
Add round trip NAT discovery unit tests

tpantelis:
Tests from `AddEndpoint` to ready channel notified.

Transition to `useLegacyNATSettings` now notifies the ready channel.

The tests required changing the duration constants like `recheckTime` to vars of type int64 and using atomic operations to access them as they can be accessed by multiple threads which can cause intermittent race errors.

Also change the endpoint field in `remoteEndpointNAT` to a non-pointer to avoid race errors.

### DIFF
--- a/pkg/natdiscovery/natdiscovery_suite_test.go
+++ b/pkg/natdiscovery/natdiscovery_suite_test.go
@@ -49,6 +49,7 @@ const (
 	testRemoteClusterID    = "cluster-b"
 	testRemotePublicIP     = "10.3.3.3"
 	testRemotePrivateIP    = "4.4.4.4"
+	testRemotePrivateIP2   = "5.5.5.5"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


pkg/natdiscovery/natdiscovery_suite_test.go
```diff
@@ -80,27 +81,53 @@ func parseProtocolResponse(buf []byte) *natproto.SubmarinerNatDiscoveryResponse
...
 
-func createTestListener(endpoint *types.SubmarinerEndpoint) (*natDiscovery, chan []byte) {
+func createTestListener(endpoint *types.SubmarinerEndpoint) (*natDiscovery, chan []byte, chan *NATEndpointInfo) {
 	listener, err := newNatDiscovery(&types.SubmarinerEndpoint{Spec: endpoint.Spec})
+	Expect(err).To(Succeed())
+
 	readyChannel := make(chan *NATEndpointInfo, 100)
 	listener.SetReadyChannel(readyChannel)
-	Expect(err).NotTo(HaveOccurred())
+	Expect(err).To(Succeed())
```
mangelajo, line 90: just opinion comment:

The error should not have "Succeeded", newNatDiscovery yes..., sounds weird to me for an error.

Also I'm spanish Succeeds sounds very similar to "Happened" (sucedido) which I guess it makes it weirder for me but that's a different topic.

For those reasons I always tend to like more .NotTo(HaveOccurred()).

mangelajo (approved):

sridhargaddam (approved):

pkg/natdiscovery/natdiscovery_suite_test.go
tpantelis: I used to do `NotTo(HaveOccurred())` when switched to `To(Succeed())` 😄  It's a matter of taste.

PR #1206, opened 2021-03-19
Modify aws prepare tool to support Gateway HA

### LINKED ISSUE #1180: Epic: Gateway HA with OpenShift
Out of the box, ensure that the Submariner gateway is deployed in active/passive HA mode on OpenShift-based clusters, with at least one additional node (other than the active one) labeled and ready to take over.

Previous related work from @sridhargaddam: 

1. https://github.com/submariner-io/submariner-operator/issues/586 
2. https://docs.google.com/spreadsheets/d/1JsXsyRDDXkp6t55Gm-NP5EggWTyYi2yo27pyuDYwlpc/edit#gid=0 

**Goal**:

When deploying with OpenShift, ensure that there are at least two nodes in two different AZs that are labeled properly and can act as gateways. The idea is that in case of a gateway node or AZ failure, data path traffic will be minimally impacted.

**Key work items**:

1. Deploy this manually on AWS and test what happens...
- Is there a way to disconnect an AZ, or shall we remove all instances on that AZ? (no normal power off but forced stop/shutdown)
- When an AZ is down, will the GW on the disconnected AZ leave leadership? Will the passive GW take leadership?
- What happens when the AZ goes back?

2. Assuming tests are OK, how should we automate the deployment piece?
- For ACM/cloud-prepare repo (for OpenShift/IPI, we need to deploy an additional node)
- MVP is AWS, but we will need to figure out the plan for other providers (GCP, Azure, IBM)

3. What happens in combination with NAT (on-prem)
- Probably each gateway node would need a set of ports (different ones), which can be mapped from a router
- For example:
```
oc label node $node1 submariner.io/ipsec-natt-port=4500
oc label node $node1 submariner.io/ipsec-ike-port=500
oc label node $node2 submariner.io/ipsec-natt-port=4501
oc label node $node2 submariner.io/ipsec-ike-port=501
```

4. Document the expected behavior. We need to set the right expectations and explain how the system work

- [x] Test with AWS
- [ ] Automate for ACM deployment
- [x]  #1206 Automate for IPI deployment
- [x] Test with NAT
- [x] https://github.com/submariner-io/submariner-website/pull/473 ,  #1232  Update Documentation

aswinsuryan:
*if ENABLE_HA is set an additional g/w capable node will
be deployed

Fixes: #1180

### DIFF
--- a/tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/machine-set-resources.tf
+++ b/tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/machine-set-resources.tf
@@ -1,5 +1,6 @@
 locals {
   machine_set_config_file = "${path.cwd}/submariner-gw-machine-set-${data.aws_subnet.target_public_subnet.availability_zone}.yaml"
+  machine_set_config_file_ha = "${path.cwd}/submariner-gw-machine-set-${data.aws_subnet.target_public_subnet_ha.availability_zone}.yaml"
 }
 
 # Get worker instances.
@@ -24,6 +25,11 @@ data "aws_instances" "worker_instances" {
 # Get worker instance.
 data "aws_instance" "worker_instance" {
   instance_id = data.aws_instances.worker_instances.ids[0]
+}
+
+# Get worker instance.
+data "aws_instance" "worker_instance_ha" {
+  instance_id = data.aws_instances.worker_instances.ids[1]
 }
 
 # Create empty machine set config file.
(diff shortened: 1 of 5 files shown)


tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/machine-set-resources.tf
```diff
@@ -52,6 +58,25 @@ data "template_file" "machine_set_template" {
   ]
 }
 
+# Render the template file.
+data "template_file" "machine_set_ha_template" {
+  template = file("${path.module}/templates/machine-set.yaml")
+
+  vars = {
+    az                   = data.aws_subnet.target_public_subnet_ha.availability_zone
```
mangelajo, line 66: Hmm, but we need to use different AZs for each, right?

mangelajo (changes requested):

tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/machine-set-resources.tf
aswinsuryan: With this we are taking AZ from a different public subnet, as the public subnet is per availability zone, they will use different zones.

mangelajo: Oh,  I didn't see the input anywhere, is that provided by OCP or AWS?

mangelajo (approved):

tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/machine-set-resources.tf
aswinsuryan: We get it from the terraform state , so should be provided by AWS.

tools/openshift/ocp-ipi-aws/ocp-ipi-aws-prep/machine-set-resources.tf
```diff
@@ -61,3 +86,15 @@ resource "local_file" "machine_set_config" {
     "null_resource.empty_file",
   ]
 }
+
+# Create machine set config file from template.
+resource "local_file" "machine_set_config_ha" {
+  count = var.enable_ha ? 1 : 0
```
sridhargaddam, line 92: where is the `count` used?

aswinsuryan: This is not to create the resource if enable_ha is false. If the resource count is zero no resource will be created. This is a way to implement if-else in terraform. I assume this count is used internally by terraform

sridhargaddam: Okay, thanks @aswinsuryan

sridhargaddam (approved):

PR #1209, opened 2021-03-22
Mods to public IP to private IP transition

tpantelis:
If the public IP respond first, the ready channel is immediately notified even if the private IP is still pending. If the private
IP responds within  the grace period, it will also notify the ready channel. It seems we should only notify the ready channel once, ie delay the selection of the public IP until the grace period elapses.

Added a new state, `selectPublicIPPending`, for the transient period until either the private IP responds or the grace period elapses. The check for the latter is part of `shouldCheck`. Since the `remoteEndpointNAT` now controls notification, the `readyChannel` is also stored in each `remoteEndpointNAT`.

### DIFF
--- a/pkg/natdiscovery/natdiscovery.go
+++ b/pkg/natdiscovery/natdiscovery.go
@@ -123,7 +123,7 @@ func (nd *natDiscovery) AddEndpoint(endpoint *types.SubmarinerEndpoint) {
 		}
 	}
 
-	remoteNAT := newRemoteEndpointNAT(endpoint)
+	remoteNAT := newRemoteEndpointNAT(endpoint, nd.readyChannel)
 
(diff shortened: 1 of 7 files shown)


dfarrell07 (approved):

mangelajo (dismissed):
Let's hold on on this one until I finish the integration.

I'm having to make multiple changes to the implementation. Also, the implementation supports the multiple endpoint sent twice or multiple times, it will just update the connection in libreswan/wireguard as if the remote IP of the endpoint had changed.

tpantelis:
bump

tpantelis:
Bump - we may still do this.

tpantelis:
bump for now

tpantelis:
bump

tpantelis:
bump

PR #1212, opened 2021-03-24
Enable Helm+cable driver test matrix

dfarrell07:
The Helm charts now support setting the cable driver, so the exclude
here is out-of-date and can be removed.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -17,11 +17,6 @@ jobs:
         cable_driver: ['libreswan', 'wireguard']
         ovn: ['', 'ovn']
         exclude:
-          # Our Helm setup doesn’t know how to deploy other cable drivers
-          - deploytool: 'helm'
-            cable_driver: 'wireguard'
-          - deploytool: 'helm'
-            cable_driver: 'libreswan'
           - ovn: 'ovn'
             deploytool: 'helm'
           - ovn: 'ovn'
--- a/.github/workflows/flake_finder.yml
+++ b/.github/workflows/flake_finder.yml
@@ -17,12 +17,6 @@ jobs:
         deploytool: ['operator', 'helm']
         globalnet: ['', 'globalnet']
         cable_driver: ['libreswan', 'wireguard']
-        exclude:
-          # Our Helm setup doesn’t know how to deploy other cable drivers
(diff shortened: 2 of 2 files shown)


tpantelis:
So this adds 4 more E2E jobs and maybe more load to the CI infra - wondering if this is desirable. Is it possible to add these jobs to _submariner-charts_ instead?

dfarrell07:
> So this adds 4 more E2E jobs and maybe more load to the CI infra - wondering if this is desirable. Is it possible to add these jobs to _submariner-charts_ instead?

That makes sense I think. If we go that route we should also move the globalnet part of the matrix to only run in the charts repo, and just have a vanilla Helm job here.

tpantelis:
>> So this adds 4 more E2E jobs and maybe more load to the CI infra - wondering if this is desirable. Is it possible to add these jobs to submariner-charts instead?

> That makes sense I think. If we go that route we should also move the globalnet part of the matrix to only run in the charts repo, and just have a vanilla Helm job here.

We want to run the operator globalnet jobs here. I propose we run no helm jobs herehere as has been the case.

dfarrell07:
I broke your comment for a bit Tom, sorry.

Ah I see, I thought we had some Helm jobs but I see they were actually all covered by this exclude. So we should just remove all this Helm config from the GHA, as it amounts to nothing currently.

tpantelis:
> I broke your comment for a bit Tom, sorry.
> 
> Ah I see, I thought we had some Helm jobs but I see they were actually all covered by this exclude. So we should just remove all this Helm config from the GHA, as it amounts to nothing currently.

yeah looks like the intent was to exclude `wireguard` and `libreswan` and only run helm with `strongswan`. But when we removed  `strongswan`, this was left  as is so helm wasn't running at all.  We should also change `deploytool: ['operator', 'helm']` -> `deploytool: ['operator']`, ie unless we want helm running here but I think it makes sense to run helm E2E in _submariner-charts_. But we can see what others think.

dfarrell07:
> We should also change `deploytool: ['operator', 'helm']` -> `deploytool: ['operator']`, ie unless we want helm running here but I think it makes sense to run helm E2E in _submariner-charts_. But we can see what others think.

I sent https://github.com/submariner-io/submariner/pull/1223 to do this.

PR #1214, opened 2021-03-25
Test K8s versions 1.18, 1.19, and 1.20 in E2E

dfarrell07:
Use the new ability of the shared E2E GHA to configure the Kubernetes
version. Add jobs that cover the latest available patch versions
(limited by kind images) from the three most recent minor versions.

Run one E2E job with all-default configuration per K8s version. Also
test the latest K8s version with features that are primarily implemented
in this repo, like the cable driver (or ovn) and using globalnet or not.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -12,15 +12,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        cable_driver: ['libreswan', 'wireguard']
         deploytool: ['operator']
         globalnet: ['', 'globalnet']
-        cable_driver: ['libreswan', 'wireguard']
+        k8s_version: ['1.17.17']
         ovn: ['', 'ovn']
         exclude:
           - ovn: 'ovn'
             globalnet: 'globalnet'
           - ovn: 'ovn'
             cable_driver: 'wireguard'
+        include:
+          # Recentness of K8s versions are limited by kindest/node image releases
+          - k8s_version: 1.18.15
+          - k8s_version: 1.19.7
+          - k8s_version: 1.20.2
+          # TODO: Remove these latest-K8s+major-feature jobs once default=latest K8s
+          - k8s_version: 1.20.2
+            cable_driver: 'wireguard'
+          - k8s_version: 1.20.2
+            globalnet: 'globalnet'
+          - k8s_version: 1.20.2
+            ovn: 'ovn'
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
@@ -28,6 +41,7 @@ jobs:
       - name: Run E2E deployment and tests
         uses: submariner-io/shipyard/gh-actions/e2e@devel
         with:
+          k8s_version: ${{ matrix.k8s_version }}
           using: ${{ matrix.cable_driver }} ${{ matrix.deploytool }} ${{ matrix.globalnet }} ${{ matrix.ovn }}
 
       - name: Post mortem


dfarrell07:
A couple of failing jobs, collected logs and re-running:

E2E (1.20.2, globalnet):
https://pastebin.com/apYwpA5w

E2E (libreswan, operator, 1.17.17, ovn):
https://pastebin.com/p9CENskb

tpantelis:
So this adds another 6 jobs that will run for every PR. Perhaps we should consider running these as a periodic job?

dfarrell07:
I think we should gate on testing our supported K8s versions. Otherwise we're just going to do more of the anti-pattern of periodic breaking, someone finally noticing, maybe some slow fix. This is a big lesson of OpenStack's CI.

I also think we should have gating latest-K8s coverage for major matrix features in at least one place. Ideally that coverage would gate out breakages by running on PRs close to where the feature is implemented (as I think would happen here). If we move our default K8s version to be the latest one, these jobs would go away.

I think we should add periodic jobs that cover features implemented here (globalnet, cable drivers, OVN) for all supported minor versions (adding 1.18 and 1.19 dimensions).

tpantelis:
> I think we should gate on testing our supported K8s versions. Otherwise we're just going to do more of the anti-pattern of periodic breaking, someone finally noticing, maybe some slow fix. This is a big lesson of OpenStack's CI.
> 
> I also think we should have gating latest-K8s coverage for major matrix features in at least one place. Ideally that coverage would gate out breakages by running on PRs close to where the feature is implemented (as I think would happen here). If we move our default K8s version to be the latest one, these jobs would go away.
> 
> I think we should add periodic jobs that cover features implemented here (globalnet, cable drivers, OVN) for all supported minor versions (adding 1.18 and 1.19 dimensions).

I hear ya.  Ideally we would gate on everything but of course we have resource limitations. We already see sporadic job timeouts and keep increasing the timeout. I'm afraid adding more PR gating jobs will make that worse.

dfarrell07:
I think these Globalnet with 1.21 failures are consistent:

> [Fail] [redundancy] Gateway fail-over tests when a new node is labeled as a gateway node and the label on the existing gateway node is removed [It] should start a submariner gateway on the new gateway node and be able to connect from another cluster 
/go/src/github.com/submariner-io/submariner/vendor/github.com/submariner-io/shipyard/test/e2e/framework/network_pods.go:174

https://github.com/submariner-io/submariner/pull/1214/checks?check_run_id=2224106299

skitt:
> I think these Globalnet with 1.21 failures are consistent:

1.20 ;-)

dfarrell07:
> We already see sporadic job timeouts and keep increasing the timeout. I'm afraid adding more PR gating jobs will make that worse.

@tpantelis Running more jobs will make us hit rare failures more, but I don't think adding more jobs will cause more flakeyness. I don't think the fix is to run fewer jobs, but instead to continue to fix flakeyness (as you've done a ton of, thanks).

dfarrell07:
It seems this Globalnet/1.20 failure is consistent. Here's a second example of the logs:

https://pastebin.com/6gL0tcFV

The OVN errors seem to be flaky failures, as they appear in different jobs each workflow run.

dfarrell07:
I guess this shouldn't be a draft, as the failures for Globlanet and 1.20 aren't introduced here, only revealed. We can mark those jobs as not-required for now.

tpantelis (approved):

dfarrell07:
Ah nice, the Globalnet+1.20 job passed after the most recent rebase.

tpantelis:
> Ah nice, the Globalnet+1.20 job passed after the most recent rebase.

It might be a fluke 😄

mangelajo:
I think we need to find a way of limiting the amount of jobs we add to PRs, we don't have infinite resources.

Periodic, although can be problematic should be a path.

What if we test the maximum & minimum versions on all PRs, and all of them in periodic as a compromise solution?

mangelajo (changes requested):
let's decide on a job matrix before merging this, it's too much for our CI capabilities IMHO

dfarrell07:
> I think we need to find a way of limiting the amount of jobs we add to PRs, we don't have infinite resources.

Are we hitting resource limits? I haven't seen so.

> Periodic, although can be problematic should be a path.

Strongly disagree. All of our CI lessons from history, and all of the discussions of this community, show us the importance of gating on PRs and not following the anti-pattern of waiting for periodic jobs to break.

> What if we test the maximum & minimum versions on all PRs, and all of them in periodic as a compromise solution.

If I understand you, this would leave us with no breakage-preventing coverage for cable drivers, globalnet, or the CNI. As I discussed above and for a big chunk of the call Monday, covering the major matrix features implemented in each repo is the right strategy IMHO. That means more jobs here (but fewer elsewhere), because more things are implemented here.

dfarrell07:
> What if we test the maximum & minimum versions on all PRs, and all of them in periodic as a compromise solution?

Actually this would result more jobs than proposed after making default=latest K8s version.

.github/workflows/e2e.yml
```diff
@@ -12,15 +12,27 @@ jobs:
...
+          # Recentness of K8s versions are limited by kindest/node image releases
+          - k8s_version: 1.18.15
+          - k8s_version: 1.19.7
+          - k8s_version: 1.20.2
+          - k8s_version: 1.20.2
+            cable_driver: 'wireguard'
+          - k8s_version: 1.20.2
+            globalnet: 'globalnet'
+          - k8s_version: 1.20.2
+            ovn: 'ovn'
```
mkolesnik, line 36:
```suggestion
```

These combinations should be checked with the default version we're using..

.github/workflows/e2e.yml
```diff
@@ -12,15 +12,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        cable_driver: ['libreswan', 'wireguard']
         deploytool: ['operator']
         globalnet: ['', 'globalnet']
-        cable_driver: ['libreswan', 'wireguard']
+        k8s_version: ['1.17.17']
```
mkolesnik, line 18: You're not actually using this, so all the jobs will actually run with whatever is default..

mkolesnik (changes requested):

.github/workflows/e2e.yml
dfarrell07: Good catch, thanks @mkolesnik. I double-checked and I didn't make this error elsewhere:

https://github.com/submariner-io/submariner-operator/pull/1183/files
https://github.com/submariner-io/lighthouse/pull/495/files
https://github.com/submariner-io/submariner-charts/pull/130/files

.github/workflows/e2e.yml
dfarrell07: I added an inline comment to clarify that these are temporary, to provide coverage for latest-K8s version with the major features implemented here, and will go away once default=latest version (but we shouldn't do that until we have these checks running to let us know what's coming).

.github/workflows/e2e.yml
dfarrell07: Double-checked here and it's pulling the correct node images now:

> 2021-04-07T15:07:20.3318828Z [cluster3]  ✓ Ensuring node image (kindest/node:v1.19.7)

Thanks again MIke, great catch.

dfarrell07:
> > Ah nice, the Globalnet+1.20 job passed after the most recent rebase.
> 
> It might be a fluke

It passed again. :sweat_smile: :partying_face: 

And thinking about @mkolesnik's catch, I checked those logs again and it wasn't even running with 1.20, it was the default 1.17.

So we actually haven't seen any failures related to the K8s version changes, flase alarm.

dfarrell07:
@mangelajo @mkolesnik @tpantelis How about https://github.com/submariner-io/submariner/pull/1242 as an alternative? It only adds one job but meets all of our coverage needs by jumping to 1.20 as the default and dropping coverage for 1.17 and 1.18 (not required as I understand it).

tpantelis:
> @mangelajo @mkolesnik @tpantelis How about #1242 as an alternative? It only adds one job but meets all of our coverage needs by jumping to 1.20 as the default and dropping coverage for 1.17 and 1.18 (not required as I understand it).

We build against the 1.17 version so perhaps we should test against that instead of 1.19. So full coverage for 1.20, sanity check for 1.17 and skip in between.

dfarrell07:
> So full coverage for 1.20, sanity check for 1.17 and skip in between.

That makes sense to me, I'll update https://github.com/submariner-io/submariner/pull/1242.

tpantelis:
Closed in favor of #1242.

PR #1217, opened 2021-03-26
Notify NAT discovery RemoveEndpoint in cable engine

tpantelis:
On `RemoveCable`, we need to call `RemoveEndpoint` on the NAT discovery.

Due to the async nature of install cable, an edge case is if `RemoveCable` occurs while NAT discovery is in progress. There's a timing window where the cable could be installed after `RemoveCable`. Added data structures and logic to prevent this. Also` driver.DisconnectFromEndpoint ` should not be called if cable installation did not successfully complete.

### DIFF
--- a/pkg/cableengine/cableengine.go
+++ b/pkg/cableengine/cableengine.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 
 	"github.com/submariner-io/admiral/pkg/log"
+	"github.com/submariner-io/admiral/pkg/stringset"
 	"k8s.io/klog"
 
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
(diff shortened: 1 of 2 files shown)


mangelajo (approved):
Thank you @tpantelis for being on top of this. Me forgetting about removeEndpoint was lame, but that corner case was very important. Thanks.

pkg/cableengine/cableengine.go
```diff
@@ -185,7 +200,12 @@ func (i *engine) InstallCable(endpoint types.SubmarinerEndpoint) error {
 		return nil
 	}
 
+	i.Lock()
+	i.natDiscoveryPending[endpoint.Spec.CableName]++
```
skitt, line 204: Should this also handle the case where `i.natDiscoveryPending[endpoint.Spec.CableName]` is already >= 1?

tpantelis: The intention of using a counter instead of simple bool (ie set) was to take into account possible multiple pending NAT discovery requests so we don't prematurely remove from `natDiscoveryPending` and miss a discovery response..

skitt: OK, so doesn’t that mean we need to check the number of pending discoveries in line 227, instead of blindly removing `i.natDiscoveryPending[endpoint.Spec.CableName]`?

tpantelis: In that case, when the endpoint is removed, we don't want to process any pending discovery responses. That covers the edge case described in the summary.

skitt: Oh yeah duh, the regular case is handled in line 141.

skitt (approved):

tpantelis:
Fixed conflict.

sridhargaddam (approved):

PR #1219, opened 2021-03-26
Allow also ESP/AH encapsulation on AWS

mangelajo:
The new nat discovery protocol can detect connectivity over private IPs
and that means that nat traversal is disabled by default, triggering ESP
protocol which is expected to be more performant than UDP encap.


skitt (approved):

tpantelis (approved):

dfarrell07 (approved):

sridhargaddam:
@mangelajo so this will be useful when VPC peering is used across the clusters in AWS?

mangelajo:
> @mangelajo so this will be useful when VPC peering is used across the clusters in AWS?

When private IP connection is possible between AWS instances (instead of elastic public IP) we
use ESP/AH by default instead of UDP encapsulation.

PR #1220, opened 2021-03-26
Custom backend config ports and single-udp-port

### LINKED ISSUE #1178: Expose the gateway ports used in the Endpoint (NAT support)

Add port information into the Endpoint which is published to the broker, in a way
that the connection ports for the cable driver aren't anymore fixed across the cluster.

**Why is this needed**:
Allowing HA configurations on-prem behind NAT.

mangelajo:
Local node gateway configuration can be applied using the following labels on
the gateway node:

 - gateway.submariner.io/udp-port=<port>
 - gateway.submariner.io/natt-discovery-port=<port>

This commit also moves the UDP implementation into single-port to
simplify the picture, as our two-port IPSEC doesn't make sense anymore
with the use of IKEv2 and manual port selection.

Fixes: #1178

### DIFF
--- a/pkg/apis/submariner.io/v1/endpoint.go
+++ b/pkg/apis/submariner.io/v1/endpoint.go
@@ -1,0 +1,47 @@
+/*
+© 2021 Red Hat, Inc. and others
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package v1
+
+import (
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+func (ep *EndpointSpec) GetBackendPort(configName string, defaultValue int32) (int32, error) {
+	if portStr := ep.BackendConfig[configName]; portStr != "" {
+		port, err := parsePort(portStr)
+		if err != nil {
+			return defaultValue, errors.Wrapf(err, "error parsing backend config %s", configName)
+		}
+
+		return port, nil
+	}
+
+	return defaultValue, nil
+}
+
+func parsePort(port string) (int32, error) {
+	if portInt, err := strconv.ParseUint(port, 10, 16); err != nil {
+		return -1, errors.Wrapf(err, "error parsing port %s", port)
+	} else if portInt < 1 {
+		return -1, errors.Errorf("port %s is < 1", port)
+	} else if portInt > 65535 {
+		return -1, errors.Errorf("port %s is > 65535", port)
+	} else {
+		return int32(portInt), nil
+	}
+}
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -70,15 +70,26 @@ type EndpointSpec struct {
 	ClusterID string `json:"cluster_id"`
 	CableName string `json:"cable_name"`
 	// +optional
-	HealthCheckIP    string            `json:"healthCheckIP,omitempty"`
-	Hostname         string            `json:"hostname"`
-	Subnets          []string          `json:"subnets"`
-	PrivateIP        string            `json:"private_ip"`
-	PublicIP         string            `json:"public_ip"`
-	NATEnabled       bool              `json:"nat_enabled"`
-	Backend          string            `json:"backend"`
-	BackendConfig    map[string]string `json:"backend_config,omitempty"`
-	NATDiscoveryPort *int32            `json:"natDiscoveryPort,omitempty"`
+	HealthCheckIP string            `json:"healthCheckIP,omitempty"`
+	Hostname      string            `json:"hostname"`
+	Subnets       []string          `json:"subnets"`
+	PrivateIP     string            `json:"private_ip"`
+	PublicIP      string            `json:"public_ip"`
+	NATEnabled    bool              `json:"nat_enabled"`
+	Backend       string            `json:"backend"`
+	BackendConfig map[string]string `json:"backend_config,omitempty"`
+}
+
+const (
+	GatewayConfigLabelPrefix = "gateway.submariner.io/"
+	UDPPortConfig            = "udp-port"
+	NATTDiscoveryPortConfig  = "natt-discovery-port"
+)
+
+// ValidGatewayNodeConfig list should contain only keys that configure node specific settings via labels
+var ValidGatewayNodeConfig = []string{
+	UDPPortConfig,
+	NATTDiscoveryPortConfig,
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/klog"
 
 	"github.com/submariner-io/admiral/pkg/log"
+
 	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cable"
 	"github.com/submariner-io/submariner/pkg/types"
@@ -47,29 +48,26 @@ func init() {
 }
 
 type libreswan struct {
-	secretKey string
-
-	debug   bool
-	logFile string
-
 	localEndpoint types.SubmarinerEndpoint
-
-	ipSecIKEPort  string
-	ipSecNATTPort string
-
 	// This tracks the requested connections
 	connections []subv1.Connection
+
+	secretKey string
+	logFile   string
+
+	ipSecNATTPort   string
+	defaultNATTPort int32
+
+	debug bool
 }
 
 type specification struct {
 	PSK      string
 	Debug    bool
 	LogFile  string
-	IKEPort  string `default:"500"`
 	NATTPort string `default:"4500"`
 }
 
-const defaultIKEPort = "500"
 const defaultNATTPort = "4500"
 const ipsecSpecEnvVarPrefix = "ce_ipsec"
 
@@ -82,14 +80,26 @@ func NewLibreswan(localEndpoint types.SubmarinerEndpoint, localCluster types.Sub
 		return nil, fmt.Errorf("error processing environment config for %s: %v", ipsecSpecEnvVarPrefix, err)
 	}
 
+	defaultNATTPort, err := strconv.ParseUint(ipSecSpec.NATTPort, 10, 16)
+	if err != nil {
+		return nil, errors.Errorf("error parsing CR_IPSEC_NATTPORT environment variable")
+	}
+
+	nattPort, err := localEndpoint.Spec.GetBackendPort(subv1.UDPPortConfig, int32(defaultNATTPort))
+	if err != nil {
+		return nil, errors.Wrapf(err, "error parsing %q from local endpoint", subv1.UDPPortConfig)
+	}
+
+	klog.Infof("Using NATT UDP port %d", nattPort)
+
 	return &libreswan{
-		secretKey:     ipSecSpec.PSK,
-		debug:         ipSecSpec.Debug,
-		logFile:       ipSecSpec.LogFile,
-		ipSecIKEPort:  ipSecSpec.IKEPort,
-		ipSecNATTPort: ipSecSpec.NATTPort,
-		localEndpoint: localEndpoint,
-		connections:   []subv1.Connection{},
+		secretKey:       ipSecSpec.PSK,
+		debug:           ipSecSpec.Debug,
+		logFile:         ipSecSpec.LogFile,
+		ipSecNATTPort:   strconv.Itoa(int(nattPort)),
+		defaultNATTPort: int32(defaultNATTPort),
+		localEndpoint:   localEndpoint,
+		connections:     []subv1.Connection{},
 	}, nil
 }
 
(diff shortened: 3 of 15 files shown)


pkg/cable/libreswan/libreswan.go
```diff
@@ -46,29 +47,26 @@ func init() {
 }
 
 type libreswan struct {
-	secretKey string
-
-	debug   bool
-	logFile string
-
 	localEndpoint types.SubmarinerEndpoint
```
mangelajo, line 51: The reorder of parameters here was enforced by golangci-lint "malign" checker.

It was complaining because if I reorder the struct fields it will need 256bytes instead of 264 for some reason. We don't care much about optimization here, but golangci-lint wouldn't pass otherwise.

mangelajo:
@skitt we were talking partly about this on the slack channel.

@tpantelis the continuation on the NATT work ^

pkg/cable/libreswan/libreswan.go
```diff
@@ -283,6 +293,16 @@ func (i *libreswan) ConnectToEndpoint(endpoint types.SubmarinerEndpoint, useIP s
 	leftSubnets := extractSubnets(i.localEndpoint.Spec)
 	rightSubnets := extractSubnets(endpoint.Spec)
 
+	rightNATTPort, err := cable.GetBackendPort(endpoint, subv1.UDPPortBackendConfig, i.ipSecNATTPort)
+	if err != nil {
+		return "", errors.Wrapf(err, "error parsing endpoint %q IKE port", endpoint.Spec.CableName)
+	}
```
tpantelis, line 303: Since `GetBackendPort` returns the specified default value, perhaps we should just log the error and not return it and use the default.  If we return the error, the engine will propagate to the tunnel controller which will queue for retry however, in this case, a retry won't help.

pkg/util/util.go
```diff
@@ -173,23 +174,48 @@ func GetLocalEndpoint(submSpec types.SubmarinerSpecification, backendConfig map[
...
+		if strings.HasPrefix(label, node.BackendConfigPrefix) {
+			config := label[len(node.BackendConfigPrefix):]
+			if !validConfigs.Contains(config) {
+				return backendConfig, errors.Errorf("unknown backend config from node %q label %q",
+					nodeObj.Name, label)
+			}
 
+			backendConfig[config] = value
+		}
+	}
```
tpantelis, line 181: Is there a use case where a user would want different backend configs for different gateway nodes? If not then they'd have to maintain these labels on every gateway node. We have a [proposal](https://github.com/submariner-io/enhancements/pull/4) for using policies to configure cable driver configs. It seems this may conflict with proposal. I don't want us stuck with supporting/migrating something legacy if we go with that proposal.

pkg/cable/libreswan/libreswan.go
mangelajo: Right the retries won't help here, let's try the standard port just in case and let's move on. It's also a very unlikely to happen problem, because the remote will go through the same parsing before publishing the port itself.

pkg/util/util.go
mangelajo: Yes we need different backend configs per gateway node, it's that use case what I'm trying to resolve. 

But limited ones, we could be very restrictive in the filters of what's accepted as a configuration specific per-gateway, for example:

* Ports
* "I prefer to be server" (that could be cluster-wide in most cases, but I can imagine situations where it wouldn't), we certainly need this to fix connectivity & NAT completely for production.


So this won't become legacy, unless the other proposal has a way to express that, but I believe this local configuration detail is probably more reasonable to do in the cluster itself.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -81,6 +81,17 @@ type EndpointSpec struct {
 	NATDiscoveryPort *int32            `json:"natDiscoveryPort,omitempty"`
 }
 
+const (
+	UDPPortBackendConfig               = "udp-port"
+	ForceUDPEncapsulationBackendConfig = "force-udp-encaps"
+)
+
+// This list should contain only keys that configure node specific settings via labels
```
mangelajo: @tpantelis following your comments I've made it clear here that we should think twice before adding other valid parameters, for example selection of the cable driver, or anything that would be cluster or inter-cluster -wide.

pkg/util/util.go
mangelajo: Based on this conversation I have updated this here: https://github.com/submariner-io/submariner/pull/1220/files#diff-c52b44daa2be3d01c651d0fe4dc5821b8da017dfabd4cc60830911b37445b0bfR89

mangelajo:
Closed in favor of https://github.com/submariner-io/submariner/pull/1220

pkg/apis/submariner.io/v1/types.go
```diff
@@ -70,15 +70,26 @@ type EndpointSpec struct {
...
+	PublicIP      string            `json:"public_ip"`
+	NATEnabled    bool              `json:"nat_enabled"`
+	Backend       string            `json:"backend"`
+	BackendConfig map[string]string `json:"backend_config,omitempty"`
+}
+
+const (
+	BackendConfigPrefix            = "gateway.submariner.io/"
+	UDPPortBackendConfig           = "udp-port"
+	NATTDiscoveryPortBackendConfig = "natt-discovery-port"
```
tpantelis: I think we can omit `Backend` from the names. Frankly I would've preferred `Backend` -> `CableDriver` and `BackendConfig` -> `CableDriverConfig` 😄 

```suggestion
	GatewayConfigLabelPrefix            = "gateway.submariner.io/"
	UDPPortConfig           = "udp-port"
	NATTDiscoveryPortConfig = "natt-discovery-port"
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -70,15 +70,26 @@ type EndpointSpec struct {
...
+)
+
+// ValidNodeSpecificBackendConfigs list should contain only keys that configure node specific settings via labels
+var ValidNodeSpecificBackendConfigs = []string{
```
```suggestion
var ValidGatewayNodeConfigs = []string{
```

pkg/cable/libreswan/libreswan.go
```diff
@@ -286,6 +291,12 @@ func (i *libreswan) ConnectToEndpoint(endpointInfo *natdiscovery.NATEndpointInfo
 	leftSubnets := extractSubnets(i.localEndpoint.Spec)
 	rightSubnets := extractSubnets(endpoint.Spec)
 
+	rightNATTPort, err := endpoint.Spec.GetBackendPortStr(subv1.UDPPortBackendConfig, i.defaultNATTPort)
+	if err != nil {
+		klog.Warningf("Error parsing remote endpoint %q IKE port: %s, using port %s instead", endpoint.Spec.CableName, err,
+			i.ipSecNATTPort)
```
tpantelis:
```suggestion
		klog.Warningf("Error parsing %q from remote endpoint %q - using port %s instead: %v", subv1.UDPPortBackendConfig, endpoint.Spec.CableName,
			i.ipSecNATTPort, err)
```

pkg/cable/libreswan/libreswan.go
```diff
@@ -82,14 +80,21 @@ func NewLibreswan(localEndpoint types.SubmarinerEndpoint, localCluster types.Sub
 		return nil, fmt.Errorf("error processing environment config for %s: %v", ipsecSpecEnvVarPrefix, err)
 	}
 
+	nattPort, err := localEndpoint.Spec.GetBackendPortStr(subv1.UDPPortBackendConfig, ipSecSpec.NATTPort)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error reading IKE port")
```
tpantelis:
```suggestion
		return nil, errors.Wrapf(err, "error parsing %q from local endpoint", subv1.UDPPortBackendConfig)
```

pkg/cable/wireguard/driver.go
```diff
@@ -131,12 +133,18 @@ func NewDriver(localEndpoint types.SubmarinerEndpoint, localCluster types.Submar
 
 	w.localEndpoint.Spec.BackendConfig[PublicKey] = pub.String()
 
+	port, err := localEndpoint.Spec.GetBackendPort(v1.UDPPortBackendConfig, int32(w.spec.NATTPort))
+	if err != nil {
+		return nil, errors.Wrapf(err, "error reading UDP port")
```
tpantelis:
```suggestion
		return nil, errors.Wrapf(err, "error parsing %q from local endpoint", subv1.UDPPortBackendConfig)
```

pkg/cable/wireguard/driver.go
```diff
@@ -246,6 +254,13 @@ func (w *wireguard) ConnectToEndpoint(endpointInfo *natdiscovery.NATEndpointInfo
 	klog.V(log.DEBUG).Infof("Adding connection for cluster %s, %v", remoteEndpoint.Spec.ClusterID, connection)
 	w.connections[remoteEndpoint.Spec.ClusterID] = connection
 
+	port, err := remoteEndpoint.Spec.GetBackendPort(v1.UDPPortBackendConfig, int32(w.spec.NATTPort))
+	if err != nil {
+		return "", errors.Wrapf(err, "error reading UDP port")
```
tpantelis: I assume we want the same behavior here as for libreswan.

```suggestion
		klog.Warningf("Error parsing %q from remote endpoint %q - using port %s instead: %v", subv1.UDPPortBackendConfig, remoteEndpoint.Spec.CableName,
			w.spec.NATTPort, err)
```

pkg/apis/submariner.io/v1/endpoint.go
```diff
@@ -0,0 +1,45 @@
...
+func (ep *EndpointSpec) GetBackendPortStr(configName, defaultValue string) (string, error) {
+	if portStr := ep.BackendConfig[configName]; portStr != "" {
+		port, err := parsePort(portStr)
+		if err != nil {
+			return defaultValue, errors.Wrapf(err, "error parsing backend config %s", configName)
+		}
+
+		return strconv.Itoa(int(port)), nil
+	}
+
```
tpantelis: This could call `GetBackendPort` but I would suggest removing this function. libreswan can simply call `GetBackendPort` and convert to string.

pkg/natdiscovery/natdiscovery.go
```diff
@@ -129,7 +140,7 @@ func (nd *natDiscovery) AddEndpoint(endpoint *types.SubmarinerEndpoint) {
 	remoteNAT := newRemoteEndpointNAT(endpoint)
 
 	// support nat discovery disabled or a remote cluster endpoint which still hasn't implemented this protocol
-	if _, err := extractNATDiscoveryPort(endpoint); err == errorNoNatDiscoveryPort || nd.serverPort == nil {
+	if _, err := extractNATDiscoveryPort(endpoint); err == errorNoNatDiscoveryPort || nd.serverPort == 0 {
```
tpantelis: We probably shouldn't ignore a parse error even tho it **should** never occur. 

```suggestion
	if _, err := extractNATDiscoveryPort(endpoint); err != nil || nd.serverPort == 0 {
	    if err != errorNoNatDiscoveryPort {
	        klog.Errorf("Error extracting NATT discovery port from endpoint %q: %v", endpoint.Spec.CableName, err)
	    }
```

pkg/natdiscovery/natdiscovery.go
```diff
@@ -88,14 +94,19 @@ func randomRequestCounter() (uint64, error) {
...
 
-var errorNoNatDiscoveryPort = errors.New("nat discovery port missing in endpoint")
+var errorNoNatDiscoveryPort = errors.New("natt discovery port missing in endpoint")
 
 func extractNATDiscoveryPort(endpoint *types.SubmarinerEndpoint) (int32, error) {
-	if endpoint.Spec.NATDiscoveryPort == nil {
-		return 0, errorNoNatDiscoveryPort
+	natDiscoveryPort, err := endpoint.Spec.GetBackendPort(v1.NATTDiscoveryPortBackendConfig, 0)
+	if err != nil {
+		return natDiscoveryPort, errors.Wrap(err, "error extracting NATT discovery port")
```
tpantelis:
```suggestion
		return natDiscoveryPort, err
```

pkg/util/util.go
```diff
@@ -164,6 +165,25 @@ func GetLocalEndpoint(submSpec types.SubmarinerSpecification, backendConfig map[
...
+func getBackendConfig(nodeObj *v1.Node) (map[string]string, error) {
+	validConfigs := stringset.New(subv1.ValidNodeSpecificBackendConfigs...)
+	backendConfig := map[string]string{}
+
+	for label, value := range nodeObj.Labels {
+		if strings.HasPrefix(label, subv1.BackendConfigPrefix) {
+			config := label[len(subv1.BackendConfigPrefix):]
+			if !validConfigs.Contains(config) {
+				return backendConfig, errors.Errorf("unknown backend config from node %q label %q",
+					nodeObj.Name, label)
```
tpantelis:
```suggestion
				return backendConfig, errors.Errorf("unknown config label %q on node %q",
					nodeObj.Name, label)
```

tpantelis (changes requested):

pkg/apis/submariner.io/v1/types.go
mangelajo: > I think we can omit `Backend` from the names. Frankly I would've preferred `Backend` -> `CableDriver` and `BackendConfig` -> `CableDriverConfig`

yup, agreed.

pkg/apis/submariner.io/v1/endpoint.go
mangelajo: Done

pkg/apis/submariner.io/v1/types.go
mangelajo: Done

pkg/cable/libreswan/libreswan.go
mangelajo: Done.

pkg/cable/wireguard/driver.go
mangelajo: Done.

pkg/cable/wireguard/driver.go
mangelajo: Done.

pkg/natdiscovery/natdiscovery.go
mangelajo: Done.

pkg/util/util.go
mangelajo: Done

tpantelis (approved):

dfarrell07:
Go unit test failure:

```
Summarizing 1 Failure:

[Fail] Util Function GetLocalEndpoint [It] should return a valid SubmarinerEndpoint object 
/go/src/github.com/submariner-io/submariner/pkg/util/util_test.go:125
```

dfarrell07:
> Go unit test failure:
> 
> ```
> Summarizing 1 Failure:
> 
> [Fail] Util Function GetLocalEndpoint [It] should return a valid SubmarinerEndpoint object 
> /go/src/github.com/submariner-io/submariner/pkg/util/util_test.go:125
> ```

Nevermind, this is happening more generally, not just in this PR.

dfarrell07 (approved):

mangelajo:
> > Go unit test failure:
> > ```
> > Summarizing 1 Failure:
> > 
> > [Fail] Util Function GetLocalEndpoint [It] should return a valid SubmarinerEndpoint object 
> > /go/src/github.com/submariner-io/submariner/pkg/util/util_test.go:125
> > ```
> 
> Nevermind, this is happening more generally, not just in this PR.

Let me check it, it could well be related to my change. I'll see if I can fix it here.

PR #1221, opened 2021-03-26
Use indirect build images and dnf roots

skitt:
This reduces the images as follows:

* gateway: 243MiB → 174MiB
* globalnet: 213MiB → 81MiB
* networkplugin-syncer: 255MiB → 189MiB
* route-agent: 212MiB → 80.7MiB

We use minimal dependencies and a separate build environment with a
dedicated installation root to reduce the number of required packages
and remove the support infrastructure from the target images.

See https://www.youtube.com/watch?v=HuE2Lrf5mRg for details.

Depends on https://github.com/submariner-io/shipyard/pull/536

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
-IMAGES ?= submariner-gateway submariner-route-agent submariner-globalnet submariner-networkplugin-syncer
+IMAGES ?= submariner-buildenv submariner-gateway submariner-route-agent submariner-globalnet submariner-networkplugin-syncer
 images: build
 
 include $(SHIPYARD_DIR)/Makefile.inc
--- a/package/Dockerfile.submariner-buildenv
+++ b/package/Dockerfile.submariner-buildenv
@@ -1,0 +1,33 @@
+ARG FEDORA_VERSION=33
+FROM fedora:${FEDORA_VERSION}
+ARG FEDORA_VERSION
+ARG TARGETPLATFORM
+
+COPY package/dnf_install /
+
+# dnf cache and base packages, shared across all build sub-directories
+RUN /dnf_install -k -v ${FEDORA_VERSION} \
+    glibc bash glibc-minimal-langpack coreutils-single
+
+# gateway
+
+# iproute and iptables are used internally
+# libreswan provides IKE
+# procps-ng is needed for libreswan (pidof)
+RUN /dnf_install -v ${FEDORA_VERSION} -r /output/gateway \
+    libcurl-minimal iproute iptables iptables-nft libreswan procps-ng
+
+# globalnet
+
+RUN /dnf_install -v ${FEDORA_VERSION} -r /output/globalnet \
+    iproute iptables iptables-nft ipset
+
+# networkplugin-syncer
+
+RUN /dnf_install -v ${FEDORA_VERSION} -r /output/networkplugin-syncer \
+    https://kojipkgs.fedoraproject.org//packages/ovn/20.06.2/4.fc33/x86_64/ovn-20.06.2-4.fc33.x86_64.rpm
+
+# route-agent
+
+RUN /dnf_install -v ${FEDORA_VERSION} -r /output/route-agent \
+    iproute iptables iptables-nft openvswitch procps-ng
--- a/package/Dockerfile.submariner-gateway
+++ b/package/Dockerfile.submariner-gateway
@@ -1,13 +1,13 @@
-FROM fedora:33
+ARG BASE_BRANCH
+FROM quay.io/submariner/submariner-buildenv:${BASE_BRANCH} AS build-env
 ARG TARGETPLATFORM
 
-WORKDIR /var/submariner
+FROM scratch
+ARG TARGETPLATFORM
 
-# iproute and iptables are used internally
-# libreswan provides IKE
-RUN dnf -y install --nodocs --setopt=install_weak_deps=0 \
-           iproute iptables iptables-nft libreswan && \
-    dnf -y clean all
+COPY --from=build-env /output/gateway /
+
+WORKDIR /var/submariner
 
 COPY package/submariner.sh package/pluto bin/${TARGETPLATFORM}/submariner-gateway /usr/local/bin/
 
--- a/package/Dockerfile.submariner-globalnet
+++ b/package/Dockerfile.submariner-globalnet
@@ -1,11 +1,13 @@
-FROM fedora:33
+ARG BASE_BRANCH
+FROM quay.io/submariner/submariner-buildenv:${BASE_BRANCH} AS build-env
 ARG TARGETPLATFORM
 
-WORKDIR /var/submariner
+FROM scratch
+ARG TARGETPLATFORM
 
-RUN dnf -y install --nodocs --setopt=install_weak_deps=0 \
-           iproute iptables iptables-nft ipset && \
-    dnf -y clean all
+COPY --from=build-env /output/globalnet /
+
+WORKDIR /var/submariner
 
 COPY package/submariner-globalnet.sh bin/${TARGETPLATFORM}/submariner-globalnet /usr/local/bin/
 
--- a/package/Dockerfile.submariner-networkplugin-syncer
+++ b/package/Dockerfile.submariner-networkplugin-syncer
@@ -1,11 +1,13 @@
-FROM fedora:33 
+ARG BASE_BRANCH
+FROM quay.io/submariner/submariner-buildenv:${BASE_BRANCH} AS build-env
 ARG TARGETPLATFORM
 
-WORKDIR /var/submariner
+FROM scratch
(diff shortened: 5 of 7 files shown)


mangelajo (commented):
Can we still login and run bash / ipsec calls after this?

If the answer is no, I propose we wait for 0.10 to start,
If the answer is yes, I'd be fine to merge now

skitt:
> Can we still login and run bash / ipsec calls after this?
> 
> If the answer is no, I propose we wait for 0.10 to start,
> If the answer is yes, I'd be fine to merge now

Yes, we still can:

```
$ docker run -it --rm --entrypoint /bin/bash quay.io/submariner/submariner-gateway:dev
bash-5.0# ipsec
Usage: ipsec {command} [argument] ...>
Use --help for a list of commands, or see the ipsec(8) manual page
Most commands have their own manual pages, e.g. ipsec_auto(8).
See <https://libreswan.org> for more general info.
Linux Libreswan 4.3 (netkey) on 5.10.16-200.fc33.x86_64
bash-5.0#
```

The only feature this patch removes from the images is the ability to use `dnf` inside them.

dfarrell07 (approved):

tpantelis (approved):

skitt:
@mkolesnik I’d appreciate knowing what you think of this!

mangelajo (approved):
Awesome!, thank you @skitt great optimization

skitt:
Please don’t merge this before @mkolesnik gets a chance to review it.

mangelajo:
> Please don’t merge this before @mkolesnik gets a chance to review it.
Ack, I thought it was ok after 4 days , but it's true we had a weekend in the middle.

skitt:
@mkolesnik, any chance you could take a look at this?

sridhargaddam (approved):
Nice optimization.

mkolesnik (changes requested):
Looks nice but AFAIK we'll loose caching (explained here [1]).

Also the base looks mostly the same for all of them.

Perhaps its best to build the base once (and publish it) and each image will rely on that?

That way IIUC we'll get shorter build times and caching will work as expected.

[1] https://pythonspeed.com/articles/faster-multi-stage-builds/

tpantelis:
bump

skitt:
> Perhaps its best to build the base once (and publish it) and each image will rely on that?

I’ve updated the PR to do this.

package/Dockerfile.submariner-buildenv
```diff
@@ -0,0 +1,43 @@
+FROM fedora:33
```
mkolesnik: Shouldn't "33" be an ARG by itself with a default of 33, this way we don't have to replace it in multiple places?
```suggestion
ARG FEDORA_VERSION
FROM fedora:${FEDORA_VERSION}
```

then in Makefile add `FEDORA_VERSION ?= 33`

Dockerfile.dapper
```diff
@@ -1,7 +1,7 @@
 ARG BASE_BRANCH
 FROM quay.io/submariner/shipyard-dapper-base:${BASE_BRANCH}
 
-ENV DAPPER_ENV="REPO MAKEFLAGS TAG QUAY_USERNAME QUAY_PASSWORD GITHUB_REF CLUSTERS_ARGS DEPLOY_ARGS CLEANUP_ARGS IMAGES_ARGS RELEASE_ARGS FOCUS SKIP PLUGIN" \
+ENV DAPPER_ENV="REPO MAKEFLAGS TAG QUAY_USERNAME QUAY_PASSWORD GITHUB_REF CLUSTERS_ARGS DEPLOY_ARGS CLEANUP_ARGS IMAGES_ARGS RELEASE_ARGS FOCUS SKIP PLUGIN BASE_BRANCH" \
```
mkolesnik: You sure this is needed?

`BASE_BRANCH` is defined outside the dapper scope and inside it, so I don't think it's necessary to pass it in the env.

Makefile
```diff
@@ -9,7 +9,7 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
-IMAGES ?= submariner-gateway submariner-route-agent submariner-globalnet submariner-networkplugin-syncer
+IMAGES ?= submariner-buildenv submariner-gateway submariner-route-agent submariner-globalnet submariner-networkplugin-syncer
```
mkolesnik, line 13: Yay more images :)

@mangelajo should be happy :D

package/Dockerfile.submariner-buildenv
```diff
@@ -0,0 +1,43 @@
...
+ARG TARGETPLATFORM
+
+# dnf cache and base packages, shared across all build sub-directories
+RUN dnf -y --setopt=install_weak_deps=0 --nodocs \
+           --installroot /output/base --releasever 33 \
+           install glibc bash glibc-minimal-langpack coreutils-single
```
mkolesnik: This seems very boiler-platey, how about adding a simple script that gets copied in before the run and calling it?

e.g.
```suggestion
# dnf cache and base packages, shared across all build sub-directories
RUN INSTALL_ROOT=base KEEP_CACHE=y BASE_RUN=y \
    /dnf_install.sh glibc bash glibc-minimal-langpack coreutils-single
```

with `dnf_install.sh` being something like:
```bash
#!/usr/bin/env bash
[[ -n $BASE_RUN ]] || cp -a /output/base /output/${INSTALL_ROOT}
dnf -y --setopt=install_weak_deps=0 --nodocs \
    --installroot /output/${INSTALL_ROOT} --releasever ${FEDORA_VERSION} \
    install ${*}
[[ -n $KEEP_CACHE ]] || dnf -y --installroot /output/${INSTALL_ROOT} --releasever  ${FEDORA_VERSION} clean all
```

package/Dockerfile.submariner-buildenv
```diff
@@ -0,0 +1,43 @@
...
+
+# iproute and iptables are used internally
+# libreswan provides IKE
+# procps-ng is needed for sysctl
+RUN cp -a /output/base /output/gateway && \
+    dnf -y --setopt=install_weak_deps=0 --nodocs \
+           --installroot /output/gateway --releasever 33 \
+           install libcurl-minimal \
+                   iproute iptables iptables-nft libreswan procps-ng && \
+    dnf -y --installroot /output/gateway --releasever 33 clean all
```
mkolesnik: With the previous change this could be as simple as:
```suggestion
RUN INSTALL_ROOT=gateway /dnf_install.sh libcurl-minimal iproute iptables iptables-nft libreswan procps-ng
```

package/Dockerfile.submariner-buildenv
```diff
@@ -0,0 +1,43 @@
...
+
+# globalnet
+
+RUN cp -a /output/base /output/globalnet && \
+    dnf -y --setopt=install_weak_deps=0 --nodocs \
+           --installroot /output/globalnet --releasever 33 \
+           install iproute iptables iptables-nft && \
+    dnf -y --installroot /output/globalnet --releasever 33 clean all
```
mkolesnik: With the previous change this could be as simple as:
```suggestion
RUN INSTALL_ROOT=globalnet /dnf_install.sh iproute iptables iptables-nft
```

package/Dockerfile.submariner-buildenv
```diff
@@ -0,0 +1,43 @@
...
+
+# networkplugin-syncer
+
+RUN cp -a /output/base /output/networkplugin-syncer && \
+    dnf -y --setopt=install_weak_deps=0 --nodocs \
+           --installroot /output/networkplugin-syncer --releasever 33 \
+           install https://kojipkgs.fedoraproject.org//packages/ovn/20.06.2/4.fc33/x86_64/ovn-20.06.2-4.fc33.x86_64.rpm && \
+    dnf -y --installroot /output/networkplugin-syncer --releasever 33 clean all
```
mkolesnik: With the previous change this could be as simple as:
```suggestion
RUN INSTALL_ROOT=networkplugin-syncer /dnf_install.sh https://kojipkgs.fedoraproject.org//packages/ovn/20.06.2/4.fc33/x86_64/ovn-20.06.2-4.fc33.x86_64.rpm
```

package/Dockerfile.submariner-buildenv
```diff
@@ -0,0 +1,43 @@
...
+
+# route-agent
+
+RUN cp -a /output/base /output/route-agent && \
+    dnf -y --setopt=install_weak_deps=0 --nodocs \
+           --installroot /output/route-agent --releasever 33 \
+           install iproute iptables iptables-nft openvswitch procps-ng && \
+    dnf -y --installroot /output/route-agent --releasever 33 clean all
```
mkolesnik: With the previous change this could be as simple as:
```suggestion
RUN INSTALL_ROOT=route-agent /dnf_install.sh iproute iptables iptables-nft openvswitch procps-ng
```

mkolesnik (changes requested):
Some ideas

skitt:
This is still relevant.

PR #1222, opened 2021-03-29
Move IsOverlappingCIDR to separate package

sridhargaddam:
IsOverlappingCIDR is used in couple of Submariner components like
route-agent and globalnet. Recently, when it was used in subctl,
we started to notice issues while building subctl on Windows
platforms as the util package was importing iptables as well.
To fix this issue, moving the API to a separate package in
Submariner.

### DIFF
--- a/pkg/cidr/iputil.go
+++ b/pkg/cidr/iputil.go
@@ -1,0 +1,40 @@
+/*
+© 2021 Red Hat, Inc. and others
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cidr
+
+import (
+	"net"
+)
+
+func IsOverlapping(cidrList []string, cidr string) (bool, error) {
+	_, newNet, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return false, err
+	}
+
+	for _, v := range cidrList {
+		_, baseNet, err := net.ParseCIDR(v)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


skitt:
It looks like a file got left out ;-).

mangelajo (changes requested):
You missed adding the new file: "github.com/submariner-io/submariner/pkg/iputils"

mangelajo:
Ah it is there now ;)

mangelajo (approved):

skitt (approved):

sridhargaddam:
One of the jobs failed, I'm looking into the logs.

dfarrell07 (approved):

pkg/iputils/iputil.go
```diff
@@ -0,0 +1,40 @@
+/*
```
tpantelis: I think we should avoid `utils` if we can as is recommended by  effective Go. I suggest just `ip` (or `net` altho that would require import aliasing if also importing Go `net`).

tpantelis (changes requested):
Preventing auto-merge so my previous comment is seen.

pkg/iputils/iputil.go
skitt: How about a more specific package, `cidr`, for CIDR-related functions? I’ve been meaning to rework the CIDR allocation anyway at some point (there’s a nice package somewhere that helps with all that).

tpantelis: yeah  something meaningfully specific w/o `util`. I like `cidr` - then the function  would become `cidr.IsOverlapping(...)`.

tpantelis (changes requested):

pkg/iputils/iputil.go
tpantelis: Perhaps we should also put this in `admiral`.

sridhargaddam: We briefly discussed (over slack) whether to put this API in admiral, but since it's used only by Operator (apart from Submariner) created a different package in the Submariner repo itself.

skitt (approved):
`iputil.go` could be `cidr.go` or `overlapping.go` depending on the intended granularity when adding more functions, but LGTM in general.

tpantelis (approved):

PR #1229, opened 2021-03-31
Add the natdiscovery.pb.go file to the repository

mangelajo:
Otherwise, while building downstream images we have issues
re-generating this file. It's simpler to commit the file
into the repository


SteveMattar (approved):

skitt (approved):

SteveMattar:
We can use this approach instead
https://github.com/submariner-io/submariner/pull/1230

mangelajo:
> We can use this approach instead
> #1230

I think it's simpler to use this, the other code is BSD :D

PR #1230, opened 2021-03-31
build: add the protoc-gen-go cmd to pkg

SteveMattar:
This plugin is missing while building downstream images.

Signed-off-by: Steve Mattar <smattar@redhat.com>

### DIFF
--- a/pkg/protoc-gen-go/main.go
+++ b/pkg/protoc-gen-go/main.go
@@ -1,0 +1,56 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The protoc-gen-go binary is a protoc plugin to generate Go code for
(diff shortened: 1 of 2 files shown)


pkg/protoc-gen-go/main.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
```
mangelajo, line 2: We can't add this code, it's BSD licensed.

I think it's better just to commit the generated .pb.go file as in the other patch.

mangelajo (approved):

mangelajo (changes requested):
oops wrong approve

mangelajo:
Ok, after talking to @SteveMattar we will stick to the other PR where we commit that code into the repo, so it doesn't need to be regenerated downstream.

PR #1236, opened 2021-04-06
Optimize NAT discovery protocol time

mangelajo:
This improves the discovery, and avoids the chances of running into
discovery timeout. When a new gateway is started it will beging
pinging via UDP to all the other endpoints, but it will not be
yet known by the other endpoints until they completely update
via broker (which can be 10-20secs sometimes), and they will
keep responding with unknown SRC endpoint. That's unecessary
for the protocol, really adds no value, only slows down, hence
this commit removes it.


skitt:
```
[Fail] when a remote Endpoint is added and the local Endpoint is not initially known to the remote process [It] should eventually notify with the correct NATEndpointInfo settings 
/go/src/github.com/submariner-io/submariner/pkg/natdiscovery/natdiscovery_test.go:133
```

mangelajo:
> ```
> [Fail] when a remote Endpoint is added and the local Endpoint is not initially known to the remote process [It] should eventually notify with the correct NATEndpointInfo settings 
> /go/src/github.com/submariner-io/submariner/pkg/natdiscovery/natdiscovery_test.go:133
> ```

woops

sridhargaddam (approved):

tpantelis (approved):

PR #1242, opened 2021-04-07
Update K8s default to 1.20, also test 1.17

dfarrell07:
Use the new ability of the shared E2E GHA to configure the Kubernetes
version to add a CI matrix dimension for K8s versions.

Change the default K8s version tested in E2E to be the latest K8s
version, providing good coverage for minimal additional jobs.

Test 1.17 as it is the version we currently build with, and doing so
provides default-E2E coverage for the oldest version we support.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,11 +16,15 @@ jobs:
         globalnet: ['', 'globalnet']
         cable_driver: ['libreswan', 'wireguard']
         ovn: ['', 'ovn']
+        k8s_version: ['1.20.2']
         exclude:
           - ovn: 'ovn'
             globalnet: 'globalnet'
           - ovn: 'ovn'
             cable_driver: 'wireguard'
+        include:
+          # This is the oldest K8s version we try to support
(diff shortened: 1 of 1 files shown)


dfarrell07:
We could easily add 1.17 or 1.18 if we want that coverage, but given the push-back to adding more jobs I wanted to propose the absolute minimum.

tpantelis:
> We could easily add 1.17 or 1.18 if we want that coverage, but given the push-back to adding more jobs I wanted to propose the absolute minimum.

As noted in the other PR, perhaps we should test against 1.17  instead of 1.19 since we build against it. So full coverage for 1.20, sanity check for 1.17 and skip in between.

dfarrell07:
I think we should try to have gating 1.19 coverage eventually, as it's a version we want to be sure to support as I understand it. I'll work on updating from 1.17 elsewhere, then we can change this 1.17 job to a 1.19 job.

mangelajo (approved):
Thanks a lot Daniel 🙏🏻 🙏🏻

tpantelis (approved):

PR #1246, opened 2021-04-08
Make the API additions in 0.9 optional to fix upgrades

mangelajo:
Otherwise, when we delete the Submariner operator object
from subctl it will never finish waiting on the orphan objects,
being Gateway an orphan object that cannot be deleted
because API checking fails with:
```
   "submariner" is invalid: [
    status.gateways.connections.usingIP: Required value,
    status.gateways.connections.usingNAT: Required value]
```


tpantelis (approved):

skitt (approved):

mangelajo:
wow all red

skitt:
The Upgrade test fails because `[external-dataplane]` is interpreted as a regex and is invalid (`l-d` isn’t a valid sequence).

PR #1247, opened 2021-04-08
Handle retries on install cable failure

### LINKED ISSUE #1243: Handle retries on install cable failure
**What happened**:

With the new NAT discovery process, the pipeline from the Tunnel Controller -> Cable Engine -> Cable Driver to install a cable is now asynchronous. Previously a failure was reported back to the Tunnel Controller which scheduled a retry. That is currently no longer the case.

Retries should still occur.

tpantelis:
Fixes #1243

The simple solution implemented is to set a resync interval in the tunnel controller. 30 sec was chosen as a reasonable interval as we don't expect failures to be common. This also means that every endpoint will be resynced every 30 sec but it should essentially be a no-op if the endpoint was already successfully installed. Prior to the change to use the resource
watcher, we had a resync interval set anyway (60 sec).

Modified `natDiscovery.AddEndpoint` to notify the ready channel if discovery had already completed. This enables retries to immediately  proceed again to the install cable phase.

The tunnel controller tests were changed to use the real cable engine and NAT discovery to more effectively test the retries round trip.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/coreos/go-iptables v0.5.0
 	github.com/ebay/go-ovn v0.1.1-0.20201007164241-da67e9744ec0
 	github.com/go-ping/ping v0.0.0-20201022122018-3977ed72668a
-	github.com/golang/protobuf v1.4.3 // indirect
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/onsi/ginkgo v1.16.0
 	github.com/onsi/gomega v1.11.0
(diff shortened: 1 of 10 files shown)


skitt (approved):

mangelajo (approved):
It looks good, @tpantelis thanks for handling this, the idea I had in mind was more complicated, so this is great.

I was looking at how libreswan & wireguard take re-installs , the retries should be good for both. 

I found that wireguard does not take well IP changes, because they override the GetActiveConnections logic of the cable engine to implement their own. I believe we will need to do work in there to allow IP switching, but unrelated to this patch.

PR #1249, opened 2021-04-09
Separate the e2e external net testing to it's own directory

mangelajo:
In this way we avoid having to deal around with the skip/focus
flags, and we have a single parameter to differentiate.

Depends on https://github.com/submariner-io/shipyard/pull/521


mangelajo:
@mkimuram  ^

.github/workflows/e2e-external.yml
```diff
@@ -17,5 +17,4 @@ jobs:
         uses: submariner-io/shipyard/gh-actions/e2e@devel
         with:
           plugin: '/go/src/github.com/submariner-io/submariner/scripts/e2e/external/hook'
-          focus: '\[external-dataplane\]'
-          skip: '\[dataplane\]'
+          testdir: 'test/e2e_external'
```
mkimuram: If `testdir` is not under `test/e2e`, we will need to add the directory to `UNIT_TEST_ARGS` in [Makefile](https://github.com/submariner-io/submariner/blob/devel/Makefile#L30) to make unit tests pass. So, easy fix will be just to move `test/e2e_external/e2e_external_test.go` to `test/e2e/external/e2e_external_test.go`, to make it under 'test/e2e' (It should also solve the linting error).

mkimuram (commented):
@mangelajo 

Thank you for your improvement.

I like this idea, for it will allow making E2E tests independent.
And focus/skip should be still useful when running specific tests manually or running tests which require more complicated conditions in CI.

.github/workflows/e2e-external.yml
tpantelis: Yeah I'd prefer `test/e2e/external` rather than `test/e2e_external` if that works.

mangelajo: ok, that would make sense, let's move the main go file under tests/e2e/external

I agree that skip / focus are still useful for manual testing, I missed that many times ':D

mangelajo: Oh wait, that doesn't work, if we give ginkgo the test/e2e base, it will explore and grab all the _test.go files including this one which will trigger it to be ran. 
I'll add to the makefile

mkimuram: In that case, we will still need to rename `e2e_external` to something like `e2eexternal` to avoid linting error. It doesn't allow to use '_'. '-', and so on.

mangelajo: that env var should be EXCLUDED_UNIT_TEST_DIRS or something, it's a bit confusing ':D

mangelajo:
Hopping that it will be better now ':D

.github/workflows/e2e-external.yml
```diff
@@ -17,5 +17,4 @@ jobs:
         uses: submariner-io/shipyard/gh-actions/e2e@devel
         with:
           plugin: '/go/src/github.com/submariner-io/submariner/scripts/e2e/external/hook'
-          focus: '\[external-dataplane\]'
-          skip: '\[dataplane\]'
+          testdir: 'test/externale2e'
```
tpantelis: How about `external/e2e` or just `external`?

mangelajo: external would be fine with me.

tpantelis (approved):

dfarrell07 (approved):

sridhargaddam (approved):

PR #1255, opened 2021-04-13
Refactor installCableWithNATInfo

tpantelis:
...to check timestamp regardless if the cable name differs.

### DIFF
--- a/pkg/cableengine/cableengine.go
+++ b/pkg/cableengine/cableengine.go
@@ -145,37 +145,41 @@ func (i *engine) installCableWithNATInfo(rnat *natdiscovery.NATEndpointInfo) err
 	for _, active := range activeConnections {
 		klog.V(log.TRACE).Infof("Analyzing currently active connection %q", active.Endpoint.CableName)
 
-		disconnect := false
-
-		if active.Endpoint.CableName == endpoint.Spec.CableName {
-			// There could be scenarios where the cableName would be the same but the
-			// PublicIP of the active GatewayNode changes.
-			if active.UsingIP == rnat.UseIP && active.UsingNAT == rnat.UseNAT {
-				klog.V(log.DEBUG).Infof("Cable %q is already installed - not installing again", active.Endpoint.CableName)
-				return nil
-			} else {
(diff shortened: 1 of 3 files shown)


pkg/cableengine/cableengine.go
```diff
@@ -145,37 +145,35 @@ func (i *engine) installCableWithNATInfo(rnat *natdiscovery.NATEndpointInfo) err
...
+		prevTimestamp := i.installedCables[active.Endpoint.CableName]
+		if endpoint.CreationTimestamp.Before(&prevTimestamp) {
+			klog.Warningf("The timestamp (%s) for new cable %q is older than the timestamp (%s) of the pre-existing "+
+				"cable %q - not replacing", endpoint.CreationTimestamp, endpoint.Spec.CableName, prevTimestamp, active.Endpoint.CableName)
+			return nil
+		}
+
+		if endpoint.CreationTimestamp.Equal(&prevTimestamp) && active.Endpoint.CableName == endpoint.Spec.CableName {
+			// There could be scenarios where the cableName would be the same but the PublicIP of the active GatewayNode changes.
 			if active.UsingIP == rnat.UseIP && active.UsingNAT == rnat.UseNAT {
```
mangelajo: Should we add also the compare for the BackendConfig? , I guess if it's going to be re-created it actually doesn't matter. As long as we don't change that behavior in admiral. But if we check the backend config it would allow us to change that behavior in the future if we wanted I guess.

Anyway, just a thought.

mangelajo (approved):
LGTM, I left you one comment.

pkg/cableengine/cableengine.go
tpantelis: yeah I was going to do that in a separate PR but I can do it with this one.

tpantelis: Done

mangelajo:
@sridhargaddam @Oats87 this one fixes some bugs, including one I just found, that updating a gateway UDP port won't get updated on the other gateways (Because the update of the BackedConfig is received, but ignored, we only check IP or nat status)

Oats87 (approved):
Looks sane to me

mangelajo:
Thanks @Oats87

PR #1256, opened 2021-04-13
preferred server timestamp

mangelajo:
inserted in the BackendConfig, forcing the resynchronization of the
object to the broker, which will indicate all clients that the
server has been restarted, and that they must re-connect to the endpoint.

This timestamp changes everytime a submariner-gateway is started,
otherwise a server will restart, and all the client connections will
remain unconnected because pluto doesn't know about them anymore.

### DIFF
--- a/pkg/endpoint/local_endpoint.go
+++ b/pkg/endpoint/local_endpoint.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


tpantelis:
The unit test failure is unrelated - random HTTP failure from _ipify.GetIp()_.

tpantelis (approved):

sridhargaddam (approved):
LGTM, thanks @mangelajo

PR #1257, opened 2021-04-13
Handle DST NAT detection in pkg/natdiscovery

mangelajo:
This patch implements destination NATT detection, which was
previously ignored. Leading on un-balanced NATT discovery
from the sides when one side has NAT and the other does not.

Extra details about the discovery are implemented in the protocol
dst_ip/src_ip/src_port flags which will help debugging connections
and could be helpful to cable drivers in the future as well.


pkg/natdiscovery/proto/natdiscovery.proto
```diff
@@ -54,9 +54,13 @@ message SubmarinerNatDiscoveryResponse {
   EndpointDetails sender = 3;
   EndpointDetails receiver = 4;
 
+  bool src_ip_nat_detected = 5;
```
mangelajo, line 57: I tried src_IP_NAT_detected , etc...
but the protobuf compiler generated very weird things on the go side... :/

pkg/natdiscovery/request_handle.go
```diff
@@ -88,7 +86,18 @@ func (nd *natDiscovery) handleRequestFromAddress(req *proto.SubmarinerNatDiscove
...
+	// if we will listen at some point on multiple addresses we will need to implement the
+	// unix.IP_RECVORIGDSTADDR on the UDP socket, and the go recvmsg implementation instead of readfrom
+	if req.UsingDst.IP != nd.localEndpoint.Spec.PrivateIP {
+		response.DstIpNatDetected = true
+	}
+
+	if response.SrcPortNatDetected || response.SrcIpNatDetected || response.DstIpNatDetected {
+		response.Response = proto.ResponseType_NAT_DETECTED
 
 		return nd.sendResponseToAddress(&response, addr)
```
mangelajo: I can remove this line nowm if I add an else branch for L105

tpantelis (approved):

sridhargaddam (approved):

PR #1254, opened 2021-04-12
Compare Endpoint timestamps when deciding to replace previous cable

tpantelis:
An `Endpoint` can become stale if a delete event is somehow missed or delayed. To protect against a stale `Endpoint` replacing a current one, cable installation now compares the `Endpoint` creation timestamps, favoring the newer one.

This also entailed refactoring `InstallCable` to take an `Endpoint` in order to access the `CreationTimestamp`.  `RemoveCable` was also changed for consistency.

Related to #1244

PR #1259, opened 2021-04-14
Revert "Compare Endpoint timestamps when deciding to replace previous cable"

mangelajo:
Reverts submariner-io/submariner#1254

Checking if this has something to do with the globalnet failures, although It seems completely unrelated in terms of changes, we have a suspiction that it could potentially be.


mangelajo:
Ok @tpantelis @sridhargaddam even reverting those patches I still see the failure:

https://github.com/submariner-io/submariner/pull/1259/checks?check_run_id=2341181102#step:3:2188

I wonder why this doesn't happen in the operator tree, thanks @sridhargaddam for investigating.

PR #1265, opened 2021-04-14
Modify iptables commands to wait for few seconds

sridhargaddam:
Sometimes iptables may not get xtables lock, so wait
for few seconds before giving up.


sridhargaddam:
@mangelajo @tpantelis As discussed with you during the call, there seems to be multiple issues with K8s 1.20.2
This PR fixes one of the problems I've seen locally on my PC where iptables fails to delete a rule as it could not get the lock.

mangelajo (approved):

tpantelis (approved):

PR #1300, opened 2021-04-30
Bump dependencies

skitt:
This upgrades a number of dependencies, in particular some identified
by Snyk as containing vulnerabilities:

* go-iptables v0.6.0
* go-ping
* golang/x/crypt
* golang/x/net
* golang/x/sys
* golang/x/text
* wgctrl
* Protobuf v1.26.0

crypt and text are transitive dependencies, so this pins them using
replace.

The MCS API dependency is updated to use its version tag (the commit
is identical).


dfarrell07 (approved):
Emphatic +1 overall, and I wonder if we should try to include this in 0.9. I see we're both watching the E2E failures and retriggering.

skitt:
> Emphatic +1 overall, and I wonder if we should try to include this in 0.9. I see we're both watching the E2E failures and retriggering.

Too late for 0.9 ;-).

tpantelis (approved):

skitt:
Backported as #1323.

PR #1307, opened 2021-05-04
Fix intermittent data race failure in pinger

tpantelis:
The "_when the pinger timeout expires_" test can fail due to a data race while accessing the pingTimeout global as seen [here](https://github.com/submariner-io/submariner/runs/2502291974?check_suite_focus=true). It updates `pingTimeout` in `BeforeEach` but the goroutine that reads it may still be running from the prior test. Store the global value in a `pingerInfo` field so the global var isn't accessed by multiple goroutines.


skitt (approved):

sridhargaddam (approved):

nyechiel (approved):

tpantelis:
Backport to https://github.com/submariner-io/submariner/pull/1316

PR #1324, opened 2021-05-12
Run full E2E matrix only when PR approved

### LINKED ISSUE #1289: Run full E2E matrix only when "ready-for-testing"
E2E test jobs take a long time to complete while the CI has a limited capacity to run concurrent jobs (currently 20 jobs across the entire *@submariner-io* organization.

As such, it would be better to run a default `make e2e` smoke test on each PR sync, and the full test matrix that we have today only when a "ready-for-testing" label is added to the PR.

In order to prevent accidental merging of unlabelled PRs, we can have the PR labelled automatically once it has approval(s).
One possible way to do it is: https://github.com/marketplace/actions/label-approved-pull-requests

**Why is this needed**:
To have faster feedback from small jobs, while still maintaining coverage of E2E.

mkolesnik:
There's no need to run the full E2E test matrix before a PR is approved
for testing, so adding a default test for most of the time and changing
the matrix job to run only when the `ready-to-test` label is present.

The label will be applied once 2 approvals are given on the PR, but can
also be added (or removed) manually as needed.

Resolves: #1289 

### DIFF
--- a/.github/workflows/e2e-full.yml
+++ b/.github/workflows/e2e-full.yml
@@ -1,0 +1,42 @@
+---
+name: End to End Full
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, reopened]
+
+jobs:
+  e2e:
+    name: E2E
+    if: contains(github.event.pull_request.labels.*.name, 'ready-to-test')
+    timeout-minutes: 45
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


mangelajo:
Sound like a good plan

dfarrell07 (approved):

mangelajo (approved):

tpantelis (approved):

PR #1338, opened 2021-05-17
Add GlobalIngressIP CRD and related generated code

### LINKED ISSUE #1337: Create GlobalIngressIP CRDs and associated objects
To support https://github.com/submariner-io/enhancements/pull/22/files we need a new `GlobalIngressIP` to allocate IngressIP for pods and services. This issue will track the required CRD changes including creation of clientset, informers and listers

vthapar:
Fixes #1337


pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,64 @@ type ClusterGlobalEgressIPList struct {
...
+	// Spec is the specification of desired behavior of GlobalIngressIP object.
+	Spec GlobalIngressIPSpec `json:"spec"`
+
+	// Observed status of GlobalIngressIP. Its a read-only field.
+	// +optional
+	Status GlobalIngressIPStatus `json:"status,omitempty"`
+}
+
+type GlobalIngressIPSpec struct {
+	// TargetType can be ClusterIPService or a Pod
```
sridhargaddam:
```suggestion
	// TargetType can be ClusterIPService or a HeadlessServicePod
```

sridhargaddam (commented):
A small nit @vthapar. LGTM otherwise. Thanks.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,64 @@ type ClusterGlobalEgressIPList struct {
...
+
+type GlobalIngressIPStatus struct {
+	// +optional
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// The GlobalIP assigned to the respective object.
```
```suggestion
	// The GlobalIP allocated to this object.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,64 @@ type ClusterGlobalEgressIPList struct {
...
+	// name of the corresponding Service.
+	// +Optional
+	ServiceName string `json:"serviceName,omitempty"`
+
+	// Since the Name of GlobalIngressIP will be prepended with "svc-" or "pod-" based on
+	// the object, we may have to truncate the name in some cases when the length exceeds the
+	// maximum length that is allowed by K8s. In such cases, the objRef will include the
+	// full name of the object.
+	// +Optional
+	ObjRef *corev1.LocalObjectReference `json:"objRef,omitempty"`
```
tpantelis: I think we should make this mandatory (ie non-optional) to simplify.

```suggestion
	// The reference to the Target object. 
	TargetRef corev1.LocalObjectReference `json:"targetRef,omitempty"`
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,64 @@ type ClusterGlobalEgressIPList struct {
...
+	// +optional
+	Status GlobalIngressIPStatus `json:"status,omitempty"`
+}
+
+type GlobalIngressIPSpec struct {
+	// TargetType can be ClusterIPService or a Pod
+	Target TargetType `json:"target"`
+
+	// If the TargetType points to a Pod that belongs to a Headless Service, it uses the
+	// name of the corresponding Service.
```
```suggestion
	// The name of the corresponding service if the Target references a headless service pod.
```

pkg/apis/submariner.io/v1/types.go
tpantelis:
```suggestion
	// The type of the object targeted by this GlobalIngressIP.
```

Another option in lieu of this field is to use `TypedLocalObjectReference` defined as:

```
type TypedLocalObjectReference struct {
	// APIGroup is the group for the resource being referenced.
	// If APIGroup is not specified, the specified Kind must be in the core API group.
	// For any other third-party types, APIGroup is required.
	// +optional
	APIGroup *string `json:"apiGroup" protobuf:"bytes,1,opt,name=apiGroup"`
	// Kind is the type of resource being referenced
	Kind string `json:"kind" protobuf:"bytes,2,opt,name=kind"`
	// Name is the name of resource being referenced
	Name string `json:"name" protobuf:"bytes,3,opt,name=name"`
}
```

tpantelis (changes requested):

pkg/apis/submariner.io/v1/types.go
sridhargaddam: omitempty can be removed then?

tpantelis: yeah we might as well remove it - it doesn't make sense for a non-pointer struct as it has no empty value.

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,64 @@ type ClusterGlobalEgressIPList struct {
...
+}
+
+type GlobalIngressIPSpec struct {
+	// TargetType can be ClusterIPService or a HeadlessServicePod
+	Target TargetType `json:"target"`
+
+	// If the TargetType points to a Pod that belongs to a Headless Service, it uses the
+	// name of the corresponding Service.
+	// +Optional
+	ServiceName string `json:"serviceName,omitempty"`
```
sridhargaddam: As discussed during the meeting, lets use the ServiceName as LocalObjectReference and store the ClusterIPServiceName or HeadlessServiceName all the time. Both ServiceRef and TargetRef will be an optional field and will be populated accordingly.

sridhargaddam: Type will be a pointer *corev1.LocalObjectReference `json:"objRef,omitempty"

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,61 @@ type ClusterGlobalEgressIPList struct {
...
+	Spec GlobalIngressIPSpec `json:"spec"`
+
+	// Observed status of GlobalIngressIP. Its a read-only field.
+	// +optional
+	Status GlobalIngressIPStatus `json:"status,omitempty"`
+}
+
+type GlobalIngressIPSpec struct {
+	// TargetType can be ClusterIPService or a HeadlessServicePod
+	Target TargetType `json:"target"`
```
tpantelis, line 293: I thought we concluded we don't need this field.

tpantelis (changes requested):

sridhargaddam (approved):

pkg/apis/submariner.io/v1/types.go
sridhargaddam: I'm afraid no. We would need it as it allows us to expand the object to support regular Pods in the future.

tpantelis: A headless service pod would be the case where both `ServiceRef` and `PodRef` are non-nil. A regular pod would have `ServiceRef` nil. I thought we said that earlier.

vthapar: It is better to have a type field than infer from presence/absence of others. Maybe I missed the bit about removing this during the call. @sridhargaddam thoughts?

sridhargaddam: I share the same opinion as @vthapar.

tpantelis: ok

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,61 @@ type ClusterGlobalEgressIPList struct {
...
+	// Spec is the specification of desired behavior of GlobalIngressIP object.
+	Spec GlobalIngressIPSpec `json:"spec"`
+
+	// Observed status of GlobalIngressIP. Its a read-only field.
+	// +optional
+	Status GlobalIngressIPStatus `json:"status,omitempty"`
+}
+
+type GlobalIngressIPSpec struct {
+	// TargetType can be ClusterIPService or a HeadlessServicePod
```
```suggestion
	// Specifies the type of the entity targeted by this object.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,61 @@ type ClusterGlobalEgressIPList struct {
...
+	Target TargetType `json:"target"`
+
+	// Reference to corresponding Service for this object.
```
```suggestion
	// The reference to a targeted Service, if applicable.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -271,3 +272,61 @@ type ClusterGlobalEgressIPList struct {
...
+	// +Optional
+	ServiceRef *corev1.LocalObjectReference `json:"serviceRef,omitempty"`
+
+	// Reference to corresponding Pod for this object. If TargetType
+	// is ClusterIPService, this will be Nil
```
```suggestion
	// The reference to a targeted Pod, if applicable.
```

tpantelis (changes requested):

tpantelis (approved):

PR #1339, opened 2021-05-17
Include ipset binary in globalnet image

sridhargaddam:
Globalnet V2.0 internally uses ipsets to support inter-cluster
communication. This PR includes the necessary binary in the
Globalnet image.


aswinsuryan (approved):

skitt (approved):
I’m going to guess that Globalnet v2 needs IP sets, and this is why we want to have these tools, but it would be nice if the commit message explained that ;-).

PR #1340, opened 2021-05-17
Create GN2.0 iptable chains in appropriate order

sridhargaddam:
This PR creates the necessary iptable chains in a specific order
and subsequent PRs will populate the iptable rules in the appropriate
chains.

Sample output of IPTables chains:
```
Chain SUBMARINER-GN-EGRESS (1 references)
num   pkts bytes target     prot opt in     out     source               destination         
1      950 59367 SUBMARINER-GN-MARK  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
2      950 59367 SM-GN-EGRESS-PODS  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
3      950 59367 SM-GN-EGRESS-HDLS-PODS  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
4      950 59367 SM-GN-EGRESS-NS  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
5      950 59367 SM-GN-EGRESS-CLUSTER  all  --  *      *       0.0.0.0/0            0.0.0.0/0 
```

### DIFF
--- a/pkg/globalnet/controllers/ipam/globalnet.go
+++ b/pkg/globalnet/controllers/ipam/globalnet.go
@@ -66,6 +66,50 @@ func (i *Controller) initIPTableChains() error {
 		klog.Errorf("error inserting iptables rule %q: %v\n", strings.Join(forwardToSubGlobalNetChain, " "), err)
 	}
 
+	klog.V(log.DEBUG).Infof("Install/ensure %s chain exists", constants.SmGlobalnetEgressChainForPods)
+
+	if err := util.CreateChainIfNotExists(i.ipt, "nat", constants.SmGlobalnetEgressChainForPods); err != nil {
+		return fmt.Errorf("error creating iptables chain %s: %v", constants.SmGlobalnetEgressChainForPods, err)
+	}
+
+	klog.V(log.DEBUG).Infof("Install/ensure %s chain exists", constants.SmGlobalnetEgressChainForHeadlessSvcPods)
+
+	if err := util.CreateChainIfNotExists(i.ipt, "nat", constants.SmGlobalnetEgressChainForHeadlessSvcPods); err != nil {
+		return fmt.Errorf("error creating iptables chain %s: %v", constants.SmGlobalnetEgressChainForHeadlessSvcPods, err)
+	}
+
+	klog.V(log.DEBUG).Infof("Install/ensure %s chain exists", constants.SmGlobalnetEgressChainForNamespace)
+
+	if err := util.CreateChainIfNotExists(i.ipt, "nat", constants.SmGlobalnetEgressChainForNamespace); err != nil {
+		return fmt.Errorf("error creating iptables chain %s: %v", constants.SmGlobalnetEgressChainForNamespace, err)
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/globalnet/controllers/ipam/globalnet.go
```diff
@@ -66,6 +66,50 @@ func (i *Controller) initIPTableChains() error {
 		klog.Errorf("error inserting iptables rule %q: %v\n", strings.Join(forwardToSubGlobalNetChain, " "), err)
 	}
 
+	klog.V(log.DEBUG).Infof("Install/ensure %s chain exists", constants.SmGlobalnetEgressChainForPods)
```
vthapar, line 69: Won't it be better to combine log msg with the util method? If we don't want to always log, maybe something like `CreateChainIfNotExistsWithLog` ?

pkg/routeagent_driver/constants/constants.go
```diff
@@ -20,6 +20,12 @@ const (
 	SmGlobalnetEgressChain  = "SUBMARINER-GN-EGRESS"
 	SmGlobalnetMarkChain    = "SUBMARINER-GN-MARK"
 
+	// The following chains are added as part of GN 2.0 implementation
+	SmGlobalnetEgressChainForPods         = "SM-GN-EGRESS-PODS"
+	SmGlobalnetEgressChainHeadlessSvcPods = "SM-GN-EGRESS-HDLS-PODS"
```
vthapar:
```suggestion
	SmGlobalnetEgressChainForHeadlessSvcPods = "SM-GN-EGRESS-HDLS-PODS"
```

pkg/routeagent_driver/constants/constants.go
```diff
@@ -20,6 +20,12 @@ const (
...
+	SmGlobalnetEgressChainForNamespace    = "SM-GN-EGRESS-NS"
+	SmGlobalnetEgressChainCluster         = "SM-GN-EGRESS-CLUSTER"
```
vthapar:
```suggestion
	SmGlobalnetEgressChainForCluster         = "SM-GN-EGRESS-CLUSTER"
```

pkg/routeagent_driver/constants/constants.go
vthapar: Just to be consistent in naming

vthapar (changes requested):

pkg/globalnet/controllers/ipam/globalnet.go
sridhargaddam: These logs are printed only once when the Globalnet Pod comes up and will help us while debugging any issues at least until we fully transition to Globalnet 2.0 and it becomes stable.

pkg/routeagent_driver/constants/constants.go
sridhargaddam: It was becoming tooo long and it's just an internal constant. But anyways, I can change.

tpantelis (approved):

vthapar (approved):

Issue #1310, opened 2021-05-05
Verify if ipsec-preferred server will work with a network LoadBalancer on AWS


mangelajo:
**What would you like to be added**:

Rebase this patch: https://github.com/submariner-io/submariner/pull/1146

And test with the preferred-server mode for IPSEC.

**Why is this needed**:

This would allow us to save a dedicated worker node just for submariner in the case of openshift deployments.

mangelajo:
It was verified and working

PR #1344, opened 2021-05-19
Implement multiple public IP resolvers

### LINKED ISSUE #1317: If ipify.org is down, submariner-gateway is unable to start
**What happened**:

We use `ipify.org` API to discover the gateway public IP, but if this service is not responding the submariner-gateway will not start.

I would expect additional services resolving the external IP if this one fails,
if all fail, use a fallback to the last known PublicIP

We could provide 'publicip.submariner.io' also as a first resolver, this would allow us to collect anonymous statistics on submariner usage (may be country + version, or just version). But that's a separate topic.

**Environment**:
- Submariner version (use `subctl version`): 0.10.0 & devel

mangelajo:
The gateway nodes now can be labeled as:
```
gateway.submariner.io/public-ip=<resolver>[,resolver..]
```

where resolvers take the form of `<method>:<parameter>`

the implemented methods are:

- api
- lb
- ipv4
- dns

Fixes-Issue: #1317
Related-Issue: #1310
Related-Issue: #1071 

Adds multiple API fallbacks:
    
    In addition to https://api.ipify.org
    - https://api.my-ip.io/ip
    - https://ip4.seeip.org

### DIFF
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -87,10 +87,19 @@ const (
 	GatewayConfigLabelPrefix = "gateway.submariner.io/"
 	UDPPortConfig            = "udp-port"
 	NATTDiscoveryPortConfig  = "natt-discovery-port"
+	PublicIP                 = "public-ip"
 )
 
 const (
 	DefaultNATTDiscoveryPort = "4490"
+)
+
+// Valid PublicIP resolvers.
+const (
+	IPv4         = "ipv4" // ipv4:1.2.3.4
+	LoadBalancer = "lb"   // lb:external-gw-lb
+	API          = "api"  // api:api.ipify.org
+	DNS          = "dns"  // dns:mygateway.dns.name.com
 )
 
 // BackendConfig entries which aren't configured via labels, but exposed from the endpoints
@@ -102,6 +111,7 @@ const (
 var ValidGatewayNodeConfig = []string{
 	UDPPortConfig,
 	NATTDiscoveryPortConfig,
+	PublicIP,
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/endpoint/local_endpoint_test.go
+++ b/pkg/endpoint/local_endpoint_test.go
@@ -18,20 +18,30 @@ limitations under the License.
 package endpoint_test
 
 import (
+	"os"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	endpoint "github.com/submariner-io/submariner/pkg/endpoint"
+
+	"github.com/submariner-io/submariner/pkg/util"
 	v1 "k8s.io/api/core/v1"
 	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
 
-	"github.com/submariner-io/submariner/pkg/endpoint"
 	"github.com/submariner-io/submariner/pkg/types"
 )
 
+const testNodeName = "this-node"
+
 var _ = Describe("GetLocal", func() {
 	var submSpec types.SubmarinerSpecification
+	var client kubernetes.Interface
+	var testPrivateIP = util.GetLocalIP()
 	var node *v1.Node
+
 	const (
-		testPrivateIP       = "127.0.0.1"
 		testUDPPort         = "1111"
 		testUDPPortLabel    = "udp-port"
 		testNATTPortLabel   = "natt-discovery-port"
@@ -49,14 +59,19 @@ var _ = Describe("GetLocal", func() {
 
 		node = &v1.Node{
 			ObjectMeta: v1meta.ObjectMeta{
+				Name: testNodeName,
 				Labels: map[string]string{
 					backendConfigPrefix + testNATTPortLabel: "1234",
 					backendConfigPrefix + testUDPPortLabel:  testUDPPort,
 				}}}
+
+		client = fake.NewSimpleClientset(node)
+
+		os.Setenv("NODE_NAME", testNodeName)
 	})
 
 	It("should return a valid SubmarinerEndpoint object", func() {
-		endpoint, err := endpoint.GetLocal(submSpec, testPrivateIP, node)
+		endpoint, err := endpoint.GetLocal(submSpec, client)
 
 		Expect(err).ToNot(HaveOccurred())
 		Expect(endpoint.Spec.ClusterID).To(Equal("east"))
@@ -72,7 +87,7 @@ var _ = Describe("GetLocal", func() {
 	When("no NAT discovery port label is set on the node", func() {
 		It("should return a valid SubmarinerEndpoint object", func() {
 			delete(node.Labels, testNATTPortLabel)
-			_, err := endpoint.GetLocal(submSpec, testPrivateIP, node)
+			_, err := endpoint.GetLocal(submSpec, client)
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
--- a/pkg/endpoint/public_ip.go
+++ b/pkg/endpoint/public_ip.go
@@ -1,0 +1,149 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endpoint
+
+import (
+	"context"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/pkg/errors"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+)
+
+type publicIPResolverFunction func(clientset kubernetes.Interface, namespace, value string) (string, error)
+
+var publicIPMethods = map[string]publicIPResolverFunction{
+	v1.API:          publicAPI,
+	v1.IPv4:         publicIP,
+	v1.LoadBalancer: publicLoadBalancerIP,
+	v1.DNS:          publicDNSIP,
+}
+
+var IPv4RE = regexp.MustCompile(`(?:\d{1,3}\.){3}\d{1,3}`)
+
+func getPublicIP(submSpec types.SubmarinerSpecification, k8sClient kubernetes.Interface, backendConfig map[string]string) (string, error) {
+	config, ok := backendConfig[v1.PublicIP]
+	if !ok {
+		config = "api:api.ipify.org,api:api.my-ip.io/ip,api:ip4.seeip.org"
+	}
+
+	resolvers := strings.Split(config, ",")
+	for _, resolver := range resolvers {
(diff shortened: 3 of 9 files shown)


pkg/endpoint/public_ip_test.go
```diff
@@ -0,0 +1,116 @@
...
+		const jsonIP = "{\"ip\": \"" + testIP + "\"}"
+
+		It("should return the IP", func() {
+			ip, err := firstIPv4InString(jsonIP)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ip).To(Equal(testIP))
+		})
+	})
+
+	When("the content hasn't a IPv4", func() {
```
skitt:
```suggestion
	When("the content doesn't have an IPv4", func() {
```

skitt (changes requested):
Minor grammar issue; more importantly, since this means we no longer need `go-ipify`, please run `go mod tidy` to remove it from `go.mod` (and `go.sum`).

pkg/endpoint/local_endpoint_suite_test.go
```diff
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package endpoint_test
+package endpoint
```
tpantelis: This doesn't need to be changed.

pkg/endpoint/local_endpoint_test.go
```diff
@@ -13,23 +13,32 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package endpoint_test
+package endpoint
```
tpantelis: ...nor this.

pkg/endpoint/public_ip.go
```diff
@@ -0,0 +1,131 @@
...
+	if !ok {
+		config = "api:api.ipify.org"
+	}
+
+	resolvers := strings.Split(config, ",")
+	for _, resolver := range resolvers {
+		resolver = strings.Trim(resolver, " ")
+		parts := strings.Split(resolver, ":")
+		if len(parts) != 2 {
+			return "", errors.Errorf("bad format for %q", config)
```
```suggestion
			return "", errors.Errorf("invalid format for %q label: %q", v1.PublicIP, config)
```

pkg/endpoint/public_ip.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		method, ok := publicIPMethods[parts[0]]
+		if !ok {
+			return "", errors.Errorf("unknown method %q in %q", parts[0], config)
```
```suggestion
			return "", errors.Errorf("unknown resolver %q in %q label: %q", parts[0], v1.PublicIP, config)
```

pkg/endpoint/public_ip.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func publicLoadBalancerIP(clientset kubernetes.Interface, namespace, loadBalancerName string) (string, error) {
+	service, err := clientset.CoreV1().Services(namespace).Get(context.TODO(), loadBalancerName, metav1.GetOptions{})
+	if err != nil {
+		return "", errors.Wrapf(err, "error getting service %q for the public IP address", loadBalancerName)
+	}
+
+	if len(service.Status.LoadBalancer.Ingress) < 1 {
+		return "", errors.Errorf("error, service %q didn't contain any LoadBalancer ingress", loadBalancerName)
```
```suggestion
		return "", errors.Errorf("service %q doesn't contain any LoadBalancer ingress", loadBalancerName)
```

pkg/endpoint/public_ip_test.go
```diff
@@ -0,0 +1,116 @@
...
+			},
+		}
+
+		client = fake.NewSimpleClientset(service)
+
+		backendConfig = map[string]string{}
+
+	})
+
+	When("using the lb prefix for a LoadBalancer with Ingress IP", func() {
```
tpantelis:
```suggestion
	When("a LoadBalancer with Ingress IP is specified", func() {
```

pkg/endpoint/public_ip_test.go
```diff
@@ -0,0 +1,116 @@
...
+		It("should return the IP", func() {
+			backendConfig[publicIPConfig] = "lb:" + testServiceName
+			service.Status.LoadBalancer.Ingress[0].IP = testIP
+			ip, err := getPublicIP(submSpec, client, backendConfig)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ip).To(Equal(testIP))
+		})
+	})
+
+	When("using the lb prefix for a LoadBalancer with Ingress hostname", func() {
```
tpantelis:
```suggestion
	When("a LoadBalancer with Ingress hostname is specified", func() {
```

pkg/endpoint/public_ip_test.go
```diff
@@ -0,0 +1,116 @@
...
+		It("should return the IP", func() {
+			backendConfig[publicIPConfig] = "lb:" + testServiceName
+			service.Status.LoadBalancer.Ingress[0].Hostname = testIPDNS + ".nip.io"
+			ip, err := getPublicIP(submSpec, client, backendConfig)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ip).To(Equal(testIPDNS))
+		})
+	})
+
+	When("using the lb prefix for a LoadBalancer with no ingress", func() {
```
tpantelis:
```suggestion
	When("a LoadBalancer with no ingress is specified", func() {
```

pkg/endpoint/public_ip_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+		backendConfig = map[string]string{}
+
+	})
+
+	When("using the lb prefix for a LoadBalancer with Ingress IP", func() {
+		It("should return the IP", func() {
+			backendConfig[publicIPConfig] = "lb:" + testServiceName
+			service.Status.LoadBalancer.Ingress[0].IP = testIP
+			ip, err := getPublicIP(submSpec, client, backendConfig)
```
tpantelis, line 80: I don't think we should modify the service after being added to the fake client. I suggest to refactor a bit into additional `BeforeEach` to `JustBeforeEach`:

At L64:

```
var retIP string
var retErr error

BeforeEach(func() {
    backendConfig = map[string]string{}
}

JustBeforeEach(func() {
    submSpec = ...

    service = ...

    client = fake.NewSimpleClientset(service)

    retIP, retErr := getPublicIP(submSpec, client, backendConfig)
})
```

After L87:

```
BeforeEach(func() {
    backendConfig[publicIPConfig] = "lb:" + testServiceName
    service.Status.LoadBalancer.Ingress[0].IP = testIP
}

Similar with the other test specs.

```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -84,12 +84,21 @@ const (
...
 	UDPPortConfig            = "udp-port"
 	NATTDiscoveryPortConfig  = "natt-discovery-port"
+	PublicIP                 = "public-ip"
 )
 
 const (
 	DefaultNATTDiscoveryPort = "4490"
 )
 
+// PublicIP possible prefixes/methods
```
```suggestion
// Valid PublicIP resolvers.
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -84,12 +84,21 @@ const (
...
+const (
+	IP           = "ipv4" // ipv4:1.2.3.4
```
```suggestion
	IPv4           = "ipv4" // ipv4:1.2.3.4
```

pkg/apis/submariner.io/v1/types.go
```diff
@@ -84,12 +84,21 @@ const (
...
+	LoadBalancer = "lb"   // lb:external-gw-lb
+	API          = "api"  // api:ipify.org
```
```suggestion
	API          = "api"  // api:api.ipify.org
```

tpantelis (changes requested):

pkg/endpoint/public_ip_test.go
mangelajo: You're right I was going something around those lines, not exactly that but the same concept.

pkg/endpoint/local_endpoint_suite_test.go
mangelajo: I had to do it to be able to access `firstIPv4InString` during testing without making it public.

pkg/apis/submariner.io/v1/types.go
mangelajo: Done

pkg/apis/submariner.io/v1/types.go
mangelajo: Done.

pkg/apis/submariner.io/v1/types.go
mangelajo: Done.

pkg/endpoint/local_endpoint_test.go
mangelajo: same

pkg/endpoint/public_ip.go
mangelajo: Done.

pkg/endpoint/public_ip.go
mangelajo: Done.

pkg/endpoint/public_ip_test.go
mangelajo: Done thanks.

mangelajo (commented):
Thanks for the review, I handled the comments and extended the tests.

skitt (changes requested):
This still needs `go mod tidy` ;-).

pkg/endpoint/local_endpoint_suite_test.go
tpantelis: That's in _public_ip_test.go_ so that needs to be in the `endpoint ` package but the others don't.

mangelajo: Yes but then I believe I would need to create a duplicate of the test_suite.go part. I think it's not worth it.

mangelajo:
> This still needs `go mod tidy` ;-).

@skitt  oops, right!

pkg/endpoint/public_ip_test.go
```diff
@@ -0,0 +1,188 @@
...
+	When("an API entry specified with ip4.seeip.org", func() {
+		It("should return some IP", func() {
+			backendConfig[publicIPConfig] = "api:ip4.seeip.org"
+			client := fake.NewSimpleClientset()
+			ip, err := getPublicIP(submSpec, client, backendConfig)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(net.ParseIP(ip)).NotTo(BeNil())
+		})
+	})
+
```
tpantelis: I'm not sure we should test these specific sites - we may get intermittent failures.  The main thing is to test the algorithm. We could start a local HTTP server for the test (on `localhost:<some port>`) and use that URL. If we want to verify these specific sites for validity as defaults then we can add a periodic test. But I don't think we should gate unrelated PRs if these sites get flaky.

pkg/endpoint/local_endpoint_suite_test.go
mangelajo: I was front, than you @tpantelis

pkg/endpoint/public_ip_test.go
mangelajo: I think we really should. Probably unit test is not the best place to test it.
But we definitely should test them periodically to make sure that the defaults we offer to customers are functional and healthy.

I can remove those tests if we agree to enable some periodic check for those resolvers in CI

skitt: I agree we need to test the remote resolvers, not to do their monitoring for them, but just to know that they are functional as you say. However their failure should never create difficulty in evaluating or merging a PR. So a periodic test is fine for this.

mangelajo: Thanks for your feedback, I agree it will be better to periodically monitor. I will create a separate PR for that, some github action + bash should suffice.

mangelajo:
last push to update the license headers, and squashed into a single commit for easier backport.

skitt (approved):

mangelajo:
@skitt sorry about that, while handling the markdown details I updated the patch according to the new public IP resolvers too, 
see the new table at the start and let me know if it looks ok.

mangelajo:
@skitt I'm an idiot, the comment and review re-request was really for: https://github.com/submariner-io/submariner-website/pull/485/files#diff-a9163e1cec4d91ef9323f3e06e5ad35552e68c096b78f5e5635607580c48e5a1R20

If you can, please re-add your vote, no changes here.

tpantelis (approved):

mangelajo:
Thank you @tpantelis & @skitt

PR #1345, opened 2021-05-20
Refactor cleanup handler for cabledrivers

aswinsuryan:
The cleanup handler is replaced with event handlers
for cabledrivers

Required for : #1171


pkg/event/registry.go
```diff
@@ -72,13 +74,57 @@ func (er *Registry) addHandler(eventHandler Handler) error {
 	return nil
 }
 
+func (er *Registry) addGenricHandler(eventHandler Handler) error {
```
mangelajo:
```suggestion
func (er *Registry) addGenericHandler(eventHandler Handler) error {
```

tpantelis:
```suggestion
func (er *Registry) addGenericHandler(eventHandler Handler) error {
```

pkg/event/handler.go
```diff
@@ -25,16 +25,27 @@ import (
...
 const AnyNetworkPlugin = ""
 
-type Handler interface {
+// ServiceImportType designates the type of a ServiceImport
+type HandlerType string
+
+const (
+	CableDriver    HandlerType = "CableDriver"
+	CNIDriver      HandlerType = "CNIDriver"
+	GenericHandler HandlerType = "Generic"
```
tpantelis: I think this is imparting too much knowledge about cable drivers and the different types of handlers into the registry. You're basically adding this for the XFRM cleanup handler but that is cable driver agnostic anyway and can simply return `AnyNetworkPlugin` from `GetNetworkPlugins` as the logger handler does. Then you don't need to change the `Handler` interface and `Registry` at all.

pkg/event/handler.go
```diff
@@ -25,16 +25,27 @@ import (
 
 const AnyNetworkPlugin = ""
 
-type Handler interface {
+// ServiceImportType designates the type of a ServiceImport
```
tpantelis: References ServiceImport, but I don't think a comment here adds any value.

```suggestion
```

pkg/routeagent_driver/main.go
```diff
@@ -73,15 +75,22 @@ func main() {
...
 	np := os.Getenv("SUBMARINER_NETWORKPLUGIN")
+	cableDriver := os.Getenv("SUBMARINER_CABLEDRIVER")
+
 	if np == "" {
 		np = "generic"
 	}
 
-	registry := event.NewRegistry("routeagent_driver", np)
+	if cableDriver == "" {
+		cableDriver = "libreswan"
```
```suggestion
		cableDriver = cable.GetDefaultCableDriver()
```

pkg/event/handler.go
```diff
@@ -25,16 +25,27 @@ import (
...
 
+	// GetName returns the name of the event handler
+	GetHandlerType() HandlerType
+
 	// GetName returns the name of the event handler
 	GetName() string
 
-	// GetNetworkPlugin returns the kubernetes network plugin that this handler supports.
-	GetNetworkPlugins() []string
+	// GetDrivers returns the kubernetes network drivers that this handler supports.
```
tpantelis:
```suggestion
	// GetDrivers returns the names of the drivers that this handler supports.
```

mangelajo (changes requested):
I started reviewing, but I had an Idea, let me know what you think.

What if, instead of making different types of event handlers we could make them expose the list of drivers as:

"network-plugin/*" or "cable-driver/*" for the "all cases", or simply "*" for all (logger)

In that way you wouldn't need different types (network plugin vs cable driver), the code structure and interfaces would be slightly simpler, the complexity is moved into the matcher.

What do you think?

pkg/routeagent_driver/cabledriver/xfrm.go
```diff
@@ -15,33 +15,42 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
-func NewXFRMCleanupHandler() cleanup.Handler {
-	return &XFRMCleanupHandler{netLink: netlink.New()}
+func NewXfrm() *Handler {
+	return &Handler{netLink: netlink.New()}
 }
 
-func (xc *XFRMCleanupHandler) GetName() string {
-	return "XFRM cleanup handler"
+func (xfrm *Handler) GetHandlerType() event.HandlerType {
```
tpantelis: It's recommended to keep receiver vars to 1 or 2 chars.

```suggestion
func (h *Handler) GetHandlerType() event.HandlerType {
```

pkg/routeagent_driver/cabledriver/xfrm.go
```diff
@@ -15,33 +15,42 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/submariner/pkg/netlink"
 	"k8s.io/klog"
-
-	"github.com/submariner-io/submariner/pkg/routeagent_driver/cleanup"
 )
 
-type XFRMCleanupHandler struct {
+type Handler struct {
```
tpantelis:
```suggestion
type handler struct {
```

pkg/routeagent_driver/cabledriver/xfrm.go
```diff
@@ -15,33 +15,42 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+	event.HandlerBase
 	netLink netlink.Interface
 }
 
-func NewXFRMCleanupHandler() cleanup.Handler {
-	return &XFRMCleanupHandler{netLink: netlink.New()}
+func NewXfrm() *Handler {
```
tpantelis:
```suggestion
func NewXFRMCleanupHandler() event.Handler {
```

tpantelis (changes requested):

tpantelis:
> I started reviewing, but I had an Idea, let me know what you think.
> 
> What if, instead of making different types of event handlers we could make them expose the list of drivers as:
> 
> "network-plugin/_" or "cable-driver/_" for the "all cases", or simply "*" for all (logger)
> 
> In that way you wouldn't need different types (network plugin vs cable driver), the code structure and interfaces would be slightly simpler, the complexity is moved into the matcher.
> 
> What do you think?

I was actually thinking of removing any knowledge of `Handler` types from the `Registry` altogether . Instead of the `Registry` filtering handlers statically on registration, we do it dynamically on notification via the pattern where the `Registry` asks the `Handler` if it can notify. So we remove the `GetNetworkPlugins` method and add `CanNotify(eventType string)`. This removes knowledge of the network plugin from the `Registry`, especially since not all handlers are related to network plugin. So if we had a `Handler` dependent on the local cable driver type then it can obtain cable driver dynamically via the `Endpoint` and react to all events dynamically. I was going to prototype this in a separate PR.

pkg/event/handler.go
aswinsuryan: Oh it was an oversight.

tpantelis (changes requested):
Globalnet has a `cleanup.Handler` as well that isn't used. We can remove that and then the  `cleanup.Handler` interface altogether.

aswinsuryan:
> Globalnet has a `cleanup.Handler` as well that isn't used. We can remove that and then the `cleanup.Handler` interface altogether.

Ya initially I removed it and realised it is being used in globalnet and hence retained it.

tpantelis:
> > Globalnet has a `cleanup.Handler` as well that isn't used. We can remove that and then the `cleanup.Handler` interface altogether.
> 
> Ya initially I removed it and realised it is being used in globalnet and hence retained it.

It exists but isn't actually used anywhere (`GetGlobalnetCleanupHandlers`).

pkg/event/handler.go
aswinsuryan: There is a  VXLAN cleanup handler that will be added as a part of VXLAN support. So we will have to choose between VXLAN/XFRM.

aswinsuryan:
> > > Globalnet has a `cleanup.Handler` as well that isn't used. We can remove that and then the `cleanup.Handler` interface altogether.
> > 
> > 
> > Ya initially I removed it and realised it is being used in globalnet and hence retained it.
> 
> It exists but isn't actually used anywhere (`GetGlobalnetCleanupHandlers`).

Oh ok I will remove the  `cleanup.Handler` interface

pkg/routeagent_driver/cabledriver/xfrm.go
aswinsuryan: done.

aswinsuryan:
> > I started reviewing, but I had an Idea, let me know what you think.
> > What if, instead of making different types of event handlers we could make them expose the list of drivers as:
> > "network-plugin/_" or "cable-driver/_" for the "all cases", or simply "*" for all (logger)
> > In that way you wouldn't need different types (network plugin vs cable driver), the code structure and interfaces would be slightly simpler, the complexity is moved into the matcher.
> > What do you think?
> 
> I was actually thinking of removing any knowledge of `Handler` types from the `Registry` altogether . Instead of the `Registry` filtering handlers statically on registration, we do it dynamically on notification via the pattern where the `Registry` asks the `Handler` if it can notify. So we remove the `GetNetworkPlugins` method and add `CanNotify(eventType string)`. This removes knowledge of the network plugin from the `Registry`, especially since not all handlers are related to network plugin. So if we had a `Handler` dependent on the local cable driver type then it can obtain cable driver dynamically via the `Endpoint` and react to all events dynamically. I was going to prototype this in a separate PR.

Removed the cleanup handlers. Now there is a single cable cleanup handler, when we add VXLAN cable handler, it will be added to the same handler and it would try cleanup both without checking the cable driver type (no-ops).

pkg/routeagent_driver/cabledriver/cabledriver.go
```diff
@@ -15,33 +15,38 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/submariner/pkg/netlink"
 	"k8s.io/klog"
-
-	"github.com/submariner-io/submariner/pkg/routeagent_driver/cleanup"
 )
 
-type XFRMCleanupHandler struct {
+type Handler struct {
```
tpantelis:
```suggestion
type xrfmCleanup struct {
```

pkg/routeagent_driver/cabledriver/cabledriver.go
```diff
@@ -15,33 +15,38 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+	event.HandlerBase
 	netLink netlink.Interface
 }
 
-func NewXFRMCleanupHandler() cleanup.Handler {
-	return &XFRMCleanupHandler{netLink: netlink.New()}
+func NewCableCleanupHandler() *Handler {
```
tpantelis:
```suggestion
func NewXRFMCleanupHandler() event.Handler {
```

pkg/routeagent_driver/cabledriver/cabledriver.go
```diff
@@ -15,33 +15,38 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package cleanup
+package cabledriver
```
tpantelis: I think it would be better to keep this as _xfrm_cleanup_handler.go_ to keep the handlers separate.

tpantelis (changes requested):

pkg/routeagent_driver/cabledriver/cabledriver.go
aswinsuryan: Done.

pkg/routeagent_driver/cabledriver/cabledriver.go
aswinsuryan: Done.

pkg/routeagent_driver/cabledriver/cabledriver.go
aswinsuryan: Done.

pkg/globalnet/controllers/ipam/gatewaymonitor.go
```diff
@@ -266,3 +267,19 @@ func configureTCPMTUProbe() {
 		klog.Warningf(err.Error())
 	}
 }
+
+func clearGlobalnetChains(ipt iptables.Interface) {
+	klog.Info("Globalnet is enabled and active gateway migrated, flushing Globalnet chains.")
```
sridhargaddam:
```suggestion
	klog.Info("Active gateway migrated, flushing Globalnet chains.")
```

As this is now called from Globalnet code itself.

pkg/routeagent_driver/cabledriver/xfrm_cleanup_handler.go
```diff
@@ -15,33 +15,38 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
-type XFRMCleanupHandler struct {
+type xrfmCleanup struct {
+	event.HandlerBase
 	netLink netlink.Interface
 }
 
-func NewXFRMCleanupHandler() cleanup.Handler {
-	return &XFRMCleanupHandler{netLink: netlink.New()}
+func NewXRFMCleanupHandler() *xrfmCleanup {
```
tpantelis: nit

```suggestion
func NewXRFMCleanupHandler() event.Handler {
```

pkg/globalnet/controllers/ipam/gatewaymonitor.go
aswinsuryan: Done.

pkg/routeagent_driver/cabledriver/xfrm_cleanup_handler.go
aswinsuryan: Done.

tpantelis (approved):

tpantelis:
@mangelajo this is ready for re-review.

mangelajo (approved):
Much cleaner, we can make things more specific (about cable drivers etc, the day we need something like that) thank you @aswinsuryan !

mangelajo:
@aswinsuryan it seems to need manual rebase.

PR #1346, opened 2021-05-20
Produce post-mortem output for all flake-finder jobs

mangelajo:
Sometimes we need to look at the log output of healthy jobs
and while it doesn't make sense (for resource reasons) to add
it all E2E runs, flake finder seems like a good place to add it.


mkolesnik (approved):

dfarrell07:
I messed around with doing this a while ago and noticed issues with post-mortem not reporting much because E2E had already cleaned up, removed most things it would report on.

PR #1348, opened 2021-05-20
IPAM ippool methods for Globalnet v2

### LINKED ISSUE #1347: IPAM: Add support to allocate a contiguous block of IPs
Add support to IPAM to allocate a contiguous block of IPs

**Why is this needed**:
[Globalnet v2](https://github.com/submariner-io/submariner/issues/652) has use case for getting a contiguous block of IPs, which is currently not supported by Globalnet IPAM.

vthapar:
Globalnet v2 has some newer use cases for IPAM.

* `RequestIpIfAvailable` If requested IP not available, return error
* `AllocateIPBlock`: Allocate a contiguous block of IPs

Fixes #1347

### DIFF
--- a/pkg/globalnet/controllers/ipam/ippool.go
+++ b/pkg/globalnet/controllers/ipam/ippool.go
@@ -20,16 +20,19 @@ package ipam
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"math"
 	"net"
 	"sync"
+
+	"github.com/psampaz/gods/maps/treemap"
 )
 
 type IPPool struct {
 	cidr      string
 	net       *net.IPNet
 	size      int
-	available map[string]bool   // IP.String() is key
+	available *treemap.Map      // IntIP is the key, StringIP is value
 	allocated map[string]string // resource "namespace/name" is key, ipString is value
 	sync.RWMutex
 }
@@ -50,14 +53,15 @@ func NewIPPool(cidr string) (*IPPool, error) {
 		cidr:      cidr,
 		net:       network,
 		size:      size,
-		available: make(map[string]bool),
+		available: treemap.NewWithIntComparator(),
 		allocated: make(map[string]string),
 	}
 	startingIP := ipToInt(pool.net.IP) + 1
 
 	for i := 0; i < pool.size; i++ {
+		intIP := startingIP + i
 		ip := intToIP(startingIP + i).String()
-		pool.available[ip] = true
+		pool.available.Put(intIP, ip)
 	}
 
 	RecordAvailability(cidr, size)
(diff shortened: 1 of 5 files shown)


pkg/globalnet/controllers/ipam/ippool.go
```diff
@@ -129,21 +140,103 @@ func (p *IPPool) GetAllocatedIP(key string) string {
...
 		RecordAllocateGlobalIP(p.cidr)
-		p.Unlock()
 
 		return ip, nil
 	}
 
-	// It is neither allocated for this key, nor available, give another.
-	return p.Allocate(key)
+	// It is neither allocated for this key, nor available, return error.
+	return "", errors.New("IPAM: Requested IP already allocated")
```
tpantelis, line 162:
```suggestion
	return "", errors.New("requested IP already allocated")
```

pkg/globalnet/controllers/ipam/ippool.go
```diff
@@ -129,21 +140,103 @@ func (p *IPPool) GetAllocatedIP(key string) string {
...
+		// It is neither allocated for this key, nor available, give another.
+		return p.Allocate(key)
+	}
+
+	return ip, err
+}
+
+func (p *IPPool) AllocateIPBlock(key string, num int) ([]string, error) {
+	if p.available.Size() < num {
+		return nil, errors.New("IPAM: Insufficient IPs available for allocation")
```
tpantelis, line 178:
```suggestion
		return nil, errors.New("insufficient IPs available for allocation")
```

pkg/globalnet/controllers/ipam/ippool.go
```diff
@@ -129,21 +140,103 @@ func (p *IPPool) GetAllocatedIP(key string) string {
...
+		intIP := iter.Key().(int)
+		intIPs[current] = intIP
+		if current == 0 {
+			current++
+			continue
+		}
+
+		prevIntIP := intIPs[prev]
+		if prevIntIP+1 != intIP {
+			intIPs = make([]int, num)
```
tpantelis, line 202: We don't need to reinitialize the array here.

pkg/globalnet/controllers/ipam/ippool.go
```diff
@@ -129,21 +140,103 @@ func (p *IPPool) GetAllocatedIP(key string) string {
...
+			prev, current = 0, 0
+		} else {
+			prev = current
+			current++
+		}
+	}
+
+	return nil, errors.New("IPAM: unable to find contiguous block for allocation")
```
tpantelis, line 210:
```suggestion
	return nil, errors.New("unable to find a contiguous block for allocation")
```

tpantelis (commented):
I was thinking these methods:

```
func Allocate(int num) ([]string, error)
func Release(ips []string)
func Reserve(ips []string) error
```

pkg/globalnet/controllers/ipam/ippool.go
vthapar: Need to delete all existing entries in array. Though we will end up overwriting it, better to re-initialize than end up with issues from stale entries.

vthapar:
To be replaced by https://github.com/submariner-io/submariner/pull/1357

PR #1351, opened 2021-05-24
Add GlobalEgressIP controller

tpantelis:
Related to https://github.com/submariner-io/submariner/issues/1164

### DIFF
--- a/pkg/globalnet/controllers/global_egressip_controller.go
+++ b/pkg/globalnet/controllers/global_egressip_controller.go
@@ -1,0 +1,221 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package controllers
+
+import (
+	"fmt"
+
+	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/log"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/util"
+	"github.com/submariner-io/admiral/pkg/watcher"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/globalnet/controllers/ipam"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+func NewGlobalEgressIPController(config syncer.ResourceSyncerConfig, pool *ipam.IPPool) (Interface, error) {
+	var err error
+
+	klog.Info("Creating GlobalEgressIP controller")
+
+	// TODO - get list of existing GlobalEgressIPs and prime the IPPool cache.
+
+	controller := &globalEgressIPController{
+		pool:        pool,
+		podWatchers: map[string]*podWatcher{},
+		watcherConfig: watcher.Config{
+			RestMapper: config.RestMapper,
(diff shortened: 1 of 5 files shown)


pkg/globalnet/controllers/ipam/global_egressip_controller.go
```diff
@@ -0,0 +1,174 @@
...
+		return false
+	}
+
+	// TODO - remove IP table rules for previous allocated IPs?
+
+	klog.Infof("Allocating %d global IP(s) for %q", *numberOfIPs, key)
+
+	status.AllocatedIPs = make([]string, 0, *numberOfIPs)
+
+	for i := 0; i < *numberOfIPs; i++ {
```
vthapar: Add a `TODO` here to use the new `AllocateIPBlock` method when it is merged. `Allocate` will not ensure contiguous allocation.

tpantelis: yeah. I have to review your PR.

pkg/globalnet/controllers/global_egressip_controller.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (c *globalEgressIPController) process(from runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	globalEgressIP := from.(*submarinerv1.GlobalEgressIP)
+
+	klog.Infof("Processing %sd %#v", op, globalEgressIP)
+
+	switch op {
+	case syncer.Create:
+		prevStatus := globalEgressIP.Status
```
sridhargaddam, line 99: If this is a Create event, the prevStatus will be empty right?

pkg/globalnet/controllers/global_egressip_controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+	When("a GlobalEgressIP is created", func() {
+		var numberOfIPs int
+
+		JustBeforeEach(func() {
+			t.createGlobalEgressIP(newGlobalEgressIP(globalEgressIPName, numberOfIPs, nil))
+		})
+
+		Context("with the NumberOfIPs unspecified", func() {
+			BeforeEach(func() {
+				numberOfIPs = -1
```
sridhargaddam: If numberOfIPs is a negative value, we should update the status with error message right?

pkg/globalnet/controllers/global_egressip_controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+		})
+
+		Context("in the same namespace as a GlobalEgressIP", func() {
+			It("should add an appropriate Pod IP tables rule", func() {
+				// TODO verify add
+			})
+
+			Context("and then deleted", func() {
+				JustBeforeEach(func() {
+					// TODO verify add
```
sridhargaddam, line 143: This is a nice unit test @tpantelis. Just like how we have a fake driver interface for iptables we will need one for ipsets. But that's in a different PR. Good work.

pkg/globalnet/controllers/pod_watcher.go
```diff
@@ -0,0 +1,106 @@
...
+func (w *podWatcher) onCreateOrUpdate(obj runtime.Object, numRequeues int) bool {
+	pod := obj.(*corev1.Pod)
+
+	if pod.Status.PodIP == "" {
+		return false
+	}
+
+	klog.V(log.DEBUG).Infof("Pod %q with IP %s created/updated", pod.Name, pod.Status.PodIP)
+
+	// TODO - IP table magic
```
sridhargaddam, line 90: While creating the PodWatcher we would have to pass either the name of the GlobalEgressIP Object (or the ipset name) as an argument. Without that info, we will not know where to program the iptable rule.

sridhargaddam (commented):
Great work @tpantelis. Some small comments.

pkg/globalnet/controllers/global_egressip_controller.go
tpantelis: it might not be. On restart, all existing objects are notified with create events. Update events will also follow this path once we implement it. `syncer.Update` is a separate case for now.

pkg/globalnet/controllers/global_egressip_controller_test.go
tpantelis: yeah we should handle that.  Also 0. Done with https://github.com/submariner-io/submariner/pull/1351/commits/ab64e0abd6c9905eee135f1978b76bb42285c156

pkg/globalnet/controllers/pod_watcher.go
tpantelis: yeah I figured we'd need some info about the GlobalEgressIP but I wasn't sure what.

sridhargaddam (approved):
Thanks @tpantelis

pkg/globalnet/controllers/pod_watcher.go
sridhargaddam: We can update it in a later PR, not an issue.

PR #1357, opened 2021-05-24
New IPAM for Globalnet v2

### LINKED ISSUE #1347: IPAM: Add support to allocate a contiguous block of IPs

vthapar:
To support new use cases for IPAM introduced by Globalnet v2,
creating a newer version of IPAM that uses a sorted treemap
for IP Pool and modifies existing Allocate/Release methods
to allow allocation/release of multiple IPs.

Fixes #1347

### DIFF
--- a/pkg/ipam/ippool.go
+++ b/pkg/ipam/ippool.go
@@ -1,0 +1,210 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package ipam
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"math"
+	"net"
+	"sync"
+
+	"github.com/emirpasic/gods/maps/treemap"
+)
+
+type IPPool struct {
+	cidr      string
+	net       *net.IPNet
+	size      int
+	available *treemap.Map // IntIP is the key, StringIP is value
+	sync.RWMutex
+}
+
+func NewIPPool(cidr string) (*IPPool, error) {
+	_, network, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return nil, err
+	}
+
+	ones, totalbits := network.Mask.Size()
+	size := int(math.Exp2(float64(totalbits-ones))) - 2 // don't count net and broadcast
+	if size < 2 {
+		return nil, fmt.Errorf("invalid prefix for CIDR %q", cidr)
+	}
+
+	pool := &IPPool{
+		cidr:      cidr,
+		net:       network,
+		size:      size,
+		available: treemap.NewWithIntComparator(),
+	}
+	startingIP := ipToInt(pool.net.IP) + 1
+
+	for i := 0; i < pool.size; i++ {
+		intIP := startingIP + i
+		ip := intToIP(intIP).String()
+		pool.available.Put(intIP, ip)
+	}
+
+	// TODO: RecordAvailability(cidr, size)
+
+	return pool, nil
+}
+
+func ipToInt(ip net.IP) int {
+	intIP := ip
+	if len(ip) == 16 {
+		intIP = ip[12:16]
+	}
+
+	return int(binary.BigEndian.Uint32(intIP))
+}
+
+func intToIP(ip int) net.IP {
+	netIP := make(net.IP, 4)
+	binary.BigEndian.PutUint32(netIP, uint32(ip))
+
+	return netIP
+}
+
+func StringIPToInt(stringIP string) int {
+	return ipToInt(net.ParseIP(stringIP))
+}
+
+func (p *IPPool) allocate() (string, error) {
+	p.Lock()
+	defer p.Unlock()
+
+	iter := p.available.Iterator()
+	if !p.available.Empty() {
+		iter.Last()
+		ip := iter.Value().(string)
+		p.available.Remove(iter.Key())
+		// TODO: RecordAllocateGlobalIP(p.cidr)
+
+		return ip, nil
+	}
+
+	return "", errors.New("insufficient IPs available for allocation")
+}
+
+func (p *IPPool) Allocate(num int) ([]string, error) {
+	if p.available.Size() < num {
+		return nil, errors.New("insufficient IPs available for allocation")
+	}
+
+	ips := make([]string, num)
+
+	if num == 1 {
+		ip, err := p.allocate()
+		if err != nil {
+			return nil, err
+		}
+
+		ips[0] = ip
+
+		return ips, nil
+	}
+
+	intIPs := make([]int, num)
+	prev, current := 0, 0
+
+	p.Lock()
+	defer p.Unlock()
+
+	iter := p.available.Iterator()
+	for iter.Next() {
+		if current == num {
+			return p.allocateBlock(intIPs), nil
+		}
+
+		intIP := iter.Key().(int)
+		intIPs[current] = intIP
+		if current == 0 {
+			current++
+			continue
+		}
+
+		prevIntIP := intIPs[prev]
+		if prevIntIP+1 != intIP {
+			prev, current = 0, 0
+		} else {
+			prev = current
+			current++
+		}
+	}
+
+	return nil, fmt.Errorf("unable to find contiguous block of %d IPs for allocation", num)
+}
+
+func (p *IPPool) Release(ips ...string) {
+	p.Lock()
+	defer p.Unlock()
+
+	for _, ip := range ips {
+		p.available.Put(StringIPToInt(ip), ip)
+	}
+}
+
+func (p *IPPool) Reserve(ips ...string) error {
+	num := len(ips)
+	intIPs := make([]int, num)
+
+	p.Lock()
+	defer p.Unlock()
+
+	for i := 0; i < num; i++ {
+		intIPs[i] = StringIPToInt(ips[i])
+		_, found := p.available.Get(intIPs[i])
+		if !found {
+			return fmt.Errorf("requested IP %s already allocated", ips[i])
+		}
+	}
+
+	for i := 0; i < num; i++ {
+		p.available.Remove(intIPs[i])
+	}
+
+	return nil
+}
+
+func (p *IPPool) IsAvailable(ip string) bool {
+	p.RLock()
+	_, result := p.available.Get(StringIPToInt(ip))
+	p.RUnlock()
+
+	return result
+}
+
+// Make sure Lock is already taken before calling this
+func (p *IPPool) allocateBlock(intIPs []int) []string {
+	num := len(intIPs)
+	ips := make([]string, num)
+
+	for i := 0; i < num; i++ {
+		ip, _ := p.available.Get(intIPs[i])
+		ips[i] = ip.(string)
+
+		p.available.Remove(intIPs[i])
+	}
+
+	// TODO: RecordAllocateGlobalIPs(p.cidr, num)
+
+	return ips
+}
--- a/pkg/ipam/ippool.go
+++ b/pkg/ipam/ippool.go
@@ -30,9 +30,9 @@ import (
 
 type IPPool struct {
 	cidr      string
-	net       *net.IPNet
+	network   *net.IPNet
 	size      int
-	available *treemap.Map // IntIP is the key, StringIP is value
+	available *treemap.Map // int IP is the key, string IP is the value
 	sync.RWMutex
 }
 
@@ -50,11 +50,12 @@ func NewIPPool(cidr string) (*IPPool, error) {
 
 	pool := &IPPool{
 		cidr:      cidr,
-		net:       network,
+		network:   network,
 		size:      size,
 		available: treemap.NewWithIntComparator(),
 	}
-	startingIP := ipToInt(pool.net.IP) + 1
+
+	startingIP := ipToInt(pool.network.IP) + 1
 
 	for i := 0; i < pool.size; i++ {
 		intIP := startingIP + i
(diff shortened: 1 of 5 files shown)


pkg/ipam/ippool.go
```diff
@@ -0,0 +1,242 @@
...
+		net:       network,
+		size:      size,
+		available: treemap.NewWithIntComparator(),
+		allocated: make(map[string]string),
+	}
+	startingIP := ipToInt(pool.net.IP) + 1
+
+	for i := 0; i < pool.size; i++ {
+		intIP := startingIP + i
+		ip := intToIP(startingIP + i).String()
```
sridhargaddam:
```suggestion
		ip := intToIP(intIP).String()
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (p *IPPool) GetAllocatedIP(key string) string {
+	p.RLock()
+	ip := p.allocated[key]
+	p.RUnlock()
+
+	return ip
+}
+
+func (p *IPPool) RequestIPIfAvailable(key, ip string) (string, error) {
```
sridhargaddam: We many not require this API in Globalnet 2.0, but it might be helpful for future enhancements.

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,242 @@
...
+	p.Lock()
+	defer p.Unlock()
+
+	iter := p.available.Iterator()
+	for iter.Next() {
+		if current == num {
+			return p.allocateBlock(key, intIPs), nil
+		}
+
+		intIP := iter.Key().(int)
```
sridhargaddam, line 129: Can you please add some comments on the logic used for allocation.

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+		if prevIntIP+1 != intIP {
+			intIPs = make([]int, num)
+			prev, current = 0, 0
+		} else {
+			prev = current
+			current++
+		}
+	}
+
+	return nil, errors.New("IPAM: unable to find contiguous block for allocation")
```
sridhargaddam:
```suggestion
	return nil, errors.New("IPAM: unable to find contiguous block of %d IPs for allocation")
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (p *IPPool) Release(ips []string) {
+	p.Lock()
+	defer p.Unlock()
+
+	for _, ip := range ips {
+		p.available.Put(StringIPToInt(ip), ip)
+		delete(p.allocated, ip)
+	}
+	p.Unlock()
```
sridhargaddam: We are already using defer in Line 160

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (p *IPPool) Reserve(ips []string) error {
+	num := len(ips)
+	intIPs := make([]int, num)
+	p.Lock()
+	defer p.Unlock()
+
+	if p.available.Size() < num {
+		errors.New("IPAM: Insufficient IPs available for allocation")
```
sridhargaddam:
```suggestion
		return errors.New("IPAM: Insufficient IPs available for allocation")
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+	for _, ip := range ips {
+		p.available.Put(StringIPToInt(ip), ip)
+		delete(p.allocated, ip)
+	}
+	p.Unlock()
+}
+
+func (p *IPPool) Reserve(ips []string) error {
+	num := len(ips)
+	intIPs := make([]int, num)
```
sridhargaddam, line 177: You can move them to Line 178

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+	defer p.Unlock()
+
+	if p.available.Size() < num {
+		errors.New("IPAM: Insufficient IPs available for allocation")
+	}
+
+	for i := 0; i < num; i++ {
+		intIPs[i] = StringIPToInt(ips[i])
+		_, found := p.available.Get(intIPs[i]); if !found {
+			return errors.New(fmt.Sprintf("IPAM: Requested IP %s already allocated", ips[i]))
```
sridhargaddam: IMO, this API should use return codes (instead of plain error) so that caller will be able to distinguish between Insufficient IPs vs already allocated IPs using the error codes.

pkg/ipam/ippool_test.go
```diff
@@ -0,0 +1,525 @@
...
+
+	BeforeEach(func() {
+		var err error
+		pool, err = NewIPPool(testCidr)
+		Expect(err).NotTo(HaveOccurred())
+
+		available = pool.available.Size()
+		allocated = len(pool.allocated)
+
+		allocatedIP, err = pool.Allocate(service1)
```
sridhargaddam: pool.Allocate() takes an integer as an argument, but we are passing a string in the unit tests?

sridhargaddam (changes requested):
Thanks for the PR @vthapar. Some comments inline.

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+	p.Lock()
+	defer p.Unlock()
+
+	if p.available.Size() < num {
+		errors.New("IPAM: Insufficient IPs available for allocation")
+	}
+
+	for i := 0; i < num; i++ {
+		intIPs[i] = StringIPToInt(ips[i])
+		_, found := p.available.Get(intIPs[i]); if !found {
```
tpantelis:
```suggestion
		_, found := p.available.Get(intIPs[i]); 
		if !found {
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+	p.Unlock()
+}
+
+func (p *IPPool) Reserve(ips []string) error {
+	num := len(ips)
+	intIPs := make([]int, num)
+	p.Lock()
+	defer p.Unlock()
+
+	if p.available.Size() < num {
```
tpantelis: We really don't need this check - it will error out below.

pkg/ipam/ippool.go
tpantelis: I think insufficient IPs is basically the same as already allocated IPs. I like this returning the actual IP in the error message.

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func intToIP(ip int) net.IP {
+	netIP := make(net.IP, 4)
+	binary.BigEndian.PutUint32(netIP, uint32(ip))
+
+	return netIP
+}
+
+func StringIPToInt(stringIP string) int {
```
tpantelis, line 87: Does this need to be exported?

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+func NewIPPool(cidr string) (*IPPool, error) {
+	_, network, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return nil, err
+	}
+
+	ones, totalbits := network.Mask.Size()
+	size := int(math.Exp2(float64(totalbits-ones))) - 2 // don't count net and broadcast
+	if size < 2 {
+		return nil, errors.New("invalid CIDR Prefix")
```
tpantelis:
```suggestion
		return nil, fmt.Errorf("invalid prefix for CIDR %q", cidr)
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"github.com/emirpasic/gods/maps/treemap"
+)
+
+type IPPool struct {
+	cidr      string
+	net       *net.IPNet
+	size      int
+	available *treemap.Map    // IntIP is the key, StringIP is value
+	allocated map[string]bool // resource IPAddress is key, value is always true
```
tpantelis: I still don't think we need this.

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+
+		return ip, nil
+	}
+
+	return "", errors.New("IPAM: Insufficient IPs available for allocation")
+}
+
+func (p *IPPool) Allocate(num int) ([]string, error) {
+	if p.available.Size() < num {
+		return nil, errors.New("IPAM: Insufficient IPs available for allocation")
```
tpantelis:
```suggestion
		return nil, errors.New("insufficient IPs available for allocation")
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+		iter.Last()
+		ip := iter.Value().(string)
+		p.allocated[ip] = true
+		p.available.Remove(iter.Key())
+		// TODO: RecordAllocateGlobalIP(p.cidr)
+
+		return ip, nil
+	}
+
+	return "", errors.New("IPAM: Insufficient IPs available for allocation")
```
tpantelis:
```suggestion
	return "", errors.New("insufficient IPs available for allocation")
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+		intIP := iter.Key().(int)
+		intIPs[current] = intIP
+		if current == 0 {
+			current++
+			continue
+		}
+
+		prevIntIP := intIPs[prev]
+		if prevIntIP+1 != intIP {
+			intIPs = make([]int, num)
```
tpantelis: We don't really need to re-allocate here since we'll overwrite all entry slots anyway since we can only return if we've allocated `num` entries.

pkg/ipam/ippool.go
tpantelis: > 

Also

```suggestion
	return nil, errors.New("unable to find contiguous block for allocation")
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+		} else {
+			prev = current
+			current++
+		}
+	}
+
+	return nil, errors.New("IPAM: unable to find contiguous block for allocation")
+}
+
+func (p *IPPool) Release(ips []string) {
```
tpantelis:
```suggestion
func (p *IPPool) Release(ips ...string) {
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+	defer p.Unlock()
+
+	for _, ip := range ips {
+		p.available.Put(StringIPToInt(ip), ip)
+		delete(p.allocated, ip)
+	}
+	p.Unlock()
+}
+
+func (p *IPPool) Reserve(ips []string) error {
```
tpantelis:
```suggestion
func (p *IPPool) Reserve(ips ...string) error {
```

pkg/ipam/ippool.go
tpantelis:
```suggestion
			return errors.New(fmt.Sprintf("requested IP %s already allocated", ips[i]))
```

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,231 @@
...
+	for i := 0; i < num; i++ {
+		intIPs[i] = StringIPToInt(ips[i])
+		_, found := p.available.Get(intIPs[i]); if !found {
+			return errors.New(fmt.Sprintf("IPAM: Requested IP %s already allocated", ips[i]))
+		}
+	}
+
+	for i := 0; i < num; i++ {
+		p.allocated[ips[i]] = true
+		p.available.Remove(intIPs[i])
```
tpantelis, line 195: I think we can move this after L183 instead of a second loop?

pkg/ipam/ippool_test.go
```diff
@@ -0,0 +1,525 @@
...
+)
+
+func testPoolCreation() {
+	When("CIDR is missing the prefix", func() {
+		It("should fail to create the IP pool with an error", func() {
+			pool, err := NewIPPool("169.254.0.0")
+			Expect(err).To(HaveOccurred())
+			Expect(pool).To(BeNil())
+		})
+	})
```
tpantelis, line 44: Can you add newlines after each `When` block?

tpantelis (changes requested):

pkg/ipam/ippool.go
tpantelis: Actually in this case where the caller provides the IPs to reserve, if not found then it is either already allocated or is not valid for the CIDR. The caller may want to distinguish these 2 cases.

pkg/ipam/ippool.go
vthapar: Not in new ipam. In old IPAM it was being used in controller code.

pkg/ipam/ippool.go
vthapar: No. First we check if requested no. of IPs can be allocated or not. Idea is to fail if even one of the requested is not available. Unless you want to support use case of partial allocation. If we move this to after L83, we will have to add another loop for rollback of ones that were allocated.

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,203 @@
...
+
+			return retIPs, nil
+		}
+	}
+
+	return nil, fmt.Errorf("unable to allocate a contiguous block of %d IPs - available pool size is %d",
+		num, p.available.Size())
+}
+
+func (p *IPPool) Release(ips ...string) {
```
vthapar: What if someone tries to release an IP which doesn't even belong to this CIDR? With available+allocated we had an implicit check. We should add a check to make sure we're not adding an IP outside of CIDR to the pool coz then it may get allocated by mistake. This can happen if consumer code is not checking change in CIDRs and cleaning old CIDR's IPs correctly.

pkg/ipam/ippool.go
tpantelis: The unit tests use it - I left it exported.

pkg/ipam/ippool.go
tpantelis: yeah I was thinking that too.

tpantelis: Done.

pkg/ipam/ippool.go
```diff
@@ -0,0 +1,209 @@
...
+		// TODO: RecordAllocateGlobalIP(p.cidr)
+
+		return []string{iter.Value().(string)}, nil
+	}
+
+	return nil, errors.New("insufficient IPs available for allocation")
+}
+
+func (p *IPPool) Allocate(num int) ([]string, error) {
+	if num < 0 {
```
vthapar: nitpick: Can't this be a switch case? Surprised linting didn't complain about it.

tpantelis (approved):

pkg/ipam/ippool.go
tpantelis: done

sridhargaddam (approved):

Issue #1163, opened 2021-03-01
Support Globalnet EgressIPs at the cluster level


sridhargaddam:
**What would you like to be added**:
As part of [Globalnet Enhancement](https://github.com/submariner-io/enhancements/pull/22),
by default a configurable number of globalIPs will be allocated to the Cluster.
This request is to allocate the required number of globalIPs and configure the necessary
IPTable Egress rules that apply at the cluster scope.

Instead of using an IPset with individual pod IPs, we can use the PodCIDR to match the sourceIPs in the iptable/IPset rules.
This way we can avoid programming any explicit rules for supporting Hostnetworking use-cases as well.

tpantelis:
bump

PR #1360, opened 2021-05-25
Add ClusterGlobalEgressIP controller

tpantelis:
Related to https://github.com/submariner-io/submariner/issues/1163

### DIFF
--- a/pkg/globalnet/controllers/cluster_egressip_controller.go
+++ b/pkg/globalnet/controllers/cluster_egressip_controller.go
@@ -1,0 +1,141 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/util"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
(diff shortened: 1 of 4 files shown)


pkg/globalnet/controllers/cluster_egressip_controller.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (c *clusterGlobalEgressIPController) process(from runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	clusterGlobalEgressIP := from.(*submarinerv1.ClusterGlobalEgressIP)
+
+	klog.Infof("Processing %sd %#v", op, clusterGlobalEgressIP)
+
+	switch op {
+	case syncer.Create:
+		prevStatus := clusterGlobalEgressIP.Status
+		requeue := c.onCreate(clusterGlobalEgressIP)
```
sridhargaddam, line 107: Here we should check the name of the object as we only support cluster-default for now.

From the enhancement PR:
```
When Submariner Globalnet is deployed in a Cluster, it auto-allocates the configured number of globalIPs and programs the necessary egress rules on the active Gateway node of the Cluster. This is done internally by creating ClusterGlobalEgressIP CR object with the name cluster-default. Users can query the CR to check the allocated GlobalIPs at the Cluster level. If admin creates additional ClusterGlobalEgressIP CRs that match the Cluster scope, the status field of those CRs will be marked as error with appropriate message. Basically only cluster-default object will be supported.
```
Link: https://github.com/submariner-io/enhancements/blob/devel/submariner/globalnet-enhancement2-0.md

sridhargaddam: Cool. I see that you are already handling it in onCreate method. Nice.

pkg/globalnet/controllers/types.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/submariner-io/submariner/pkg/globalnet/controllers/ipam"
 )
 
+const ClusterGlobalEgressIPName = "cluster-egress-ip.submariner.io"
```
sridhargaddam: In the enhancement PR we have used `cluster-default`, but if you think `cluster-egress-ip` is a better name, we can do that. However, since it could be more than a single IP, what do you suggest?

tpantelis: We can omit the "_ip_" part, "_cluster-egress.submariner.io_".  I qualified it with our domain to further ensure uniqueness as we do with label names.  I saw that u/s qualifies well-known CR names with "_k8s.io_".

tpantelis: Updated in last push.

sridhargaddam (approved):
Thanks @tpantelis

PR #1362, opened 2021-05-26
Update e2e to support globalnet 2.0 changes in future PRs

sridhargaddam:
This PR does not modify the e2e tests as such but modifies the
way connector Pods are scheduled. This will allow us to modify
tests easily for Globalnet 2.0 use-case validation.

### DIFF
--- a/test/e2e/redundancy/gateway_failover.go
+++ b/test/e2e/redundancy/gateway_failover.go
@@ -25,7 +25,6 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
 	"github.com/submariner-io/shipyard/test/e2e/tcp"
-	"github.com/submariner-io/submariner/test/e2e/dataplane"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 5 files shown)


test/e2e/redundancy/gateway_failover.go
```diff
@@ -25,6 +25,7 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
 	"github.com/submariner-io/shipyard/test/e2e/tcp"
+	"github.com/submariner-io/submariner/test/e2e/dataplane"
```
tpantelis: We don't want the `redundancy` package referencing `dataplane` as that will cause the ginkgo test specs to get pulled in as well. Perhaps we should put the new functions in shipyard as well? Will we need `tcp.RunConnectivityTest` anymore?

tpantelis (changes requested):

test/e2e/redundancy/gateway_failover.go
sridhargaddam: Done

test/e2e/framework/resources.go
```diff
@@ -0,0 +1,142 @@
+/*
```
tpantelis: _resources.go_ -> _dataplane.go_?

tpantelis (changes requested):

test/e2e/framework/resources.go
sridhargaddam: Done

tpantelis (approved):

vthapar (approved):

Issue #1331, opened 2021-05-14
Use the event handler in the Globalnet GatewayMonitor


tpantelis:
**What would you like to be added**:

The Globalnet GatewayMonitor has overlap with the event handler Controller so we should use the latter for the gateway transition logic.

**Why is this needed**:

Code reuse.

mkolesnik:
@astoycos ping

PR #1363, opened 2021-05-26
Add Globalnet v2 gateway monitor

tpantelis:
Based on the previous one and starts the `GlobalEgressIP` and `ClusterGlobalEgressIP` controllers.

### DIFF
--- a/pkg/globalnet/controllers/gateway_monitor.go
+++ b/pkg/globalnet/controllers/gateway_monitor.go
@@ -1,0 +1,429 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package controllers
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/log"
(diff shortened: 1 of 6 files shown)


pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -0,0 +1,429 @@
...
+	gatewayMonitor := &gatewayMonitor{
+		baseController: newBaseController(),
+		spec:           spec,
+		isGatewayNode:  false,
+		remoteSubnets:  stringset.NewSynchronized(),
+	}
+
+	var err error
+
+	gatewayMonitor.ipt, err = iptables.New()
```
mangelajo, line 56: Can we use the event handler architecture in here?.

In the case of OVN we have the chance of eventually creating an OVN-specific version of the handler which would not be iptables based. Also would smoothen the transition to other architectures like nftables eventually.

mangelajo, line 56: Can we use the event handler architecture in here?.

In the case of OVN we have the chance of eventually creating an OVN-specific version of the handler which would not be iptables based. Also would smoothen the transition to other architectures like nftables eventually.

mangelajo (changes requested):
@sridhargaddam @tpantelis some comments about this. Is it possible to construct the gateway monitor like an event handler? 

https://github.com/submariner-io/submariner/blob/devel/pkg/routeagent_driver/main.go#L84

pkg/globalnet/controllers/gateway_monitor.go
sridhargaddam: Right, we are also thinking of using the Event handler framework and based on TransitionToGateway and TransitionToNonGateway, we can start/stop the other Globalnet V2 controllers

tpantelis: yes that is the [plan](https://github.com/submariner-io/submariner/issues/1331). The new controllers are higher priority - just want to get those working first.

mangelajo: ack :+1:

mangelajo (approved):
ok I'm fine with moving along, as long as we move into event handler before releasing

sridhargaddam (approved):

PR #1371, opened 2021-05-27
Use short Kubernetes versions

skitt:
Depends on https://github.com/submariner-io/shipyard/pull/568


.github/workflows/e2e-full.yml
```diff
@@ -18,15 +18,15 @@ jobs:
         globalnet: ['', 'globalnet']
         cable_driver: ['libreswan', 'wireguard']
         ovn: ['', 'ovn']
-        k8s_version: ['1.20.2']
+        k8s_version: ['1.20']
```
mangelajo, line 21: Side comment not related to this change:
Makes sense, I wonder if we should probably govern the default k8s version from shipyard itself. So then the developers get the same default experience and we can rotate them all from a single place.

@mkolesnik @dfarrell07 what do you think?

mangelajo (approved):

.github/workflows/e2e-full.yml
mkolesnik: It's already the default - https://github.com/submariner-io/shipyard/blob/devel/gh-actions/e2e/action.yaml#L7

What you're missing is how GHA matrices work, where apparently you have to have the variable defined in the matrix as a single value list for includes to work

mkolesnik (approved):

dfarrell07 (approved):

tpantelis (approved):

PR #1376, opened 2021-06-01
Label gateway pod with the host and HA status

### LINKED ISSUE #1370: Make submariner-gateway label it's own Pod as gateway=active and host=$host when active
As part of #1071

This way we can point a single Load Balancer (or multiple in the future) to the right active gateway.

Depends on https://github.com/submariner-io/submariner-operator/issues/1389

mangelajo:
Fixes-Issue: #1370

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
 	"github.com/submariner-io/admiral/pkg/watcher"
+	"github.com/submariner-io/submariner/pkg/pod"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -83,6 +84,7 @@ const (
 
 var VERSION = "not-compiled-properly"
 
+// nolint:gocyclo // the main function has a lot of error checking that increases cyclomatic complexity
 func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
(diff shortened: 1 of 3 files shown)


sridhargaddam (approved):
LGTM, thanks @mangelajo

mangelajo:
Thank you @sridhargaddam

mangelajo:
@skitt @tpantelis I need another review here :pray:  :)

pkg/cableengine/cableengine.go
```diff
@@ -67,16 +68,25 @@ type engine struct {
...
 
 // NewEngine creates a new Engine for the local cluster
-func NewEngine(localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint) Engine {
+func NewEngine(localCluster types.SubmarinerCluster, localEndpoint types.SubmarinerEndpoint,
+	gwPod pod.GatewayPodInterface) (Engine, error) {
+	if gwPod != nil {
+		if err := gwPod.SetHALabels(v1.HAStatusPassive); err != nil {
+			return nil, err
+		}
+	}
```
tpantelis: This seems like an unnecessary coupling, ie having the caller pass an interface for this function to call when the caller could directly call it.  Same in `StartEngine`.

tpantelis (changes requested):

Jaanki (approved):

pkg/cableengine/cableengine.go
mangelajo: Yeah, but we store it to call it on tear down.

mangelajo: Yes, but it's not unnecessary.

This allows for better testing. Otherwise if we do it all in main, we can't test the behavior as we do in unit tests.

In fact I think I should probably go further and make the responsibility of declaring itself "active" or "passive" part of the cable driver. So when we have A/A cable drivers they can declare themselves active in multiple places, or not.

I can agree that the initial "passive" labeling can be moved into main. But the "active" probably belongs more inside the engine itself.

Anyway, it's easy to move into main, this is all probably very blurry and subjective, so let me know if you believe it's still better in main.

tpantelis: Another solution is to update the pod in the GatewaySyncer - that already polls and updates the status in the Gateway object. Yet another way is to add a simple Gateway watcher that updates the pod label.

mangelajo: That makes sense to me, I'll update to handle it from the gateway syncer and see how it looks.

mangelajo:
@tpantelis I updated the patch to do it from main.go

I figured out that the gateway Syncer is not acceptable because we the redirection of the loadbalancer will work based on those labels, and from the syncer we have a refresh period which will increase the HA failover time.

mangelajo:
It's updated.

mangelajo:
I broke it ':D

mangelajo:
@tpantelis can you re-review? Thanks

tpantelis (approved):

PR #1379, opened 2021-06-02
Use update-only Federator in GN controllers

tpantelis:
- Use update-only `Federator` as the update status `Federator` was problematic.
- The `Reason` field in `metav1.Conditon` must be non-empty
- The `AllocatedIPs` field needs to specify "omitempty" as it's not required.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/onsi/gomega v1.13.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.10.0
-	github.com/submariner-io/admiral v0.10.0-m1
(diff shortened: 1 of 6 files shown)


sridhargaddam (approved):

sridhargaddam:
@tpantelis After this PR is merged, we should update submariner-operator to use latest Submariner repo right?

tpantelis:
> @tpantelis After this PR is merged, we should update submariner-operator to use latest Submariner repo right?

I don't think we need to yet - we haven't converted to the new controllers.

sridhargaddam:
> > @tpantelis After this PR is merged, we should update submariner-operator to use latest Submariner repo right?
> 
> I don't think we need to yet - we haven't converted to the new controllers.

Yes, I meant to say when we move to new controllers.

vthapar (approved):

PR #1390, opened 2021-06-07
Add GN 2.0 ServiceExport and GlobalIngressIP controllers

tpantelis:

### DIFF
--- a/pkg/globalnet/controllers/controllers_suite_test.go
+++ b/pkg/globalnet/controllers/controllers_suite_test.go
@@ -48,9 +48,11 @@ import (
 )
 
 const (
-	namespace          = "submariner"
-	localCIDR          = "169.254.1.0/24"
-	globalEgressIPName = "east-region"
+	namespace           = "submariner"
+	localCIDR           = "169.254.1.0/24"
+	globalEgressIPName  = "east-region"
+	globalIngressIPName = "db-service-ingress-ip"
+	serviceName         = "nginx"
 )
 
 func init() {
@@ -75,13 +77,16 @@ type testDriverBase struct {
 	globalCIDR             string
 	globalEgressIPs        dynamic.ResourceInterface
 	clusterGlobalEgressIPs dynamic.ResourceInterface
+	globalIngressIPs       dynamic.ResourceInterface
+	services               dynamic.ResourceInterface
+	serviceExports         dynamic.ResourceInterface
 	pods                   dynamic.NamespaceableResourceInterface
 }
 
 func newTestDriverBase() *testDriverBase {
 	t := &testDriverBase{
 		restMapper: test.GetRESTMapperFor(&submarinerv1.Endpoint{}, &corev1.Service{}, &corev1.Node{}, &corev1.Pod{},
-			&submarinerv1.GlobalEgressIP{}, &submarinerv1.ClusterGlobalEgressIP{}, &mcsv1a1.ServiceExport{}),
+			&submarinerv1.GlobalEgressIP{}, &submarinerv1.ClusterGlobalEgressIP{}, &submarinerv1.GlobalIngressIP{}, &mcsv1a1.ServiceExport{}),
 		scheme:     runtime.NewScheme(),
 		ipt:        fakeIPT.New(),
 		globalCIDR: localCIDR,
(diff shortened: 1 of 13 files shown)


pkg/globalnet/controllers/controllers_suite_test.go
```diff
@@ -48,9 +48,11 @@ import (
 )
 
 const (
-	namespace          = "submariner"
-	localCIDR          = "169.254.1.0/24"
-	globalEgressIPName = "east-region"
+	namespace           = "submariner"
+	localCIDR           = "169.254.1.0/24"
```
sridhargaddam, line 52: As part of adding support for GlobalIPs at Cluster scope, we require localCIDRs (i.e., Pod/Service CIDRs). So maybe we can rename localCIDR to globalCIDR to avoid ambiguity.

pkg/globalnet/controllers/controllers_suite_test.go
```diff
@@ -148,20 +170,49 @@ func (t *testDriverBase) deletePod(p *corev1.Pod) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      s.GetName(),
+			Namespace: s.GetNamespace(),
+		},
+	})
+}
+
+func (t *testDriverBase) getGlobalIngressIPStatus(name string) *submarinerv1.GlobalIngressIPStatus {
+	status := &submarinerv1.GlobalIngressIPStatus{}
+	getStatus(t.globalIngressIPs, name, status)
```
sridhargaddam, line 189:
```suggestion
	if !getStatus(t.globalIngressIPs, name, status) {
	    return nil
	}
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -0,0 +1,159 @@
...
+	client := config.SourceClient.Resource(*gvr).Namespace(corev1.NamespaceAll)
+	list, err := client.List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	federator := federate.NewUpdateFederator(config.SourceClient, config.RestMapper, corev1.NamespaceAll)
+
+	for i := range list.Items {
+		err = controller.reserveAllocatedIPs(federator, &list.Items[i])
```
sridhargaddam, line 59: Just for info: Along with reserving the IPs we also have to sync the IPTable rules, hence in my PR [https://github.com/submariner-io/submariner/pull/1387], I removed this code and included similar logic in OnCreate method. We can leave it this way, when I work on the IPTables stuff, I'll do the necessary changes.

pkg/globalnet/controllers/service_export_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	"k8s.io/klog"
+	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
+)
+
+func NewServiceExportController(config syncer.ResourceSyncerConfig) (Interface, error) {
+	var err error
+
+	klog.Info("Creating ServiceExport controller")
+
+	_, gvr, err := util.ToUnstructuredResource(&corev1.Service{}, config.RestMapper)
```
sridhargaddam, line 41:
```suggestion
	_, gvr, err := util.ToUnstructuredResource(&mcsv1a1.ServiceExport, config.RestMapper)
```
Should it be mcsv1a1.ServiceExport?

pkg/globalnet/controllers/service_export_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	return nil, false
+}
+
+func (c *serviceExportController) onCreate(serviceExport *mcsv1a1.ServiceExport) (runtime.Object, bool) {
+	key, _ := cache.MetaNamespaceKeyFunc(serviceExport)
+
+	service, exists, err := c.getService(serviceExport.Name, serviceExport.Namespace)
+	if err != nil || !exists {
+		klog.Infof("Exported Service %q does not exist yet - re-queueing", key)
+		return nil, true
```
sridhargaddam, line 89: When we requeue, how frequently/periodically does it retry?

sridhargaddam (commented):
Thanks for working on the PR @tpantelis

pkg/globalnet/controllers/controllers_suite_test.go
tpantelis: This naming came from the previous tests. It's used for `Endpoints` and is meant to distinguish between a CIDR for a local cluster and one for a remote cluster.

pkg/globalnet/controllers/controllers_suite_test.go
tpantelis: I did this to avoid having to check for nil in all call sites.  If the `status`  field was missing, returning an empty struct is equivalent to nil. Hence I removed the return value from `getStatus`.

pkg/globalnet/controllers/service_export_controller.go
tpantelis: No, this is used to retrieve `Service` objects.

pkg/globalnet/controllers/service_export_controller.go
tpantelis: Up to 30 sec.

pkg/globalnet/controllers/global_ingressip_controller.go
tpantelis: This was done on startup so we can ensure each object will be able to reserve its previously allocated IPs (unless the global CIDR changed). Otherwise, in doing this on callbacks, we may process a newly created object first which could allocate a previously used block, in which case we'd need to reallocate.

vthapar (approved):

pkg/globalnet/controllers/service_export_controller.go
vthapar: iirc, it is exponential backoff with cap of 30 secs.

pkg/globalnet/controllers/controllers_suite_test.go
sridhargaddam: Okay, I can change it in my next PR

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: Good point.

sridhargaddam (approved):
Thanks @tpantelis

PR #1394, opened 2021-06-08
Fix CODEOWNERS file to include sub-directories

sridhargaddam:
The `docs/*` pattern will match files like `docs/getting-started.md`
but not further nested files. To include all files including sub-directories,
we should be using `docs/`


skitt (approved):

mangelajo (approved):

sridhargaddam:
@skitt Oops did not see that you modified the auto-rebase strategy. Anyways, we are good now :D

PR #1398, opened 2021-06-09
Follow-up to #1387

tpantelis:
#1387

- Enhanced unit tests to cover more cases including updates, `NumberOfIPs` negative/zero, deletion et al
- Release the previous allocated IPs if `NumberOfIPs` is updated to 0
-  Modified `validate` to return a success/fail bool instead of error. It's a user error so we don't need to explicitly log a warning/error since we append a `Status` condition.
- Changed `localSubnets` to an array since that is effectively the way it's used and it's not intended to change.

Related to https://github.com/submariner-io/submariner/issues/1163

### DIFF
--- a/pkg/globalnet/controllers/cluster_egressip_controller.go
+++ b/pkg/globalnet/controllers/cluster_egressip_controller.go
@@ -24,7 +24,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/federate"
 	"github.com/submariner-io/admiral/pkg/log"
-	"github.com/submariner-io/admiral/pkg/stringset"
 	"github.com/submariner-io/admiral/pkg/syncer"
 	"github.com/submariner-io/admiral/pkg/util"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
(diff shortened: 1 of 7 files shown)


pkg/globalnet/controllers/cluster_egressip_controller.go
```diff
@@ -173,11 +171,9 @@ func (c *clusterGlobalEgressIPController) validate(numberOfIPs int, egressIP *su
 			Reason:  "ZeroInput",
 			Message: "No global IPs to allocate",
```
sridhargaddam: The message could be confusing. Shall we use "GlobalIP not allocated"

tpantelis: I think "GlobalIP not allocated" might convey we tried to allocate but couldn't. We're trying to say that the user has specified to not allocate any so perhaps "The specified NumberOfIPs is 0"?

sridhargaddam (approved):
LGTM, thanks @tpantelis

pkg/globalnet/controllers/cluster_egressip_controller.go
sridhargaddam: Yes, "The specified NumberOfIPs is 0" looks good.

tpantelis: Done

vthapar (approved):

mangelajo (approved):

PR #1413, opened 2021-06-14
Add GN 2.0 Service controller

tpantelis:
...to delete the associated `GlobalIngressIP`, if any, to handle the case where a `Service` is deleted but not the `ServiceExport`.


sridhargaddam (approved):
Thanks for quickly addressing this requirement @tpantelis

vthapar (approved):

PR #1415, opened 2021-06-15
Reallocate global IPs if IP pool reserve fails

tpantelis:
If it can't reserve the previously allocated IP(s) on  startup, likely due to global CIDR change, empty the previously allocated IP(s) and update the Status condition appropriately so the create callback will re-allocate.


sridhargaddam (approved):
Thanks for fixing this @tpantelis

vthapar (approved):

PR #1416, opened 2021-06-15
Fix panic in GetKubeProxyClusterIPServiceChainName

tpantelis:
"_index out of range [0] with length 0_" - due to no `Port` in the `Service` spec. Thus occurs intermittently during the headless service test case if it runs long enough to reach the call to `GetKubeProxyClusterIPServiceChainName`. A `Service` can't be created without ports in K8s so modified the test accordingly.

### DIFF
--- a/pkg/globalnet/controllers/controllers_suite_test.go
+++ b/pkg/globalnet/controllers/controllers_suite_test.go
@@ -420,16 +420,11 @@ func newClusterIPService() *corev1.Service {
 	}
 }
 
-func newHeadlessService(name string) *corev1.Service {
-	return &corev1.Service{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
-		},
-		Spec: corev1.ServiceSpec{
-			ClusterIP: corev1.ClusterIPNone,
-			Type:      corev1.ServiceTypeClusterIP,
-		},
-	}
+func newHeadlessService() *corev1.Service {
+	s := newClusterIPService()
+	s.Spec.ClusterIP = corev1.ClusterIPNone
+
+	return s
 }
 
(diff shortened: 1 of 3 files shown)


vthapar:
Headless services do have ports. Can you share where exactly you ran into this error? But going by `ClusterIP` in the name, we shouldn't likely be here if it is headless service. So maybe another check that checks for type and returns if it isn't ClusterIP.

pkg/globalnet/controllers/iptables/iface.go
```diff
@@ -130,6 +130,10 @@ func (i *ipTables) GetKubeProxyClusterIPServiceChainName(service *corev1.Service
 	// CNIs that use kube-proxy with iptables for loadbalancing create an iptables chain for each service
 	// and incoming traffic to the clusterIP Service is directed into the respective chain.
 	// Reference: https://bit.ly/2OPhlwk
+	if len(service.Spec.Ports) == 0 {
```
sridhargaddam: Ideally this should never happen because we cannot create a Service without ports in K8s. 

I think we should rather fix the Headless service definition to include the following 
https://github.com/submariner-io/submariner/blob/b444546153c539d19effb3a477c45f0653b0f946/pkg/globalnet/controllers/controllers_suite_test.go#L428

```
			Ports: []corev1.ServicePort{{
				Name:       serviceName,
				Port:       int32(8080),
				TargetPort: intstr.FromInt(8080),
				Protocol:   corev1.ProtocolTCP,
			}},
```

sridhargaddam (changes requested):

sridhargaddam (approved):

vthapar (approved):

Issue #1410, opened 2021-06-14
LoadBalancer support for AWS NLB doesn't sync properly


mangelajo:
**What happened**:

The load balancer we create or the gateway doesn't sync properly to AWS:

```
  Warning  SyncLoadBalancerFailed  43s                 service-controller  Error syncing load balancer: failed to ensure load balancer: error creating load balancer target group: "DuplicateTargetGroupName: A target group with the same name 'k8s-submarin-submarin-081853df64' exists, but with different settings\n\tstatus code: 400, request id: 564d9c37-e939-46e5-9399-c774281f70a5"
  Warning  SyncLoadBalancerFailed  43s                 service-controller  Error syncing load balancer: failed to ensure load balancer: error adding tags for targetGroup arn:aws:elasticloadbalancing:us-east-2:269733383066:targetgroup/k8s-submarin-submarin-4aef34b009/84e95f12ba7d44e5 due to "TargetGroupNotFound: Target groups 'arn:aws:elasticloadbalancing:us-east-2:269733383066:targetgroup/k8s-submarin-submarin-4aef34b009/84e95f12ba7d44e5' not found\n\tstatus code: 400, request id: e6bf0782-22bc-42fc-a74c-95d867a573a5"
  Warning  SyncLoadBalancerFailed  34s                 service-controller  Error syncing load balancer: failed to ensure load balancer: error adding tags for targetGroup arn:aws:elasticloadbalancing:us-east-2:269733383066:targetgroup/k8s-submarin-submarin-55367631ea/4bb4b59590445989 due to "TargetGroupNotFound: Target groups 'arn:aws:elasticloadbalancing:us-east-2:269733383066:targetgroup/k8s-submarin-submarin-55367631ea/4bb4b59590445989' not found\n\tstatus code: 400, request id: 1556f319-6012-4291-9b6c-7a5c58b893c6"
```
I'm working on a reproducer here https://github.com/mangelajo/k8s-udp-lb-test/blob/main/reproducer.yaml still not ready

One of the issues being found is that most of the time the 2nd port (NATT discovery port) won't work or will take a long time to work. I'm thinking of a workaround for this.

sridhargaddam:
Is this a problem with AWS NLB itself or the controller wrongly configuring the AWS NLB when a Service of type Loadbalancer is created?

mangelajo:
> Is this a problem with AWS NLB itself or the controller wrongly configuring the AWS NLB when a Service of type Loadbalancer is created?

[...]

PR #1418, opened 2021-06-16
Speedup NAT detection when behind a cloud load balancer

mangelajo:
As a result of #1410, the NATT port, most of the time is not properly
exposed by the load balancer. But it's ok to assume in this case
that packets will traverse NAT.

The discovery when the other side is a load balancer backed endpoint
times out after 6 seconds, allowing 2 tries on private IP to give it
a chance. If timed out it will default to Public IP and NAT.

Related-Issue: #1410

### DIFF
--- a/pkg/natdiscovery/remote_endpoint.go
+++ b/pkg/natdiscovery/remote_endpoint.go
@@ -38,6 +38,7 @@ const (
 var (
 	recheckTime                    = (2 * time.Second).Nanoseconds()
 	totalTimeout                   = (60 * time.Second).Nanoseconds()
+	totalTimeoutLoadBalancer       = (6 * time.Second).Nanoseconds()
 	publicToPrivateFailoverTimeout = time.Second.Nanoseconds()
 )
 
@@ -47,10 +48,12 @@ type remoteEndpointNAT struct {
 	lastCheck              time.Time
 	lastTransition         time.Time
 	started                time.Time
-	useNAT                 bool
+	timeout                time.Duration
 	useIP                  string
 	lastPublicIPRequestID  uint64
 	lastPrivateIPRequestID uint64
+	useNAT                 bool
+	usingLoadBalancer      bool
 }
 
 type NATEndpointInfo struct {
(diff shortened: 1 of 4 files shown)


sridhargaddam (approved):

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -86,17 +102,26 @@ func (rn *remoteEndpointNAT) sinceLastTransition() time.Duration {
...
 func (rn *remoteEndpointNAT) hasTimedOut() bool {
-	return time.Since(rn.started) > toDuration(&totalTimeout)
+	return time.Since(rn.started) > rn.timeout
 }
 
 func (rn *remoteEndpointNAT) useLegacyNATSettings() {
-	rn.useNAT = rn.endpoint.Spec.NATEnabled
-	if rn.endpoint.Spec.NATEnabled {
+	switch {
+	case rn.usingLoadBalancer:
```
sridhargaddam, line 110: We are giving priority to LoadBalancer first, followed by Spec.NATEnabled and finally falling back to default. Since we are using a switch it might not be immediately evident for people looking at the code. Can you please add a small comment.

mangelajo: But that's an inner characteristic of switch in go. I think we don't need to explain go to the reader, I don't know ':D  let's see what other reviewers think of this.

@sridhargaddam thanks a lot for reviewing

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -86,17 +102,26 @@ func (rn *remoteEndpointNAT) sinceLastTransition() time.Duration {
...
+
+	case rn.endpoint.Spec.NATEnabled:
+		rn.useIP = rn.endpoint.Spec.PublicIP
+		rn.useNAT = rn.endpoint.Spec.NATEnabled
+		rn.transitionToState(selectedPublicIP)
 		klog.V(log.DEBUG).Infof("using NAT legacy settings for endpoint %q, using public IP %q", rn.endpoint.Spec.CableName,
 			rn.useIP)
-	} else {
+
+	default:
```
tpantelis, line 124:
```suggestion
	default:
	    rn.useNAT = false
```

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -86,17 +102,26 @@ func (rn *remoteEndpointNAT) sinceLastTransition() time.Duration {
...
+	case rn.usingLoadBalancer:
+		rn.useNAT = true
 		rn.useIP = rn.endpoint.Spec.PublicIP
 		rn.transitionToState(selectedPublicIP)
+		klog.V(log.DEBUG).Infof("using NAT for the load balancer backed endpoint %q, using public IP %q", rn.endpoint.Spec.CableName,
+			rn.useIP)
+
+	case rn.endpoint.Spec.NATEnabled:
+		rn.useIP = rn.endpoint.Spec.PublicIP
+		rn.useNAT = rn.endpoint.Spec.NATEnabled
```
mangelajo:
```suggestion
		rn.useNAT = true
```
To match @tpantelis comments, and at this point rn.endpointSpec.NATEnabled has to be true... so..

pkg/natdiscovery/remote_endpoint.go
mangelajo: Right, at this point it was working because it's false by default, but better let's be explicit.

tpantelis (approved):

PR #1425, opened 2021-06-19
Add Node controller unit tests

tpantelis:
Also fixed issues found and other minor changes. See individual commits for details,.

### DIFF
--- a/pkg/globalnet/controllers/node_controller.go
+++ b/pkg/globalnet/controllers/node_controller.go
@@ -28,6 +28,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/ipam"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


vthapar (approved):

pkg/globalnet/controllers/node_controller.go
```diff
@@ -89,19 +90,25 @@ func (n *nodeController) process(from runtime.Object, numRequeues int, op syncer
...
-				_ = n.pool.Release(existingGlobalIP)
-				return nil, false
-			} else {
-				_ = n.pool.Release(existingGlobalIP)
-				return n.updateNodeAnnotation(node, "")
-			}
+		if existingGlobalIP := node.GetAnnotations()[constants.SmGlobalIP]; existingGlobalIP != "" {
+			cniIfaceIP := node.GetAnnotations()[constants.CNIInterfaceIP]
+
+			requeue := n.flushRulesAndReleaseIPs(node.Name, numRequeues, func(allocatedIPs []string) error {
```
sridhargaddam: We dont have to flush rules when the event corresponds to a different node as those rules would have been present on that OLD Gateway node and NOT on the current Active Gateway node. We just have to release the IP back to Pool and remove the annotation.

sridhargaddam (changes requested):
Thanks @tpantelis for working on the unit tests. A small change is required.

pkg/globalnet/controllers/node_controller.go
tpantelis: ahh - right

tpantelis: Removed the commit that changed it.

sridhargaddam (approved):

PR #1427, opened 2021-06-22
Enable HeadlessService with globalnet v2

### LINKED ISSUE #732: Globalnet changes to support Headless Services
Globalnet changes to support headless services

**Why is this needed**:
Support for Headless services will be added to lighthouse in 0.6.0 This requires pod IPs to be reachable across clusters. Currently globalnet only annotates services with globalIP on the ingress side. To support Headless Services it will need to annotate pods behind a Headless Service too.

vthapar:
Fixes #732

Depends on submariner-io/submariner-operator#1443

### DIFF
--- a/pkg/globalnet/controllers/global_ingressip_controller.go
+++ b/pkg/globalnet/controllers/global_ingressip_controller.go
@@ -199,3 +199,14 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
 		return nil
 	}, ingressIP.Status.AllocatedIP)
 }
+
+func getIngressIPName(obj runtime.Object) string {
+	switch typedObj := obj.(type) {
+	case *corev1.Service:
+		return fmt.Sprintf("svc-%.59s", typedObj.Name)
+	case *corev1.Pod:
+		return fmt.Sprintf("pod-%.59s", typedObj.Name)
+	}
+
+	return ""
+}
--- a/pkg/globalnet/controllers/ingress_pod_controller.go
+++ b/pkg/globalnet/controllers/ingress_pod_controller.go
@@ -1,0 +1,141 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package controllers
+
+import (
+	"context"
+
+	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/stringset"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/util"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+const labelServiceRef = "submariner.io/serviceRef"
+
+func startIngressPodController(svc *corev1.Service, config syncer.ResourceSyncerConfig) (*ingressPodController, error) {
+	var err error
+
+	_, gvr, err := util.ToUnstructuredResource(&submarinerv1.GlobalIngressIP{}, config.RestMapper)
+	if err != nil {
+		return nil, err
+	}
+
+	controller := &ingressPodController{
+		baseSyncerController: newBaseSyncerController(),
+		ingressIPs:           config.SourceClient.Resource(*gvr),
+		svcName:              svc.Name,
+		namespace:            svc.Namespace,
+		ingressIPMap:         stringset.NewSynchronized(),
+	}
+
+	labelSelector := labels.Set(svc.Spec.Selector).AsSelector().String()
+
+	controller.resourceSyncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:                "Ingress Pod syncer",
+		ResourceType:        &corev1.Pod{},
+		SourceClient:        config.SourceClient,
+		SourceNamespace:     svc.Namespace,
+		RestMapper:          config.RestMapper,
+		Federator:           federate.NewCreateFederator(config.SourceClient, config.RestMapper, corev1.NamespaceAll),
+		Scheme:              config.Scheme,
+		Transform:           controller.process,
+		SourceLabelSelector: labelSelector,
+		ResourcesEquivalent: arePodsEqual,
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	if err := controller.Start(); err != nil {
+		return nil, err
+	}
+
+	klog.Infof("Created Pod controller for (%s/%s) with selector %q", svc.Namespace, svc.Name, labelSelector)
+
+	return controller, nil
+}
+
+func (c *ingressPodController) process(from runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	pod := from.(*corev1.Pod)
+	key, _ := cache.MetaNamespaceKeyFunc(pod)
+
+	ingressIP := &submarinerv1.GlobalIngressIP{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getIngressIPName(pod),
+			Namespace: pod.Namespace,
+			Labels: map[string]string{
+				labelServiceRef: c.svcName,
+			},
+		},
+	}
+
+	if op == syncer.Delete {
+		c.ingressIPMap.Remove(ingressIP.Name)
+		klog.Infof("ingress Pod %s for service %s deleted", key, c.svcName)
+
+		return ingressIP, false
+	}
+
+	// TODO: handle phase and podIP changes?
+	if c.ingressIPMap.Contains(ingressIP.Name) || pod.Status.Phase != corev1.PodRunning || pod.Status.PodIP == "" {
+		// Avoid assigning ingressIPs to pods that are not ready with an endpoint IP
+		return nil, false
+	}
+
+	klog.Infof("%q ingress Pod %s for service %s", op, key, c.svcName)
+
+	ingressIP.Spec = submarinerv1.GlobalIngressIPSpec{
+		Target:     submarinerv1.HeadlessServicePod,
+		ServiceRef: &corev1.LocalObjectReference{Name: c.svcName},
+		PodRef:     &corev1.LocalObjectReference{Name: pod.Name},
+	}
+
+	c.ingressIPMap.Add(ingressIP.Name)
+
+	return ingressIP, false
+}
+
+func (c *ingressPodController) cleanupIngressIPs() {
+	svcSelector := labels.SelectorFromSet(map[string]string{labelServiceRef: c.svcName}).String()
+	err := c.ingressIPs.Namespace(c.namespace).DeleteCollection(context.TODO(), metav1.DeleteOptions{},
+		metav1.ListOptions{LabelSelector: svcSelector})
+
+	if err != nil && !errors.IsNotFound(err) {
+		klog.Errorf("error deleting GlobalIngressIPs for service %s/%s: %v", c.namespace, c.svcName, err)
+	}
+}
+
+func arePodsEqual(obj1, obj2 *unstructured.Unstructured) bool {
+	phase1, _, _ := unstructured.NestedString(obj1.Object, "status", "phase")
+	phase2, _, _ := unstructured.NestedString(obj2.Object, "status", "phase")
+	podIP1, _, _ := unstructured.NestedString(obj1.Object, "status", "podIP")
+	podIP2, _, _ := unstructured.NestedString(obj2.Object, "status", "podIP")
+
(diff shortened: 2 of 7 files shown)


pkg/globalnet/controllers/pod_controller.go
```diff
@@ -0,0 +1,108 @@
+/*
```
tpantelis: How about we name the file _ingress_pod_controller_ to distinguish? Later on we can name the the other one _egress_pod_controller_.

pkg/globalnet/controllers/pod_controller.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/submariner-io/admiral/pkg/util"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog"
+)
+
+func startPodController(svc *corev1.Service, config syncer.ResourceSyncerConfig) (*podController, error) {
```
tpantelis:
```suggestion
func startIngressPodController(svc *corev1.Service, config syncer.ResourceSyncerConfig) (*podController, error) {
```

pkg/globalnet/controllers/pod_controller.go
```diff
@@ -0,0 +1,108 @@
...
+		scheme:               config.Scheme,
+		svcName:              svc.Name,
+		namespace:            svc.Namespace,
+		ingressIPMap:         make(map[string]bool),
+	}
+
+	labelSelector := labels.Set(svc.Spec.Selector).AsSelector().String()
+
+	controller.resourceSyncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:                "Pod syncer",
```
tpantelis:
```suggestion
		Name:                "Ingress Pod syncer",
```

pkg/globalnet/controllers/pod_controller.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if err := controller.Start(); err != nil {
+		return nil, err
+	}
+
+	klog.Infof("Creating Pod controller for (%s/%s) with selector %q", svc.Namespace, svc.Name, labelSelector)
```
tpantelis:
```suggestion
	klog.Infof("Created Pod controller for (%s/%s) with selector %q", svc.Namespace, svc.Name, labelSelector)
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -199,3 +199,14 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
 		return nil
 	}, ingressIP.Status.AllocatedIP)
 }
+
+func GetIngressIPName(obj runtime.Object) string {
```
tpantelis:
```suggestion
func getIngressIPName(obj runtime.Object) string {
```

pkg/globalnet/controllers/pod_controller.go
```diff
@@ -0,0 +1,108 @@
...
+	ingressIP := &submarinerv1.GlobalIngressIP{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GetIngressIPName(pod),
+			Namespace: pod.Namespace,
+		},
+	}
+
+	if op == syncer.Delete {
+		delete(c.ingressIPMap, pod.Name)
+		klog.Infof("Delete Pod %s/%s", pod.Namespace, pod.Name)
```
tpantelis:
```suggestion
		klog.Infof("Ingress pod %s/%s deleted", pod.Namespace, pod.Name)
```

pkg/globalnet/controllers/pod_controller.go
```diff
@@ -0,0 +1,108 @@
...
+
+		return ingressIP, false
+	}
+
+	if c.ingressIPMap[pod.Name] || pod.Status.Phase != corev1.PodRunning || pod.Status.PodIP == "" {
+		// Avoid assigning ingressIPs to pods that are not ready with an endpoint IP
+		return nil, false
+	}
+
+	klog.Infof("%q Pod %s/%s", op, pod.Namespace, pod.Name)
```
tpantelis:
```suggestion
	klog.Infof("%q ingress Pod %s/%s", op, pod.Namespace, pod.Name)
```

pkg/globalnet/controllers/pod_controller.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	if op == syncer.Delete {
+		delete(c.ingressIPMap, pod.Name)
+		klog.Infof("Delete Pod %s/%s", pod.Namespace, pod.Name)
+
+		return ingressIP, false
+	}
+
+	if c.ingressIPMap[pod.Name] || pod.Status.Phase != corev1.PodRunning || pod.Status.PodIP == "" {
```
tpantelis: What is the intent of the `ingressIPMap` check here?

pkg/globalnet/controllers/types.go
```diff
@@ -129,3 +130,12 @@ type nodeController struct {
 	nodeName string
 	nodes    dynamic.ResourceInterface
 }
+
+type podController struct {
```
tpantelis:
```suggestion
type ingressPodController struct {
```

pkg/globalnet/controllers/types.go
```diff
@@ -129,3 +130,12 @@ type nodeController struct {
...
+	*baseSyncerController
+	pods         dynamic.NamespaceableResourceInterface
+	scheme       *runtime.Scheme
+	svcName      string
+	namespace    string
+	ingressIPMap map[string]bool
```
tpantelis: Looks like this can be a set.

```suggestion
	ingressIPs        stringset.Interface
```

tpantelis (changes requested):

pkg/globalnet/controllers/global_ingressip_controller.go
vthapar: We'll need to call it from `_test.go` code too.

pkg/globalnet/controllers/pod_controller.go
vthapar: This is to check if we have already created globalIngressIP for this pod. A pod may get further updates but we're not interested if we already have assigned it a globalIngressIP

tpantelis: We can ignore pod updates - you're using the create-only federator anyway as there's nothing to update on the `GlobalIngressIP`.

pkg/globalnet/controllers/types.go
```diff
@@ -129,3 +130,12 @@ type nodeController struct {
 	nodeName string
 	nodes    dynamic.ResourceInterface
 }
+
+type podController struct {
+	*baseSyncerController
+	pods         dynamic.NamespaceableResourceInterface
+	scheme       *runtime.Scheme
```
tpantelis: `ingressIPs` and  `scheme` are unused.

pkg/globalnet/controllers/ingress_pod_controller.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return ingressIP, false
+}
+
+func (c *ingressPodController) CleanupIngressIPs() {
+	// TODO: Use DeleteCollection()
+	for _, ingressIPName := range c.ingressIPMap.Elements() {
+		err := c.ingressIPs.Namespace(c.namespace).Delete(context.TODO(), ingressIPName, metav1.DeleteOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			klog.Errorf("error deleting globalIngressIP %s/%s %v", c.namespace, ingressIPName, err)
```
tpantelis:
```suggestion
			klog.Errorf("Error deleting GlobalIngressIP %s/%s: %v", c.namespace, ingressIPName, err)
```

pkg/globalnet/controllers/ingress_pod_controller.go
```diff
@@ -0,0 +1,123 @@
...
+		ServiceRef: &corev1.LocalObjectReference{Name: c.svcName},
+		PodRef:     &corev1.LocalObjectReference{Name: pod.Name},
+	}
+
+	c.ingressIPMap.Add(ingressIP.Name)
+
+	return ingressIP, false
+}
+
+func (c *ingressPodController) CleanupIngressIPs() {
```
tpantelis:
```suggestion
func (c *ingressPodController) cleanupIngressIPs() {
```

pkg/globalnet/controllers/ingress_pod_controller.go
```diff
@@ -0,0 +1,123 @@
...
+	// TODO: Use DeleteCollection()
```
tpantelis: There is a small window where this could get called prior to the ingressIPMap being (fully) populated. You could use the resource syncer cache (`GetResources`) which is fully populated on startup.

tpantelis (changes requested):

pkg/globalnet/controllers/ingress_pod_controller.go
```diff
@@ -0,0 +1,127 @@
...
+			Namespace: pod.Namespace,
+			Labels: map[string]string{
+				labelServiceRef: c.svcName,
+			},
+		},
+	}
+
+	if op == syncer.Delete {
+		c.ingressIPMap.Remove(ingressIP.Name)
+		klog.Infof("ingress Pod %s/%s deleted", pod.Namespace, pod.Name)
```
sridhargaddam: nit: We can use the following in Line 83 and 
`key, _ := cache.MetaNamespaceKeyFunc(pod)`
and print the key at Line 96, 106

sridhargaddam: Also, please include service name as well in the log message.

pkg/globalnet/controllers/ingress_pod_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+		return ingressIP, false
+	}
+
+	if c.ingressIPMap.Contains(ingressIP.Name) || pod.Status.Phase != corev1.PodRunning || pod.Status.PodIP == "" {
+		// Avoid assigning ingressIPs to pods that are not ready with an endpoint IP
+		return nil, false
+	}
+
+	klog.Infof("%q ingress Pod %s/%s", op, pod.Namespace, pod.Name)
```
sridhargaddam: Might help to have service name as well while debugging issues.

pkg/globalnet/controllers/ingress_pod_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	if op == syncer.Delete {
+		c.ingressIPMap.Remove(ingressIP.Name)
+		klog.Infof("ingress Pod %s/%s deleted", pod.Namespace, pod.Name)
+
+		return ingressIP, false
+	}
+
+	if c.ingressIPMap.Contains(ingressIP.Name) || pod.Status.Phase != corev1.PodRunning || pod.Status.PodIP == "" {
```
sridhargaddam, line 106: We should also use `ResourcesEquivalent: syncer.AreSpecsEquivalent` while creating the resourceSyncer to ignore certain update events for the Pod. Otherwise, even when a Pod is running and has an IP, when a label/annotation on a Pod is updated, we will be notified of the event.

pkg/globalnet/controllers/service_export_controller.go
```diff
@@ -168,11 +184,23 @@ func (c *serviceExportController) getService(name, namespace string) (*corev1.Se
...
 }
+
+func (c *serviceExportController) onCreateHeadless(key string, service *corev1.Service) (runtime.Object, bool) {
+	controller, err := startIngressPodController(service, c.config)
+	if err != nil {
+		klog.Errorf("Failed to create pod controller for service %q", key)
+		return nil, true
+	}
+
+	c.podControllers.Store(key, controller)
```
sridhargaddam, line 197: I think we should use Lock here

pkg/globalnet/controllers/service_export_controller.go
```diff
@@ -148,6 +157,13 @@ func (c *serviceExportController) onDelete(serviceExport *mcsv1a1.ServiceExport)
 
 	klog.Infof("ServiceExport %q deleted", key)
 
+	if obj, found := c.podControllers.Load(key); found {
```
sridhargaddam: Lock needed

pkg/globalnet/controllers/service_export_controller.go
```diff
@@ -76,6 +76,16 @@ func NewServiceExportController(config syncer.ResourceSyncerConfig) (Interface,
 	return controller, nil
 }
 
+func (c *serviceExportController) Stop() {
+	c.baseController.Stop()
+
```
sridhargaddam, line 81: I think we should use Lock/Unlock here

sridhargaddam (changes requested):
Thanks for working on this use-case @vthapar

pkg/globalnet/controllers/service_export_controller.go
vthapar: This is a syncmap, so explicit locking not needed.

pkg/globalnet/controllers/service_export_controller.go
vthapar: It is a syncmap.

pkg/globalnet/controllers/ingress_pod_controller.go
tpantelis: Wouldn't we want to  check equivalence of `Status.Phase` and `Status.PodIP`?

vthapar: Added pod equivalency check. Only fields that matter for us are podIP and phase. Changes to any other fields, including spec are irrelevant if they don't impact these two fields.

tpantelis (approved):

pkg/globalnet/controllers/service_controller.go
```diff
@@ -70,6 +71,10 @@ func (c *serviceController) onDelete(service *corev1.Service) (runtime.Object, b
 
 	klog.Infof("Service %q deleted", key)
 
+	if _, exists, _ := c.svcExController.resourceSyncer.GetResource(service.Name, service.Namespace); exists {
+		c.svcExController.cleanupPodController(key)
+	}
```
tpantelis, line 76: I think we should avoid (direct) coupling of the controllers. I think it would be better to create a pod controllers type to manage the map that is shared between the two controllers. But this is fine for now - let's get this in - I can look into refactoring after.

sridhargaddam (commented):
Thanks for addressing the review comments.

PR #1453, opened 2021-07-01
Fix race condition in globalEgressIPController tests

tpantelis:
The `podWatcher` is created asynchronously and may still be starting in the background when the next test is started causing a race when the `podWatcher` tries to create the ipset interface via the NewFunc hook.

To prevent this, pass the ipset interface to the podWatcher. The `podWatcher` already is passed the ipset name so instead of also passing the interface, a new `Named` ipset interface was added to encapsulate the named ipset and the general interface. Also modified the `globalEgressIPController` to use the `Named` ipset.


skitt (approved):

vthapar (approved):

sridhargaddam (approved):
LGTM, thanks @tpantelis

PR #1454, opened 2021-07-02
Fix issues with MTU settings

aswinsuryan:
When the clusters have different MTU configures,  some of
the e2e tests are failing.  TCPMTU probe is now enabled to
fix this on all route agents.

### DIFF
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
@@ -1,0 +1,58 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package mtu
+
+import (
+	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner/pkg/event"
+	"github.com/submariner-io/submariner/pkg/netlink"
+)
+
+type mtuHandler struct {
+	event.HandlerBase
+}
+
+func NewMTUHandler() event.Handler {
+	return &mtuHandler{}
+}
+
+func (h *mtuHandler) GetNetworkPlugins() []string {
+	return []string{event.AnyNetworkPlugin}
+}
+
+func (h *mtuHandler) GetName() string {
+	return "MTU handler"
+}
+
+func (h *mtuHandler) Init() error {
+	klog.Infof("Updating configurations for Path MTU")
+	configureTCPMTUProbe()
+
+	return nil
+}
+
+func configureTCPMTUProbe() {
+	mtuProbe := "2"
+	baseMss := "1024"
+
+	err := netlink.New().ConfigureTCPMTUProbe(mtuProbe, baseMss)
+	if err != nil {
+		klog.Warningf(err.Error())
+	}
+}
--- a/pkg/routeagent_driver/main.go
+++ b/pkg/routeagent_driver/main.go
@@ -39,6 +39,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/environment"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/kubeproxy"
+	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/mtu"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/ovn"
 )
 
@@ -88,6 +89,7 @@ func main() {
 		ovn.NewHandler(env, smClientset),
 		cabledriver.NewXRFMCleanupHandler(),
 		cabledriver.NewVXLANCleanup(),
+		mtu.NewMTUHandler(),
 	); err != nil {
 		klog.Fatalf("Error registering the handlers: %s", err.Error())
 	}


nyechiel:
@mangelajo @sridhargaddam appreciate your review here. We also need to see if it makes sense to implement this on other cable drivers and backport this to release-0.9 and 0.8?

mangelajo (changes requested):
But the issue we had was OVN specific, right?

We have this here for OVN, which will only be triggered with that one, and not others:

https://github.com/submariner-io/submariner/blob/devel/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go#L118

Also --clamp-mss-to-pmtu, did you check that it fixes it?, In the case of OVN that didn't fly (at least the way I tried).

aswinsuryan:
> he issue we had was OVN specific, right?

We have a similar issue in 0.9/0.8 OpenshiftSDN downstream CI which uses an OCP cluster and AWS cluster. I did not see this issue when I tried with two AWS clusters with d/s build, may because both sides had the same MTU configurations.

But when I tried with a KIND cluster with  u/s build and AWS with u/s build I was able to reproduce the issue, 3 test cases were failing due to the MTU issue and after adding these rules it got fixed.

With OVN if it is d/s CI that fails, may this fix along with your fix may solve it as one side of the downstream deployments uses OpenshiftSDN

pkg/cable/libreswan/libreswan.go
```diff
@@ -131,6 +135,11 @@ func (i *libreswan) Init() error {
 		return fmt.Errorf("error starting Pluto: %v", err)
 	}
 
+	i.ipt, err = iptables.New()
```
sridhargaddam: It would be better to include a wait for 5 seconds which will help in situations when lock cannot be immediately acquired.

You can do the following
`iptables.New(iptables.IPFamily(iptables.ProtocolIPv4), iptables.Timeout(5))`

pkg/cable/libreswan/libreswan.go
```diff
@@ -584,3 +596,33 @@ func (i *libreswan) runPluto() error {
...
 }
+
+func (i *libreswan) addClampMssToPathMTU(subnets string) {
+	ruleSpecSource := []string{"-s", subnets, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j",
+		"TCPMSS", "--clamp-mss-to-pmtu"}
+	ruleSpecDest := []string{"-d", subnets, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j",
+		"TCPMSS", "--clamp-mss-to-pmtu"}
+
+	if err := i.ipt.AppendUnique(mangleTable, postRoutingChain, ruleSpecSource...); err != nil {
+		klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpecSource, " "), err)
```
sridhargaddam: Although not frequent, we have seen instances where programming iptables rules fail sometimes. It would be better if we include some retry mechanism.

sridhargaddam (changes requested):

pkg/cable/libreswan/libreswan.go
aswinsuryan: I am using the API from `github.com/submariner-io/submariner/pkg/iptables`. Seems like it is already being done there.

pkg/cable/libreswan/libreswan.go
aswinsuryan: Done.

pkg/cable/libreswan/libreswan.go
```diff
@@ -584,3 +607,55 @@ func (i *libreswan) runPluto() error {
...
+			}
+		}
+
+		if err == nil {
+			break
+		}
+	}
+
+	return err
+}
```
skitt: Please don’t use `bool` arguments to control behaviour. Instead, pass in the function:

```suggestion
func (i *libreswan) updateMangleTableEntryWithRetry(updater func(table, chain string, rulespec ...string) error, rulespec ...string) error {
	var err error

	for retries := 1; retries < 5; retries++ {
		if err = updater(mangleTable, subMarinerPostRoutingChain, rulespec...); err != nil {
			time.Sleep(5 * time.Second)
		} else {
			break
		}
	}

	return err
}
```

Call this using either `i.updateMangleTableEntryWithRetry(i.ipt.AppendUnique, ...)` or `i.updateMangleTableEntryWithRetry(i.ipt.DeleteUnique, ...)`.

skitt (changes requested):

pkg/cable/libreswan/libreswan.go
```diff
@@ -584,3 +607,55 @@ func (i *libreswan) runPluto() error {
...
+		} else {
+			if err = i.ipt.Delete(mangleTable, subMarinerPostRoutingChain, rulespec...); err != nil {
+				time.Sleep(5 * time.Second)
+			}
+		}
+
+		if err == nil {
+			break
+		}
+	}
```
tpantelis: We could use `retry.OnError` here. However perhaps we should just propagate the error to the cable engine which will retry.

tpantelis (changes requested):

pkg/cable/libreswan/libreswan.go
```diff
@@ -584,3 +607,55 @@ func (i *libreswan) runPluto() error {
...
+		klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpecSource, " "), err)
+	}
+
+	if err := i.updateMangleTableEntryWithRetry(true, ruleSpecDest...); err != nil {
+		klog.Errorf("error appending iptables rule \"%s\": %v\n", strings.Join(ruleSpecDest, " "), err)
+	}
+}
+
+func (i *libreswan) removeClampMssToPathMTU(subnets string) {
+	ruleSpecSource := []string{"-s", subnets, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j",
```
sridhargaddam: If I remember correctly, @mangelajo wanted to verify the behavior once why PMTU is not working (aka ICMP packets are not delivered). Are we planning to get this PR in or wait for @mangelajo to update his findings?

Another thing we discussed during the call was to check if this is applicable to all the CableDrivers?

Anyways, we also need a similar fix in submariner-operator code as we are using iptables for the first time in gateway pod - https://github.com/submariner-io/submariner-operator/commit/0448fcd60972da03c84c345c24c2cf1c34b6b42b

sridhargaddam (changes requested):

pkg/cable/libreswan/libreswan.go
aswinsuryan: Done

pkg/cable/libreswan/libreswan.go
aswinsuryan: > If I remember correctly, @mangelajo wanted to verify the behavior once why PMTU is not working (aka ICMP packets are not delivered). Are we planning to get this PR in or wait for @mangelajo to update his findings?

Maybe since the d/s release is nearing we can try this fix as this should be harmless. @mangelajo @nyechiel thoughts?
> 
> Another thing we discussed during the call was to check if this is applicable to all the CableDrivers?
I think the idea was to see if this fix works in downstream CI before porting it to other cable drivers.
> 
> Anyways, we also need a similar fix in submariner-operator code as we are using iptables for the first time in gateway pod - [submariner-io/submariner-operator@0448fcd](https://github.com/submariner-io/submariner-operator/commit/0448fcd60972da03c84c345c24c2cf1c34b6b42b)

I will check and raise a PR if required.

pkg/cable/libreswan/libreswan.go
aswinsuryan: Done.

pkg/cable/libreswan/libreswan.go
```diff
@@ -131,6 +139,21 @@ func (i *libreswan) Init() error {
 		return fmt.Errorf("error starting Pluto: %v", err)
 	}
 
+	i.ipt, err = iptables.New()
+	if err != nil {
+		return fmt.Errorf("error initializing iptables: %v", err)
+	}
+
+	if err := util.CreateChainIfNotExists(i.ipt, mangleTable, subMarinerPostRoutingChain); err != nil {
+		klog.Errorf("error creating iptables chain %s: %v", subMarinerPostRoutingChain, err)
```
tpantelis: It seems like we should propagate this error. Same on L154.

pkg/cable/libreswan/libreswan.go
```diff
@@ -584,3 +610,51 @@ func (i *libreswan) runPluto() error {
...
+		Factor:   2,
+		Steps:    4,
+	}
+
+	err := retry.OnError(retryConfig, func(err error) bool {
+		return true
+	}, func() error {
+		err := updater(mangleTable, subMarinerPostRoutingChain, rulespec...)
+		return err
+	})
```
tpantelis: As indicated earlier, I think it would be better to propagate the error and let the cable engine handle retry async so we don't block it. In that case, we don't need really this function.

tpantelis (changes requested):

pkg/cable/libreswan/libreswan.go
aswinsuryan: Done.

pkg/cable/libreswan/libreswan.go
aswinsuryan: Done.

mangelajo:
I believe we can merge now but we should put it in the route agent I believe .

sridhargaddam:
> I believe we can merge now but we should put it in the route agent I believe .

We can keep it in route-agent if we are sure that its required for all the CableDrivers.

sridhargaddam:
> > I believe we can merge now but we should put it in the route agent I believe .
> 
> We can keep it in route-agent if we are sure that its required for all the CableDrivers.

I think we can merge it based on the current understanding of the issue and improve it in subsequent PRs

pkg/cable/libreswan/libreswan.go
```diff
@@ -584,3 +618,37 @@ func (i *libreswan) runPluto() error {
...
+	ruleSpecSource := []string{"-s", subnets, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j",
+		"TCPMSS", "--clamp-mss-to-pmtu"}
+	ruleSpecDest := []string{"-d", subnets, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j",
+		"TCPMSS", "--clamp-mss-to-pmtu"}
+
+	if err := i.ipt.Delete(mangleTable, postRoutingChain, ruleSpecSource...); err != nil {
+		return fmt.Errorf("error deleting iptables rule \"%s\": %v", strings.Join(ruleSpecSource, " "), err)
+	}
+
+	if err := i.ipt.Delete(mangleTable, postRoutingChain, ruleSpecDest...); err != nil {
```
sridhargaddam, line 649: looks like its duplicated

tpantelis (dismissed):

sridhargaddam (dismissed):

nyechiel (changes requested):
After looking more at the problem reported, please let's not merge this change without additional tests first. I feel like it's too risky.

We need to test this in the following scenarios:

1. On-prem (OpenShiftSDN) to AWS (OpenShiftSDN)
2. AWS Region1/MTU1 (OpenShiftSDN) to AWS Region 2/MTU2 (OpenShiftSDN)

mangelajo:
Let's refactor it as a route-agent handler (so we can include it with ovn) and so it works with other cable drivers.

skitt (dismissed):

skitt (approved):

mangelajo (approved):

mangelajo:
I'm not sure if this affects all the pods using networking, or only the hostNetwork.

Let's see how it flies.

nyechiel (approved):

sridhargaddam:
@aswinsuryan @mangelajo do we really need this fix on all the nodes in the cluster? Is it not sufficient if its updated on the active Gateway node?

aswinsuryan:
> @aswinsuryan @mangelajo do we really need this fix on all the nodes in the cluster? Is it not sufficient if it's updated on the active Gateway node?

Yes, we need the fix in all the clusters as host networking traffic from non-g/w node failed otherwise.

PR #1455, opened 2021-07-03
Add E2E test for external connectivity with globalnet

mkimuram:
This PR adds E2E test for external connectivity with globalnet.


mkimuram:
@sridhargaddam 

This is still WIP. TODOs are:
- [x] Move codes for creating Service without selector and endpoints to shipyard framework ([here](https://github.com/submariner-io/shipyard/pull/595))
- [x] Support Service without selector in globalnet v2
- [x] Make External network work with globalnet v2

Note that Service without selector is expected to be used for accessing from pod to external app, accessing from external app to pod looks already working fine.

mkimuram:
Updated this PR's codes like [this](https://github.com/mkimuram/submariner/tree/extglobalnet2) in different branch. I confirmed that it works with https://github.com/submariner-io/shipyard/pull/595. Once the dependent PR is merged, I will apply the same change to this PR.

As for the TODOs in the first comment:
- Support Service without selector in globalnet v2 --> works well as is.
- Make External network work with globalnet v2 --> works well on other than the cluster directly connected to external network (Source IP isn't GlobalIP in the cluster).

Also, network pod needs to be scheduled on GatewayNode in the cluster1 like [this](https://github.com/mkimuram/submariner/blob/extglobalnet2/test/external/dataplane/connectivity.go#L148) to work properly. Is it just because there is only one worker node in cluster1?

mangelajo (approved):

mangelajo (approved):

tpantelis (approved):

skitt (approved):

PR #1456, opened 2021-07-05
Datapath support for headless services

sridhargaddam:
This PR implements datapath support for Headless Service Pods.
For each of the backend Pod we will create a Ingress object
and the corresponding globalIP can be read from the object.
This PR also handles the reconsiliation code which will validate
if any of the Headless Service backend pods were deleted while
the globalnet Pod was down and handles it. Unit tests and e2e tests
will be added in subsequent PRs.

Fixes issue: https://github.com/submariner-io/submariner/issues/732


mangelajo (approved):
I haven't following all patches, but with reduced context it looks good to me.

tpantelis (approved):

PR #1458, opened 2021-07-07
Remove makefile special args

mkolesnik:
Instead, use `using` notation to run external network tests by specifying
`using=external-net`.


Makefile
```diff
@@ -24,6 +21,12 @@ else
 override CLUSTER_SETTINGS_FLAG = --cluster_settings $(DAPPER_SOURCE)/scripts/cluster_settings
 endif
 
+ifneq (,$(filter external-net,$(_using)))
+override E2E_ARGS += --testdir test/external
+override DEPLOY_ARGS += --plugin scripts/e2e/external/hook
+override CLEANUP_ARGS += --plugin scripts/e2e/external/hook
```
skitt, line 27: I know this practice isn’t new to this patch, but what’s the reason behind using `override` when appending to a variable?

mkolesnik: Without it weird stuff happens, I think this is the explanation (but since we started using it a long time ago it might be something else):
https://www.gnu.org/software/make/manual/html_node/Override-Directive.html

skitt: Ah yes, sorry, I’d forgotten about the requirement when we want to add to a command-line-specified variable!

skitt (approved):

tpantelis (approved):

PR #1464, opened 2021-07-09
Use SHAs for all GHA versions

dfarrell07:
Per GitHub's security guidelines, GHAs should be pinned using full
length commit SHAs instead of tags.

The SHAs are from the latest commit of each GHA, which also updates the
GHAs from their tagged versions.

### DIFF
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@25a956c84d5dd820d28caab9f86b8d183aeeff3d
       - name: Delete current CODEOWNERS file
         run: rm CODEOWNERS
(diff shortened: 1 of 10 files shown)


skitt (approved):

mkolesnik (commented):
For ease of use perhaps we can trust actions from `actions` as it's an official github repo and governed by them?

tpantelis (approved):

dfarrell07:
> For ease of use perhaps we can trust actions from `actions` as it's an official github repo and governed by them?

We could trust that action, but on the other hand:

1. It's [basically-never-substantially updated](https://github.com/actions/checkout/commits/main). It's unlikely we will need to update the SHA.
2. Is there really a big difference between trusting [GitHub employees](https://github.com/orgs/actions/people) not to get hacked and other popular GHA authors? I don't think we're only defending against their intentions. Even if we trust them, we should verify we're getting what they mean to be shipping.
3. It's not more usable to copy "v2" around than to copy a SHA (except for the free updates, but see points 1 and 2).

PR #1473, opened 2021-07-12
Add "Updated" GlobalEgressIPConditionType

tpantelis:
If the pod selector is updated we add a `Condition` to the status indicating it's not supported but with type `GlobalEgressIPAllocated` and status `False`. This is misleading b/c the allocated status did not change. So added a new type `GlobalEgressIPUpdated` for this case and, at the same time, removed the unused `GlobalEgressIPConflict` type.


skitt (approved):

pkg/apis/submariner.io/v1/types.go
```diff
@@ -228,7 +228,7 @@ type GlobalEgressIPConditionType string
 
 const (
 	GlobalEgressIPAllocated GlobalEgressIPConditionType = "Allocated"
-	GlobalEgressIPConflict  GlobalEgressIPConditionType = "Conflict"
+	GlobalEgressIPUpdated   GlobalEgressIPConditionType = "Updated"
```
dfarrell07, line 231: The GitHub diff UI might be tricking me, but it seems like you're loosing `GlobalEgressIPConflict` here? I also see it doesn't seem to be used elsewhere, which I guess is why nothing breaks, but that also seems weird because I imagined an unused variable would be flagged. :confused:

tpantelis: It was unused and I don't think we'll have a use for it. It's not flagged b/c it's public/exported so the linter can 't know if it's used or not.

dfarrell07: Okay, make sense. Might be good to flag the removal in the commit message IMHO.

dfarrell07 (approved):

sridhargaddam (approved):

PR #1474, opened 2021-07-12
Move the iptables tools to the iptables package

skitt:
... to avoid a dependency from the util package on iptables.


tpantelis (approved):

tpantelis:
@skitt Thanks - this was on my list of things to do.

skitt:
> @skitt Thanks - this was on my list of things to do.

Excellent! I tried to make it better still by having the functions tied to the interface but the mix of uses means it’s not straightfoward, unless we force everyone to use the wrapper, and then we might run into cycles (I didn’t push far enough to check).

tpantelis:
> > @skitt Thanks - this was on my list of things to do.
> 
> Excellent! I tried to make it better still by having the functions tied to the interface but the mix of uses means it’s not straightfoward, unless we force everyone to use the wrapper, and then we might run into cycles (I didn’t push far enough to check).

They're utilities that use the Interface. We add could add them to the interface but we wouldn't want the fake to also have to implement them. In that case we could have a common base implementation but i don't think that's necessary. I think it's good enough that they're in the proper package now 😄

dfarrell07 (approved):

PR #1479, opened 2021-07-13
Configure TCPMTU probe instead of iptable rules

aswinsuryan:
This enables TCP MTU probing by default with a base mss of 1024.

### DIFF
--- a/pkg/routeagent_driver/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/mtu/mtuhandler.go
@@ -1,0 +1,72 @@
+package mtu
+
+import (
+	"bytes"
+	"io/ioutil"
+
+	"k8s.io/klog"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/submariner/pkg/event"
+)
+
+type mtuHandler struct {
+	event.HandlerBase
+}
+
+func NewMTUHandler() event.Handler {
+	return &mtuHandler{}
+}
+
+func (h *mtuHandler) GetNetworkPlugins() []string {
+	return []string{event.AnyNetworkPlugin}
+}
+
+func (h *mtuHandler) GetName() string {
+	return "MTU handler"
+}
+
+func (h *mtuHandler) Init() error {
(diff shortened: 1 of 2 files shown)


pkg/routeagent_driver/mtu/mtuhandler.go
```diff
@@ -0,0 +1,72 @@
...
+	// Ignore leading and terminating newlines
+	existing = bytes.Trim(existing, "\n")
+
+	if bytes.Equal(existing, contents) {
+		return nil
+	}
+	// Permissions are already 644, the files are never created
+	// #nosec G306
+	return ioutil.WriteFile(path, contents, 0644)
+}
```
skitt, line 72: Can these be shared with `netlink.go`?

skitt (changes requested):
Is this “instead of iptable rules”? I don’t see any removal of `iptables` calls.

pkg/routeagent_driver/mtu/mtuhandler.go
skitt: Or just use `netlink.New().ConfigureTCPMTUProbe(...)`.

aswinsuryan: @skitt I guess the API is missing in 0.8.0, I am using that for other branches.

aswinsuryan: I took this  from https://github.com/submariner-io/submariner/tree/devel/pkg/netlink

skitt: Oops, sorry, I missed that this was for 0.8!

skitt (approved):

dfarrell07 (approved):

tpantelis (approved):

aswinsuryan:
Need to wait for the decision on https://github.com/submariner-io/submariner/pull/1480, before merge.

aswinsuryan:
This is PR does not seems to fix the Openshift AWS- OSP MTU issue. Hence closing.

PR #1482, opened 2021-07-13
E2E: Headless Services with globalnet

vthapar:

### DIFF
--- a/test/e2e/dataplane/tcp_gn_pod_connectivity.go
+++ b/test/e2e/dataplane/tcp_gn_pod_connectivity.go
@@ -51,7 +51,7 @@ var _ = Describe("[dataplane-globalnet] Basic TCP connectivity tests across over
 
 	When("a pod connects via TCP to the globalIP of a remote service", func() {
 		BeforeEach(func() {
-			toEndpointType = tcp.GlobalIP
+			toEndpointType = tcp.GlobalServiceIP
 			networking = framework.PodNetworking
 		})
 
@@ -74,7 +74,7 @@ var _ = Describe("[dataplane-globalnet] Basic TCP connectivity tests across over
 
 	When("a pod with HostNetworking connects via TCP to the globalIP of a remote service", func() {
 		BeforeEach(func() {
-			toEndpointType = tcp.GlobalIP
+			toEndpointType = tcp.GlobalServiceIP
 			networking = framework.HostNetworking
 		})
 
@@ -86,4 +86,20 @@ var _ = Describe("[dataplane-globalnet] Basic TCP connectivity tests across over
 			verifyInteraction(framework.GatewayNode, framework.NonGatewayNode)
 		})
 	})
+
+	When("a pod connects via TCP to the globalIP of a remote headless service", func() {
+		BeforeEach(func() {
+			toEndpointType = tcp.GlobalPodIP
+			networking = framework.PodNetworking
+		})
+
+		When("the pod is not on a gateway and the remote service is not on a gateway", func() {
+			verifyInteraction(framework.NonGatewayNode, framework.NonGatewayNode)
+		})
+
+		When("the pod is on a gateway and the remote service is on a gateway", func() {
+			verifyInteraction(framework.GatewayNode, framework.GatewayNode)
+		})
+	})
+
 })
--- a/test/e2e/framework/dataplane.go
+++ b/test/e2e/framework/dataplane.go
@@ -28,6 +28,11 @@ import (
 	"github.com/submariner-io/shipyard/test/e2e/tcp"
 	"github.com/submariner-io/submariner/pkg/globalnet/constants"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+const (
+	testAppLabel = "test-app"
 )
 
 func VerifyDatapathConnectivity(p tcp.ConnectivityTestParams, globalnetEnabled bool) {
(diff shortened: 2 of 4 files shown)


test/e2e/framework/dataplane.go
```diff
@@ -125,3 +141,74 @@ func execCmdInBash(p tcp.ConnectivityTestParams, cmd []string, pod *v1.Pod) (str
...
 	return p.Framework.ExecWithOptions(execOptions, p.FromCluster)
 }
+
+func createTCPHeadlessService(cluster framework.ClusterIndex, selectorName string, port int, namespace string) *v1.Service {
+	tcpService := v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("test-headless-svc-%s", selectorName),
+		},
+		Spec: v1.ServiceSpec{
+			Type:      "ClusterIP",
```
sridhargaddam:
```suggestion
			Type:      v1.ServiceTypeClusterIP,
```

test/e2e/dataplane/tcp_gn_pod_connectivity.go
```diff
@@ -85,4 +85,28 @@ var _ = Describe("[dataplane-globalnet] Basic TCP connectivity tests across over
...
+			verifyInteraction(framework.NonGatewayNode, framework.NonGatewayNode)
+		})
+
+		When("the pod is not on a gateway and the remote service is on a gateway", func() {
+			verifyInteraction(framework.NonGatewayNode, framework.GatewayNode)
+		})
+
+		When("the pod is on a gateway and the remote service is not on a gateway", func() {
+			verifyInteraction(framework.GatewayNode, framework.NonGatewayNode)
+		})
```
sridhargaddam: Since 
```suggestion
```

Since we are adding many new tests with GN2.0, we can avoid certain tests to limit the e2e test duration, particularly the ones we believe will be covered by other tests. So, lets remove these two tests.

sridhargaddam (changes requested):
Thanks for the PR @vthapar. Some minor comments, otherwise it looks good.

test/e2e/framework/dataplane.go
```diff
@@ -125,3 +141,74 @@ func execCmdInBash(p tcp.ConnectivityTestParams, cmd []string, pod *v1.Pod) (str
...
+			err = services.Delete(context.TODO(), tcpService.Name, metav1.DeleteOptions{})
+			if err != nil {
+				return nil, err
+			}
+
+			service, err = services.Create(context.TODO(), &tcpService, metav1.CreateOptions{})
+		}
+
+		return service, err
+	}, framework.NoopCheckResult).(*v1.Service)
```
tpantelis: You should be able to use the `util.CreateAnew` function in admiral.

test/e2e/framework/dataplane.go
```diff
@@ -125,3 +141,74 @@ func execCmdInBash(p tcp.ConnectivityTestParams, cmd []string, pod *v1.Pod) (str
...
+		podList := awaitPodsByLabelSelector(p.ToCluster, labels.Set(service.Spec.Selector).AsSelector().String(),
+			service.Namespace, 1)
+		ingressIPName := fmt.Sprintf("pod-%s", podList.Items[0].Name)
+		return p.Framework.AwaitGlobalIngressIP(p.ToCluster, ingressIPName, service.Namespace)
+	}
+
+	return ""
+}
+
+func awaitPodsByLabelSelector(cluster framework.ClusterIndex, labelSelector, namespace string, expectedCount int) *v1.PodList {
```
tpantelis: I assume this will be moved to shipyard?

test/e2e/framework/dataplane.go
```diff
@@ -97,7 +113,7 @@ func verifyGlobalnetDatapathConnectivity(p tcp.ConnectivityTestParams) {
 	Expect(listenerPod.TerminationMessage).To(ContainSubstring(connectorPod.Config.Data))
 	Expect(stdOut).To(ContainSubstring(listenerPod.Config.Data))
 
-	if p.ToEndpointType == tcp.GlobalIP {
+	if p.ToEndpointType == tcp.GlobalServiceIP || p.ToEndpointType == tcp.GlobalPodIP {
```
tpantelis: nit but L72-80 above assume `p.ToEndpointType` is one of these two so these checks shouldn't be necessary.

test/e2e/framework/dataplane.go
vthapar: Yes. This and maybe HeadlessService too.

test/e2e/framework/dataplane.go
vthapar: > You should be able to use the `util.CreateAnew` function in admiral.

`CreateAnew` takes `resource.Interface` as client argument while services here is `ServiceInterface` which is just a generic interface. Also, `CreateAnew` doesn't return the created object [which not an issue for now as we don't use any generated fields]

tpantelis: > `CreateAnew` takes `resource.Interface` as client argument while services here is `ServiceInterface` which is just a generic interface. 

You can use `resource.ForService` to get a Service `resource.Interface`.

> Also, `CreateAnew` doesn't return the created object [which not an issue for now as we don't use any generated fields]

https://github.com/submariner-io/admiral/pull/253

sridhargaddam (approved):

test/e2e/framework/dataplane.go
tpantelis: I think we should just put it in shipyard now rather than adding technical debt - there's no rush in getting this PR in.

test/e2e/framework/dataplane.go
tpantelis: Actually if we put this in shipyard then we can't use admiral. https://github.com/submariner-io/shipyard/pull/595 adds a function for this so we can use that.

tpantelis (approved):

PR #1485, opened 2021-07-14
Add UT for MTUHandler

aswinsuryan:


pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -24,6 +24,11 @@ import (
 	"github.com/submariner-io/submariner/pkg/netlink"
 )
 
+const (
+	mtuProbe = "2"
```
sridhargaddam, line 30: nit: A comment about what "2" means would help users understand what we are trying to achieve.

sridhargaddam (approved):

pkg/routeagent_driver/handlers/mtu/mtuhandler_suite_test.go
```diff
@@ -0,0 +1,35 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/klog"
+)
+
+func init() {
+	klog.InitFlags(nil)
+}
+
+func TestKubeProxyIPTables(t *testing.T) {
```
tpantelis:
```suggestion
func TestMTU(t *testing.T) {
```

pkg/netlink/fake/netlink.go
```diff
@@ -231,6 +234,13 @@ func (n *NetLink) EnableLooseModeReversePathFilter(interfaceName string) error {
 }
 
 func (n *NetLink) ConfigureTCPMTUProbe(mtuProbe, baseMss string) error {
+	if mtuProbe == "" || baseMss == "" {
+		return fmt.Errorf("invalid arguments passed")
+	}
+
+	n.mtuProbe = mtuProbe
+	n.baseMss = baseMss
+
```
tpantelis, line 246: For consistency with the other functions, I think we should lock and not assume all calls will be single-threaded.

pkg/netlink/fake/netlink.go
```diff
@@ -342,3 +352,11 @@ func (n *NetLink) AwaitNoRule(table int) {
...
 	}, 5).Should(BeNil(), "Rule for %v exists", table)
 }
+
+func (n *NetLink) VerifyMtuProbe(mtuProbe string) {
+	Expect(mtuProbe).To(Equal(n.mtuProbe))
+}
+
+func (n *NetLink) VerifyBaseMss(baseMss string) {
+	Expect(baseMss).To(Equal(n.baseMss))
+}
```
tpantelis: Since we set these atomically we should really verify them atomically and also assume multi-threaded. 

```suggestion
func (n *NetLink) VerifyTCPMTUProbe(mtuProbe, baseMss string) {
        format := "MTU probe: %s, Base MSS: %s"
	Eventually(func() string {
            n.Lock()
            defer n.Unlock()
	    
             return fmt.Sprintf(format, n.mtuProbe, n.baseMss)
	}, 5).Should(Equal(fmt.Sprintf(format, mtuProbe, baseMss)))
}
```

dfarrell07 (approved):

pkg/routeagent_driver/handlers/mtu/mtuhandler_test.go
```diff
@@ -0,0 +1,68 @@
...
+func testInit() {
+	t := newTestDriver()
+
+	When("MTUHandler is initialized", func() {
+		It("Expect the mtuProbe and baseMss to be configured", func() {
+			t.netLink.VerifyMtuProbe(mtuProbe)
+			t.netLink.VerifyBaseMss(baseMss)
+		})
+	})
+}
```
tpantelis: The `MTUHandler` is simple so I think the tests can be simplified, ie I think the use of a `testDriver` in this case is overkill (unless it's expected for `MTUHandler` to become more complex, in which case we can add it later if needed). 

```
var _ = Describe("MTUHandler", func() {
    var (
        netLink *fakeNetlink.NetLink
	handler event.Handler
    )

    BeforeEach(func() {
        netLink = fakeNetlink.New()
        netlinkAPI.NewFunc = func() netlinkAPI.Interface {
            return netLink
        }
		
        handler = NewMTUHandler()
    })

    AfterEach(func() {
       netlinkAPI.NewFunc = nil
    })

    When("initialized", func() {
        It("should configure the MTU Probe correctly", func() {
            Expect(t.handler.Init()).To(Succeed())
            netLink.Verify...
        })
    })
})

```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler_test.go
aswinsuryan: Done.

pkg/netlink/fake/netlink.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/mtu/mtuhandler_suite_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
aswinsuryan: Done.

pkg/netlink/fake/netlink.go
```diff
@@ -231,6 +234,13 @@ func (n *NetLink) EnableLooseModeReversePathFilter(interfaceName string) error {
 }
 
 func (n *NetLink) ConfigureTCPMTUProbe(mtuProbe, baseMss string) error {
+	if mtuProbe == "" || baseMss == "" {
+		return fmt.Errorf("invalid arguments passed")
+	}
+
+	n.mtuProbe = mtuProbe
+	n.baseMss = baseMss
```
tpantelis, line 245:
```suggestion
        n.Lock()
        defer n.Unlock()
		
	n.mtuProbe = mtuProbe
	n.baseMss = baseMss
```

tpantelis (changes requested):

pkg/netlink/fake/netlink.go
aswinsuryan: Sorry for missing this ,I misread it.

tpantelis (approved):

PR #1496, opened 2021-07-19
Update default number of cluster egress IPs

tpantelis:
It was decided to change it to 8 instead of 1.


mangelajo (approved):

sridhargaddam (changes requested):
Thanks for the PR @tpantelis. We also require changes in our e2e validation which currently assumes only a single IP here - https://github.com/submariner-io/submariner/blob/1441a805db2055628366791d480ca494fe7f0923/test/e2e/framework/dataplane.go#L119

sridhargaddam (approved):
Thanks for fixing the issue @tpantelis

PR #1499, opened 2021-07-21
Ignore CNI Interface when healthcheck is disabled

sridhargaddam:
On some platforms, it was seen that CNI Interface on the node
was missing. Submariner uses the CNIInterface IP for certain use-cases
like HostNetwork to remoteCluster support and for Healthcheck use-cases.
The current code was expecting the CNIInterface to be present even
when the healthcheck is disabled, this PR fixes it.

### DIFF
--- a/pkg/endpoint/local_endpoint.go
+++ b/pkg/endpoint/local_endpoint.go
@@ -93,14 +93,15 @@ func GetLocal(submSpec types.SubmarinerSpecification, k8sClient kubernetes.Inter
 
 	endpoint.Spec.PublicIP = publicIP
 
-	if !globalnetEnabled {
+	if submSpec.HealthCheckEnabled && !globalnetEnabled {
 		// When globalnet is enabled, HealthCheckIP will be the globalIP assigned to the Active GatewayNode.
 		// In a fresh deployment, globalIP annotation for the node might take few seconds. So we listen on NodeEvents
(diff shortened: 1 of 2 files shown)


aswinsuryan (approved):

pkg/endpoint/local_endpoint.go
```diff
@@ -93,7 +93,7 @@ func GetLocal(submSpec types.SubmarinerSpecification, k8sClient kubernetes.Inter
 
 	endpoint.Spec.PublicIP = publicIP
 
-	if !globalnetEnabled {
+	if submSpec.HealthCheckEnabled && !globalnetEnabled {
```
mangelajo, line 96: Let's also explain on the error message L103:
```
return types.SubmarinerEndpoint{}, fmt.Errorf("error getting CNI Interface IP address: %v, please disable the health check if your CNI does not expose a pod IP on the host", err)
```

or something like that.

mangelajo (approved):

sridhargaddam:
@nyechiel @mangelajo Verified the PR locally disabling health-check and it uncovered an issue with e2e. Fixed the issue in e2e and verified that all e2e tests are passing fine. This is good for merge now.

nyechiel (approved):

tpantelis (approved):

PR #1500, opened 2021-07-21
Add clamp rules to fix TCP MTU issue

aswinsuryan:

### DIFF
--- a/pkg/routeagent_driver/constants/constants.go
+++ b/pkg/routeagent_driver/constants/constants.go
@@ -21,6 +21,10 @@ package constants
 const (
 	// IPTable chains used by RouteAgent
 	SmPostRoutingChain = "SUBMARINER-POSTROUTING"
+	PostRoutingChain   = "POSTROUTING"
+	MangleTable        = "mangle"
+	RemoteCIDRIPSet    = "remoteCIDRs"
+	LocalCIDRIPSet     = "localCIDRs"
 
 	// In order to support connectivity from HostNetwork to remoteCluster, route-agent tries
 	// to discover the CNIInterface[#] on the respective node and does SNAT of outgoing
--- a/pkg/routeagent_driver/constants/constants.go
+++ b/pkg/routeagent_driver/constants/constants.go
@@ -23,8 +23,8 @@ const (
 	SmPostRoutingChain = "SUBMARINER-POSTROUTING"
 	PostRoutingChain   = "POSTROUTING"
 	MangleTable        = "mangle"
-	RemoteCIDRIPSet    = "remoteCIDRs"
-	LocalCIDRIPSet     = "localCIDRs"
+	RemoteCIDRIPSet    = "SUBMARINER-REMOTECIDRS"
+	LocalCIDRIPSet     = "SUBMARINER-LOCALCIDRS"
 
 	// In order to support connectivity from HostNetwork to remoteCluster, route-agent tries
 	// to discover the CNIInterface[#] on the respective node and does SNAT of outgoing
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
@@ -1,0 +1,144 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mtu
+
+import (
+	"fmt"
+	"strings"
+
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/event"
+	"github.com/submariner-io/submariner/pkg/iptables"
+)
+
+const (
+	postRoutingChain           = "POSTROUTING"
+	subMarinerPostRoutingChain = "SUBMARINER-POSTROUTING"
+	mangleTable                = "mangle"
+)
+
+type mtuHandler struct {
+	event.HandlerBase
+	ipt iptables.Interface
+}
+
+func NewMTUHandler() event.Handler {
+	return &mtuHandler{}
+}
+
+func (h *mtuHandler) GetNetworkPlugins() []string {
+	return []string{event.AnyNetworkPlugin}
+}
+
+func (h *mtuHandler) GetName() string {
+	return "MTU handler"
+}
+
+func (h *mtuHandler) Init() error {
+	var err error
+	h.ipt, err = iptables.New()
+	if err != nil {
+		return fmt.Errorf("error initializing iptables: %v", err)
+	}
+
+	if err := iptables.CreateChainIfNotExists(h.ipt, mangleTable, subMarinerPostRoutingChain); err != nil {
+		return fmt.Errorf("error creating iptables chain %s: %v", subMarinerPostRoutingChain, err)
+	}
+
+	forwardToSubMarinerPostRoutingChain := []string{"-j", subMarinerPostRoutingChain}
+
+	if err := iptables.PrependUnique(h.ipt, mangleTable, postRoutingChain, forwardToSubMarinerPostRoutingChain); err != nil {
+		return fmt.Errorf("error inserting iptables rule %q: %v",
+			strings.Join(forwardToSubMarinerPostRoutingChain, " "), err)
+	}
+
+	return nil
+}
+
+func (h *mtuHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	subnets := extractSubnets(endpoint.Spec)
+	for _, subnet := range subnets {
+		err := h.addClampMssToPathMTU(subnet)
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (h *mtuHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	subnets := extractSubnets(endpoint.Spec)
+	for _, subnet := range subnets {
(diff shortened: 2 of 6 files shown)


sridhargaddam:
@aswinsuryan Not sure why the following file is part of this PR - natdiscovery.pb.go ?

aswinsuryan:
> @aswinsuryan Not sure why the following file is part of this PR - natdiscovery.pb.go ?

Oh it was accidental , removed it.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,144 @@
...
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/event"
+	"github.com/submariner-io/submariner/pkg/iptables"
+)
+
+const (
+	postRoutingChain           = "POSTROUTING"
+	subMarinerPostRoutingChain = "SUBMARINER-POSTROUTING"
+	mangleTable                = "mangle"
+)
```
sridhargaddam: We already have a macro for SUBMARINER-POSTROUTING. Also, please move the others to the pkg/routeagent_driver/constants/constants.go file.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return nil
+}
+
+func (h *mtuHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	subnets := extractSubnets(endpoint.Spec)
+	for _, subnet := range subnets {
+		err := h.addClampMssToPathMTU(subnet)
+
+		if err != nil {
```
```suggestion
		if err != nil {
```

New line is not required.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return nil
+}
+
+func (h *mtuHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	subnets := extractSubnets(endpoint.Spec)
+	for _, subnet := range subnets {
+		err := h.removeClampMssToPathMTU(subnet)
+
+		if err != nil {
```
```suggestion
		if err != nil {
```

Same

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,144 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (h *mtuHandler) addClampMssToPathMTU(subnets string) error {
```
```suggestion
func (h *mtuHandler) addClampMssToPathMTU(subnet string) error {
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,144 @@
...
+	ruleSpecSource := []string{"-s", subnets, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j",
```
sridhargaddam: Normally it's a good practice to include a match on both the sourceCIDR and destinationCIDR to avoid any leaks. This is what we do in the route-agent rules programmed in the NAT table (SUBMARINER-POSTROUTING chain).

Since this is the first time we are adding these rules, personally I feel that it's good to go directly with ipsets (just like how we are using them in the new Globalnet 2.0). It would be pretty straightforward for this use-case.
 
1. Create two IPset chains with a specific name (say remoteCIDRs and localCIDRs)
2. Add an iptable rule in the mange table, that matches on the ipset source localCIDR and ipset destination remoteCIDR - perform the mss-clamp-action.
3. Add an iptable rule in the mangle table, that matches on the ipset source remoteCIDR and ipset destination localCIDR - perform the mss-clamp-action.
3. When a new remoteEndpoint is added, update the remoteCIDR ipset chain and when the remoteEndpoint is deleted remote the entry.

So irrespective of how many clusters are connected, we will only be using two iptable rules in the mangle table.
It will be very optimal and less error-prone.

CC: @mangelajo @nyechiel

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,144 @@
...
+	ruleSpecDest := []string{"-d", subnets, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j",
+		"TCPMSS", "--clamp-mss-to-pmtu"}
+
+	if err := h.ipt.Delete(mangleTable, subMarinerPostRoutingChain, ruleSpecSource...); err != nil {
+		return fmt.Errorf("error deleting iptables rule \"%s\": %v", strings.Join(ruleSpecSource, " "), err)
+	}
+
+	if err := h.ipt.Delete(mangleTable, subMarinerPostRoutingChain, ruleSpecDest...); err != nil {
+		return fmt.Errorf("error deleting iptables rule \"%s\": %v", strings.Join(ruleSpecDest, " "), err)
+	}
```
mangelajo: Let's print those error if they happen but let's not return fmt.Errorf

only log.Errorf(

In this way the error is soft, we log it, we can continue attempting to delete the rules (for example if there was an issue creating them initially).

pkg/routeagent_driver/handlers/mtu/mtuhandler_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+var _ = Describe("MTUHandler", func() {
+	var (
+		ipt     *fakeIPT.IPTables
+		handler event.Handler
+	)
+
+	BeforeEach(func() {
+
+		ipt = fakeIPT.New()
```
sridhargaddam:
```suggestion
		ipt = fakeIPT.New()
```

sridhargaddam (changes requested):
Thanks for the PR @aswinsuryan. I have a suggestion. Please let me know what you think.

mangelajo (changes requested):
small change request, beyond that it looks good.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
mangelajo: Yes, I like that, but may be @aswinsuryan may need some help with that. Can you give him some pointers? (I believe we do it in globalnet, right?)

sridhargaddam: Yes, we are using it for GlobalEgressIPs with namespace/podSelectors here - https://github.com/submariner-io/submariner/blob/debfc47870e4e9239dbaa6035f3a38c0c1882ad3/pkg/globalnet/controllers/global_egressip_controller.go#L65

Feel free to reach out if you have any questions @aswinsuryan

pkg/routeagent_driver/constants/constants.go
```diff
@@ -21,6 +21,10 @@ package constants
 const (
 	// IPTable chains used by RouteAgent
 	SmPostRoutingChain = "SUBMARINER-POSTROUTING"
+	PostRoutingChain   = "POSTROUTING"
+	MangleTable        = "mangle"
+	RemoteCIDRIPSet    = "remoteCIDRs"
+	LocalCIDRIPSet     = "localCIDRs"
```
mangelajo: Just a nit, so it's more obvious when somebody inspect the system to who do those belong.
```suggestion
	RemoteCIDRIPSet    = "subm-remoteCIDRs"
	LocalCIDRIPSet     = "subm-localCIDRs"
```

mangelajo (changes requested):
Just one nit, very nit

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return nil
+}
+
+func (h *mtuHandler) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
+	subnets := extractSubnets(endpoint.Spec)
+	for _, subnet := range subnets {
+		err := h.localIPSet.DelEntry(subnet)
+		if err != nil {
+			klog.Errorf("Error deleting the subnet %q from the local IPSet: %v", subnet, err)
```
sridhargaddam, line 118: Shall we return err here, so that it will be retried?

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return nil
+}
+
+func (h *mtuHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	subnets := extractSubnets(endpoint.Spec)
+	for _, subnet := range subnets {
+		err := h.remoteIPSet.DelEntry(subnet)
+		if err != nil {
+			klog.Errorf("Error deleting the subnet %q from the remote IPSet: %v", subnet, err)
```
sridhargaddam, line 142: same as above

pkg/routeagent_driver/constants/constants.go
sridhargaddam: @aswinsuryan The max length for an IPset can be 31 chars, so we can use the full name itself.
i.e., SUBMARINER-REMOTECIDRS, SUBMARINER-LOCALCIDRS

sridhargaddam (changes requested):
Thanks for addressing the review comments @aswinsuryan. Some minor comments.

pkg/routeagent_driver/constants/constants.go
```diff
@@ -21,6 +21,10 @@ package constants
 const (
 	// IPTable chains used by RouteAgent
 	SmPostRoutingChain = "SUBMARINER-POSTROUTING"
+	PostRoutingChain   = "POSTROUTING"
+	MangleTable        = "mangle"
+	RemoteCIDRIPSet    = "SUBMARINER-REMOTECIDRS,"
```
sridhargaddam:
```suggestion
	RemoteCIDRIPSet    = "SUBMARINER-REMOTECIDRS"
```

Typo in the name, extra "," is added.

sridhargaddam (changes requested):
There is a typo in the ipset chain name

pkg/routeagent_driver/constants/constants.go
aswinsuryan: Done.

sridhargaddam (approved):
Thanks for addressing all the comments @aswinsuryan

mangelajo (approved):

mangelajo:
I marked it for squash, so if we all agree merge will be quicker.

PR #1501, opened 2021-07-22
Revert MTU probe changes

aswinsuryan:
This did not fix the issue with Openshift clusters


pkg/netlink/fake/netlink.go
```diff
@@ -235,16 +232,6 @@ func (n *NetLink) EnableLooseModeReversePathFilter(interfaceName string) error {
 }
 
 func (n *NetLink) ConfigureTCPMTUProbe(mtuProbe, baseMss string) error {
```
mangelajo, line 234: Shouldn't we remove this function altogether ? or is it called somewhere?

mangelajo (approved):

pkg/netlink/fake/netlink.go
aswinsuryan: it is used in global  gateway monitor.

sridhargaddam (approved):

pkg/netlink/fake/netlink.go
mangelajo: Then we should not delete code, right.

mangelajo:
This revert is urgent so we can continue with rebasing #1500

pkg/netlink/fake/netlink.go
aswinsuryan: No, I just reverted the changes the MTU probe PR did. This was a function returning nil before those changes. I am not sure deleting it will have any impact in globalnet UT.

PR #1505, opened 2021-07-27
Include procps-ng which is internally used by libreswan 

sridhargaddam:
Currently in gateway pod logs, we see two errors

...
libreswan.go:525] Starting Pluto
/usr/libexec/ipsec/_stackmanager: line 111: pidof: command not found
Initializing NSS database

/usr/sbin/ipsec: line 175: iptables: command not found
nflog ipsec capture disabled
...

pidof is used in libreswan api start_xfrm, so adding procps-ng binary.

On the other hand, iptables utility is mainly used to initialise
iptables rules for the nflog devices so that we can capture pre-encrypted
or post-decrypted packets using tcpdump. Currently we are not configuring
any nflog params while starting pluto process, so we do not require this
binary. This is validated as part of "--checknflog" command, so commenting
it out.

### DIFF
--- a/package/Dockerfile.submariner-gateway
+++ b/package/Dockerfile.submariner-gateway
@@ -6,7 +6,7 @@ WORKDIR /var/submariner
 # iproute is used internally
 # libreswan provides IKE
 RUN dnf -y install --nodocs --setopt=install_weak_deps=0 \
-           iproute libreswan && \
+           iproute libreswan procps-ng && \
     dnf -y clean all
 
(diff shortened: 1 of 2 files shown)


nyechiel:
@sridhargaddam I understand that this is only a draft, but can you add a little bit more context on this one?

sridhargaddam:
> @sridhargaddam I understand that this is only a draft, but can you add a little bit more context on this one?

Yes @nyechiel. I will add it. Looking into libreswan codebase to understand where the itpables utility is used.

mangelajo (approved):

nyechiel (approved):

PR #1512, opened 2021-08-02
Delete iptable chains as soon as service is deleted

sridhargaddam:
In the new implementation of Globalnet, as soon as a service is deleted, we
are initially deleting the corresponding GlobalIngressIP Object and in the
respective controller.OnDelete method, we are flushing out the iptable rules.
This is causing some delay and in the meantime when kubeproxy is trying to
remove the corresponding service related iptable chains, its erroring out
as there are still some references to the chain. Kubeproxy continues to
retry the delete operation. This PR modifies the behavior to delete
the itpable chains as soon as the service is deleted.


sridhargaddam:
@tpantelis @vthapar While debugging some flaky errors seen in CI, noticed some error logs in kubeproxy even though the tests pass. This should help in reducing the errors. I'm marking it as draft as its too late in the release cycle.

skitt:
@sridhargaddam this seems useful to me, what’s the plan for it?

sridhargaddam:
> @sridhargaddam this seems useful to me, what’s the plan for it?

Yes @skitt. I'll have to revisit this.

PR #1513, opened 2021-08-03
E2e, delete ServiceExport before Service

sridhargaddam:
Since Globalnet internally uses kubeproxy iptable chain, if we delete
the serviceExport first, we will give Globalnet opportunity to remove
the iptable references ahead of kubeproxy. This will inturn reduce the
possibilities of flaky errors to some extent.

### DIFF
--- a/test/e2e/framework/dataplane.go
+++ b/test/e2e/framework/dataplane.go
@@ -178,8 +178,8 @@ func verifyGlobalnetDatapathConnectivity(p tcp.ConnectivityTestParams, egressIPT
 
 	Expect(listenerPod.TerminationMessage).To(matchIP)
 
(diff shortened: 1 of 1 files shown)


sridhargaddam:
@dfarrell07 @skitt @tpantelis @nyechiel I was looking into some of the flaky errors including the passing tests and noticed some error logs in kubeproxy (which are not fatal and will be retried). Until we completely get rid of kubeproxy dependency we will have this kind of issue. Changing the order of resource deletion allows Globalnet to delete the references slightly ahead of kubeproxy. 

So, can we get this PR in for release-0.10 so that we can avoid some flaky errors that correspond to a known issue?

tpantelis (approved):

skitt (approved):

PR #1555, opened 2021-09-16
Fix E2E test for globalnet w/ external network to handle multiple egressIPs

mkimuram:
This PR fixes `[external-dataplane] Connectivity` `should be able to connect from an external app to a pod in a cluster` E2E test.

fixes: #1554

PR #1558, opened 2021-09-17
Implement globalnet for headless service without selector

### LINKED ISSUE #1537: Headless service without selector doesn't work
**What happened**:
Headless service without selector doesn't work.

Headless service without selector does work

- Create headless service without selector
```
export KUBECONFIG=kubeconfig.cluster-a
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: test-vm
spec:
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
  clusterIP: None
EOF

cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Endpoints
metadata:
  name: test-vm
subsets:
  - addresses:
      - ip: 192.168.122.142
    ports:
      - port: 80
EOF

subctl export service -n default test-vm
```
- Check that globalingressip for it is assigned
```
kubectl get globalingressip test-vm
Error from server (NotFound): globalingressips.submariner.io "test-vm" not found
```

- Logs from globalnet pod at that time
```
kubectl logs -n submariner-operator -l app=submariner-globalnet
I0819 14:24:20.249605       1 service_export_controller.go:145] Processing ServiceExport "default/test-vm"
I0819 14:24:20.350946       1 ingress_pod_controller.go:91] Created Pod controller for (default/test-vm) with selector ""
I0819 14:24:20.442889       1 ingress_pod_controller.go:123] "create" ingress Pod default/cl1 for service test-vm
I0819 14:24:20.480555       1 global_ingressip_controller.go:124] Processing created &v1.GlobalIngressIP{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"pod-cl1", GenerateName:"", Namespace:"default", SelfLink:"", UID:"2ec8fde4-e79d-4244-b01a-6dded600b580", ResourceVersion:"1347889", Generation:1, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63764979860, loc:(*time.Location)(0x2021ea0)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string{"submariner.io/serviceRef":"test-vm"}, Annotations:map[string]string{"submariner.io/headless-svc-pod-ip":"10.42.0.13"}, OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:"submariner-globalnet", Operation:"Update", APIVersion:"submariner.io/v1", Time:(*v1.Time)(0xc0000a7d58), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0000a7de8)}}}, Spec:v1.GlobalIngressIPSpec{Target:"HeadlessServicePod", ServiceRef:(*v1.LocalObjectReference)(0xc000d46ae0), PodRef:(*v1.LocalObjectReference)(0xc000d46af0)}, Status:v1.GlobalIngressIPStatus{Conditions:[]v1.Condition(nil), AllocatedIP:""}}
I0819 14:24:20.480976       1 global_ingressip_controller.go:147] Allocating global IP for "default/pod-cl1"
I0819 14:24:20.481089       1 iface.go:145] Installing iptables rule for Headless SVC Pod -d 242.0.255.253 -j DNAT --to 10.42.0.13
I0819 14:24:20.598690       1 iface.go:227] Installing iptable egress rules for HDLS SVC Pod "default/pod-cl1": -p all -s 10.42.0.13 -m mark --mark 0xC0000/0xC0000 -j SNAT --to 242.0.255.253
I0819 14:24:20.614227       1 base_controllers.go:194] Updated: &v1.GlobalIngressIPStatus{Conditions:[]v1.Condition{v1.Condition{Type:"Allocated", Status:"True", ObservedGeneration:0, LastTransitionTime:v1.Time{Time:time.Time{wall:0xc03fb845249ba30e, ext:851262680326300, loc:(*time.Location)(0x2021ea0)}}, Reason:"Success", Message:"Allocated global IP"}}, AllocatedIP:"242.0.255.253"}
```

According to the above log, it seems to be handled as [ingress pod](https://github.com/submariner-io/submariner/blob/devel/pkg/globalnet/controllers/ingress_pod_controller.go#L123), not a service without selector. Maybe we need to add a separete logic for headless service without selector, [here](https://github.com/submariner-io/submariner/blob/59f656555db0c2f18b691b7ab1a28edb34f2f43f/pkg/globalnet/controllers/service_export_controller.go#L218)? (There seems the logic only for headless service __with__ selector).

**Environment**:
- Install tools:
- Others:

mkimuram:
This PR adds support for globalnet for headless service without selector.

Fixes: #1537 

Dependency:
- RBAC:
  - [x] submariner-io/submariner-operator#1582
  - [x] submariner-io/submariner-charts#169
  - [x] #1582

### DIFF
--- a/pkg/globalnet/constants/constants.go
+++ b/pkg/globalnet/constants/constants.go
@@ -29,6 +29,7 @@ const (
 	// The following chains are added as part of GN 2.0 implementation.
 	SmGlobalnetEgressChainForPods            = "SM-GN-EGRESS-PODS"
 	SmGlobalnetEgressChainForHeadlessSvcPods = "SM-GN-EGRESS-HDLS-PODS"
+	SmGlobalnetEgressChainForHeadlessSvcEPs  = "SM-GN-EGRESS-HDLS-EPS"
 	SmGlobalnetEgressChainForNamespace       = "SM-GN-EGRESS-NS"
 	SmGlobalnetEgressChainForCluster         = "SM-GN-EGRESS-CLUSTER"
 
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -73,11 +73,21 @@ func (c *baseSyncerController) Start() error {
 	return c.resourceSyncer.Start(c.stopCh) // nolint:wrapcheck  // Let the caller wrap it
 }
 
-func (c *baseSyncerController) reconcile(client dynamic.ResourceInterface, labelSelector string,
+func (c *baseSyncerController) reconcile(client dynamic.ResourceInterface, labelSelector, fieldSelector string,
 	transform func(obj *unstructured.Unstructured) runtime.Object,
 ) {
+	listOptions := metav1.ListOptions{}
+
+	if labelSelector != "" {
+		listOptions.LabelSelector = labelSelector
+	}
+
+	if fieldSelector != "" {
+		listOptions.FieldSelector = fieldSelector
+	}
+
 	c.resourceSyncer.Reconcile(func() []runtime.Object {
-		objList, err := client.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+		objList, err := client.List(context.TODO(), listOptions)
 		if err != nil {
 			klog.Errorf("Error listing resources for reconciliation: %v", err)
 			return nil
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -76,18 +76,11 @@ func (c *baseSyncerController) Start() error {
 func (c *baseSyncerController) reconcile(client dynamic.ResourceInterface, labelSelector, fieldSelector string,
 	transform func(obj *unstructured.Unstructured) runtime.Object,
 ) {
-	listOptions := metav1.ListOptions{}
-
-	if labelSelector != "" {
-		listOptions.LabelSelector = labelSelector
-	}
-
-	if fieldSelector != "" {
-		listOptions.FieldSelector = fieldSelector
-	}
-
 	c.resourceSyncer.Reconcile(func() []runtime.Object {
-		objList, err := client.List(context.TODO(), listOptions)
+		objList, err := client.List(context.TODO(), metav1.ListOptions{
+			LabelSelector: labelSelector,
+			FieldSelector: fieldSelector,
+		})
 		if err != nil {
 			klog.Errorf("Error listing resources for reconciliation: %v", err)
 			return nil
--- a/pkg/globalnet/controllers/gateway_monitor.go
+++ b/pkg/globalnet/controllers/gateway_monitor.go
@@ -292,7 +292,13 @@ func (g *gatewayMonitor) startControllers() error {
 		return errors.Wrap(err, "error creating the Endpoints controller")
 	}
 
-	c, err = NewServiceExportController(g.syncerConfig, podControllers, endpointsControllers)
+	ingressEndpointsControllers, err := NewIngressEndpointsControllers(g.syncerConfig)
+	if err != nil {
+		return errors.WithMessage(err, "error creating the IngressEndpointsControllers")
+	}
+
+	c, err = NewServiceExportController(g.syncerConfig, podControllers, endpointsControllers,
+		ingressEndpointsControllers)
 	if err != nil {
 		return errors.Wrap(err, "error creating the ServiceExport controller")
 	}
@@ -390,6 +396,12 @@ func (g *gatewayMonitor) createGlobalnetChains() error {
 		return errors.Wrapf(err, "error creating iptables chain %s", constants.SmGlobalnetEgressChainForHeadlessSvcPods)
 	}
 
+	klog.V(log.DEBUG).Infof("Install/ensure %s chain exists", constants.SmGlobalnetEgressChainForHeadlessSvcEPs)
+
+	if err := iptables.CreateChainIfNotExists(g.ipt, "nat", constants.SmGlobalnetEgressChainForHeadlessSvcEPs); err != nil {
+		return errors.Wrapf(err, "error creating iptables chain %s", constants.SmGlobalnetEgressChainForHeadlessSvcEPs)
+	}
+
 	klog.V(log.DEBUG).Infof("Install/ensure %s chain exists", constants.SmGlobalnetEgressChainForNamespace)
 
 	if err := iptables.CreateChainIfNotExists(g.ipt, "nat", constants.SmGlobalnetEgressChainForNamespace); err != nil {
@@ -412,13 +424,18 @@ func (g *gatewayMonitor) createGlobalnetChains() error {
 		klog.Errorf("error inserting iptables rule %q: %v\n", strings.Join(forwardToSubGlobalNetChain, " "), err)
 	}
 
-	forwardToSubGlobalNetChain = []string{"-j", constants.SmGlobalnetEgressChainForNamespace}
+	forwardToSubGlobalNetChain = []string{"-j", constants.SmGlobalnetEgressChainForHeadlessSvcEPs}
 	if err := iptables.InsertUnique(g.ipt, "nat", constants.SmGlobalnetEgressChain, 4, forwardToSubGlobalNetChain); err != nil {
 		klog.Errorf("error inserting iptables rule %q: %v\n", strings.Join(forwardToSubGlobalNetChain, " "), err)
 	}
 
-	forwardToSubGlobalNetChain = []string{"-j", constants.SmGlobalnetEgressChainForCluster}
+	forwardToSubGlobalNetChain = []string{"-j", constants.SmGlobalnetEgressChainForNamespace}
 	if err := iptables.InsertUnique(g.ipt, "nat", constants.SmGlobalnetEgressChain, 5, forwardToSubGlobalNetChain); err != nil {
+		klog.Errorf("error inserting iptables rule %q: %v\n", strings.Join(forwardToSubGlobalNetChain, " "), err)
+	}
+
+	forwardToSubGlobalNetChain = []string{"-j", constants.SmGlobalnetEgressChainForCluster}
+	if err := iptables.InsertUnique(g.ipt, "nat", constants.SmGlobalnetEgressChain, 6, forwardToSubGlobalNetChain); err != nil {
 		klog.Errorf("error inserting iptables rule %q: %v\n", strings.Join(forwardToSubGlobalNetChain, " "), err)
 	}
 
--- a/pkg/globalnet/controllers/global_ingressip_controller.go
+++ b/pkg/globalnet/controllers/global_ingressip_controller.go
@@ -225,31 +225,24 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
 			return false
 		}
 	} else {
-		var target string
+		var annotationKey string
 		var tType iptables.TargetType
 
 		if ingressIP.Spec.Target == submarinerv1.HeadlessServicePod {
-			target = ingressIP.GetAnnotations()[headlessSvcPodIP]
-			if target == "" {
-				_ = c.pool.Release(ips...)
-
-				klog.Warningf("%q annotation is missing on %q", headlessSvcPodIP, key)
-
-				return true
-			}
-
+			annotationKey = headlessSvcPodIP
 			tType = iptables.PodTarget
 		} else if ingressIP.Spec.Target == submarinerv1.HeadlessServiceEndpoints {
-			target = ingressIP.GetAnnotations()[headlessSvcEndpointsIP]
-			if target == "" {
-				_ = c.pool.Release(ips...)
-
-				klog.Warningf("%q annotation is missing on %q", headlessSvcEndpointsIP, key)
-
-				return true
-			}
-
+			annotationKey = headlessSvcEndpointsIP
 			tType = iptables.EndpointsTarget
+		}
+
+		target := ingressIP.GetAnnotations()[annotationKey]
+		if target == "" {
+			_ = c.pool.Release(ips...)
+
+			klog.Warningf("%q annotation is missing on %q", annotationKey, key)
+
+			return true
 		}
 
 		err = c.iptIface.AddIngressRulesForHeadlessSvc(ips[0], target, tType)
--- a/pkg/globalnet/controllers/global_ingressip_controller.go
+++ b/pkg/globalnet/controllers/global_ingressip_controller.go
@@ -82,20 +82,30 @@ func NewGlobalIngressIPController(config *syncer.ResourceSyncerConfig, pool *ipa
 
 		// nolint:wrapcheck  // No need to wrap these errors.
 		err = controller.reserveAllocatedIPs(federator, obj, func(reservedIPs []string) error {
+			var target string
+			var tType iptables.TargetType
+
 			metrics.RecordAllocateGlobalIngressIPs(pool.GetCIDR(), len(reservedIPs))
+
 			if gip.Spec.Target == submarinerv1.ClusterIPService {
 				return controller.ensureInternalServiceExists(gip)
 			} else if gip.Spec.Target == submarinerv1.HeadlessServicePod {
-				target := gip.GetAnnotations()[headlessSvcPodIP]
-				err := controller.iptIface.AddIngressRulesForHeadlessSvcPod(reservedIPs[0], target)
-				if err != nil {
-					return err
-				}
-
-				key, _ := cache.MetaNamespaceKeyFunc(obj)
-				return controller.iptIface.AddEgressRulesForHeadlessSVCPods(key, target, reservedIPs[0], globalNetIPTableMark)
+				target = gip.GetAnnotations()[headlessSvcPodIP]
+				tType = iptables.PodTarget
+			} else if gip.Spec.Target == submarinerv1.HeadlessServiceEndpoints {
+				target = gip.GetAnnotations()[headlessSvcEndpointsIP]
+				tType = iptables.EndpointsTarget
+			} else {
+				return nil
 			}
-			return nil
+
+			err := controller.iptIface.AddIngressRulesForHeadlessSvc(reservedIPs[0], target, tType)
+			if err != nil {
+				return err
+			}
+
+			key, _ := cache.MetaNamespaceKeyFunc(obj)
+			return controller.iptIface.AddEgressRulesForHeadlessSvc(key, target, reservedIPs[0], globalNetIPTableMark, tType)
 		})
 
 		if err != nil {
@@ -214,24 +224,42 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
 
 			return false
 		}
-	} else if ingressIP.Spec.Target == submarinerv1.HeadlessServicePod {
-		podIP := ingressIP.GetAnnotations()[headlessSvcPodIP]
-		if podIP == "" {
-			_ = c.pool.Release(ips...)
+	} else {
+		var target string
+		var tType iptables.TargetType
 
-			klog.Warningf("%q annotation is missing on %q", headlessSvcPodIP, key)
+		if ingressIP.Spec.Target == submarinerv1.HeadlessServicePod {
+			target = ingressIP.GetAnnotations()[headlessSvcPodIP]
+			if target == "" {
+				_ = c.pool.Release(ips...)
 
-			return true
+				klog.Warningf("%q annotation is missing on %q", headlessSvcPodIP, key)
+
+				return true
+			}
+
+			tType = iptables.PodTarget
+		} else if ingressIP.Spec.Target == submarinerv1.HeadlessServiceEndpoints {
+			target = ingressIP.GetAnnotations()[headlessSvcEndpointsIP]
+			if target == "" {
+				_ = c.pool.Release(ips...)
+
+				klog.Warningf("%q annotation is missing on %q", headlessSvcEndpointsIP, key)
+
+				return true
+			}
+
+			tType = iptables.EndpointsTarget
 		}
 
-		err = c.iptIface.AddIngressRulesForHeadlessSvcPod(ips[0], podIP)
+		err = c.iptIface.AddIngressRulesForHeadlessSvc(ips[0], target, tType)
 		if err != nil {
-			klog.Errorf("Error while programming Service %q ingress rules for Pod: %v", key, err)
+			klog.Errorf("Error while programming Service %q ingress rules for %v: %v", key, tType, err)
 			err = errors.WithMessage(err, "Error programming ingress rules")
 		} else {
-			err = c.iptIface.AddEgressRulesForHeadlessSVCPods(key, podIP, ips[0], globalNetIPTableMark)
+			err = c.iptIface.AddEgressRulesForHeadlessSvc(key, target, ips[0], globalNetIPTableMark, tType)
 			if err != nil {
-				_ = c.iptIface.RemoveIngressRulesForHeadlessSvcPod(ips[0], podIP)
+				_ = c.iptIface.RemoveIngressRulesForHeadlessSvc(ips[0], target, tType)
 				err = errors.WithMessage(err, "Error programming egress rules")
 			}
 		}
@@ -303,16 +331,25 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
 	}
 
 	return c.flushRulesAndReleaseIPs(key, numRequeues, func(allocatedIPs []string) error {
-		metrics.RecordDeallocateGlobalIngressIPs(c.pool.GetCIDR(), len(allocatedIPs))
-		if ingressIP.Spec.Target == submarinerv1.HeadlessServicePod {
-			podIP := ingressIP.GetAnnotations()[headlessSvcPodIP]
-			if podIP != "" {
-				if err := c.iptIface.RemoveIngressRulesForHeadlessSvcPod(ingressIP.Status.AllocatedIP, podIP); err != nil {
-					return err
-				}
+		var target string
+		var tType iptables.TargetType
 
-				return c.iptIface.RemoveEgressRulesForHeadlessSVCPods(key, podIP, ingressIP.Status.AllocatedIP, globalNetIPTableMark)
+		metrics.RecordDeallocateGlobalIngressIPs(c.pool.GetCIDR(), len(allocatedIPs))
+
+		if ingressIP.Spec.Target == submarinerv1.HeadlessServicePod {
+			target = ingressIP.GetAnnotations()[headlessSvcPodIP]
+			tType = iptables.PodTarget
+		} else if ingressIP.Spec.Target == submarinerv1.HeadlessServiceEndpoints {
+			target = ingressIP.GetAnnotations()[headlessSvcEndpointsIP]
+			tType = iptables.EndpointsTarget
+		}
+
+		if target != "" {
+			if err := c.iptIface.RemoveIngressRulesForHeadlessSvc(ingressIP.Status.AllocatedIP, target, tType); err != nil {
+				return err
 			}
+
+			return c.iptIface.RemoveEgressRulesForHeadlessSvc(key, target, ingressIP.Status.AllocatedIP, globalNetIPTableMark, tType)
 		}
 
 		return nil
--- a/pkg/globalnet/controllers/ingress_endpoints_controller.go
+++ b/pkg/globalnet/controllers/ingress_endpoints_controller.go
@@ -115,11 +115,11 @@ func (c *ingressEndpointsController) onCreateOrUpdate(endpoints *corev1.Endpoint
 	klog.Infof("%q ingress Endpoints %s for service %s", op, key, c.svcName)
 
 	// Get list of current endpoint IPs
-	epIPs := map[string]bool{}
+	usedEpIPs := map[string]bool{}
 
 	for _, subset := range endpoints.Subsets {
 		for _, addr := range subset.Addresses {
-			epIPs[addr.IP] = true
+			usedEpIPs[addr.IP] = true
 		}
 	}
 
@@ -136,9 +136,9 @@ func (c *ingressEndpointsController) onCreateOrUpdate(endpoints *corev1.Endpoint
 		annotations := ingressIP.GetAnnotations()
 		if epIP, ok := annotations[headlessSvcEndpointsIP]; ok {
 			// ingressIP is still used
-			if _, used := epIPs[epIP]; used {
+			if _, used := usedEpIPs[epIP]; used {
 				// Delete from the hash to avoid recreating existing ingressIP
-				delete(epIPs, epIP)
+				delete(usedEpIPs, epIP)
 				// Skip deleting ingressIP
 				continue
 			}
@@ -151,7 +151,7 @@ func (c *ingressEndpointsController) onCreateOrUpdate(endpoints *corev1.Endpoint
 	}
 
 	// Create new ingressIPs
-	for epIP := range epIPs {
+	for epIP := range usedEpIPs {
 		ingressIP := newIngressIP(endpoints.Name, endpoints.Namespace, epIP)
 
 		ingressIP.ObjectMeta.Annotations = map[string]string{
--- a/pkg/globalnet/controllers/ingress_endpoints_controller.go
+++ b/pkg/globalnet/controllers/ingress_endpoints_controller.go
@@ -1,0 +1,206 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/util"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+func startIngressEndpointsController(svc *corev1.Service, config *syncer.ResourceSyncerConfig) (*ingressEndpointsController, error) {
+	var err error
+
+	_, gvr, err := util.ToUnstructuredResource(&submarinerv1.GlobalIngressIP{}, config.RestMapper)
+	if err != nil {
+		return nil, errors.Wrap(err, "error converting resource")
+	}
+
+	controller := &ingressEndpointsController{
+		baseSyncerController: newBaseSyncerController(),
+		svcName:              svc.Name,
+		namespace:            svc.Namespace,
+		config:               *config,
+		ingressIPs:           config.SourceClient.Resource(*gvr),
+	}
+
+	fieldSelector := fields.Set(map[string]string{"metadata.name": svc.Name}).AsSelector().String()
+
+	controller.resourceSyncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:                "Ingress Endpoints syncer",
+		ResourceType:        &corev1.Endpoints{},
+		SourceClient:        config.SourceClient,
+		SourceNamespace:     svc.Namespace,
+		RestMapper:          config.RestMapper,
+		Federator:           federate.NewCreateFederator(config.SourceClient, config.RestMapper, corev1.NamespaceAll),
+		Scheme:              config.Scheme,
+		Transform:           controller.process,
+		SourceFieldSelector: fieldSelector,
+		ResourcesEquivalent: areEndpointsEqual,
+	})
+
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating the endpoints syncer")
+	}
+
+	if err := controller.Start(); err != nil {
+		return nil, errors.Wrap(err, "error starting the endpoint syncer")
+	}
+
+	ingressIPs := config.SourceClient.Resource(*gvr).Namespace(corev1.NamespaceAll)
+	controller.reconcile(ingressIPs, "" /* labelSelector */, fieldSelector, func(obj *unstructured.Unstructured) runtime.Object {
+		endpointsName, exists, _ := unstructured.NestedString(obj.Object, "spec", "serviceRef", "name")
+		if exists {
+			return &corev1.Endpoints{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      endpointsName,
+					Namespace: obj.GetNamespace(),
+				},
+			}
+		}
+
+		return nil
+	})
+
+	klog.Infof("Created Endpoints controller for (%s/%s) with selector %q", svc.Namespace, svc.Name, fieldSelector)
+
+	return controller, nil
+}
+
+func (c *ingressEndpointsController) process(from runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	endpoints := from.(*corev1.Endpoints)
+
+	switch op {
+	case syncer.Create, syncer.Update:
+		return c.onCreateOrUpdate(endpoints, op)
+	case syncer.Delete:
+		return c.onDelete(endpoints)
+	}
+
+	return nil, false
+}
+
+func (c *ingressEndpointsController) onCreateOrUpdate(endpoints *corev1.Endpoints, op syncer.Operation) (runtime.Object, bool) {
+	key, _ := cache.MetaNamespaceKeyFunc(endpoints)
+
+	klog.Infof("%q ingress Endpoints %s for service %s", op, key, c.svcName)
+
+	// Get list of current endpoint IPs
+	epIPs := map[string]bool{}
+
+	for _, subset := range endpoints.Subsets {
+		for _, addr := range subset.Addresses {
+			epIPs[addr.IP] = true
+		}
+	}
+
+	// Get List of existing ingressIPs
+	selector := labels.SelectorFromSet(map[string]string{ServiceRefLabel: endpoints.Name}).String()
+
+	existingIngressIPs, err := c.ingressIPs.Namespace(endpoints.Namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return nil, true
+	}
+
+	// Delete unused ingressIPs
+	for _, ingressIP := range existingIngressIPs.Items {
+		annotations := ingressIP.GetAnnotations()
+		if epIP, ok := annotations[headlessSvcEndpointsIP]; ok {
+			// ingressIP is still used
+			if _, used := epIPs[epIP]; used {
+				// Delete from the hash to avoid recreating existing ingressIP
+				delete(epIPs, epIP)
+				// Skip deleting ingressIP
+				continue
+			}
+		}
+
+		err := c.ingressIPs.Namespace(ingressIP.GetNamespace()).Delete(context.TODO(), ingressIP.GetName(), metav1.DeleteOptions{})
+		if err != nil {
+			return nil, true
+		}
+	}
+
+	// Create new ingressIPs
+	for epIP := range epIPs {
+		ingressIP := newIngressIP(endpoints.Name, endpoints.Namespace, epIP)
+
+		ingressIP.ObjectMeta.Annotations = map[string]string{
+			headlessSvcEndpointsIP: epIP,
+		}
+
+		ingressIP.Spec = submarinerv1.GlobalIngressIPSpec{
+			Target:     submarinerv1.HeadlessServiceEndpoints,
+			ServiceRef: &corev1.LocalObjectReference{Name: c.svcName},
+		}
+
+		uIngressIP, _, err := util.ToUnstructuredResource(ingressIP, c.config.RestMapper)
+		if err != nil {
+			return nil, true
+		}
+
+		_, err = c.ingressIPs.Namespace(ingressIP.GetNamespace()).Create(context.TODO(), uIngressIP, metav1.CreateOptions{})
+		if err != nil {
+			return nil, true
+		}
+	}
+
+	return nil, false
+}
+
+func (c *ingressEndpointsController) onDelete(endpoints *corev1.Endpoints) (runtime.Object, bool) {
+	key, _ := cache.MetaNamespaceKeyFunc(endpoints)
+
+	selector := labels.SelectorFromSet(map[string]string{ServiceRefLabel: endpoints.Name}).String()
+
+	err := c.ingressIPs.Namespace(endpoints.Namespace).DeleteCollection(context.TODO(), metav1.DeleteOptions{},
+		metav1.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return nil, true
+	}
+
+	klog.Infof("Ingress Endpoints %s for service %s deleted", key, c.svcName)
+
+	return nil, false
+}
+
+func newIngressIP(name, namespace, epIP string) *submarinerv1.GlobalIngressIP {
+	return &submarinerv1.GlobalIngressIP{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("ep-%.44s-%.15s", name, epIP),
+			Namespace: namespace,
+			Labels: map[string]string{
+				ServiceRefLabel: name,
+			},
+		},
+	}
+}
--- a/pkg/globalnet/controllers/ingress_endpoints_controllers.go
+++ b/pkg/globalnet/controllers/ingress_endpoints_controllers.go
@@ -9,14 +9,14 @@ You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
 */
 
-//nolint:dupl // same logic to ingress_endpoints_controllers, but for a different class
+//nolint:dupl // same logic to ingress_pod_controllers, but for a different class
 package controllers
 
 import (
@@ -33,21 +33,20 @@ import (
 	"k8s.io/klog"
 )
 
-func NewIngressPodControllers(config *syncer.ResourceSyncerConfig) (*IngressPodControllers, error) {
-	// We'll panic if config is nil, this is intentional
+func NewIngressEndpointsControllers(config *syncer.ResourceSyncerConfig) (*IngressEndpointsControllers, error) {
 	_, gvr, err := util.ToUnstructuredResource(&submarinerv1.GlobalIngressIP{}, config.RestMapper)
 	if err != nil {
 		return nil, errors.Wrap(err, "error converting resource")
 	}
 
-	return &IngressPodControllers{
-		controllers: map[string]*ingressPodController{},
+	return &IngressEndpointsControllers{
+		controllers: map[string]*ingressEndpointsController{},
 		config:      *config,
 		ingressIPs:  config.SourceClient.Resource(*gvr),
 	}, nil
 }
 
-func (c *IngressPodControllers) start(service *corev1.Service) error {
+func (c *IngressEndpointsControllers) start(service *corev1.Service) error {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
@@ -56,7 +55,7 @@ func (c *IngressPodControllers) start(service *corev1.Service) error {
 		return nil
 	}
 
-	controller, err := startIngressPodController(service, &c.config)
+	controller, err := startIngressEndpointsController(service, &c.config)
 	if err != nil {
 		return err
 	}
@@ -66,7 +65,7 @@ func (c *IngressPodControllers) start(service *corev1.Service) error {
 	return nil
 }
 
-func (c *IngressPodControllers) stopAll() {
+func (c *IngressEndpointsControllers) stopAll() {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
@@ -74,16 +73,16 @@ func (c *IngressPodControllers) stopAll() {
 		controller.Stop()
 	}
 
-	c.controllers = map[string]*ingressPodController{}
+	c.controllers = map[string]*ingressEndpointsController{}
 }
 
-func (c *IngressPodControllers) stopAndCleanup(serviceName, serviceNamespace string) {
+func (c *IngressEndpointsControllers) stopAndCleanup(serviceName, serviceNamespace string) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
 	key := c.key(serviceName, serviceNamespace)
-
 	controller, exists := c.controllers[key]
+
 	if exists {
 		controller.Stop()
 		delete(c.controllers, key)
@@ -98,6 +97,6 @@ func (c *IngressPodControllers) stopAndCleanup(serviceName, serviceNamespace str
 	}
 }
 
-func (c *IngressPodControllers) key(n, ns string) string {
+func (c *IngressEndpointsControllers) key(n, ns string) string {
 	return ns + "/" + n
 }
--- a/pkg/globalnet/controllers/iptables/iface.go
+++ b/pkg/globalnet/controllers/iptables/iface.go
@@ -35,13 +35,13 @@ import (
 type Interface interface {
 	AddClusterEgressRules(sourceIP, snatIP, globalNetIPTableMark string) error
 	RemoveClusterEgressRules(sourceIP, snatIP, globalNetIPTableMark string) error
-	AddIngressRulesForHeadlessSvcPod(globalIP, podIP string) error
-	RemoveIngressRulesForHeadlessSvcPod(globalIP, podIP string) error
+	AddIngressRulesForHeadlessSvc(globalIP, podIP string, targetType TargetType) error
+	RemoveIngressRulesForHeadlessSvc(globalIP, podIP string, targetType TargetType) error
 	GetKubeProxyClusterIPServiceChainName(service *corev1.Service, kubeProxyServiceChainPrefix string) (string, bool, error)
 	AddIngressRulesForHealthCheck(cniIfaceIP, globalIP string) error
 	RemoveIngressRulesForHealthCheck(cniIfaceIP, globalIP string) error
-	AddEgressRulesForHeadlessSVCPods(key, sourceIP, snatIP, globalNetIPTableMark string) error
-	RemoveEgressRulesForHeadlessSVCPods(key, sourceIP, snatIP, globalNetIPTableMark string) error
+	AddEgressRulesForHeadlessSvc(key, sourceIP, snatIP, globalNetIPTableMark string, targetType TargetType) error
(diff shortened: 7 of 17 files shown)


mkimuram:
@sridhargaddam @vthapar  @aswinsuryan 

This is still a prototype that only supports a single IP of the endpoints (and many quick & dirty hacks remaining). However, it seems to work (See below for details). Let's make this prototype as a start point of the further discussion.

<details>

1. Deploy this branch

```bash
make e2e using=external-net,globalnet
```

2. Update RBAC to allow handling core/v1 endpoints for both cluster1 and cluster2 ([this](https://github.com/submariner-io/submariner-operator/blob/devel/bundle/manifests/submariner-globalnet_rbac.authorization.k8s.io_v1_clusterrole.yaml#L13) should be modified, later.)

(Just kubectl edit the ClusterRole submariner-globalnet to add `endpoints` in `resources` of `apiGroups == ""`)

```bash
export KUBECONFIG=output/kubeconfigs/kind-config-cluster1
kubectl get ClusterRole submariner-globalnet -o yaml > /tmp/rbac-before.yaml 
kubectl edit ClusterRole submariner-globalnet
kubectl get ClusterRole submariner-globalnet -o yaml > /tmp/rbac-after.yaml 
diff -urN /tmp/rbac-before.yaml /tmp/rbac-after.yaml 
--- /tmp/rbac-before.yaml       2021-09-16 23:34:30.754644486 +0000
+++ /tmp/rbac-after.yaml        2021-09-16 23:35:13.890396448 +0000
@@ -3,7 +3,7 @@
 metadata:
   creationTimestamp: "2021-09-16T23:31:06Z"
   name: submariner-globalnet
-  resourceVersion: "859"
+  resourceVersion: "1952"
   uid: 14770711-cda4-47a7-85d3-8cb750a394be
 rules:
 - apiGroups:
@@ -13,6 +13,7 @@
   - services
   - namespaces
   - nodes
+  - endpoints
   verbs:
   - get
   - list
```

(Do the same to cluster2)

```bash
export KUBECONFIG=output/kubeconfigs/kind-config-cluster2
kubectl edit ClusterRole submariner-globalnet
```

3. Test headless service without selector

- Create headless service without selector and endpoints, and export the service
```bash
docker inspect "ext-app" -f '{{(index .NetworkSettings.Networks "'pseudo-ext'").IPAddress}}'
ip=$(docker inspect "ext-app" -f '{{(index .NetworkSettings.Networks "'pseudo-ext'").IPAddress}}')
export KUBECONFIG=output/kubeconfigs/kind-config-cluster1 
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: test-vm
spec:
  clusterIP: None
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
EOF

cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Endpoints
metadata:
  name: test-vm
subsets:
  - addresses:
      - ip: ${ip}
    ports:
      - port: 80
EOF

subctl export service -n default test-vm
```

- Check the globalnet log
```bash
kubectl logs -n submariner-operator -l app=submariner-globalnet

I0917 16:42:44.860635       1 service_export_controller.go:147] Processing ServiceExport "default/test-vm"
I0917 16:42:44.961770       1 ingress_endpoints_controller.go:93] Created Endpoints controller for (default/test-vm) with selector "metadata.name=test-vm"
I0917 16:42:44.962547       1 ingress_endpoints_controller.go:125] "create" ingress Endpoints default/test-vm for service test-vm
I0917 16:42:44.972396       1 global_ingressip_controller.go:135] Processing created &v1.GlobalIngressIP{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"ep-test-vm", GenerateName:"", Namespace:"default", SelfLink:"", UID:"a163ca0d-125c-4f80-a0ad-c471c0cf6536", ResourceVersion:"2307", Generation:1, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63767493764, loc:(*time.Location)(0x2026ea0)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string{"submariner.io/serviceRef":"test-vm"}, Annotations:map[string]string{"submariner.io/headless-svc-endpoints-ip":"172.117.0.4"}, OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:"submariner-globalnet", Operation:"Update", APIVersion:"submariner.io/v1", Time:(*v1.Time)(0xc0006a0db0), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc0006a0dc8)}}}, Spec:v1.GlobalIngressIPSpec{Target:"HeadlessServiceEndpoints", ServiceRef:(*v1.LocalObjectReference)(0xc0005c7070), PodRef:(*v1.LocalObjectReference)(0xc0005c7080)}, Status:v1.GlobalIngressIPStatus{Conditions:[]v1.Condition(nil), AllocatedIP:""}}
I0917 16:42:44.972455       1 global_ingressip_controller.go:158] Allocating global IP for "default/ep-test-vm"
I0917 16:42:44.972476       1 iface.go:145] Installing iptables rule for Headless SVC Pod -d 242.254.1.253 -j DNAT --to 172.117.0.4
I0917 16:42:44.976890       1 iface.go:227] Installing iptable egress rules for HDLS SVC Pod "default/ep-test-vm": -p all -s 172.117.0.4 -m mark --mark 0xC0000/0xC0000 -j SNAT --to 242.254.1.253
I0917 16:42:44.981070       1 base_controllers.go:194] Updated: &v1.GlobalIngressIPStatus{Conditions:[]v1.Condition{v1.Condition{Type:"Allocated", Status:"True", ObservedGeneration:0, LastTransitionTime:v1.Time{Time:time.Time{wall:0xc0494f413a79ca9a, ext:311044842452, loc:(*time.Location)(0x2026ea0)}}, Reason:"Success", Message:"Allocated global IP"}}, AllocatedIP:"242.254.1.253"}
```

- Confirm that globalingress IP is assigned to the service
```bash
kubectl get globalingressip ep-test-vm -o jsonpath='{.status.allocatedIP}{"\n"}'
242.254.1.253
```

- Test Access from cluster1 and cluster2, and check the log

```bash
export KUBECONFIG=output/kubeconfigs/kind-config-cluster1 
kubectl -n default run tmp-shell --rm -i --tty --image quay.io/submariner/nettest -- bash
curl 242.254.1.253
```

```bash
export KUBECONFIG=output/kubeconfigs/kind-config-cluster2 
kubectl -n default run tmp-shell --rm -i --tty --image quay.io/submariner/nettest -- bash
curl 242.254.1.253
```

```bash
docker logs ext-app
172.117.0.3 - - [17/Sep/2021 16:44:17] "GET / HTTP/1.1" 200 -
242.254.2.6 - - [17/Sep/2021 16:44:37] "GET / HTTP/1.1" 200 -
```

</details>

mkimuram:
Also, manually confirmed that reverse access also works as expected.

<details>

- Deploy StatefulSet in cluster2

```bash
export KUBECONFIG=output/kubeconfigs/kind-config-cluster2
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: nginx-ss
  labels:
    app.kubernetes.io/instance: nginx-ss
    app.kubernetes.io/name: nginx-ss
spec:
  ports:
  - port: 80
    name: web
  clusterIP: None
  selector:
    app.kubernetes.io/instance: nginx-ss
    app.kubernetes.io/name: nginx-ss
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  serviceName: "nginx-ss"
  replicas: 2
  selector:
    matchLabels:
        app.kubernetes.io/instance: nginx-ss
        app.kubernetes.io/name: nginx-ss
  template:
    metadata:
      labels:
        app.kubernetes.io/instance: nginx-ss
        app.kubernetes.io/name: nginx-ss
    spec:
      containers:
      - name: nginx-ss
        image: k8s.gcr.io/nginx-slim:0.8
        ports:
        - containerPort: 80
          name: web
EOF
subctl export service -n default nginx-ss
```

- Check globalingressip for each pod managed by the StatefulSet
```bash
kubectl get globalingressip pod-web-0 -o jsonpath='{.status.allocatedIP}{"\n"}'
242.254.2.252
```

```bash
kubectl get globalingressip pod-web-1 -o jsonpath='{.status.allocatedIP}{"\n"}'
242.254.2.251
```

- Test access from external app to each pod

```bash
docker exec -it ext-app bash
curl 242.254.2.252
curl 242.254.2.251
```

- Check the access log for each pod (source IPs match ext-app's globalIngressIP)
```bash
kubectl logs web-0
242.254.1.253 - - [17/Sep/2021:19:26:29 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
kubectl logs web-1
242.254.1.253 - - [17/Sep/2021:19:26:33 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
```

- Test access from each pod to external app

```bash
kubectl exec -it web-0 -- bash
curl 242.254.1.253
```

```bash
kubectl exec -it web-1 -- bash
curl 242.254.1.253
```

- Check the access log from each pod (source IPs match each pod's globalIngressIP)

```bash
docker logs ext-app
242.254.2.252 - - [17/Sep/2021 19:29:55] "GET / HTTP/1.1" 200 -
242.254.2.251 - - [17/Sep/2021 19:30:15] "GET / HTTP/1.1" 200 -
```

</details>

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	return controller, nil
+}
+
+func (c *ingressEndpointsController) process(from runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	endpoints := from.(*corev1.Endpoints)
+	key, _ := cache.MetaNamespaceKeyFunc(endpoints)
+
+	ingressIP := &submarinerv1.GlobalIngressIP{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("ep-%.59s", endpoints.Name),
```
sridhargaddam:
```suggestion
			Name:      fmt.Sprintf("ep-%.60s", endpoints.Name),
```

I think we can go with 60 here.

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,153 @@
...
+			Namespace: endpoints.Namespace,
+			Labels: map[string]string{
+				ServiceRefLabel: c.svcName,
+			},
+		},
+	}
+
+	if op == syncer.Delete {
+		c.ingressIPMap.Remove(ingressIP.Name)
+		klog.Infof("ingress Endpoints %s for service %s deleted", key, c.svcName)
```
sridhargaddam:
```suggestion
		klog.Infof("Ingress Endpoints %s for service %s deleted", key, c.svcName)
```

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	// TODO: consider other conditions
+	if c.ingressIPMap.Contains(ingressIP.Name) {
+		// Avoid assigning ingressIPs to endpoints that are not ready with an endpoint IP
+		return nil, false
+	}
+
+	klog.Infof("%q ingress Endpoints %s for service %s", op, key, c.svcName)
+
+	// TODO: consider handling multiple endpoints IPs
+	if len(endpoints.Subsets) == 0 || len(endpoints.Subsets[0].Addresses) == 0 {
```
sridhargaddam: As per this implementation, if a user manually creates an endpoint with more than one ipAddress, we will only program rules for the first entry right?

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -88,6 +90,15 @@ func NewGlobalIngressIPController(config syncer.ResourceSyncerConfig, pool *ipam
...
 
+				key, _ := cache.MetaNamespaceKeyFunc(obj)
+				return controller.iptIface.AddEgressRulesForHeadlessSVCPods(key, target, reservedIPs[0], globalNetIPTableMark)
+			} else if gip.Spec.Target == submarinerv1.HeadlessServiceEndpoints {
+				// TODO: fixme (Maybe only rename and change messages?)
+				err := controller.iptIface.AddIngressRulesForHeadlessSvcPod(reservedIPs[0], target)
+				if err != nil {
+					return err
+				}
+
```
sridhargaddam: This is same as L88-L94. I agree, we can rename it and match both Target with submarinerv1.HeadlessServicePod || submarinerv1.HeadlessServiceEndpoints

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -207,6 +218,40 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
...
+			klog.Errorf("Error while programming Service %q ingress rules for Pod: %v", key, err)
+			err = errors.WithMessage(err, "Error programming ingress rules")
+		} else {
+			err = c.iptIface.AddEgressRulesForHeadlessSVCPods(key, endpointsIP, ips[0], globalNetIPTableMark)
+			if err != nil {
+				_ = c.iptIface.RemoveIngressRulesForHeadlessSvcPod(ips[0], endpointsIP)
+				err = errors.WithMessage(err, "Error programming egress rules")
+			}
+		}
+
```
sridhargaddam: This is same as submarinerv1.HeadlessServicePod. We can remove the duplication

sridhargaddam (commented):
Thank you for the nice work @mkimuram.

Few things
1. Are you planning to push the Unit tests as part of the same PR?
2. We need changes to the RBAC to support listening for endpoints both in Submariner-operator and submariner-charts. I think you can go ahead and push these PRs first.
3. Did you test the end-to-end functionality and is it working fine?

pkg/globalnet/controllers/global_ingressip_controller.go
tpantelis: yeah it differs by annotation key. I suggest to extract a method:

 `func (c *globalIngressIPController) addRulesForHeadlessService(ingressIP *submarinerv1.GlobalIngressIP, annotationKey string, ips []string) bool`

```
if addRulesForHeadlessService(ingressIP, headlessSvcEndpointsIP, ips) {
    return true
}
```

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	ingressIP.ObjectMeta.Annotations = map[string]string{
+		headlessSvcEndpointsIP: endpoints.Subsets[0].Addresses[0].IP,
+	}
+
+	ingressIP.Spec = submarinerv1.GlobalIngressIPSpec{
+		Target:     submarinerv1.HeadlessServiceEndpoints,
+		ServiceRef: &corev1.LocalObjectReference{Name: c.svcName},
+		// TODO: Consider adding endpointsRef field to GlobalIngressIPSpec
```
tpantelis: I think we should definitely do this.

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if op == syncer.Delete {
+		c.ingressIPMap.Remove(ingressIP.Name)
+		klog.Infof("ingress Endpoints %s for service %s deleted", key, c.svcName)
+
+		return ingressIP, false
+	}
+
+	// TODO: consider other conditions
```
tpantelis: I'm not clear on this comment.

tpantelis (changes requested):

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Yes. The current implementation only uses the first entry.

This commit meant to be just a quick prototype. I expect that multiple entries can be added, however I'm not sure how multiple `GlobalIngressIP`s can be handled by using admiral's syncer framework. Can it just return `GlobalIngressIPList`? (and then need to create a controller for it?)

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Just meant to remind considering other conditions to return here, like [this](https://github.com/submariner-io/submariner/blob/devel/pkg/globalnet/controllers/ingress_pod_controller.go#L118). Removed.

mkimuram (commented):
@sridhargaddam @tpantelis  Thank you for your review. I will reflect the comments and add unit tests.

mkimuram:
> We need changes to the RBAC to support listening for endpoints both in Submariner-operator and submariner-charts. I think you can go ahead and push these PRs first.

Created issues and PRs, and added descriptions on dependencies in [the first comment](https://github.com/submariner-io/submariner/pull/1558#issue-999555499) of this PR.

mkimuram:
@sridhargaddam @tpantelis 

Addressed review comments. PTAL

> Are you planning to push the Unit tests as part of the same PR?

Unit tests are added.

> Did you test the end-to-end functionality and is it working fine?

End-to-end functionality is tested manually as described in [here](https://github.com/submariner-io/submariner/pull/1558#issuecomment-921950088) and [here](https://github.com/submariner-io/submariner/pull/1558#issuecomment-922041105). I will add automated E2E tests later on top of https://github.com/submariner-io/submariner/pull/1555.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Updated to add `endpointsRef`, however I started to think that it won't be necessary because name of Service and Endpoints are the same (I think that these fields are to track the full name for the case that it is truncated as a part of GlobalIngressIP's name, so ServiceRef field will be enough). 

If it is agreed, I will delete `endpointsRef` field.

pkg/globalnet/controllers/global_ingressip_controller.go
mkimuram: Fixed in a bit different way (to be aligned to the other place). `targetType` is passed to add to different chains for pod ("SM-GN-EGRESS-HDLS-PODS") and endpoints ("SM-GN-EGRESS-HDLS-EPS") to keep existing behavior (and it helps  distinguishing the logs).

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Note that this is still TODO.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Fixed.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Fixed.

mkimuram:
Refactored and added E2E tests. The last three commits are for E2E tests, but only the tests in the last commit are for this PR's change. I will move the other E2E commits to another PR to make it clear.

Note that the tests in the last commits fails due to the lack of access right to endpointslice (and it seems that there is a bug in the cleanup logic of headless service without selector, because unrelated tests become fail after the tests). I will investigate and fix it.

sridhargaddam:
When this PR is ready for review, please remove the status from draft.

dfarrell07:
I see this is marked as ready to review, thanks @mkimuram! It needs a manuel rebase.

pkg/globalnet/controllers/ingress_endpoints_controllers.go
```diff
@@ -16,16 +16,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//nolint:dupl // same logic to ingress_pod_controllers, but for a different class
```
mkimuram, line 19: Let me consider a bit more if I can fix it properly.

mkimuram (commented):
Rebased.

pkg/globalnet/constants/constants.go
```diff
@@ -28,6 +28,7 @@ const (
 	// The following chains are added as part of GN 2.0 implementation.
 	SmGlobalnetEgressChainForPods            = "SM-GN-EGRESS-PODS"
 	SmGlobalnetEgressChainForHeadlessSvcPods = "SM-GN-EGRESS-HDLS-PODS"
+	SmGlobalnetEgressChainForHeadlessSvcEPs  = "SM-GN-EGRESS-HDLS-EPS"
```
sridhargaddam, line 32: Can we re-use the same chain SmGlobalnetEgressChainForHeadlessSvcPods?

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -287,7 +288,12 @@ func (g *gatewayMonitor) startControllers() error {
 		return errors.Wrap(err, "error creating the IngressPodControllers")
 	}
 
-	c, err = NewServiceExportController(g.syncerConfig, podControllers)
+	endpointsControllers, err := NewIngressEndpointsControllers(g.syncerConfig)
```
sridhargaddam: nit: endpointsController and similarly NewIngressEndpointsController

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,149 @@
...
+
+	if op == syncer.Delete {
+		c.ingressIPMap.Remove(ingressIP.Name)
+		klog.Infof("Ingress Endpoints %s for service %s deleted", key, c.svcName)
+
+		return ingressIP, false
+	}
+
+	if c.ingressIPMap.Contains(ingressIP.Name) {
+		// Avoid assigning ingressIPs to endpoints that are not ready with an endpoint IP
```
sridhargaddam: I think this debug message is not relevant for endpoints

pkg/globalnet/controllers/ingress_endpoints_controller.go
sridhargaddam: > Yes. The current implementation only uses the first entry.
> 
> This commit meant to be just a quick prototype. I expect that multiple entries can be added, however I'm not sure how multiple `GlobalIngressIP`s can be handled by using admiral's syncer framework. Can it just return `GlobalIngressIPList`? (and then need to create a controller for it?)

@tpantelis can you please advise?

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,149 @@
...
+		ServiceRef: &corev1.LocalObjectReference{Name: c.svcName},
+	}
+
+	c.ingressIPMap.Add(ingressIP.Name)
+
+	return ingressIP, false
+}
+
+func areEndpointsEqual(obj1, obj2 *unstructured.Unstructured) bool {
+	subsets1, _, _ := unstructured.NestedSlice(obj1.Object, "subsets")
```
sridhargaddam: Here we are checking the entire subsets whereas while creating the IngressIP, we only consider IPAddress from subsets[0] (Line 131). Should we check only subset[0]?

pkg/globalnet/controllers/service_export_controller_test.go
```diff
@@ -253,6 +254,37 @@ func testHeadlessService() {
...
+	var service *corev1.Service
+	var endpoints *corev1.Endpoints
+
+	BeforeEach(func() {
+		service = newHeadlessServiceWithoutSelector()
+		endpoints = newHeadlessServiceEndpoints(service.Name)
+		t.createServiceExport(t.createService(service))
+	})
+
+	When("an endpoints for an exported Service is created", func() {
```
sridhargaddam: nit: endpoints -> endpoint

nyechiel:
@mkimuram can you look at the conflicts please?

mkimuram:
@nyechiel Thank you for taking a look at this PR. Actually, this PR has a dependency to other PR that is described in the first comment (and the depended PR has another depended PR that I'm working on). I will update this PR once they are merged.

mkimuram:
Created another branch https://github.com/mkimuram/submariner/commits/issue/1537v2 that is rebased on https://github.com/submariner-io/submariner/pull/1582#issuecomment-1043713012. I will update this PR based on it, once the PR is merged.

sridhargaddam:
@mkimuram PR that adds support in GN for services without selectors is now merged.

mkimuram:
I've rebased 1537v2 branch with the latest #1582 commits (E2E test PR) and will copy the branch once the E2E PR is merged.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: @sridhargaddam @anfredette @nerdalert 

> As per this implementation, if a user manually creates an endpoint with more than one ipAddress, we will only program rules for the first entry right?

The last commit adds support for multiple IPs in endpoints of headless service without selector. PTAL
(Note that first 2 commits will be merged through #1582 )

Current E2E test doesn't cover this specific case, but I've manually tested by below way and worked well:

<details>

1. Deploy the test environment

```
make deploy using=external-net,globalnet
```

2. Create headless service without selector with one endpoint and test connectivity and source IPs

- In cluster1, create headless service without selector with one endpoint
```
export KUBECONFIG=output/kubeconfigs/kind-config-cluster1
 
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: test-vm
spec:
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
  clusterIP: None
EOF
 
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Endpoints
metadata:
  name: test-vm
subsets:
  - addresses:
      - ip: 172.117.0.4
    ports:
      - port: 80
EOF
 
 
kubectl get svc,endpoints,globalingressip
NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
service/kubernetes   ClusterIP   100.0.0.1    <none>        443/TCP   23m
service/test-vm      ClusterIP   None         <none>        80/TCP    27s
 
NAME                   ENDPOINTS         AGE
endpoints/kubernetes   172.18.0.5:6443   23m
endpoints/test-vm      172.117.0.4:80    27s
 
 
subctl export service -n default test-vm
```

- globalingressip is created after the service is exported
```
kubectl get svc,endpoints,globalingressip
NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
service/kubernetes   ClusterIP   100.0.0.1    <none>        443/TCP   24m
service/test-vm      ClusterIP   None         <none>        80/TCP    55s
 
NAME                   ENDPOINTS         AGE
endpoints/kubernetes   172.18.0.5:6443   24m
endpoints/test-vm      172.117.0.4:80    55s
 
NAME                                                   IP
globalingressip.submariner.io/ep-test-vm-172.117.0.4   242.254.1.253
 
kubectl get -o yaml globalingressip.submariner.io/ep-test-vm-172.117.0.4
apiVersion: submariner.io/v1
kind: GlobalIngressIP
metadata:
  annotations:
    submariner.io/headless-svc-endpoints-ip: 172.117.0.4
  creationTimestamp: "2022-03-08T22:48:58Z"
  generation: 2
  labels:
    submariner-io/originatingNamespace: default
    submariner.io/serviceRef: test-vm
  name: ep-test-vm-172.117.0.4
  namespace: default
  resourceVersion: "5297"
  uid: 46ed1f13-e65c-4240-8374-0935313d134b
spec:
  serviceRef:
    name: test-vm
  target: HeadlessServiceEndpoints
status:
  allocatedIP: 242.254.1.253
  conditions:
  - lastTransitionTime: "2022-03-08T22:48:58Z"
    message: Allocated global IP
    reason: Success
    status: "True"
    type: Allocated
``` 
 
- In cluster2, create statefulset, service without selector and export the service.
```
export KUBECONFIG=output/kubeconfigs/kind-config-cluster2
 
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: nginx-ss
  labels:
    app.kubernetes.io/instance: nginx-ss
    app.kubernetes.io/name: nginx-ss
spec:
  ports:
  - port: 80
    name: web
  clusterIP: None
  selector:
    app.kubernetes.io/instance: nginx-ss
    app.kubernetes.io/name: nginx-ss
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  serviceName: "nginx-ss"
  replicas: 2
  selector:
    matchLabels:
        app.kubernetes.io/instance: nginx-ss
        app.kubernetes.io/name: nginx-ss
  template:
    metadata:
      labels:
        app.kubernetes.io/instance: nginx-ss
        app.kubernetes.io/name: nginx-ss
    spec:
      containers:
      - name: nginx-ss
        image: k8s.gcr.io/nginx-slim:0.8
        ports:
        - containerPort: 80
          name: web
EOF
 
subctl export service -n default nginx-ss
```

- Test access from pod to external
```
kubectl get globalingressip
NAME        IP
pod-web-0   242.254.2.252
pod-web-1   242.254.2.253
 
kubectl exec -it web-0 -- curl -m 10 242.254.1.253
 
docker logs ext-app 2>&1 | tail -n 1
242.254.2.252 - - [08/Mar/2022 22:56:55] "GET / HTTP/1.1" 200 -
 
kubectl exec -it web-1 -- curl -m 10 242.254.1.253
 
docker logs ext-app 2>&1 | tail -n 1
242.254.2.253 - - [08/Mar/2022 22:57:40] "GET / HTTP/1.1" 200 -
```

- Test access from external to pod
```
docker exec -it ext-app curl -m 10 242.254.2.252
 
kubectl logs web-0
242.254.1.253 - - [08/Mar/2022:22:59:11 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
 
 
docker exec -it ext-app curl -m 10 242.254.2.253
kubectl logs web-1
242.254.1.253 - - [08/Mar/2022:23:00:12 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
```

3. Create another test container, add it to endpoints, and test connectivity and source IPs

- Create another test container
```
docker run --cap-add=NET_ADMIN --name ext-app2 --network pseudo-ext \
-d registry.access.redhat.com/ubi7/ubi python -m SimpleHTTPServer 80
 
docker exec -i ext-app2 yum install -y iproute
docker exec -i ext-app2 ip r add 242.254.1.0/24 via 172.117.0.3
docker exec -i ext-app2 ip r add 242.254.2.0/24 via 172.117.0.3
```

- Add it to endpoints (Then new ingressglobalip ep-test-vm-172.117.0.5 is added)
```
docker inspect ext-app2 -f '{{(index .NetworkSettings.Networks "pseudo-ext").IPAddress}}'
172.117.0.5
 
export KUBECONFIG=output/kubeconfigs/kind-config-cluster1
 
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Endpoints
metadata:
  name: test-vm
subsets:
  - addresses:
      - hostname: ext-app
        ip: 172.117.0.4
      - hostname: ext-app2
        ip: 172.117.0.5
    ports:
      - port: 80
EOF
 
kubectl get svc,endpoints,globalingressip
NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
service/kubernetes   ClusterIP   100.0.0.1    <none>        443/TCP   52m
service/test-vm      ClusterIP   None         <none>        80/TCP    28m
 
NAME                   ENDPOINTS                       AGE
endpoints/kubernetes   172.18.0.5:6443                 52m
endpoints/test-vm      172.117.0.4:80,172.117.0.5:80   28m
 
NAME                                                   IP
globalingressip.submariner.io/ep-test-vm-172.117.0.4   242.254.1.253
globalingressip.submariner.io/ep-test-vm-172.117.0.5   242.254.1.252
```
 
- Test access from pod to external
```
export KUBECONFIG=output/kubeconfigs/kind-config-cluster2
 
kubectl exec -it web-0 -- curl -m 10 242.254.1.252
 
docker logs ext-app2 2>&1 | tail -n 1
242.254.2.252 - - [08/Mar/2022 23:19:05] "GET / HTTP/1.1" 200 -
 
kubectl exec -it web-1 -- curl -m 10 242.254.1.252
 
docker logs ext-app2 2>&1 | tail -n 1
242.254.2.253 - - [08/Mar/2022 23:19:25] "GET / HTTP/1.1" 200 -
```

- Test access from external to pod
```
docker exec -it ext-app2 curl -m 10 242.254.2.252
 
kubectl logs web-0
242.254.1.253 - - [08/Mar/2022:22:59:11 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
242.254.1.252 - - [08/Mar/2022:23:20:31 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
 
docker exec -it ext-app2 curl -m 10 242.254.2.253
kubectl logs web-1
242.254.1.253 - - [08/Mar/2022:23:00:12 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
242.254.1.252 - - [08/Mar/2022:23:20:42 +0000] "GET / HTTP/1.1" 200 612 "-" "curl/7.29.0"
```

</details>

mkimuram:
@nerdalert @anfredette

Please check https://github.com/submariner-io/submariner/pull/1558#discussion_r822977638 
Any feedback, like testing, review, and comments on use cases, is welcome.

mkimuram:
Rebased and resolved conflict.

mkimuram:
Let me check why the test failed and fix it.

mkimuram:
@dfarrell07 @sridhargaddam  I confirmed that accessing [via Service](https://github.com/submariner-io/submariner/blob/devel/test/external/dataplane/connectivity.go#L226) from external docker container [here](https://github.com/submariner-io/submariner/blob/devel/test/external/dataplane/connectivity.go#L233) is flaky in non-globalnet cases, even without this PR. The error happens because it takes some time before the service actually become ready to be accessed. Currently, there seems [no official way](https://github.com/kubernetes/kubernetes/issues/80828) to check if a service is actually ready. So, I've tried checking if there are any endpoints like [this workaround](https://github.com/openshift/cluster-logging-operator/pull/794/commits/d7c5aaa3cb0820908818605ad3c041b10808ce4d#diff-a4a97950ca409680f090d60ca871efdc3637f80567b3153d9d5af083b11e1cf4R158-R159) (see details below), but it still seems to fail with the change (maybe there is still a small amount of time between the endpoints field is updated and iptables rules are reflected). 

Therefore, I just added a sleep for this path in the last commit and it doesn't seem to flake now. Note that as for globalnet cases, they wait for globalIP to be assigned (and at that time they seem to be ready to be accessed), so they don't seem to flake.

<details>

```golang
diff --git a/test/external/dataplane/connectivity.go b/test/external/dataplane/connectivity.go
index 747b3ee..7a63750 100644
--- a/test/external/dataplane/connectivity.go
+++ b/test/external/dataplane/connectivity.go
@@ -19,6 +19,7 @@ limitations under the License.
 package dataplane
 
 import (
+	"context"
 	"fmt"
 	"sort"
 
@@ -26,6 +27,9 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/submariner-io/shipyard/test/e2e/tcp"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 const (
@@ -204,6 +208,7 @@ func testExternalConnectivity(p testParams) {
 		Command:       simpleHTTPServerCommand,
 	})
 	svc := np.CreateService()
+	awaitServiceReady(p.Cluster, svc.Namespace, svc.Name)
 
 	// Get handle for existing docker
 	docker := framework.New(extAppName)
@@ -298,3 +303,37 @@ func getNonGatewayClusterIndex(names []string) (framework.ClusterIndex, error) {
 	return framework.ClusterIndex(0),
 		fmt.Errorf("failed to find non-gateway-cluster (gateway-cluster name: %s, cluster names: %v)", clusterName, names)
 }
+
+func awaitServiceReady(cluster framework.ClusterIndex, namespace, svcName string) {
+	framework.AwaitUntil("await non-empty endpoints",
+		func() (interface{}, error) {
+			ep, err := framework.KubeClients[cluster].CoreV1().Endpoints(namespace).Get(context.TODO(), svcName, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return nil, nil // nolint:nilnil // Returning nil value is intentional
+			}
+
+			return ep, err
+		},
+		func(result interface{}) (bool, string, error) {
+			if result == nil {
+				return false, fmt.Sprintf("endpoints %s/%s not found", namespace, svcName), nil
+			}
+
+			endpoints := result.(*v1.Endpoints)
+
+			var found bool
+			for _, subset := range endpoints.Subsets {
+				for _, addr := range subset.Addresses {
+					if addr.IP != "" {
+						found = true
+					}
+				}
+			}
+
+			if !found {
+				return false, fmt.Sprintf("no IP found in endpoints %s/%s", namespace, svcName), nil
+			}
+
+			return true, "", nil
+		})
+}
diff --git a/test/external/dataplane/gn_connectivity.go b/test/external/dataplane/gn_connectivity.go
index cea64e7..c9bfea4 100644
--- a/test/external/dataplane/gn_connectivity.go
+++ b/test/external/dataplane/gn_connectivity.go
@@ -332,6 +332,7 @@ func testGlobalNetExternalConnectivity(p testParams, g globalnetTestParams) {
 		Command:       simpleHTTPServerCommand,
 	})
 	svc := createSvc(p, np)
+	awaitServiceReady(p.Cluster, svc.Namespace, svc.Name)
 	p.Framework.CreateServiceExport(np.Config.Cluster, svc.Name)
 
 	// Get globalIPs for the network pod to use later
```

</details>

mkimuram:
Removed the workaround commit and rebased to include the fix for the flakey tests. Now it should be ready to be merged.

mkimuram:
@sridhargaddam @tpantelis Could you approve this PR?

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -390,6 +396,12 @@ func (g *gatewayMonitor) createGlobalnetChains() error {
 		return errors.Wrapf(err, "error creating iptables chain %s", constants.SmGlobalnetEgressChainForHeadlessSvcPods)
 	}
 
+	klog.V(log.DEBUG).Infof("Install/ensure %s chain exists", constants.SmGlobalnetEgressChainForHeadlessSvcEPs)
+
+	if err := iptables.CreateChainIfNotExists(g.ipt, "nat", constants.SmGlobalnetEgressChainForHeadlessSvcEPs); err != nil {
```
sridhargaddam, line 401: Does this create issues during upgrades (i.e., what if the existing chains are already present and we are trying to insert a new chain in between)?

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,142 @@
...
+		svcName:              svc.Name,
+		namespace:            svc.Namespace,
+		ingressIPMap:         stringset.NewSynchronized(),
+	}
+
+	fieldSelector := fields.Set(map[string]string{"metadata.name": svc.Name}).AsSelector().String()
+
+	controller.resourceSyncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:                "Ingress Endpoints syncer",
+		ResourceType:        &corev1.Endpoints{},
```
sridhargaddam, line 60: Can you please explain the impact on this code when the endpoints are deprecated/removed and replaced with EndpointSlices?

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,142 @@
...
+		RestMapper:          config.RestMapper,
+		Federator:           federate.NewCreateFederator(config.SourceClient, config.RestMapper, corev1.NamespaceAll),
+		Scheme:              config.Scheme,
+		Transform:           controller.process,
+		SourceFieldSelector: fieldSelector,
+		ResourcesEquivalent: areEndpointsEqual,
+	})
+
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating the syncer")
```
sridhargaddam: nit: "error creating the endpoints syncer"

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if err := controller.Start(); err != nil {
+		return nil, errors.Wrap(err, "error creating the syncer")
```
sridhargaddam: "error starting the endpoint syncer"

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,142 @@
...
+		Transform:           controller.process,
+		SourceFieldSelector: fieldSelector,
+		ResourcesEquivalent: areEndpointsEqual,
+	})
+
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating the syncer")
+	}
+
+	if err := controller.Start(); err != nil {
```
sridhargaddam, line 75: Any specific reason for starting the controller here instead of in gateway_monitor.go Line 316?

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return nil, true
+	}
+
+	// Delete unused ingressIPs
+	for _, ingressIP := range existingIngressIPs.Items {
+		annotations := ingressIP.GetAnnotations()
+		if epIP, ok := annotations[headlessSvcEndpointsIP]; ok {
+			if _, used := epIPs[epIP]; used {
+				// ingressIP is still used
```
sridhargaddam: If the ingressIP is still used, should we remove it from the epIP list? As otherwise we seem to create new IngressIPs in Line 152. Am I missing something?

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,203 @@
...
+
+	klog.Infof("Ingress Endpoints %s for service %s deleted", key, c.svcName)
+
+	return nil, false
+}
+
+func newIngressIP(name, namespace, epIP string) *submarinerv1.GlobalIngressIP {
+	return &submarinerv1.GlobalIngressIP{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("ep-%.44s-%.15s", name, epIP),
```
sridhargaddam, line 199: So if the user creates two different endpoints for different services where the initial 44 chars of names match and epIP matches, we would have a problem here. Is this correct? Its a very corner case, but just want to hear from you if I'm reading the code correctly.

pkg/globalnet/controllers/iptables/iface.go
```diff
@@ -105,13 +112,13 @@ func (i *ipTables) ipTableChainExists(table, chain string) (bool, error) {
 	return false, nil
 }
 
-func (i *ipTables) AddIngressRulesForHeadlessSvcPod(globalIP, podIP string) error {
-	if globalIP == "" || podIP == "" {
-		return fmt.Errorf("globalIP %q or podIP %q cannot be empty", globalIP, podIP)
+func (i *ipTables) AddIngressRulesForHeadlessSvc(globalIP, ip string, targetType TargetType) error {
+	if globalIP == "" || ip == "" {
+		return fmt.Errorf("globalIP %q or %s IP %q cannot be empty", targetType, globalIP, ip)
```
sridhargaddam:
```suggestion
		return fmt.Errorf("globalIP %q or %s IP %q cannot be empty", globalIP, targetType, ip)
```

pkg/globalnet/controllers/iptables/iface.go
```diff
@@ -120,14 +127,14 @@ func (i *ipTables) AddIngressRulesForHeadlessSvcPod(globalIP, podIP string) erro
 	return nil
 }
 
-func (i *ipTables) RemoveIngressRulesForHeadlessSvcPod(globalIP, podIP string) error {
-	if globalIP == "" || podIP == "" {
-		return fmt.Errorf("globalIP %q or podIP %q cannot be empty", globalIP, podIP)
+func (i *ipTables) RemoveIngressRulesForHeadlessSvc(globalIP, ip string, targetType TargetType) error {
+	if globalIP == "" || ip == "" {
+		return fmt.Errorf("globalIP %q or %s IP %q cannot be empty", targetType, globalIP, ip)
```
sridhargaddam:
```suggestion
		return fmt.Errorf("globalIP %q or %s IP %q cannot be empty", globalIP, targetType, ip)
```

sridhargaddam (changes requested):
Thanks for the PR @mkimuram.

pkg/globalnet/controllers/gateway_monitor.go
mkimuram: I think that the jump rule to this chain is added by using `iptables.InsertUnique` with specifying the number to be inserted, so where this chain is inserted would be consistent regardless of whether the other chains already present.

However, sharing `SmGlobalnetEgressChainForHeadlessSvcPods` chain may be the other way to workaround it, as you mentioned.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Correct. It's similar restriction to [the one for pod](https://github.com/submariner-io/submariner/blob/devel/pkg/globalnet/controllers/ingress_pod_controller.go#L108), but much shorter than than the pod one. We might be able to use hash, but it would make it harder to recognize what endpoints this ingress is for. WDYT?

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: I thought that it is safe to ensure the ingressIP resource to exist, but I noticed that calling `create` without checking already exist error doesn't do the intended works. So, it would be better removing it from epIP as you mentioned. Thank you for pointing it out.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: If `Endpoints` are deprecated, we would need to re-implement this to use `EndpointSlices`. However, I'm not sure if there are fixed plan to deprecate it.

For this particular headless service without selector case, I think that we should use `Endpoints`, because `"The control plane automatically creates EndpointSlices for any Kubernetes Service that has a selector specified."` as described [here](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#endpointslice-resource). As I tested as below, `EndpointSlices` doesn't seem to be created automatically, for manually created `Endpoints`. 

<details>

```bash
kubectl get endpoints,endpointslices
NAME                   ENDPOINTS         AGE
endpoints/kubernetes   172.18.0.2:6443   2m39s

NAME                                        ADDRESSTYPE   PORTS   ENDPOINTS    AGE
endpointslice.discovery.k8s.io/kubernetes   IPv4          6443    172.18.0.2   2m39s

cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Endpoints
metadata:
  name: test-vm
subsets:
  - addresses:
      - ip: 192.168.4.2
    ports:
      - port: 80
EOF

kubectl get endpoints,endpointslices
NAME                   ENDPOINTS         AGE
endpoints/kubernetes   172.18.0.2:6443   3m48s
endpoints/test-vm      192.168.4.2:80    4s

NAME                                        ADDRESSTYPE   PORTS   ENDPOINTS    AGE
endpointslice.discovery.k8s.io/kubernetes   IPv4          6443    172.18.0.2   3m48s
```
</details>

pkg/globalnet/constants/constants.go
mkimuram: I think that we can. However, the name is a bit confusing and it makes rules for different purpose exist in one chain. If there is no side effect(like update issue), it would be clean to add a separate chain.

mkimuram (commented):
@sridhargaddam Thank you for your review and comments. I put some comments back and will update the code latter.

pkg/globalnet/constants/constants.go
sridhargaddam: I had a look at the code and thought about it for sometime and I think it should be okay even if we introduce a new chain in between. Can you please verify once in your local setup running this updated Globalnet image on an existing (v0.12.0) Globalnet setup. If you find that there is no issue, we can go with the new chain.

pkg/globalnet/controllers/ingress_endpoints_controller.go
sridhargaddam: Ok, thanks @mkimuram

pkg/globalnet/controllers/ingress_endpoints_controller.go
sridhargaddam: Agree, using a hast makes it difficult to recognise it. Since its a very corner case, lets continue this way.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: In my understanding, it needs to be started here, because this controller is started via `IngressEndpointsControllers.start()` per Service basis. It is using the similar approach that ingress_pod_controllers.go does for ingress_pod_controller.go.

pkg/globalnet/controllers/service_export_controller_test.go
mkimuram: fixed.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: Fixed as suggested.

pkg/globalnet/controllers/iptables/iface.go
mkimuram: Fixed as suggested.

pkg/globalnet/controllers/iptables/iface.go
mkimuram: Fixed as suggested.

pkg/globalnet/controllers/ingress_endpoints_controller.go
mkimuram: In my understanding, it needs to be started here, because this controller is started via `IngressEndpointsControllers.start()` per Service basis. It is using the similar approach that ingress_pod_controllers.go does for ingress_pod_controller.go.

pkg/globalnet/constants/constants.go
mkimuram: @sridhargaddam Confirmed that it worked as expected. Please also see below details:

<details>

- Build the container image for this PR

```bash
git checkout issue/1537
make images
docker tag quay.io/submariner/submariner-globalnet:dev quay.io/submariner/submariner-globalnet:issue1537
```

- Deploy v0.12.0 submariner

```bash
git checkout v0.12.0

make deploy using=external-net,globalnet
```

- Confirm the nat chain for v0.12.0 (`SM-GN-EGRESS-HDLS-EPS` doesn't exist)

```bash
docker exec -it cluster1-worker iptables -t nat -S SUBMARINER-GN-EGRESS
# Warning: iptables-legacy tables present, use iptables-legacy to see them
-N SUBMARINER-GN-EGRESS
-A SUBMARINER-GN-EGRESS -j SUBMARINER-GN-MARK
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-HDLS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-NS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-CLUSTER

docker exec -it cluster2-worker iptables -t nat -S SUBMARINER-GN-EGRESS
# Warning: iptables-legacy tables present, use iptables-legacy to see them
-N SUBMARINER-GN-EGRESS
-A SUBMARINER-GN-EGRESS -j SUBMARINER-GN-MARK
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-HDLS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-NS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-CLUSTER
```

- Test that existing test pass
```bash
make e2e using=external-net,globalnet

• [SLOW TEST:46.771 seconds]
[external-dataplane] Connectivity
/go/src/github.com/submariner-io/submariner/test/external/dataplane/connectivity
.go:41
  should be able to connect from an external app to a pod in a cluster
  /go/src/github.com/submariner-io/submariner/test/external/dataplane/connectivi
ty.go:44
------------------------------

JUnit path was configured: /go/src/github.com/submariner-io/submariner/output/e2
e-junit.xml

JUnit report was created: /go/src/github.com/submariner-io/submariner/output/e2e
-junit.xml

Ran 1 of 1 Specs in 46.822 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2E (46.82s)
PASS
```

- Update the globalnet image to the one built from this PR
(submariner-operator doesn't allow directly updating globalnet daemonset because it [reconciles the image](https://github.com/submariner-io/submariner-operator/blob/devel/controllers/submariner/globalnet_resources.go#L78). So, it seems that we need to update the `Submariner` CR).

```bash
kind load docker-image quay.io/submariner/submariner-globalnet:issue1537 --name cluster1
kind load docker-image quay.io/submariner/submariner-globalnet:issue1537 --name cluster2

export KUBECONFIG=output/kubeconfigs/kind-config-cluster1
kubectl get daemonset submariner-globalnet -n submariner-operator -o jsonpath='{range .spec.template.spec.containers[*]}{.name}{" "}{.image}{"\n"}{end}'
submariner-globalnet localhost:5000/submariner-globalnet:local

kubectl get submariner -n submariner-operator -o yaml > /tmp/submariner_cr.txt
kubectl edit submariner -n submariner-operator
kubectl get submariner -n submariner-operator -o yaml > /tmp/submariner_cr_after.txt

diff -urN  /tmp/submariner_cr.txt /tmp/submariner_cr_after.txt | grep -C 2 "imageOverrides"
     debug: false
     globalCIDR: 242.254.1.0/24
+    imageOverrides:
+      submariner-globalnet: quay.io/submariner/submariner-globalnet:issue1537
     namespace: submariner-operator

kubectl get daemonset submariner-globalnet -n submariner-operator -o jsonpath='{range .spec.template.spec.containers[*]}{.name}{" "}{.image}{"\n"}{end}'
submariner-globalnet quay.io/submariner/submariner-globalnet:issue1537
```

```bash
export KUBECONFIG=output/kubeconfigs/kind-config-cluster2
kubectl edit submariner -n submariner-operator

kubectl get daemonset submariner-globalnet -n submariner-operator -o jsonpath='{range .spec.template.spec.containers[*]}{.name}{" "}{.image}{"\n"}{end}'
submariner-globalnet quay.io/submariner/submariner-globalnet:issue1537
```

- Confirm the nat chain for this PR (`SM-GN-EGRESS-HDLS-EPS` exists in the right order)

```bash
docker exec -it cluster1-worker iptables -t nat -S SUBMARINER-GN-EGRESS
# Warning: iptables-legacy tables present, use iptables-legacy to see them
-N SUBMARINER-GN-EGRESS
-A SUBMARINER-GN-EGRESS -j SUBMARINER-GN-MARK
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-HDLS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-HDLS-EPS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-NS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-CLUSTER

docker exec -it cluster2-worker iptables -t nat -S SUBMARINER-GN-EGRESS
# Warning: iptables-legacy tables present, use iptables-legacy to see them
-N SUBMARINER-GN-EGRESS
-A SUBMARINER-GN-EGRESS -j SUBMARINER-GN-MARK
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-HDLS-PODS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-HDLS-EPS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-NS
-A SUBMARINER-GN-EGRESS -j SM-GN-EGRESS-CLUSTER
```

- Test that e2e test for this PR succeeds

```bash
git checkout issue/1537

go mod vendor

make "e2e" using="external-net globalnet"


Ran 15 of 28 Specs in 281.317 seconds
SUCCESS! -- 15 Passed | 0 Failed | 7 Pending | 6 Skipped
--- PASS: TestE2E (281.34s)
PASS

```

</details>

pkg/globalnet/controllers/uninstall.go
```diff
@@ -52,6 +52,7 @@ func UninstallDataPath() {
 		// The chains have to be deleted in a specific order.
 		constants.SmGlobalnetEgressChainForCluster,
 		constants.SmGlobalnetEgressChainForHeadlessSvcPods,
+		constants.SmGlobalnetEgressChainForHeadlessSvcEPs,
```
mkimuram, line 59: Also, added uninstall code here. 

By this change,  this version of subctl comand starts to try deleting this chain, while old globalnet deployment doesn't have this chain. However,  uninstallation continues even `FlushIPTableChain` fails, so there won't be issue by version mismatch. 

Actually, [`ClearChain`](https://github.com/coreos/go-iptables/blob/master/iptables/iptables.go#L427) called from `FlushIPTableChain` doesn't seem to return error even when the chain doesn't exist already (it creates new one and flush). So, it would be safe to just add the chain to the list to be deleted.

sridhargaddam (approved):
Thanks for addressing the review comments @mkimuram.

@tpantelis @anfredette can you please review.

sridhargaddam:
@mkimuram The last commit can be squashed with the previous one as there is a linting error.

pkg/globalnet/controllers/base_controllers.go
```diff
@@ -73,11 +73,21 @@ func (c *baseSyncerController) Start() error {
...
+		listOptions.LabelSelector = labelSelector
+	}
+
+	if fieldSelector != "" {
+		listOptions.FieldSelector = fieldSelector
+	}
+
 	c.resourceSyncer.Reconcile(func() []runtime.Object {
-		objList, err := client.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+		objList, err := client.List(context.TODO(), listOptions)
```
tpantelis: We can simplify this.

```suggestion
		objList, err := client.List(context.TODO(), metav1.ListOptions{
		    LabelSelector: labelSelector,
		    FieldSelector: fieldSelector,
		})
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -214,24 +224,42 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
...
+			target = ingressIP.GetAnnotations()[headlessSvcEndpointsIP]
+			if target == "" {
+				_ = c.pool.Release(ips...)
+
+				klog.Warningf("%q annotation is missing on %q", headlessSvcEndpointsIP, key)
+
+				return true
+			}
+
+			tType = iptables.EndpointsTarget
```
tpantelis: We can simplify to:

```Go
var annotationKey string
var tType iptables.TargetType

if ingressIP.Spec.Target == submarinerv1.HeadlessServicePod {
    annotationKey = headlessSvcPodIP
    tType = iptables.PodTarget
} else if ingressIP.Spec.Target == submarinerv1.HeadlessServiceEndpoints {
    annotationKey = headlessSvcEndpointsIP
    tType = iptables.EndpointsTarget
}

target = ingressIP.GetAnnotations()[annotationKey]
if target == "" {
    _ = c.pool.Release(ips...)

    klog.Warningf("%q annotation is missing on %q", annotationKey, key)

    return true
}
```

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,206 @@
...
+	return nil, false
+}
+
+func (c *ingressEndpointsController) onCreateOrUpdate(endpoints *corev1.Endpoints, op syncer.Operation) (runtime.Object, bool) {
+	key, _ := cache.MetaNamespaceKeyFunc(endpoints)
+
+	klog.Infof("%q ingress Endpoints %s for service %s", op, key, c.svcName)
+
+	// Get list of current endpoint IPs
+	epIPs := map[string]bool{}
```
tpantelis:
```suggestion
	usedEpIPs := map[string]bool{}
```

pkg/globalnet/controllers/ingress_endpoints_controller.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return nil, true
+	}
+
+	// Delete unused ingressIPs
+	for _, ingressIP := range existingIngressIPs.Items {
+		annotations := ingressIP.GetAnnotations()
+		if epIP, ok := annotations[headlessSvcEndpointsIP]; ok {
+			// ingressIP is still used
+			if _, used := epIPs[epIP]; used {
```
tpantelis:
```suggestion
			if usedEpIPs[epIP] {
```

tpantelis (changes requested):
Looks good - just a couple minor things.

mkimuram:
@tpantelis Thank you for your review. Fixed as suggested.

Note that `"Linting / Commit Message(s)"` and `"Linting / Markdown Links (modified files)"` are constantly failing, so "not required". Also, `"Upgrade / Latest Release to Latest Version"` seems flaking (but succeeded this time).

tpantelis (approved):

PR #1569, opened 2021-10-15
Wrap validation error instead of recreating

wangchenglong01:
## Which problem is this PR solving?
- Wrap validation error instead of recreating

## Short description of the changes
- Wrap validation error instead of recreating

Depends on submariner-io/releases#241

### DIFF
--- a/pkg/routeagent_driver/cni/cni_iface.go
+++ b/pkg/routeagent_driver/cni/cni_iface.go
@@ -20,7 +20,6 @@ package cni
 
 import (
 	"context"
-	"fmt"
 	"net"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


skitt (approved):

dfarrell07 (approved):

wangchenglong01:
How to fix the failing of E2E?

dfarrell07:
> How to fix the failing of E2E?

That vxlan+ovn test isn't marked as required because they are both new features and seem to still need work. Your PR very likely didn't impact that test, cause the failure. So no need to worry about it, we're just giving some time for folks to review. 

Although folks might be thinking about if this should go in now, right at the end of validating 0.11, or wait until after 0.11.  I'll leave that for others. :) /cc @tpantelis @maayanf24

maayanf24:
> > How to fix the failing of E2E?
> 
> That vxlan+ovn test isn't marked as required because they are both new features and seem to still need work. Your PR very likely didn't impact that test, cause the failure. So no need to worry about it, we're just giving some time for folks to review.
> 
> Although folks might be thinking about if this should go in now, right at the end of validating 0.11, or wait until after 0.11. I'll leave that for others. :) /cc @tpantelis @maayanf24

Let's wait and merge this after the release of 0.11

pkg/routeagent_driver/cni/cni_iface.go
```diff
@@ -91,13 +91,13 @@ func discover(clusterCIDR string) (*Interface, error) {
 func AnnotateNodeWithCNIInterfaceIP(nodeName string, clientSet kubernetes.Interface, clusterCidr []string) error {
 	cniIface, err := Discover(clusterCidr[0])
 	if err != nil {
-		return fmt.Errorf("DiscoverCNIInterface returned error %v", err)
+		return errors.Wrapf(err, "DiscoverCNIInterface returned error")
```
tpantelis:
```suggestion
		return errors.Wrap(err, "Discover returned error")
```

tpantelis (changes requested):

tpantelis (approved):

PR #1571, opened 2021-10-18
Ignore endpoint events if processed already.

### LINKED ISSUE #1544: Gateway migration in one cluster is causing failures in remote cluster
**What happened**:
In a two-cluster setup (perf1 and perf2), install Submariner (with NAT disabled) to connect both the clusters and ensure that connections are successfully established. In such a setup, when the active gateway migrates on the Perf1 cluster, the routes in table 150 on the Active Gateway node of Perf2 are accidentally deleted. Because of this, pinger (aka Healthcheck) which runs on Perf2 gateway is unable to reach Perf1 health-check IP. Along with this, Hostnetworking to remoteCluster from Perf2 is broken.

When the active Gateway migrates, the HostNetworking use-cases should continue to work.

A temporary work-around is to restart the route-agent running on the active Gateway node of the Perf2 cluster.

**Environment**:
Submariner version:	v0.9
Kubernetes Server version: v1.21.1+9807387
CNI: OpenShiftSDN on VmWare Clusters

Logs from route-agent running on the Active Gateway node of Perf2 cluster are attached:
[api-perf2-chris-ocs-ninja_6443_submariner-routeagent-xzx8f.log](https://github.com/submariner-io/submariner/files/7096459/api-perf2-chris-ocs-ninja_6443_submariner-routeagent-xzx8f.log)

aswinsuryan:
In deployments, if we delete submariner-gateway machineset and then
later delete the submariner namespace in a cluster. It would lead
to a stale endpoint in other clusters. Now if we create a machineset
and join the cluster again, the old endpoint gets cleared. But the old
endpoint delete event comes after the new endpoint is created and the
host network routing rules gets deleted.

With this fix, a delete event will be ignored if it has timestamp older
than the current endpoint.

Fixes: #1544

### DIFF
--- a/pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
+++ b/pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
@@ -105,6 +105,14 @@ func (kp *SyncHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
 	kp.syncHandlerMutex.Lock()
 	defer kp.syncHandlerMutex.Unlock()
 
+	lastProcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.ClusterName]
+
+	if ok && lastProcessedTime.After(endpoint.CreationTimestamp.Time) {
+		klog.Infof("Ignoring new remote %#v since a later endpoint was already"+
+			"processed", endpoint)
+		return nil
+	}
+
 	for _, inputCidrBlock := range endpoint.Spec.Subnets {
 		if !kp.remoteSubnets.Contains(inputCidrBlock) {
 			kp.remoteSubnets.Add(inputCidrBlock)
@@ -124,6 +132,8 @@ func (kp *SyncHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
 	kp.updateRoutingRulesForHostNetworkSupport(endpoint.Spec.Subnets, Add)
 	kp.updateIptableRulesForInterClusterTraffic(endpoint.Spec.Subnets, Add)
 
+	kp.remoteEndpointTimeStamp[endpoint.ClusterName] = endpoint.CreationTimestamp
+
 	return nil
 }
 
@@ -134,6 +144,16 @@ func (kp *SyncHandler) RemoteEndpointUpdated(endpoint *submV1.Endpoint) error {
 func (kp *SyncHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 	kp.syncHandlerMutex.Lock()
 	defer kp.syncHandlerMutex.Unlock()
+
+	lastProcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.ClusterName]
+
+	if ok && lastProcessedTime.After(endpoint.CreationTimestamp.Time) {
+		klog.Infof("Ignoring deleted remote %#v since a later endpoint was already"+
+			"processed", endpoint)
+		return nil
+	}
+
+	delete(kp.remoteEndpointTimeStamp, endpoint.ClusterName)
 
 	for _, inputCidrBlock := range endpoint.Spec.Subnets {
 		kp.remoteSubnets.Remove(inputCidrBlock)
--- a/pkg/routeagent_driver/handlers/kubeproxy/kp_iptables.go
+++ b/pkg/routeagent_driver/handlers/kubeproxy/kp_iptables.go
@@ -29,6 +29,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/netlink"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/cni"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 )
 
@@ -38,10 +39,11 @@ type SyncHandler struct {
 	localClusterCidr []string
 	localServiceCidr []string
 
-	remoteSubnets    stringset.Interface
-	remoteSubnetGw   map[string]net.IP
-	remoteVTEPs      stringset.Interface
-	routeCacheGWNode stringset.Interface
+	remoteSubnets           stringset.Interface
+	remoteSubnetGw          map[string]net.IP
+	remoteVTEPs             stringset.Interface
+	routeCacheGWNode        stringset.Interface
+	remoteEndpointTimeStamp map[string]v1.Time
 
 	syncHandlerMutex     sync.Mutex
 	isGatewayNode        bool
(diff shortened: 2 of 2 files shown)


pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -101,6 +101,15 @@ func (kp *SyncHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
 
 	kp.syncHandlerMutex.Lock()
 	defer kp.syncHandlerMutex.Unlock()
+
+	lastPorcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.Name]
```
tpantelis:
```suggestion
	lastProcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.Name]
```

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -101,6 +101,15 @@ func (kp *SyncHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
...
+
+	if ok && lastPorcessedTime.After(endpoint.CreationTimestamp.Time) {
+		klog.Infof("Ingoring new remote %#v since a later endoint was already"+
```
tpantelis: I'm not sure how noisy this could be but It seems this should be DEBUG or TRACE level.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -129,6 +140,17 @@ func (kp *SyncHandler) RemoteEndpointUpdated(endpoint *submV1.Endpoint) error {
 func (kp *SyncHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 	kp.syncHandlerMutex.Lock()
 	defer kp.syncHandlerMutex.Unlock()
+
+	lastPorcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.Name]
```
tpantelis:
```suggestion
	lastProcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.Name]
```

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -129,6 +140,17 @@ func (kp *SyncHandler) RemoteEndpointUpdated(endpoint *submV1.Endpoint) error {
...
+
+	if ok && lastPorcessedTime.After(endpoint.CreationTimestamp.Time) {
+		klog.Infof("Ingoring deleted remote %#v since a later endoint was already"+
```
tpantelis: DEBUG or TRACE?

tpantelis (commented):
Can you explain in the PR summary the proposed solution and how it fixes the issue?

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -129,6 +140,17 @@ func (kp *SyncHandler) RemoteEndpointUpdated(endpoint *submV1.Endpoint) error {
...
+			"processed", endpoint)
+		return nil
```
sridhargaddam, line 153: The exact problem that we noticed in OCP deployment was that a stale/old endpoint event coming after an updated event. Here we are skipping the old one, but what about deleting any resources (like iptables, routes etc) which point to the old endpoint? Can you please point me where we will handle it with this change?

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
sridhargaddam: Ideally, these events should happen only during migration and if there is out-of-order events. So, I would guess it should be less noisy.

sridhargaddam (changes requested):

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
tpantelis: This code correlates the time stamp between endpoints with the same name. I'm curious what sequence of events can occur for endpoint(s) with the same name that would cause an issue? Can the exact scenario sequence that we're trying to solve be outlined in the PR?

tpantelis (changes requested):

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
aswinsuryan: Added more details to the commit message.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
aswinsuryan: Ya it should happen only in corner conditions and should be good for us to know if it happens.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
aswinsuryan: Retaining Info as mentioned above.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
sridhargaddam: @aswinsuryan, thanks for the PR. I see that you are now skipping the endpoint processing if its arriving after the new endpoint. Can you please check that on the old gateway node, the following entries are removed. We are seeing this more often in OCP setup.

1. IP route table 150 entries
2. vx-submariner interface 
3. IP xfrm policy/state rules

sridhargaddam (commented):
Thanks for the PR @aswinsuryan. I have one comment.

sridhargaddam:
@aswinsuryan ping.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
aswinsuryan: @sridhargaddam This issue occurs when the old g/w node(machine set) is deleted and a new one is created. So in the remote cluster, the old one does not exist. In the local cluster, g/w node did not change, but it received a stale event. So we need to keep that rules.

Is  that what you meant to do?

aswinsuryan:
> @aswinsuryan ping.
Replied in the comments.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
dfarrell07: > This issue occurs when the old g/w node(machine set) is deleted and a new one is created.

Is it possible to add tests for this scenario? I'm thinking of the [CII Silver](https://bestpractices.coreinfrastructure.org/en/criteria) requirement: 

> The project MUST add regression tests to an automated test suite for at least 50% of the bugs fixed within the last six months.

aswinsuryan: @dfarrell07 I am not sure if it can be done in kind,may be removing a worker node and adding it back help.

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -135,6 +145,16 @@ func (kp *SyncHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 	kp.syncHandlerMutex.Lock()
 	defer kp.syncHandlerMutex.Unlock()
 
+	lastProcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.ClusterName]
+
+	if ok && lastProcessedTime.After(endpoint.CreationTimestamp.Time) {
+		klog.Infof("Ignoring deleted remote %#v since a later endoint was already"+
+			"processed", endpoint)
```
sridhargaddam:
```suggestion
		klog.Infof("Ignoring deleted remote %#v since a later endpoint was already"+
			"processed", endpoint)
```

pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go
```diff
@@ -105,6 +105,14 @@ func (kp *SyncHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
 	kp.syncHandlerMutex.Lock()
 	defer kp.syncHandlerMutex.Unlock()
 
+	lastProcessedTime, ok := kp.remoteEndpointTimeStamp[endpoint.ClusterName]
+
+	if ok && lastProcessedTime.After(endpoint.CreationTimestamp.Time) {
+		klog.Infof("Ignoring new remote %#v since a later endoint was already"+
```
sridhargaddam:
```suggestion
		klog.Infof("Ignoring new remote %#v since a later endpoint was already"+
```

sridhargaddam (commented):
Thanks @aswinsuryan. Some minor nits

tpantelis (approved):

dfarrell07 (approved):

sridhargaddam (approved):

PR #1575, opened 2021-10-25
Support e2e on single node deployments

sridhargaddam:
Currently, in e2e, we schedule pods sometimes on gateway nodes and sometimes
on non-gateway nodes to cover a wide range of use-cases. However, when the
same tests are executed on a single node deployments it returns an unfriendly
error message which would be hard to understand. This PR modifies the code
to run only the applicable test-scenarios.

Depends on: https://github.com/submariner-io/shipyard/pull/654

### DIFF
--- a/test/e2e/framework/dataplane.go
+++ b/test/e2e/framework/dataplane.go
@@ -241,3 +241,22 @@ func GetGlobalnetEgressParams(egressIP GlobalEgressIPType) GlobalnetTestParams {
 		GlobalEgressIP:   egressIP,
 	}
 }
+
+func IsSingleNodeDeployment(sourceNode, destNode framework.NetworkPodScheduling,
+	sourceCluster, destCluster framework.ClusterIndex) bool {
+	if sourceNode == framework.NonGatewayNode &&
+		framework.TestContext.NumNodesInCluster[sourceCluster] == 1 {
+		framework.Skipf("Skipping the test as cluster %q has only a single node...",
+			framework.TestContext.ClusterIDs[sourceCluster])
+		return true
+	}
+
+	if destNode == framework.NonGatewayNode &&
+		framework.TestContext.NumNodesInCluster[destCluster] == 1 {
+		framework.Skipf("Skipping the test as cluster %q has only a single node...",
+			framework.TestContext.ClusterIDs[destCluster])
+		return true
+	}
+
+	return false
+}
--- a/test/e2e/framework/dataplane.go
+++ b/test/e2e/framework/dataplane.go
@@ -242,21 +242,21 @@ func GetGlobalnetEgressParams(egressIP GlobalEgressIPType) GlobalnetTestParams {
 	}
 }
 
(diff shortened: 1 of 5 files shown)


sridhargaddam:
Marking this PR as draft as it's not required for 0.11 release

nyechiel:
@sridhargaddam is still something we still want to merge?

sridhargaddam:
> @sridhargaddam is still something we still want to merge?

Yes @nyechiel, its ready for review/merge.

CC: @tpantelis @skitt

tpantelis (commented):
The gateway fail-over test requires at least one node - should check and skip that one too?

test/e2e/framework/dataplane.go
```diff
@@ -241,3 +241,22 @@ func GetGlobalnetEgressParams(egressIP GlobalEgressIPType) GlobalnetTestParams {
 		GlobalEgressIP:   egressIP,
 	}
 }
+
+func IsSingleNodeDeployment(sourceNode, destNode framework.NetworkPodScheduling,
```
tpantelis: The function name is kind of a misnomer - it's really checking if a connectivity test can be run involving a non-gateway node so how about:

```suggestion
func CanExecuteNonGatewayConnectivityTest(sourceNode, destNode framework.NetworkPodScheduling,
```

sridhargaddam:
> The gateway fail-over test requires at least one node - should check and skip that one too?

Sure, adding the check. Thanks @tpantelis

tpantelis (approved):

test/e2e/framework/dataplane.go
```diff
@@ -241,3 +241,22 @@ func GetGlobalnetEgressParams(egressIP GlobalEgressIPType) GlobalnetTestParams {
 		GlobalEgressIP:   egressIP,
 	}
 }
+
+func CanExecuteNonGatewayConnectivityTest(sourceNode, destNode framework.NetworkPodScheduling,
```
skitt, line 245: I’m somewhat surprised by this — the function name says we’re determining whether we _can_ run a given test, yet when this function returns `true` we _skip_ tests. Should the booleans be reversed?

skitt (changes requested):

test/e2e/framework/dataplane.go
sridhargaddam: Good observation. Originally this API was IsSingleNodeCluster, so the boolean was apt for that. Let me update it. Thanks @skitt

skitt (approved):

sridhargaddam:
Unit test failure does not seem to be related to this PR

```
Summarizing 1 Failure:

[Fail] Pinger when the pinger is stopped [It] should no longer update the statistics 
/go/src/github.com/submariner-io/submariner/pkg/cableengine/healthchecker/pinger_test.go:110
```

tpantelis:
> Unit test failure does not seem to be related to this PR
> 

Right - likely a timing issue with the test.  I'll take a peek at it.

PR #1576, opened 2021-10-26
Enable desired linters and document disabled ones

tpantelis:
See individual commits for details. A separate commit was created for each linter that yielded errors.


main.go
```diff
@@ -290,7 +292,7 @@ func startHTTPServer() *http.Server {
 	http.Handle("/metrics", promhttp.Handler())
 
 	go func() {
-		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
+		if err := srv.ListenAndServe(); errors.Is(err, http.ErrServerClosed) {
```
skitt: This is inverting the conditional.

```suggestion
		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
```

pkg/globalnet/main.go
```diff
@@ -140,7 +141,7 @@ func startHTTPServer() *http.Server {
 	http.Handle("/metrics", promhttp.Handler())
 
 	go func() {
-		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
+		if err := srv.ListenAndServe(); errors.Is(err, http.ErrServerClosed) {
```
skitt:
```suggestion
		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
```

pkg/natdiscovery/natdiscovery.go
```diff
@@ -148,7 +148,7 @@ func (nd *natDiscovery) AddEndpoint(endpoint *v1.Endpoint) {
 
 	// support nat discovery disabled or a remote cluster endpoint which still hasn't implemented this protocol
 	if _, err := extractNATDiscoveryPort(&endpoint.Spec); err != nil || nd.serverPort == 0 {
-		if err != errorNoNATDiscoveryPort {
+		if errors.Is(err, errorNoNATDiscoveryPort) {
```
skitt:
```suggestion
		if !errors.Is(err, errorNoNATDiscoveryPort) {
```

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
```diff
@@ -20,12 +20,12 @@ package kubeproxy
 
 import (
 	"errors"
-	"fmt"
 	"net"
 	"strconv"
 	"strings"
 	"syscall"
 
+	extErrors "github.com/pkg/errors"
```
skitt: I think it would be clearer long term if we come up with a consistent set of package shorthands: _e.g._ `errors` for Golang `errors`, `exterrors` for the GH package, and `apierrors` for the K8s package.

pkg/routeagent_driver/handlers/ovn/bridge.go
```diff
@@ -81,7 +81,7 @@ func (ovn *Handler) configureSubmarinerInterface() error {
 		Mask: ipNet.Mask,
 	}}
 
-	if err := ovn.netlink.AddrAdd(ovnLink, ipConfig); err != nil && err != syscall.EEXIST {
+	if err := ovn.netlink.AddrAdd(ovnLink, ipConfig); err != nil && errors.Is(err, syscall.EEXIST) {
```
skitt:
```suggestion
	if err := ovn.netlink.AddrAdd(ovnLink, ipConfig); err != nil && !errors.Is(err, syscall.EEXIST) {
```

skitt (changes requested):

main.go
tpantelis: Oh right - thanks for noticing that.

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
tpantelis: I use `apierrors` (or `apiErrors`) anyway. Re: the other two, we've mostly used the GH package so `errors` made sense. But with the newer APIs (eg `Is`, `As` etc) in the Golang package, we may be using that more. However I think GH will still be used much more so perhaps we should keep using `errors` for that and use `goerrors` (or `goErrors`) for the other.

Either way I think that should be a separate PR.

tpantelis:
@skitt Thanks for the first review. This is still in draft b/c I still have more linters to enable. I thought it was easier to put them all in one PR rather than handling conflicts and rebases across many PRs.

skitt:
> @skitt Thanks for the first review. This is still in draft b/c I still have more linters to enable. I thought it was easier to put them all in one PR rather than handling conflicts and rebases across many PRs.

You’re welcome, I agree it makes sense to enable everything in one PR! I’ll update my performance PR...

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
skitt: Yeah, I was listing `apierrors` for completeness. `errors` for the GH package makes sense, with `goerrors` for the Golang package. Let’s make it a separate PR, yes.

.golangci.yml
```diff
@@ -10,81 +10,129 @@ linters-settings:
...
       - unnamedResult
   gocyclo:
-    min-complexity: 20
+    min-complexity: 15
   govet:
     enable:
       - fieldalignment
   lll:
     line-length: 140
+  nlreturn:
```
Jaanki: Not needed since we are not enabling nlreturn

Jaanki (changes requested):

.golangci.yml
tpantelis: I left it just in case we do but might as well remove it.

skitt (approved):

Jaanki (approved):

PR #1597, opened 2021-11-23
fix error return

jinglina:
fix error return format


skitt:
Could you clarify what this fixes? [`errors.Wrapf()`](https://pkg.go.dev/github.com/pkg/errors#Wrapf) returns `nil` if `err` is `nil`, which is what you’re doing manually here.

tpantelis:
> Could you clarify what this fixes? [`errors.Wrapf()`](https://pkg.go.dev/github.com/pkg/errors#Wrapf) returns `nil` if `err` is `nil`, which is what you’re doing manually here.

`errors.WithMessagef` also behaves that way. In fact, `Wrapf` just wraps `withMessage` under hood to add the stack trace. So this proposed change doesn't really change any behavior other than to lose the stack trace, which isn't necessary.

jinglina:
> > Could you clarify what this fixes? [`errors.Wrapf()`](https://pkg.go.dev/github.com/pkg/errors#Wrapf) returns `nil` if `err` is `nil`, which is what you’re doing manually here.
> 
> `errors.WithMessagef` also behaves that way. In fact, `Wrapf` just wraps `withMessage` under hood to add the stack trace. So this proposed change doesn't really change any behavior other than to lose the stack trace, which isn't necessary.

yes. you are right. I just feel confused when errors return without any judgement about `nil`

PR #1606, opened 2021-12-02
Formalize local build settings

elevran:
This PR formalizes the settings needed for a "local build", outside of a Dapper container.

A developer should run (for example) `LOCAL_BUILD=1 make images` to build all images on their machine.
If `LOCAL_BUILD` is defined, the Makefile sets the needed variables to defaults under the following assumptions:
- Shipyard location is `../shipyard` (this would be the typical case if both repos are cloned under the same (e.g., `submariner-io`) directory); and
- `DAPPER_HOST_ARCH` is derived from `go env GOHOSTARCH`

All values are picked from outside the Makefile, if already defined

I've also added an (unrelated) generic target to print make variables - it was useful in ensuring correct values are being set in the Makefile or picked up from the external environment.

Signed-off-by: Etai Lev Ran <elevran@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,17 @@ BASE_BRANCH ?= devel
 PROTOC_VERSION=3.17.3
 export BASE_BRANCH
 
+# Define LOCAL_BUILD to build directly on the host and not inside a Dapper container
+ifdef LOCAL_BUILD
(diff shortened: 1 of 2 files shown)


elevran:
will resubmit without offending commit message

PR #1607, opened 2021-12-02
Formalize local build settings

elevran:
This PR formalizes the settings needed for a "local build", outside of a Dapper container.

A developer should run (for example) `LOCAL_BUILD=1 make images` to build all images on their machine.
If `LOCAL_BUILD` is defined, the Makefile sets the needed variables to defaults under the following assumptions:
- Shipyard location is `../shipyard` (this would be the typical case if both repos are cloned under the same (e.g., `submariner-io`) directory); and
- `DAPPER_HOST_ARCH` is derived from `go env GOHOSTARCH`

All values are picked from outside the Makefile, if already defined.

Signed-off-by: Etai Lev Ran <elevran@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,3 +140,5 @@ endif
 
 # Disable rebuilding Makefile
 Makefile Makefile.inc: ;
+
+print-%: ; @echo $* = $($*) 
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ export BASE_BRANCH
 
 # Define LOCAL_BUILD to build directly on the host and not inside a Dapper container
 ifdef LOCAL_BUILD
-DAPPER_HOST_ARCH ?= $(shell go env GOHOSTARCH | cut -d= -f2 | tr -d '"')
+DAPPER_HOST_ARCH ?= $(shell go env GOHOSTARCH)
 SHIPYARD_DIR ?= ../shipyard
 SCRIPTS_DIR ?= $(SHIPYARD_DIR)/scripts/shared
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,6 +3,17 @@ BASE_BRANCH ?= devel
 PROTOC_VERSION=3.17.3
 export BASE_BRANCH
 
+# Define LOCAL_BUILD to build directly on the host and not inside a Dapper container
+ifdef LOCAL_BUILD
+DAPPER_HOST_ARCH ?= $(shell go env GOHOSTARCH | cut -d= -f2 | tr -d '"')
```
skitt:
```suggestion
DAPPER_HOST_ARCH ?= $(shell go env GOHOSTARCH)
```

`go env GOHOSTARCH` only outputs that specific value, not `GOHOSTARCH="foo"` (as with `go env`).

skitt (changes requested):
Thanks, I have a minor tweak to the `DAPPER_HOST_ARCH` declaration.

The `print-%` target would be more generally useful in Shipyard’s `Makefile.inc`, could you submit it there?

Makefile
elevran: thanks for the catch on `go env` - amended
will submit the print-% target on shipyard

skitt (approved):

tpantelis (approved):

elevran:
@skitt unit tests are failing (though I'm not sure how they would be affected by this PR...)
Is there a way to tell the Bot (or GH actions) to rerun a specific failing task?

skitt:
> @skitt unit tests are failing (though I'm not sure how they would be affected by this PR...) Is there a way to tell the Bot (or GH actions) to rerun a specific failing task?

Project owners can restart them, unfortunately other users can’t as far as I’m aware.

PR #1613, opened 2021-12-07
Drop obsolete security replaces

skitt:
Our dependency tree pulls in the same or newer versions of /x/crypto
and /x/text, remove the replace directives.

This helps with go install (see
<https://github.com/golang/go/issues/44840>).


tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
Do we know what's going on with the failing unit tests? Seems like it's unrelated to this PR, as it's also happening on other recent PRs like https://github.com/submariner-io/submariner/pull/1614.

Issue #1634, opened 2021-12-22
Exporting multiple headless services with Globalnet does not work


kevinanderson1:
**What happened**: Exporting multiple services with globalnet does not allocate a globalnet ingress ip `giip` per exported service and instead only creates one and in some cases overwrites it and in others deletes it. This breaks connectivity when using globalnet and exporting more than one service per cluster. If globalnet is restarted you can see the globalnet controller iterating through the exported services and reusing the same globalnet ingress ip and deleting objects without allocating an ip.

**What you expected to happen**: I expect each exported service to get a unique globalnet ingress ip which is published to the broker and synced to other clusters.

**How to reproduce it (as minimally and precisely as possible)**: 
1. Create kind-dup.yaml
```yaml
# kind-dup.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  apiServerAddress: "192.168.100.244" # Update this to your local IP for you system. Had to do this to work around some other issues
  podSubnet: "10.42.0.0/16"
  serviceSubnet: "10.43.0.0/16"
nodes:
- role: control-plane
- role: worker
```
2. `kind create cluster --name cluster-dup-1 --config ./kind-dup.yaml `
3. `kind create cluster --name cluster-dup-2 --config ./kind-dup.yaml `
4. `kubectl config set-context cluster-dup-1`
5. `subctl deploy-broker --globalnet --globalnet-cidr-range 10.24.0.0/16 --globalnet-cluster-size 2048`
6. `subctl join broker-info.subm `
7. `kubectl config set-context cluster-dup-2`
8. `subctl join broker-info.subm`
9.  Deploy multiple services. I used this sample YAML (stateful-set.yaml)
`kubectl apply -f stateful-set.yaml`
```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-r0
  labels:
    app: nginx-r0
spec:
  ports:
  - port: 80
    name: web
  clusterIP: None
  selector:
    app: nginx-r0
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: nginx-r0
spec:
  selector:
    matchLabels:
      app: nginx-r0 # has to match .spec.template.metadata.labels
  serviceName: "nginx-r0"
  replicas: 1 # by default is 1
  template:
    metadata:
      labels:
        app: nginx-r0 # has to match .spec.selector.matchLabels
    spec:
      terminationGracePeriodSeconds: 10
      containers:
      - name: nginx
[...]

PR #1640, opened 2021-12-24
Globalnet fix for issues with multiple headless svs

sridhargaddam:
The reconciler method used in startIngressPodController was listing all the
globalIngressIP objects in the cluster and was deleting the previous ones.
Because of this, at any point of time, only one globalIngressIP object was
existing for an Headless Service that is exported. This PR updates the
reconciler to use labelSelector such that it only tries to list the pods
that match the associated label of the headless svc.

Fixes issue: https://github.com/submariner-io/submariner/issues/1634
(cherry picked from commit 79a0f2f27aa2d31aa8de059e4537048c473c6257)

### DIFF
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -66,9 +66,10 @@ func (c *baseSyncerController) Start() error {
 	return c.resourceSyncer.Start(c.stopCh)
 }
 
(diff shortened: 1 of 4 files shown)


skitt (approved):

nyechiel (approved):
Thanks for the quick resolution @sridhargaddam

Issue #1166, opened 2021-03-01
Avoid dependency on the iptable chains programmed by IPtables kube-proxy driver.


sridhargaddam:
**What would you like to be added**:
Currently, Globalnet code uses the kube-proxy iptable chains for redirecting the incoming traffic destined to the globalIP of the Service. While this has its own advantages, the drawback is that when an exported service is deleted, until globalnet deletes the reference to the chain, kube-proxy will not be able to delete the corresponding chain.
This sometimes can cause kube-proxy to log an error and retry the operation after some specified interval.

Instead of depending on the kube-proxy IPTable chains for Services, this enhancement request is to modify Globalnet to create its own load-balancing chains for exported services.

tpantelis:
bump

mkimuram:
@sridhargaddam 

As you commented in the meeting, chain won't be recreated, if no [endpoint](https://github.com/kubernetes/kubernetes/blob/0a865b6e6e4a814c52313a2c24978843a756c400/pkg/proxy/iptables/proxier.go#L1048) exists. So, it won't be enough just to block deletion of service to avoid chain deletion until deleting globalnet's referencing chain.

sridhargaddam:
> As you commented in the meeting, chain won't be recreated, if no [endpoint](https://github.com/kubernetes/kubernetes/blob/0a865b6e6e4a814c52313a2c24978843a756c400/pkg/proxy/iptables/proxier.go#L1048) exists. So, it won't be enough just to block deletion of service to avoid chain deletion until deleting globalnet's referencing chain.

Right @mkimuram. We will have to look for a better solution to this requirement. Thanks for looking into it in parallel.

mkimuram:
To create submariner's own chain, there will be two approaches:
1) Copy the kube-proxy chain (Like [this](https://github.com/mkimuram/submariner/tree/issue/1166v1))
2) Implement the same logic as kube-proxy (Like [this](https://github.com/mkimuram/submariner/tree/issue/1166v2))

mkimuram:
> 1. Copy the kube-proxy chain (Like this)
> 2. Implement the same logic as kube-proxy (Like this)

We are currently thinking about taking 2nd approach.

[...]

PR #1643, opened 2021-12-30
Enhance GN to support SVCs without KP dependency

sridhargaddam:
In the current implementation, Globalnet controller uses the
kubeproxy iptable chains for supporting globalIngressIPs (i.e.,
for services). While this works fine, its prone to issues and is
causing race conditions when the exported service is deleted.

This PR modifies the Globalnet implementation to use internal
services for every exported service. The internal service created
by Globalnet controller is configured with the assigned globalIP
as an external-ip. Once the internal service is created, kubeproxy
running in the cluster will treat such service as a regular K8s
service with external-ip and programs the necessary rules to
support connectivity to the external-ip (aka globalIP).

This PR also includes the associated unit tests to validate
various scenarios which could potentially arise while using this
feature.

Depends on: submariner-io/submariner-operator#1732
Depends on: submariner-io/submariner-charts#194
Fixes issue: #1166

### DIFF
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -237,12 +237,7 @@ func createService(svc *corev1.Service,
 		err = nil
 	}
 
-	if err != nil {
-		klog.Errorf("Error creating the internal Service %s/%s: %v", svc.Namespace, svc.Name, err)
-		return nil, err
-	}
-
-	return obj, nil
+	return obj, errors.Wrapf(err, "error creating the internal Service %s/%s", svc.Namespace, svc.Name)
 }
 
 func deleteService(namespace, name string,
@@ -250,14 +245,10 @@ func deleteService(namespace, name string,
 	err := client.Namespace(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
 	if apierrors.IsNotFound(err) {
 		klog.Warningf("Could not find internal Service %s/%s: %v", namespace, name, err)
-		err = nil
+		return nil
 	}
 
-	if err != nil {
-		klog.Errorf("Error deleting the internal Service %s/%s: %v", namespace, name, err)
-	}
-
-	return err
+	return errors.Wrapf(err, "error deleting the internal Service %s/%s", namespace, name)
 }
 
 func GetInternalSvcName(name string) string {
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -20,14 +20,21 @@ package controllers
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/base32"
 	"fmt"
+	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/federate"
+	resourceUtil "github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/util"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	iptiface "github.com/submariner-io/submariner/pkg/globalnet/controllers/iptables"
 	"github.com/submariner-io/submariner/pkg/ipam"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -195,4 +202,68 @@ func checkStatusChanged(oldStatus, newStatus interface{}, retObj runtime.Object)
 	klog.Infof("Updated: %#v", newStatus)
 
 	return retObj
+}
+
+func getService(name, namespace string,
+	client dynamic.NamespaceableResourceInterface, scheme *runtime.Scheme) (*corev1.Service, bool, error) {
+	obj, err := client.Namespace(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil, false, nil
+	}
+
+	if err != nil {
+		return nil, false, errors.Wrapf(err, "error retrieving Service %s/%s", namespace, name)
+	}
+
+	service := &corev1.Service{}
+	err = scheme.Convert(obj, service, nil)
+
+	if err != nil {
+		return nil, false, errors.Wrapf(err, "error converting %#v to Service", obj)
+	}
+
+	return service, true, nil
+}
+
+func createService(svc *corev1.Service,
+	client dynamic.NamespaceableResourceInterface) (*unstructured.Unstructured, error) {
+	gnService, err := resourceUtil.ToUnstructured(svc)
+	if err != nil {
+		return nil, err // nolint:wrapcheck  // Let the caller wrap it
+	}
+
+	obj, err := client.Namespace(gnService.GetNamespace()).Create(context.TODO(), gnService, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		err = nil
+	}
+
+	if err != nil {
+		klog.Errorf("Error creating the internal Service %s/%s: %v", svc.Namespace, svc.Name, err)
+		return nil, err
+	}
+
+	return obj, nil
+}
+
+func deleteService(namespace, name string,
+	client dynamic.NamespaceableResourceInterface) error {
+	err := client.Namespace(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if apierrors.IsNotFound(err) {
+		klog.Warningf("Could not find internal Service %s/%s: %v", namespace, name, err)
+		err = nil
+	}
+
+	if err != nil {
+		klog.Errorf("Error deleting the internal Service %s/%s: %v", namespace, name, err)
+	}
+
+	return err
+}
+
+func GetInternalSvcName(name string) string {
+	hash := sha256.Sum256([]byte(name))
+	encoded := base32.StdEncoding.EncodeToString(hash[:])
+	svcName := globalnetInternalServicePrefix + encoded[:32]
+
+	return strings.ToLower(svcName)
 }
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -254,7 +254,7 @@ func deleteService(namespace, name string,
 func GetInternalSvcName(name string) string {
 	hash := sha256.Sum256([]byte(name))
 	encoded := base32.StdEncoding.EncodeToString(hash[:])
-	svcName := globalnetInternalServicePrefix + encoded[:32]
+	svcName := InternalServicePrefix + encoded[:32]
 
 	return strings.ToLower(svcName)
 }
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -244,11 +244,11 @@ func deleteService(namespace, name string,
 	client dynamic.NamespaceableResourceInterface) error {
 	err := client.Namespace(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
 	if apierrors.IsNotFound(err) {
-		klog.Warningf("Could not find internal Service %s/%s: %v", namespace, name, err)
+		klog.Warningf("Could not find Service %s/%s to delete", namespace, name)
 		return nil
 	}
 
-	return errors.Wrapf(err, "error deleting the internal Service %s/%s", namespace, name)
+	return errors.Wrapf(err, "error deleting Service %s/%s", namespace, name)
 }
 
 func GetInternalSvcName(name string) string {
--- a/pkg/globalnet/controllers/global_ingressip_controller.go
+++ b/pkg/globalnet/controllers/global_ingressip_controller.go
@@ -188,9 +188,6 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
 					InternalServiceLabel: service.Name,
 				},
 				Finalizers: []string{InternalServiceFinalizer},
-				Annotations: map[string]string{
-					GlobalIngressIP: ips[0],
-				},
 			},
 		}
 
@@ -283,13 +280,13 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
 			if err = finalizer.Remove(context.TODO(), resource.ForDynamic(c.services.Namespace(ingressIP.Namespace)), intSvc,
 				InternalServiceFinalizer); err != nil {
 				klog.Errorf("Error while removing the finalizer from service %q: %v", key, err)
-				return shouldRequeue(numRequeues)
+				return true
 			}
 
 			err = deleteService(ingressIP.Namespace, intSvcName, c.services)
 			if err != nil {
 				klog.Errorf("Error while deleting the internal %q: %v", key, err)
-				return shouldRequeue(numRequeues)
+				return true
 			}
 		}
 
@@ -326,7 +323,7 @@ func (c *globalIngressIPController) ensureInternalServiceExists(ingressIP *subma
 		return fmt.Errorf("internal service created by Globalnet controller %q does not exist", key)
 	}
 
-	if service.Spec.ExternalIPs[0] != ingressIP.Status.AllocatedIP {
+	if len(service.Spec.ExternalIPs) == 0 || service.Spec.ExternalIPs[0] != ingressIP.Status.AllocatedIP {
 		// A user is ideally not supposed to modify the external-ip of the Globalnet internal service, but
 		// in-case its done accidentally, as part of controller start/re-start scenario, this code will fix
 		// the issue by deleting and re-creating the internal service with valid configuration.
--- a/pkg/globalnet/controllers/global_ingressip_controller.go
+++ b/pkg/globalnet/controllers/global_ingressip_controller.go
@@ -24,6 +24,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/finalizer"
+	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/syncer"
 	"github.com/submariner-io/admiral/pkg/util"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
@@ -32,6 +34,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 )
@@ -183,7 +186,11 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
 				Name:      GetInternalSvcName(service.Name),
 				Namespace: service.Namespace,
 				Labels: map[string]string{
-					ServiceRefLabel: service.Name,
+					InternalServiceLabel: service.Name,
+				},
+				Finalizers: []string{InternalServiceFinalizer},
+				Annotations: map[string]string{
+					GlobalIngressIP: ips[0],
 				},
 			},
 		}
@@ -264,17 +271,30 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
 	key, _ := cache.MetaNamespaceKeyFunc(ingressIP)
 
 	if ingressIP.Spec.Target == submarinerv1.ClusterIPService {
-		intSvc := GetInternalSvcName(ingressIP.Spec.ServiceRef.Name)
-		klog.Infof("Deleting the internal service %q created by Globalnet controller", intSvc)
+		intSvcName := GetInternalSvcName(ingressIP.Spec.ServiceRef.Name)
+		klog.Infof("Deleting the internal service %q created by Globalnet controller", intSvcName)
 
-		err := deleteService(ingressIP.Namespace, intSvc, c.services)
+		intSvc, exists, err := getService(intSvcName, ingressIP.Namespace, c.services, c.scheme)
 		if err != nil {
-			if shouldRequeue(numRequeues) {
-				return true
+			klog.Errorf("Error retrieving the internal service created by Globalnet controller %q: %v", key, err)
+			return shouldRequeue(numRequeues)
+		}
+
+		if exists {
+			if err = finalizer.Remove(context.TODO(), ForService(c.services, ingressIP.Namespace), intSvc,
+				InternalServiceFinalizer); err != nil {
+				klog.Errorf("Error while removing the finalizer from service %q: %v", key, err)
+				return shouldRequeue(numRequeues)
+			}
+
+			err = deleteService(ingressIP.Namespace, intSvcName, c.services)
+			if err != nil {
+				klog.Errorf("Error while deleting the internal %q: %v", key, err)
+				return shouldRequeue(numRequeues)
 			}
 		}
 
-		if err := c.pool.Release(ingressIP.Status.AllocatedIP); err != nil {
+		if err = c.pool.Release(ingressIP.Status.AllocatedIP); err != nil {
 			klog.Errorf("Error while releasing the global IPs for %q: %v", key, err)
 		}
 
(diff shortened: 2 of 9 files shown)


pkg/globalnet/controllers/base_controllers.go
```diff
@@ -196,3 +203,67 @@ func checkStatusChanged(oldStatus, newStatus interface{}, retObj runtime.Object)
...
+	if apierrors.IsNotFound(err) {
+		klog.Warningf("Could not find internal Service %s/%s: %v", namespace, name, err)
+		err = nil
+	}
+
+	if err != nil {
+		klog.Errorf("Error deleting the internal Service %s/%s: %v", namespace, name, err)
+	}
+
+	return err
```
tpantelis:
```suggestion
	if apierrors.IsNotFound(err) {
		klog.Warningf("Could not find internal Service %s/%s: %v", namespace, name, err)
		return nil
	}

	return errors.Wrapf(err, "error deleting the internal Service %s/%s: %v", namespace, name)
```

pkg/globalnet/controllers/base_controllers.go
```diff
@@ -196,3 +203,67 @@ func checkStatusChanged(oldStatus, newStatus interface{}, retObj runtime.Object)
...
+	if apierrors.IsAlreadyExists(err) {
+		err = nil
+	}
+
+	if err != nil {
+		klog.Errorf("Error creating the internal Service %s/%s: %v", svc.Namespace, svc.Name, err)
+		return nil, err
+	}
+
+	return obj, nil
```
tpantelis:
```suggestion
	
	return obj, errors.Wrapf(err, "error creating the internal Service %s/%s: %v", svc.Namespace, svc.Name)
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -165,29 +162,47 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
...
 	if ingressIP.Spec.Target == submarinerv1.ClusterIPService {
-		chainName := ingressIP.GetAnnotations()[kubeProxyIPTableChainAnnotation]
-		if chainName == "" {
+		serviceRef := ingressIP.Spec.ServiceRef
+
+		service, exists, err := getService(serviceRef.Name, ingressIP.Namespace, c.services, c.scheme)
+		if err != nil || !exists {
 			_ = c.pool.Release(ips...)
+			key := fmt.Sprintf("%s/%s", ingressIP.Namespace, serviceRef.Name)
+			klog.Warningf("Exported Service %q does not exist yet - re-queueing", key)
```
tpantelis: We should log the error.

```suggestion
                         if err != nil {
                             klog.Errorf("Error retrieving exported Service %q - re-queueing", key)
                         } else {
		            klog.Warningf("Exported Service %q does not exist yet - re-queueing", key)
		        }
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -165,29 +162,47 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
...
+			key := fmt.Sprintf("%s/%s", internalService.Namespace, internalService.Name)
+			klog.Warningf("Failed to create the internal Service %q ", key)
 
+			_ = c.pool.Release(ips...)
 			ingressIP.Status.Conditions = util.TryAppendCondition(ingressIP.Status.Conditions, &metav1.Condition{
 				Type:    string(submarinerv1.GlobalEgressIPAllocated),
 				Status:  metav1.ConditionFalse,
-				Reason:  "ProgramIPTableRulesFailed",
-				Message: fmt.Sprintf("Error programming ingress rules: %v", err),
+				Reason:  "FailedToCreateInternalGlobalnetService",
```
tpantelis: We can shorten this a bit.

```suggestion
				Reason:  "InternalServiceCreationFailed",
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -165,29 +162,47 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
...
+		internalService.Spec.Selector = service.Spec.Selector
+		internalService.Spec.ExternalIPs = extIPs
 
+		_, err = createService(internalService, c.services)
+		if err != nil {
 			_ = c.pool.Release(ips...)
+			key := fmt.Sprintf("%s/%s", internalService.Namespace, internalService.Name)
+			klog.Warningf("Failed to create the internal Service %q ", key)
 
+			_ = c.pool.Release(ips...)
```
tpantelis: This is a duplicate of L193.

```suggestion
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -165,29 +162,47 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
...
+		extIPs := []string{ips[0]}
+		internalService.Spec.Ports = service.Spec.Ports
+		internalService.Spec.Selector = service.Spec.Selector
+		internalService.Spec.ExternalIPs = extIPs
 
+		_, err = createService(internalService, c.services)
+		if err != nil {
 			_ = c.pool.Release(ips...)
+			key := fmt.Sprintf("%s/%s", internalService.Namespace, internalService.Name)
+			klog.Warningf("Failed to create the internal Service %q ", key)
```
tpantelis:
```suggestion
			klog.Errorf("Failed to create the internal Service %q ", key)
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -264,3 +292,22 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
...
+
+	service, exists, err := getService(internalSvc, ingressIP.Namespace, c.services, c.scheme)
+	if err != nil || !exists {
+		return fmt.Errorf("internal service created by Globalnet controller %q does not exist", key)
+	}
+
+	if service.Spec.ExternalIPs[0] != ingressIP.Status.AllocatedIP {
+		_ = deleteService(ingressIP.Namespace, internalSvc, c.services)
+
+		return fmt.Errorf("globalIP assigned to %q does not match with Internal Service ExternalIP", key)
```
tpantelis, line 337: How would this scenario occur?

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -70,22 +77,12 @@ func NewGlobalIngressIPController(config *syncer.ResourceSyncerConfig, pool *ipa
...
-
-		if target == "" {
-			continue
-		}
-
 		// nolint:wrapcheck  // No need to wrap these errors.
 		err = controller.reserveAllocatedIPs(federator, obj, func(reservedIPs []string) error {
 			if gip.Spec.Target == submarinerv1.ClusterIPService {
-				return controller.iptIface.AddIngressRulesForService(reservedIPs[0], target)
+				return controller.ensureInternalServiceExists(gip)
```
tpantelis, line 85: Is this necessary on startup? We only save the allocated IP in the `GlobalIngressIP` if the internal service was successfully created so we **shouldn't** run into the case where the expected internal service is missing on startup. Also if the service was missing then we'd return an error and abort startup with no recovery. If the concern is an unexpected out-of-band delete of the service (eg user "shooting themselves in the foot") then I think we should handle that in the control loop, if at all.

tpantelis (changes requested):

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: > Is this necessary on startup? We only save the allocated IP in the `GlobalIngressIP` if the internal service was successfully created so we **shouldn't** run into the case where the expected internal service is missing on startup. 

In an ideal scenario, the internal service will be present, but if the user accidentally (or intentionally) deletes the submariner-internalService, we will be clearing the AllocatedGlobalIP from the GlobalIngressIP and update its status accordingly.

> Also if the service was missing then we'd return an error and abort startup with no recovery. 

I don't think we will abort the startup. We will clear the AllocatedGlobalIP and update the status of giip - https://github.com/submariner-io/submariner/blob/62927d92c2d4c387bd6c2fe6b62ca7a6acb136d7/pkg/globalnet/controllers/base_controllers.go#L116

> If the concern is an unexpected out-of-band delete of the service (eg user "shooting themselves in the foot") then I think we should handle that in the control loop, if at all.

Yes, if the user deletes the submariner-internalService, we are handling that in the service_controller.go -> onDelete method.

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: Good catch @tpantelis

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: In an ideal scenario this will not happen, but if user intentionally (or accidentally) modifies the external-ip (aka globalIP) of the submariner-internalService we delete that service and re-create it with correct ip.

sridhargaddam (commented):
Thank you @tpantelis for your review comments.

sridhargaddam:
Please remember to squash the commits at the time of merge.

pkg/globalnet/controllers/global_ingressip_controller.go
tpantelis: I guess it doesn't hurt to check here but this would only run on a controller restart which should be very infrequent. Ideally we would handle such erroneous changes in the control loop but I'm not sure it's really worth trying to handle such cases - I'm sure there's other potential erroneous scenarios elsewhere in globalnet et al which we don't handle.

pkg/globalnet/controllers/global_ingressip_controller.go
tpantelis: > I don't think we will abort the startup. We will clear the AllocatedGlobalIP and update the status of giip

Yes you're right, I missed that. Actually I wrote it a while back - it's hard to remember every detail 😄

pkg/globalnet/controllers/service_controller.go
```diff
@@ -119,6 +119,11 @@ func (c *serviceController) onDelete(service *corev1.Service) (runtime.Object, b
 		return nil, false
 	}
 
+	origService := service.GetLabels()[ServiceRefLabel]
+	if origService != "" {
+		service.Name = origService
+	}
```
tpantelis: So for an internal service delete this will delete the associated `GlobalIngressIP`. This doesn't seem ideal. First, the `GlobalIngressIP`"owns" the internal service and controls its lifecycle not the other way around. Second, it differs from the startup behavior in the `globalIngressIPController` where we simply clear the `GlobalIngressIP`'s allocated global IP if the internal service doesn't exist. It seems we should do the same here however, again, I really don't think it's worth trying to handle such unexpected, erroneous out-of-band edge cases.

tpantelis (changes requested):

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: I see your POV @tpantelis, thanks. Agree that these are not common scenarios and will only happen if user does something crazy. By handling these scenarios (whenever possible) we are going to save our time while troubleshooting issues particularly when people don't mention the steps they performed (before they land into issues) and expect support to resolve the situation.

I'd like to get some feedback from others on this topic - @nyechiel @skitt @vthapar @aswinsuryan @mkolesnik @mkimuram 

Short summary: Globalnet controller now creates an internal Service for every exported Service. The internal service that is created by Globalnet Controller will be configured with an external-ip which will be nothing but the globalIP assigned to the exported service. Ideally the internal service is owned by Globalnet and users should NOT modify/delete it. But if users modify the external-ip (in some platforms like OCP, this is not possible) or delete the service (possible), should Globalnet controller handle such scenarios by trying to keep the associated GlobalIngressIP in Sync or can we treat this as an configuration issue from the users (with end result as undefined)?

sridhargaddam (commented):
Thanks again @tpantelis for your review comments.

pkg/globalnet/controllers/global_ingressip_controller.go
tpantelis: > Ideally the internal service is owned by Globalnet and users should NOT modify/delete it. But if users modify the external-ip (in some platforms like OCP, this is not possible) or delete the service (possible), should Globalnet controller handle such scenarios by trying to keep the associated GlobalIngressIP in Sync or can we treat this as an configuration issue from the users (with end result as undefined)?

I understand your concerns and trying to mitigate user mistakes. I think at the very least, where reasonable, we could detect such cases and log error messages. Eg, if an internal service is deleted but the associated `GlobalIngressIP` still exists then log a message. Similarly if the `ExternalIP` of an internal service is modified, log it. Also, `subclt diagnose` could detect these inconsistencies.

mkimuram: I think that @sridhargaddam once shared an idea to create internal Services for exported Services in submariner namespace (maybe for other reasons). By doing so, we would be able to prevent users from accidentally deleting them. However, drawback will be that we need an extra step to sync the endpoints for the service to submariner namespace. Also, we will need to consider how to avoid name collision for the services.

It might be too much just to avoid accidental deletion, but it should address this issue.

tpantelis: > prevent users from accidentally deleting them

We can actually do this by adding a finalizer to the internal service. We can also make the `GlobalIngressIP` the owner so K8s automatically deletes it (after we remove the finalizer).

sridhargaddam: Thanks @tpantelis @mkimuram

We discussed this topic in today's external network meeting and @aswinsuryan also suggested the same approach. So, lets use the finalizer on the internal-service to stop users from deleting the internal services. Also, I'll think about configuring the owner for the internal service with `GlobalIngressIP` so that K8s will take care of it. 

With regards to users modifying the external-ip in the internal-services I'll log a message and will enhance subctl diagnose to detech the inconsistencies. Good points @tpantelis

tpantelis: 👍 For cleanup on uninstall, we'll have to remove the finalizers.

tpantelis: BTW - admiral has `finalizer.Add `and `finalizer.Remove ` functions for convenience.

pkg/globalnet/controllers/base_controllers.go
```diff
@@ -196,3 +203,58 @@ func checkStatusChanged(oldStatus, newStatus interface{}, retObj runtime.Object)
...
+	}
+
+	return obj, errors.Wrapf(err, "error creating the internal Service %s/%s", svc.Namespace, svc.Name)
+}
+
+func deleteService(namespace, name string,
+	client dynamic.NamespaceableResourceInterface) error {
+	err := client.Namespace(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if apierrors.IsNotFound(err) {
+		klog.Warningf("Could not find internal Service %s/%s: %v", namespace, name, err)
```
tpantelis:
```suggestion
		klog.Warningf("Could not find Service %s/%s to delete", namespace, name)
```

pkg/globalnet/controllers/base_controllers.go
```diff
@@ -196,3 +203,58 @@ func checkStatusChanged(oldStatus, newStatus interface{}, retObj runtime.Object)
...
+		return nil
+	}
+
+	return errors.Wrapf(err, "error deleting the internal Service %s/%s", namespace, name)
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error deleting Service %s/%s", namespace, name)
```

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -244,13 +270,39 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
...
+		if err != nil {
+			klog.Errorf("Error retrieving the internal service created by Globalnet controller %q: %v", key, err)
+			return shouldRequeue(numRequeues)
+		}
+
+		if exists {
+			if err = finalizer.Remove(context.TODO(), ForService(c.services, ingressIP.Namespace), intSvc,
+				InternalServiceFinalizer); err != nil {
+				klog.Errorf("Error while removing the finalizer from service %q: %v", key, err)
+				return shouldRequeue(numRequeues)
```
tpantelis: It seems we should try indefinitely and not limit the requeues here. I don't think we'd hit an error condition that isn't transient.

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -244,13 +270,39 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
...
 			}
-		} else if ingressIP.Spec.Target == submarinerv1.HeadlessServicePod {
+
+			err = deleteService(ingressIP.Namespace, intSvcName, c.services)
+			if err != nil {
+				klog.Errorf("Error while deleting the internal %q: %v", key, err)
+				return shouldRequeue(numRequeues)
```
tpantelis: Same here.

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -264,3 +316,59 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
 		return nil
 	}, ingressIP.Status.AllocatedIP)
 }
+
+func (c *globalIngressIPController) ensureInternalServiceExists(ingressIP *submarinerv1.GlobalIngressIP) error {
```
tpantelis, line 316: I thought we decided we don't need these checks/remediations in lieu of using a finalizer et al?

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -264,3 +316,59 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
...
+
+		_ = deleteService(ingressIP.Namespace, internalSvc, c.services)
+
+		return fmt.Errorf("globalIP assigned to %q does not match with Internal Service ExternalIP", key)
+	}
+
+	return nil
+}
+
+func ForService(client dynamic.NamespaceableResourceInterface, namespace string) resource.Interface {
```
tpantelis: Can't we use admiral's `resource.ForDynamic` in lieu this?

pkg/globalnet/controllers/global_ingressip_controller_test.go
```diff
@@ -82,32 +79,94 @@ var _ = Describe("GlobalIngressIP controller", func() {
...
+
+	It("should successfully create an internal submariner service", func() {
+		internalSvcName := controllers.GetInternalSvcName(serviceName)
+		intSvc := t.awaitService(internalSvcName)
+		externalIP := intSvc.Spec.ExternalIPs[0]
+		Expect(externalIP).ToNot(BeEmpty())
+		finalizer := intSvc.GetFinalizers()[0]
+		Expect(finalizer).To(Equal(controllers.InternalServiceFinalizer))
+		gIP := intSvc.GetAnnotations()[controllers.GlobalIngressIP]
+		Expect(gIP).To(Equal(externalIP))
```
tpantelis: To improve readability please use newlines between logical sections.

```suggestion
	        intSvc := t.awaitService(controllers.GetInternalSvcName(serviceName))
                externalIP := intSvc.Spec.ExternalIPs[0]
                Expect(externalIP).ToNot(BeEmpty())
		 
                 Expect(intSvc.GetFinalizers).To(Equal([]string{controllers.InternalServiceFinalizer}))

                 gIP := intSvc.GetAnnotations()[controllers.GlobalIngressIP]
                 Expect(gIP).To(Equal(externalIP))
```

pkg/globalnet/controllers/service_controller.go
```diff
@@ -105,6 +105,28 @@ func (c *serviceController) process(from runtime.Object, numRequeues int, op syn
...
+	// Log an error if the external-ip of the Globalnet internal service is modified.
+	origService := service.GetLabels()[InternalServiceLabel]
+	if origService != "" {
+		if service.Spec.ExternalIPs[0] != "" {
+			globalIPFromAnnotation := service.GetAnnotations()[GlobalIngressIP]
+			if globalIPFromAnnotation != service.Spec.ExternalIPs[0] {
+				klog.Errorf("ExternalIP %q of Globalnet internal service %q does not match with allocated globalIP %q",
+					service.Spec.ExternalIPs[0], key, globalIPFromAnnotation)
+			}
+		}
```
tpantelis: I assume the external IP shouldn't ever be empty, also we should check the array length as a user could have removed it entirely.

This validates the external IP against the global IP stored as an annotation. Ideally we should read the associated `GlobalIngressIP` resource and retrieve its allocated IP as it's possible the user could change the annotation as well, as nonsensical as that seems, or change the `GlobalIngressIP`'s allocated IP so they don't match.  However I really think we should forego the integrity checking here and do it in subctl diagnose.

```suggestion
		if len(service.Spec.ExternalIPs) == 0 {
		    klog.Errorf("The ExternalIP for Globalnet internal service %q is missing", key)
		} else {
                     globalIPFromAnnotation := service.GetAnnotations()[GlobalIngressIP]
                     if globalIPFromAnnotation != service.Spec.ExternalIPs[0] {
                         klog.Errorf("ExternalIP %q of Globalnet internal service %q does not match with allocated globalIP %q",
                         service.Spec.ExternalIPs[0], key, globalIPFromAnnotation)
                     }
		}
```

tpantelis (changes requested):

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: Currently the requeue count is set to 20. So, as you said, in all likelihood we may not hit this, but instead of indefinitely trying lets try for 20 mins.

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: same ;D

pkg/globalnet/controllers/global_ingressip_controller.go
sridhargaddam: Currently in most of the Submariner pods like route-agent, gateway and globalnet, if something is wrong at run-time, we try to fix it on pod restart. So, lets keep it to address such issues as explained in line 331 to 333

pkg/globalnet/controllers/service_controller.go
sridhargaddam: > I assume the external IP shouldn't ever be empty, also we should check the array length as a user could have removed it entirely.

Ack, done.

> This validates the external IP against the global IP stored as an annotation. Ideally we should read the associated `GlobalIngressIP` resource and retrieve its allocated IP as it's possible the user could change the annotation as well,

I thought about it, but it involves making an API call just to log an error for mismatch. So took this approach.

> as nonsensical as that seems, or change the `GlobalIngressIP`'s allocated IP so they don't match. However I really think we should forego the integrity checking here and do it in subctl diagnose.

As we are just logging the issue and not taking any action, please lets keep this for sometime to get feedback (we can remove it later). Also, I will add support in subctl diagnose in a future PR to flag these issues.

sridhargaddam (commented):

pkg/globalnet/controllers/service_controller.go
tpantelis: But it adds more complexity and untested code that needs to be maintained for little value except it can provide a timestamp if the external IP was changed. But we can do that easily by intercepting change updates and comparing the old and new values and logging if changed.  This code also requires maintaining a label just for this purpose which is unlikely to ever happen in the first place. Plus the check is incomplete b/c it doesn't handle unexpected out-of-band changes to the allocated IP.

I strongly think we should keep such integrity checks in subctl diagnose as we can do a more complete job there and doesn't require scraping logs to find a problem. I also think it's inconsistent to perform this one integrity check in code but not for other cases where integrity can be compromised via user mistake (eg what if the user randomly changes an allocated IP).

pkg/globalnet/controllers/global_ingressip_controller.go
tpantelis: It doesn't hurt to try indefinitely unless it's a non-transient issue that will never resolve,  like "not found" but we handle that.

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -264,3 +315,30 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
...
+	serviceRef := ingressIP.Spec.ServiceRef
+	internalSvc := GetInternalSvcName(serviceRef.Name)
+	key := fmt.Sprintf("%s/%s", ingressIP.Namespace, internalSvc)
+
+	service, exists, err := getService(internalSvc, ingressIP.Namespace, c.services, c.scheme)
+	if err != nil || !exists {
+		return fmt.Errorf("internal service created by Globalnet controller %q does not exist", key)
+	}
+
+	if service.Spec.ExternalIPs[0] != ingressIP.Status.AllocatedIP {
```
tpantelis: `ExternalIPs` could be nil here.

tpantelis (approved):

tpantelis:
The commits should be squashed before merging.

vthapar (approved):

Issue #1652, opened 2022-01-06
submariner should accept "Calico" (with capital C) as network plugin


dberg1:
My OpenShift cluster is configured to use Calico as CNI network plugin instead of the default OpenShiftSDN:
```
apiVersion: config.openshift.io/v1
kind: Network
metadata:
  name: cluster
spec:
  clusterNetwork:
    - cidr: 172.17.0.0/18
      hostPrefix: 23
  externalIP:
    policy: {}
  networkType: Calico
  serviceNetwork:
    - 172.21.0.0/16
```
unfortunately, submariner does not recognize "Calico" as a supported CNI network plugin because of the capital C:
```
# subctl diagnose all
...
 • Checking Submariner support for the CNI network plugin  ...
 ✗ Checking Submariner support for the CNI network plugin
 ✗ The detected CNI network plugin ("Calico") is not supported by Submariner. Supported network plugins: [generic canal-flannel weave-net OpenShiftSDN OVNKubernetes calico]
```

**What happened**:
`submariner-route-agent` does not create the required iptables rules because it does not recognize the "Calico" network plugin.

**What you expected to happen**:
submariner should support "Calico" as this is the suggested name to use in the calico documentation:
https://projectcalico.docs.tigera.io/getting-started/openshift/installation

> $ sed -i 's/OpenShiftSDN/Calico/' install-config.yaml

**How to reproduce it (as minimally and precisely as possible)**:
Create an OpenShift cluster using "Calico" as network plugin.

**Anything else we need to know?**:

**Environment**:

dberg1:
Changing pkg/routeagent_driver/constants/constants.go:
```
NetworkPluginCalico        = "calico"
```
to
```
NetworkPluginCalico        = "Calico"
```
seems to fix the issue for me (but you might want to accept both "calico" and "Calico").
At least the `submariner-route-agent` now creates the iptables rules properly.

sridhargaddam:
> seems to fix the issue for me (but you might want to accept both "calico" and "Calico"). At least the `submariner-route-agent` now creates the iptables rules properly.

Thanks for reporting the issue @dberg1. Can you please confirm that after you modified the name to "Calico" the connectivity between the clusters is working fine?

dberg1:
My setup is a bit special as I am trying to connect an OpenShift cluster using OpenShiftSDN to another OpenShift cluster using Calico.
[...]

PR #1666, opened 2022-01-19
Update Calico network plugin name

### LINKED ISSUE #1852: submariner should accept "Calico" (with capital C) as network plugin
My OpenShift cluster is configured to use Calico as CNI network plugin instead of the default OpenShiftSDN:
```
apiVersion: config.openshift.io/v1
kind: Network
metadata:
  name: cluster
spec:
  clusterNetwork:
    - cidr: 172.17.0.0/18
      hostPrefix: 23
  externalIP:
    policy: {}
  networkType: Calico
  serviceNetwork:
    - 172.21.0.0/16
```
unfortunately, submariner does not recognize "Calico" as a supported CNI network plugin because of the capital C:
```
# subctl diagnose all
...
 • Checking Submariner support for the CNI network plugin  ...
 ✗ Checking Submariner support for the CNI network plugin
 ✗ The detected CNI network plugin ("Calico") is not supported by Submariner. Supported network plugins: [generic canal-flannel weave-net OpenShiftSDN OVNKubernetes calico]
```

**What happened**:
`submariner-route-agent` does not create the required iptables rules because it does not recognize the "Calico" network plugin.

submariner should support "Calico" as this is the suggested name to use in the calico documentation:
https://projectcalico.docs.tigera.io/getting-started/openshift/installation

> $ sed -i 's/OpenShiftSDN/Calico/' install-config.yaml

Create an OpenShift cluster using "Calico" as network plugin.

**Environment**:

yboaron:
According to documentation [1], calico should be configured as
an OpenShift third-party network plugin named 'Calico'.

Fixes: submariner-io/submariner-operator#1852 
[1]
https://projectcalico.docs.tigera.io/getting-started/openshift/installation

### DIFF
--- a/pkg/event/registry.go
+++ b/pkg/event/registry.go
@@ -19,6 +19,8 @@ limitations under the License.
 package event
 
 import (
+	"strings"
+
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/stringset"
@@ -53,10 +55,10 @@ func (er *Registry) addHandler(eventHandler Handler) error {
 	evNetworkPlugins := stringset.New()
 
 	for _, np := range eventHandler.GetNetworkPlugins() {
-		evNetworkPlugins.Add(np)
+		evNetworkPlugins.Add(strings.ToLower(np))
 	}
 
-	if evNetworkPlugins.Contains(AnyNetworkPlugin) || evNetworkPlugins.Contains(er.networkPlugin) {
+	if evNetworkPlugins.Contains(AnyNetworkPlugin) || evNetworkPlugins.Contains(strings.ToLower(er.networkPlugin)) {
 		if err := eventHandler.Init(); err != nil {
 			return errors.Wrapf(err, "Event handler %q failed to initialize", eventHandler.GetName())
 		}


pkg/routeagent_driver/constants/constants.go
```diff
@@ -54,5 +54,5 @@ const (
 	NetworkPluginWeaveNet      = "weave-net"
 	NetworkPluginOpenShiftSDN  = "OpenShiftSDN"
 	NetworkPluginOVNKubernetes = "OVNKubernetes"
-	NetworkPluginCalico        = "calico"
+	NetworkPluginCalico        = "Calico"
```
skitt: In non-OpenShift setups, the plugin is named `calico`; we can’t just replace the name here.

skitt (changes requested):

pkg/routeagent_driver/constants/constants.go
yboaron: Thanks @skitt, 
Hmm, how bout adding a new entry to the constants list (something like OcpNetworkPluginCalico = "Calico" )?

sridhargaddam: Rather than adding a new constant, how about doing a case insensitive comparison at the place where its validated?

yboaron: Sounds good to me, I'll update the PR

dberg1: @yboaron If you decide not to change the `NetworkPluginCalico` constant and do a case insensitive comparison, make sure to also change the comparison in:
https://github.com/submariner-io/submariner-operator/blob/5eb19252606e888f444bcb79ccf78c3ec4331105/pkg/subctl/cmd/diagnose/cni.go#L73

yboaron: Thanks for the feedback @dberg1 ,
I'm new to Submariner, maybe I'm missing something here ...

So, the operator is running auto discovery process and updates 
submariner.Status.NetworkPlugin with clusterNetwork.NetworkPlugin [1]
and for Calico the clusterNetwork.NetworkPlugin is set to constant from submariner code [2]

It isn't clear to me how we can have a mismatch in plugin name in that case, 
could you please elaborate on how you hit the case described in [3] ?

[1]
https://github.com/submariner-io/submariner-operator/blob/devel/controllers/submariner/submariner_networkdiscovery.go#L66
[2]
https://github.com/submariner-io/submariner-operator/blob/devel/pkg/discovery/network/calico.go#L55
[3]
https://github.com/submariner-io/submariner/issues/1652

dberg1: @yboaron `discoverNetwork` returns with `err != nil` since Calico does not match any supported plugin:
https://github.com/submariner-io/submariner-operator/blob/2287e445e48bc1ebcc7f9648ec5f80eed3af9ea5/controllers/submariner/submariner_networkdiscovery.go#L51
so it returns here: 
https://github.com/submariner-io/submariner-operator/blob/2287e445e48bc1ebcc7f9648ec5f80eed3af9ea5/pkg/discovery/network/calico.go#L51

yboaron: Not sure that I fully understood your comment, and I don't see why the issue you described above is specific to calico/Calico.

From what I can see, 'discoverCalicoNetwork' function [1] checks for config map named 'calico-config'   [2] and then calls to 'discoverNetwork'  [3].
The discoverNetwork function [4] retrieves pod and service cidrs from kube-api, I can't see any code here that is specific to networking plugin.

Could it be that calico config-map name was != 'calico-config' in your case?
Could you please describe the exact steps you run to hit this case ( I can try to reproduce it in my env)?

[1] 
https://github.com/submariner-io/submariner-operator/blob/2287e445e48bc1ebcc7f9648ec5f80eed3af9ea5/pkg/discovery/network/calico.go#L30

[2]
https://github.com/submariner-io/submariner-operator/blob/2287e445e48bc1ebcc7f9648ec5f80eed3af9ea5/pkg/discovery/network/calico.go#L39

[3]
https://github.com/submariner-io/submariner-operator/blob/2287e445e48bc1ebcc7f9648ec5f80eed3af9ea5/pkg/discovery/network/calico.go#L49

[4] 
https://github.com/submariner-io/submariner-operator/blob/2287e445e48bc1ebcc7f9648ec5f80eed3af9ea5/pkg/discovery/network/generic.go#L51

tpantelis (approved):

pkg/routeagent_driver/constants/constants.go
yboaron: @dberg1  If you can still access your Ocp cluster with calico it will be useful if you can share the following:
- NetworkPlugin env var from Routeagent daemonset manifest [1]
- logs from one of submariner-routeagent pods  [2]

it may hint at the source of the problem


[1] kubectl get ds -n submariner-operator  submariner-routeagent  -o yaml  | grep -A 2 NETWORKPLU

[2] kubectl logs  -n submariner-operator submariner-routeagent-j8jf4

dberg1: https://github.com/submariner-io/submariner/issues/1652#issuecomment-1020248034

yboaron: Please check https://github.com/submariner-io/submariner/issues/1652#issuecomment-1020356088

yboaron: Following the discussion in https://github.com/submariner-io/submariner/issues/1652 , I think we should address this issue in the network-discovery code in the operator.

I'm going to close this PR and create a new PR for the operator repo

yboaron:
https://github.com/submariner-io/submariner-operator/pull/1802 should address Calico network plugin in OCP

PR #1676, opened 2022-01-24
Build images without Dapper

skitt:
We can achieve the same effect as running with Dapper, without Dapper,
by building images in multiple stages. The first stage runs in a
container built from our Shipyard base image, and subsequent stages
build our deliverable images.

This brings the following benefits:

* our container image builds are driven by our container descriptors,
  not by a combination of our Makefiles and container descriptors;
  this is more familiar to developers used to containers
* providing container descriptor-driven builds allows more third-party
  tools to be used directly, e.g. repository-hosted container image
  builds (on Quay or GitHub notably) and container descriptor analysis
  tools
* the risk of differences between "proper" container builds and local
  builds (with LOCAL_BUILD=1) is eliminated
* our build no longer depends on being able to build containers inside
  containers (which isn't supported everywhere)
* by building the binaries we want to ship in our container images
  during the container image build, we can ensure that the binaries
  are built for the architecture expected by the container image

The default target platform is linux/amd64 for the time being.

Depends on https://github.com/submariner-io/shipyard/pull/713

### DIFF
--- a/package/Dockerfile.submariner-gateway
+++ b/package/Dockerfile.submariner-gateway
@@ -1,4 +1,17 @@
+ARG BASE_BRANCH
+ARG SOURCE=/go/src/github.com/submariner-io/submariner
+ARG TARGETPLATFORM=linux/amd64
+
+FROM quay.io/submariner/shipyard-dapper-base:${BASE_BRANCH} AS builder
+ARG SOURCE
+ARG TARGETPLATFORM
+
+COPY . ${SOURCE}
+
+RUN make -C ${SOURCE} LOCAL_BUILD=1 bin/${TARGETPLATFORM}/submariner-gateway
+
 FROM fedora:34
+ARG SOURCE
 ARG TARGETPLATFORM
 
 WORKDIR /var/submariner
@@ -10,6 +23,6 @@ RUN dnf -y install --nodocs --setopt=install_weak_deps=0 \
            iproute libreswan kmod && \
     dnf -y clean all
 
-COPY package/submariner.sh package/pluto bin/${TARGETPLATFORM}/submariner-gateway /usr/local/bin/
+COPY --from=builder ${SOURCE}/package/submariner.sh ${SOURCE}/package/pluto ${SOURCE}/bin/${TARGETPLATFORM}/submariner-gateway /usr/local/bin/
 
 ENTRYPOINT submariner.sh
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,6 @@ else
 SETTINGS ?= $(DAPPER_SOURCE)/.shipyard.e2e.yml
 endif
 
-images: build
-
 include $(SHIPYARD_DIR)/Makefile.inc
 
 TARGETS := $(shell ls -p scripts | grep -v -e /)
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,8 @@ dockertogoarch = $(patsubst arm/v7,arm,$(1))
 
 deploy: images
 
+e2e: vendor/modules.txt
+
 golangci-lint: pkg/natdiscovery/proto/natdiscovery.pb.go
 
 unit: pkg/natdiscovery/proto/natdiscovery.pb.go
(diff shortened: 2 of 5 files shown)


dfarrell07 (dismissed):
LGTM

If anyone else is wondering about the repeating ARGs: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

skitt:
To allow building in Buildah in rootless mode, this needs to run without Dapper at all, not just allow building without Dapper.

package/Dockerfile.submariner-gateway
```diff
@@ -1,4 +1,17 @@
...
+ARG SOURCE=/go/src/github.com/submariner-io/submariner
+ARG TARGETPLATFORM=linux/amd64
+
+FROM quay.io/submariner/shipyard-dapper-base:${BASE_BRANCH} AS builder
+ARG SOURCE
+ARG TARGETPLATFORM
+
+COPY . ${SOURCE}
+
+RUN make -C ${SOURCE} LOCAL_BUILD=1 bin/${TARGETPLATFORM}/submariner-gateway
```
mkolesnik, line 11: Problem with multi-stage builds is inability to cache layers, i.e. this layer is not stored anywhere so it'll always have to be recreated, affecting any layers after it.

mkolesnik (commented):
This still leaves building the binaries outside the dapper as well, which leads to them being built twice on each change.

I also have a concern about caching.

package/Dockerfile.submariner-gateway
skitt: We can use `--cache-to` with `mode=max` to cache all the layers.

skitt:
> This still leaves building the binaries outside the dapper as well, which leads to them being built twice on each change.

We can avoid that by dropping the `images: build` dependency; IIRC that causes issues elsewhere, which need some fixes in Shipyard, but it’s been a while so I’ve forgotten the details. I’ll look into it again.

package/Dockerfile.submariner-gateway
skitt: Note too that multi-stage builds are handled in a tree, not as a sequential list of layers. Layers in the final image are cached, as are layers from images explicitly loaded into the cache; so changes in one stage only affect layers in subsequent stage which depend on that stage (_i.e._ `COPY --from` lines and subsequent layers).

skitt: I’m not sure caching _everything_ is desirable, since that would skip rebuilds in circumstances where we actually want a rebuild.

skitt:
> > This still leaves building the binaries outside the dapper as well, which leads to them being built twice on each change.
> 
> We can avoid that by dropping the `images: build` dependency; IIRC that causes issues elsewhere, which need some fixes in Shipyard, but it’s been a while so I’ve forgotten the details. I’ll look into it again.

The problem is that `e2e` needs vendoring; I’ve added that as a separate commit.

dfarrell07:
@sridhargaddam Can you take a look at this one?

dfarrell07 (approved):

mkolesnik:
I think the main question here is whats the benefit of doing this?

This seems to add a lot of complexity (and some degree of copy pasta) and perhaps some penalty on caching to the dockerfiles, but what are the benefits?

I feel the way we're building now is simple and also can be done without dapper anyhow with `LOCAL_BUILD` so I don't see that as a gain.

BTW we can remove `images: build` make dependency anyhow as it should be taken care of the Shipyard auto-dependencies (if that doesnt work we should fix it).

mkolesnik (commented):
I think the main question here is whats the benefit of doing this?

This seems to add a lot of complexity (and some degree of copy pasta) and perhaps some penalty on caching to the dockerfiles, but what are the benefits?

I feel the way we're building now is simple and also can be done without dapper anyhow with `LOCAL_BUILD` so I don't see that as a gain.

BTW we can remove `images: build` make dependency anyhow as it should be taken care of the Shipyard auto-dependencies (if that doesnt work we should fix it).

dfarrell07:
> can be done without dapper anyhow with LOCAL_BUILD

Does that introduce more requirements for the configuration of a dev's machine, making the shared dev env less portable/sharable?

> but what are the benefits?

Not using Dapper seems like a clear benefit.

skitt:
See the updated PR description for a list of benefits (probably not exhaustive). See also the tech exchange where this was presented in context.

dfarrell07:
@mkolesnik can you take a look at the benefits @skitt provided?

mkolesnik (approved):

PR #1687, opened 2022-02-07
Add more globalnet metrics

### LINKED ISSUE #1471: Support metrics in Globalnet 2.0
In Globalnet 1.0, we used to have the following metrics. This issue is to support metrics with Globalnet 2.0 code introducing new metrics as necessary.

```
submariner_global_IP_availability | cidr | Count of available global IPs per CIDR
submariner_global_IP_allocated   | cidr | Count of global IPs allocated for Pods/Services per CIDR
```

maayanf24:

### DIFF
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -121,7 +121,6 @@ func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Fede
 
 	if err == nil && len(reservedIPs) > 0 {
 		err = postReserve(reservedIPs)
-
 		if err != nil {
 			_ = c.pool.Release(reservedIPs...)
 		}
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -96,7 +96,7 @@ func (c *baseSyncerController) reconcile(client dynamic.ResourceInterface, label
 }
 
 func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Federator, obj *unstructured.Unstructured,
-	postReserve func(allocatedIPs []string) error) error {
+	postReserve func(allocatedIPs []string) error, metricsAllocate func(string, int)) error {
 	var reservedIPs []string
 
 	clearAllocatedIPs := func() {}
@@ -121,10 +121,13 @@ func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Fede
 
 	if err == nil && len(reservedIPs) > 0 {
 		err = postReserve(reservedIPs)
+
 		if err != nil {
 			_ = c.pool.Release(reservedIPs...)
 		}
 	}
+
+	metricsAllocate(c.pool.GetCider(), len(reservedIPs))
 
 	if err != nil {
 		key, _ := cache.MetaNamespaceKeyFunc(obj)
@@ -153,7 +156,7 @@ func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Fede
 }
 
 func (c *baseIPAllocationController) flushRulesAndReleaseIPs(key string, numRequeues int, flushRules func(allocatedIPs []string) error,
-	allocatedIPs ...string) bool {
+	metricsDeallocate func(string, int), allocatedIPs ...string) bool {
 	if len(allocatedIPs) == 0 {
 		return false
 	}
@@ -172,6 +175,8 @@ func (c *baseIPAllocationController) flushRulesAndReleaseIPs(key string, numRequ
 	if err := c.pool.Release(allocatedIPs...); err != nil {
 		klog.Errorf("Error while releasing the global IPs for %q: %v", key, err)
 	}
+
+	metricsDeallocate(c.pool.GetCider(), len(allocatedIPs))
 
 	return false
 }
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -96,7 +96,7 @@ func (c *baseSyncerController) reconcile(client dynamic.ResourceInterface, label
 }
 
 func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Federator, obj *unstructured.Unstructured,
-	postReserve func(allocatedIPs []string) error, metricsAllocate func(string, int)) error {
+	postReserve func(allocatedIPs []string) error) error {
 	var reservedIPs []string
 
 	clearAllocatedIPs := func() {}
@@ -127,8 +127,6 @@ func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Fede
 		}
 	}
 
-	metricsAllocate(c.pool.GetCider(), len(reservedIPs))
-
 	if err != nil {
 		key, _ := cache.MetaNamespaceKeyFunc(obj)
 
@@ -156,7 +154,7 @@ func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Fede
 }
 
 func (c *baseIPAllocationController) flushRulesAndReleaseIPs(key string, numRequeues int, flushRules func(allocatedIPs []string) error,
-	metricsDeallocate func(string, int), allocatedIPs ...string) bool {
+	allocatedIPs ...string) bool {
 	if len(allocatedIPs) == 0 {
 		return false
 	}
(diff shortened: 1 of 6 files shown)


pkg/globalnet/controllers/base_controllers.go
```diff
@@ -120,7 +120,9 @@ func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Fede
 	err := c.pool.Reserve(reservedIPs...)
 
 	if err == nil && len(reservedIPs) > 0 {
+		metricsAllocate(c.pool.GetCider(), len(reservedIPs))
```
sridhargaddam: We should move this to Line 130 as there is a possibility for error during postReserve and the allocated IP can be released in Line127

pkg/globalnet/controllers/base_controllers.go
```diff
@@ -96,7 +96,7 @@ func (c *baseSyncerController) reconcile(client dynamic.ResourceInterface, label
 }
 
 func (c *baseIPAllocationController) reserveAllocatedIPs(federator federate.Federator, obj *unstructured.Unstructured,
-	postReserve func(allocatedIPs []string) error) error {
+	postReserve func(allocatedIPs []string) error, metricsAllocate func(string, int)) error {
```
sridhargaddam: @tpantelis are you fine with this approach or do you have any suggestion?

pkg/globalnet/metrics/metrics.go
```diff
@@ -43,10 +43,38 @@ var (
 			cidrLabel,
 		},
 	)
+	globalEgressIPsAllocatedGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "submariner_global_Egress_IP_allocated",
```
tpantelis:
```suggestion
			Name: "submariner_global_egress_IP_allocated",
```

pkg/globalnet/metrics/metrics.go
```diff
@@ -43,10 +43,38 @@ var (
...
+			Help: "Count of global Egress IPs allocated for Pods/Services per CIDR",
+		},
+		[]string{
+			cidrLabel,
+		},
+	)
+	clusterGlobalEgressIPsAllocatedGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "submariner_cluster_global_Egress_IP_allocated",
```
tpantelis:
```suggestion
			Name: "submariner_cluster_global_egress_IP_allocated",
```

pkg/globalnet/metrics/metrics.go
```diff
@@ -43,10 +43,38 @@ var (
...
+			Help: "Count of global Egress IPs allocated for clusters per CIDR",
+		},
+		[]string{
+			cidrLabel,
+		},
+	)
+	globalIngressIPsAllocatedGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "submariner_global_Ingress_IP_allocated",
```
tpantelis:
```suggestion
			Name: "submariner_global_ingress_IP_allocated",
```

pkg/ipam/ippool.go
```diff
@@ -214,3 +214,7 @@ func (p *IPPool) Size() int {
 
 	return p.available.Size()
 }
+
+func (p *IPPool) GetCider() string {
```
tpantelis:
```suggestion
func (p *IPPool) GetCIDR() string {
```

pkg/globalnet/controllers/base_controllers.go
tpantelis: I assume you mean adding a second callback function. I think it's simpler to also use the `postReserve` function for the metrics.

tpantelis (changes requested):

sridhargaddam:
@maayanf24 can you please look into the review comments from @tpantelis and also rebase the PR.

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -310,7 +319,7 @@ func (c *globalIngressIPController) onDelete(ingressIP *submarinerv1.GlobalIngre
 		}
 
 		return nil
-	}, ingressIP.Status.AllocatedIP)
+	})
```
maayanf24: this line removed by mistake, I'll add it back

pkg/globalnet/controllers/cluster_egressip_controller.go
```diff
@@ -254,6 +259,8 @@ func (c *clusterGlobalEgressIPController) allocateGlobalIPs(key string, numberOf
 		return true
 	}
 
+	metrics.RecordAllocateClusterGlobalEgressIPs(c.pool.GetCIDR(), numberOfIPs)
```
tpantelis: I think we should move this after L279 then we don't need to record deallocated on L276 (it's a noop).

pkg/globalnet/controllers/global_egressip_controller.go
```diff
@@ -81,6 +82,7 @@ func NewGlobalEgressIPController(config *syncer.ResourceSyncerConfig, pool *ipam
 
 	for i := range list.Items {
 		err = controller.reserveAllocatedIPs(federator, &list.Items[i], func(reservedIPs []string) error {
+			metrics.RecordAllocateGlobalEgressIPs(pool.GetCIDR(), len(reservedIPs))
```
tpantelis, line 85: This is on startup to reserve previously allocated IPs so it I don't think we should record this as we're not actually allocating new block. @sridhargaddam WDYT?

pkg/globalnet/controllers/global_egressip_controller.go
```diff
@@ -226,6 +228,8 @@ func (c *globalEgressIPController) allocateGlobalIPs(key string, numberOfIPs int
 		return true
 	}
 
+	metrics.RecordAllocateGlobalEgressIPs(c.pool.GetCIDR(), numberOfIPs)
```
tpantelis: Similar to above, I think we should move this after L249 and remove L245.

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -81,6 +82,7 @@ func NewGlobalIngressIPController(config *syncer.ResourceSyncerConfig, pool *ipa
 
 		// nolint:wrapcheck  // No need to wrap these errors.
 		err = controller.reserveAllocatedIPs(federator, obj, func(reservedIPs []string) error {
+			metrics.RecordAllocateGlobalIngressIPs(pool.GetCIDR(), len(reservedIPs))
```
tpantelis, line 85: Similar as above - we're reserving not allocating here.

pkg/globalnet/controllers/global_ingressip_controller.go
```diff
@@ -163,12 +165,15 @@ func (c *globalIngressIPController) onCreate(ingressIP *submarinerv1.GlobalIngre
 		return true
 	}
 
+	metrics.RecordAllocateGlobalIngressIPs(c.pool.GetCIDR(), 1)
```
tpantelis: Same as above (move after L255 and don't deallocate on error).

tpantelis (changes requested):

pkg/globalnet/controllers/global_egressip_controller.go
sridhargaddam: Good point. 

1. If the metrics value is like a time series information on when the globalIPs are allocated then, as you said, we are just reserving the previously allocated globalIP here, so its not like a new allocation.
2. But if it shows the current number of globalIPs allocated, then we can record it (as the pod was just restarted).

 Lets hear from @skitt

skitt: The metric is maintained per run of the controller: whenever the controller starts, the metric is 0. If there are already-allocated IPs, we need to adjust the metric accordingly, so this seems correct to me.

nyechiel:
@sridhargaddam @skitt it looks like Tom's comments have been addressed. Can you please review?

tpantelis (approved):

skitt (approved):

sridhargaddam:
@skitt I remember you having some plans to have some CI job which checks for commit messages that includes message like "Address review comments", "Fix review comments" etc. The commits in this PR should have been squashed.

skitt:
> @skitt I remember you having some plans to have some CI job which checks for commit messages that includes message like "Address review comments", "Fix review comments" etc. The commits in this PR should have been squashed.

Aargh, they should have, I missed that. The check already exists, it’s “No "Apply suggestions from code review" Commits”, but it only checks for GitHub suggestions and fix-ups.

PR #1689, opened 2022-02-07
Fix using ovn kubernetes as the CNI 

### LINKED ISSUE #1101: no destroy-recreate objects in ovn under updates.
Avoid creating and destroying some ovn resources under the networkplugin-syncer as

**Why is this needed**:
that creates small downtime and can be handled with proper update support in go-ovn

Depends on #1100

### LINKED ISSUE #1608: Broken service connectivity across cluster with ovn-kubernetes
**What happened**:
Exposed services are not reachable from remote cluster when using ovn-kubernetes (in a OKD 4.9 environment). 

The clusters are properly connected and the services are properly exported. The traffic reaches the remote gateway and it gets inyected into br-submariner ovs bridge as expected, however the traffic gets dropped there (I check the geneve tunnel between the gateway node and the node hosting the service pod does not have any traffic).

Checking further, it was observed that the loadbalancer is not associated to the submariner router at the OVN NB DB:
ovn-nbctl lr-lb-list SUBMARINER_ROUTER_ID
(empty return)

If the loadbalancer id get gets manually added to the submariner router (lr-lb-add SUBMARINER_ROUTER_ID LOADBALANCER_ID) then the connectivity works as expected

Connectivity to work out of the box

Deploy 2 clusters witn ovn-kubernetes, expose a service in one of them, trying to reach it from the other.

I deployed with OKD and saw that the next function seems to look for something into the external_ids to find the loadbalancers to attach to the submariner router: https://github.com/submariner-io/submariner/blob/devel/pkg/networkplugin-syncer/handlers/ovn/services.go#L56,
so I'm not sure if this is related to how those external ids are handle in this type of environment or would happen the same in any environment using ovn-kubernetes

**Environment**:
All green here

Environment is running in VMs in a cloud provider (OpenStack based)

- Install tools:
Using okd IPI installer for the k8s cluster.

- Others:

astoycos:
This PR re-writes the submariner networkplugin-syncer for ovn-kubernetes.   Mainly it converts from using go-ovn/nbctl to libovsdb for interacting with the ovn-dbs. I will update this PR with more commits/explanations as I finish tidying things up. 

Fixes https://github.com/submariner-io/submariner/issues/1608

Depends on https://github.com/ovn-org/ovn-kubernetes/pull/2747
Depends on https://github.com/submariner-io/shipyard/pull/704


astoycos:
I would expect E2E to fail until the dependent PRs (updating our KIND setup) works

astoycos:
I've tested this locally by running E2E and things look good so far, so I'll go ahead and open this up for review.

mangelajo:
Very cool, @astoycos pong me to see if I can do a review. I'm not sure if I have enough bandwidth currently to go through the whole patch / do a detailed review, I will try do diagonally-read as much as possible.

I'd make sure it works with OCP + kind.

astoycos:
@skitt @dfarrell07 Currently in the `End to End Full / E2E (operator, vxlan, ovn) (pull_request)` workflow... a major false positive considering ovn-k isn't running at all

```
[14:35:21.497] [submariner]$ [cluster2] cut -f 2- -d=
Cluster settings::
  broker - 'cluster1'
  clusters - 'cluster1' 'cluster2'
  cni - ([cluster2]="weave" [cluster1]="weave" )
  nodes per cluster - ([cluster2]="control-plane worker worker" [cluster1]="control-plane worker" )
```

skitt:
The dependencies have been merged, @astoycos this should be good to go once you rebase it!

astoycos:
```
    nginx-demo:
      Container ID:  containerd://5f33ec7720d53392dc1ffbaf76c83df53a293e41aa9dba5ec0a2f0fe32d05eba
      Image:         localhost:5000/nettest:local
      Image ID:      localhost:5000/nettest@sha256:63f21b9d8065d60870c2bca980dbc4dc6358edf669661576b8cbee39e3024a2d
      Port:          8080/TCP
      Host Port:     0/TCP
      Command:
        /app/simpleserver
      State:          Waiting
        Reason:       CrashLoopBackOff
      Last State:     Terminated
        Reason:       StartError
        Message:      failed to create containerd task: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/app/simpleserver": stat /app/simpleserver: no such file or directory: unknown
``` 

Is breaking upgrades currently but not related to this PR 

This should be ready for review, the code is much simpler now due to some ovn-k changes that went in this cycle.

astoycos:
Just a heads up @mkolesnik, I'm assuming this ^ is to be expected?

main.go
```diff
@@ -252,6 +252,11 @@ func main() {
 	}()
 
 	<-stopCh
+
+	if err := cableEngine.Cleanup(); err != nil {
+		klog.Errorf("error cleaning up cableEngine resources before removeing Gateway")
```
dfarrell07:
```suggestion
		klog.Errorf("error cleaning up cableEngine resources before removing Gateway")
```

pkg/networkplugin-syncer/handlers/ovn/ovn_logical_router_policy.go
```diff
@@ -0,0 +1,82 @@
...
+	lrpStalePredicate := func(item *nbdb.LogicalRouterPolicy) bool {
+		subnet := strings.Split(item.Match, " ")[2]
+
+		return item.Priority == ovnRoutePoliciesPrio && !remoteSubnets.Contains(subnet)
+	}
+
+	// Cleanup any existing lrps not representing the correct set of remote subnets
+	err := libovsdbops.DeleteLogicalRouterPoliciesWithPredicate(ovn.nbdb, ovnClusterRouter, lrpStalePredicate)
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete stale submariner logical route policies")
```
dfarrell07, line 40:
```suggestion
		return errors.Wrapf(err, "failed to delete stale submariner logical router policies")
```

You call these "router policies" everywhere else, finding "route policies" in the logs might not be so clear.

pkg/networkplugin-syncer/handlers/ovn/ovn_logical_router_policy.go
```diff
@@ -0,0 +1,82 @@
...
+		lrpSubPredicate := func(item *nbdb.LogicalRouterPolicy) bool {
+			subnet1 := strings.Split(item.Match, " ")[2]
+			subnet2 := strings.Split(lrp.Match, " ")[2]
+
+			return item.Priority == ovnRoutePoliciesPrio && subnet1 == subnet2
+		}
+
+		if err := libovsdbops.CreateOrUpdateLogicalRouterPolicyWithPredicate(ovn.nbdb,
+			ovnClusterRouter, lrp, lrpSubPredicate); err != nil {
+			return errors.Wrapf(err, "failed to create submariner logical Router policy %v and add it to the ovn cluster router", lrp)
```
dfarrell07, line 55:
```suggestion
			return errors.Wrapf(err, "failed to create submariner logical router policy %v and add it to the ovn cluster router", lrp)
```

Super-nit but while I'm comparing these error messages.

go.mod
```diff
@@ -1,30 +1,32 @@
...
 	golang.zx2c4.com/wireguard/wgctrl v0.0.0-20211215182854-7a385b3431de
 	google.golang.org/protobuf v1.27.1
-	k8s.io/api v0.21.0
-	k8s.io/apimachinery v0.21.0
+	k8s.io/api v0.23.3
+	k8s.io/apimachinery v0.23.3
 	k8s.io/client-go v1.5.2
 	k8s.io/klog v1.0.0
-	k8s.io/utils v0.0.0-20210305010621-2afb4311ab10
+	k8s.io/klog/v2 v2.30.0
```
dfarrell07, line 28: I think this klog bump is breaking CI?

```
#11 8.480 [15:49:43.881] [submariner]$ CGO_ENABLED=0 go build -trimpath -ldflags -s -w -X main.VERSION=devel-1a5ac81 -o bin/linux/amd64/submariner-gateway .
#11 80.41 # sigs.k8s.io/controller-runtime/pkg/log
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go:50:31: cannot use NullLogger{} (value of type NullLogger) as type logr.Logger in argument to NewDelegatingLogger
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go:54:24: cannot use Log (variable of type *DelegatingLogger) as type logr.Logger in variable declaration
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go:56:16: assignment mismatch: 1 variable but logr.FromContext returns 2 values
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/null.go:30:21: cannot use NullLogger{} (value of type NullLogger) as type logr.Logger in variable declaration
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/null.go:48:37: undefined: logr.InfoLogger
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/null.go:54:9: cannot use log (variable of type NullLogger) as type logr.Logger in return statement
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/null.go:59:9: cannot use log (variable of type NullLogger) as type logr.Logger in return statement
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/deleg.go:103:9: cannot use res (variable of type *DelegatingLogger) as type logr.Logger in return statement
#11 80.41 vendor/sigs.k8s.io/controller-runtime/pkg/log/deleg.go:116:9: cannot use res (variable of type *DelegatingLogger) as type logr.Logger in return statement
#11 105.6 make: *** [Makefile:85: bin/linux/amd64/submariner-gateway] Error 2
```

astoycos: Yep Looking at fixing it now :)

tpantelis: yeah we need to keep `logr` at 0.4.0 so we need to keep K8s at 0.21.11 and `klog/v2` at v2.9.0 (or pin it).

pkg/networkplugin-syncer/handlers/ovn/connection.go
```diff
@@ -113,3 +113,57 @@ func getOVNTLSConfig(pkFile, certFile, caFile string) (*tls.Config, error) {
 		MinVersion:   tls.VersionTLS12,
 	}, nil
 }
+
+func createLibovsdbClient(dbAddress string, tlsConfig *tls.Config, dbModel model.ClientDBModel) (libovsdbclient.Client, error) {
```
tpantelis, line 117:
```suggestion
func createLibOvsdbClient(dbAddress string, tlsConfig *tls.Config, dbModel model.ClientDBModel) (libovsdbclient.Client, error) {
```

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	klog.V(log.DEBUG).Infof("Reconciling these raw remote subnets %v to logical router policies", remoteSubnets.Elements())
+
+	err := ovn.reconcileSubOvnLogicalRouterPolicies(remoteSubnets)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return ovn.reconcileSubOvnLogicalRouterPolicies(remoteSubnets)
```

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	submarinerRouter.Options["chassis"] = chassis.Name
+
+	err = libovsdbops.CreateOrUpdateLogicalRouter(ovn.nbdb, submarinerRouter)
+	if err != nil {
+		return errors.Wrap(err, "failed to set chassis option on ovn submariner router")
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return errors.Wrap(err, "failed to set chassis option on ovn submariner router")
```

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
```diff
@@ -0,0 +1,254 @@
...
+		MAC:      submarinerUpstreamMAC,
+		Networks: []string{submarinerUpstreamNET},
+	}
+
+	err = libovsdbops.CreateOrUpdateLogicalRouterPorts(ovn.nbdb, &subLogicalRouter, &subRouterTosubGatewayLrp)
+	if err != nil {
+		return errors.Wrap(err, "failed to add gateway to submariner router LSP to the submariner gateway switch")
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return errors.Wrap(err, "failed to add gateway to submariner router LSP to the submariner gateway switch")
```

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
```diff
@@ -0,0 +1,254 @@
...
+
+	klog.V(log.DEBUG).Infof("reconciling north static routes on %q router for subnets %v", submarinerLogicalRouter,
+		remoteSubnets.Elements())
+
+	err := ovn.reconcileSubOvnLogicalRouterStaticRoutes(submarinerUpstreamRPort, HostUpstreamIP, remoteSubnets)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return ovn.reconcileSubOvnLogicalRouterStaticRoutes(submarinerUpstreamRPort, HostUpstreamIP, remoteSubnets)
```

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
```diff
@@ -0,0 +1,254 @@
...
+
+	klog.V(log.DEBUG).Infof("reconciling south static routes on %q for subnets %v", submarinerLogicalRouter,
+		localSubnets)
+
+	err := ovn.reconcileSubOvnLogicalRouterStaticRoutes(submarinerDownstreamRPort, ovnClusterSubmarinerIP, localSubnets)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return ovn.reconcileSubOvnLogicalRouterStaticRoutes(submarinerDownstreamRPort, ovnClusterSubmarinerIP, localSubnets)
```

go.mod
astoycos: So the OVN-k Bits require we have k8s 23.3..... Thankfully I was able to bump k8s version (unpinning all pinned versions) without any issues

pkg/networkplugin-syncer/handlers/ovn/connection.go
astoycos: `Libovsdb` is a single word here :)

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
astoycos: Wouldn't this result in us always returning an error? And in this case we need to ensure that the `libovsdbops` function call  is wrapped here so that if it's encountered the user can understand the context

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
astoycos: Same comment as above ^

go.mod
tpantelis: We ran into issues in **submariner-operator** with **lighthouse** bumped to v0.23.5 - we had to pin a bunch of things. We're going to have to do the same here (ie pin k8s 0.21.11). `sigs.k8s.io/controller-runtime` v0.11.2 uses the backwards incompatible `logr` 1.2.0 - we need to directly set or pin it to `sigs.k8s.io/controller-runtime` v0.9.7. Also `github.com/go-logr/logr` pinned to v0.4.0 and `k8s.io/klog/v2` to v2.9.0.

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
tpantelis: No, `errors.Wrap` et al conveniently check if `err` is nil.

astoycos: Ah Nice I didn't know that :)

astoycos:
I still need to work on vendoring issues, since now they're breaking unit tests

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
astoycos: Done

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
astoycos: Done

pkg/globalnet/controllers/controllers_suite_test.go
```diff
@@ -114,6 +116,10 @@ func newTestDriverBase() *testDriverBase {
 	Expect(submarinerv1.AddToScheme(t.scheme)).To(Succeed())
 	Expect(corev1.AddToScheme(t.scheme)).To(Succeed())
 
+	// TODO: Remove this workaround for https://github.com/kubernetes/client-go/issues/949 once
+	// admiral has been updated
+	t.scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "fake-dynamic-client-group", Version: "v1", Kind: "List"}, &unstructured.UnstructuredList{})
+
```
tpantelis, line 122: I think we should pin to K8s 0.21.11 to avoid having to do this. The rest of the submariner projects use 0.21.11 so I think we should be consistent.

tpantelis (changes requested):

pkg/globalnet/controllers/controllers_suite_test.go
astoycos: @tpantelis  I was more just trying to figure out why that was happening :) ..... Also I believe [this problem](https://github.com/kubernetes/client-go/issues/949) occurred from 0.19 -> 0.20 so we'll probably still run into this specific issue 

> We ran into issues in submariner-operator with lighthouse bumped to v0.23.5 - we had to pin a bunch of things. We're going to have to do the same here (ie pin k8s 0.21.11). sigs.k8s.io/controller-runtime v0.11.2 uses the backwards incompatible logr 1.2.0 - we need to directly set or pin it to sigs.k8s.io/controller-runtime v0.9.7. Also github.com/go-logr/logr pinned to v0.4.0 and k8s.io/klog/v2 to v2.9.0.

I'll give pinning to 0.21.11 a try...But currently there's some dependencies with the imported OVN-K libraries which are forcing the version up to v0.23.3 since that's what's defined here -> https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/go.mod#L38.  Also something to keep in mind... We don't use `logr` in this repo so I don't think that's a problem anymore... I updated klog/v2 to v2.30 (in this current iteration) and it looks like everything is running alright 

But I do need to ensure that we can import those ovn-k libraries without depending on k8s version.. Or else then every time ovn-k bumps their K8s version we'll have to, creating an unnecessary dependency (unless we are always just keeping it pinned...)

astoycos: For example when I pin to 0.21.11 

```
#14 29.60 # github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdbops
#14 29.60 vendor/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdbops/transact.go:21:20: undefined: wait.PollImmediateUntilWithContext
```

tpantelis: > Also something to keep in mind... We don't use `logr` in this repo so that's not a problem anymore... 

But I intend/want to use `logr`, specifically for zerolog.

> But I do need to ensure that we can import those ovn-k libraries without depending on k8s version.. Or else then every time ovn-k bumps their K8s version we'll have to, creating an unnecessary dependency.

I see that you're using the `master` branch from `github.com/ovn-org/ovn-kubernetes`. Not sure how stable that is but I guess we have to since no prior version has a `go.mod` file.

tpantelis: `wait.PollImmediateUntilWithContext ` was added in K8s 0.22 so perhaps try pinning to 0.22.9, which uses `k8s.io/klog/v2` v2.9.0 and `logr` v0.4.0.

astoycos: > I see that you're using the master branch from github.com/ovn-org/ovn-kubernetes. Not sure how stable that is but I guess we have to since no prior version has a go.mod file.

Right they have no upstream release versioning process so it is a bit unstable.... granted these bits were already pretty unstable, and regardless I don't think what we have here is necessarily the best long-term ovn-k integration solution since any way we cut it there will always be a tightly coupled dependency with the ovn-k version

tpantelis: Ideally we would bump to K8s 0.23 across the board. At one point **lighthouse** was using 0.23 due to `coredns` bump but we downgraded due to **submariner-operator** issues.  Obviously **submariner** is building fine with 0.23 with this PR.  I'm not sure what's holding us back. @skitt? Is it the older operator framework in **submariner-operator**?

tpantelis: I tried pinning to K8s 0.22.9 et al but hit a showstopper:

```
github.com/submariner-io/submariner/pkg/networkplugin-syncer/handlers/ovn imports
        github.com/ovn-org/libovsdb/client imports
        github.com/go-logr/stdr imports
        github.com/go-logr/logr/funcr: no required module provides package github.com/go-logr/logr/funcr; to add it:
        go get github.com/go-logr/logr/funcr
~/go/src/github.com/submariner-io/submariner>go get github.com/go-logr/logr/funcr
go: module github.com/go-logr/logr@upgrade found (v1.2.3, replaced by github.com/go-logr/logr@v0.4.0), but does not contain package github.com/go-logr/logr/funcr
```

So `github.com/go-logr/logr/funcr` was added in `logr` v1.0.0 which also introduces the backwards incompatible `Logger` change from interface to struct. So in order to use `libovsdb`, we need K8s 0.23 et al. We'll see what happens when the **submariner** version is bumped in **submariner-operator** 😄 Of course bumping everything to K8s 0.23 would help.

astoycos: Yeah I experienced the exact same thing :(  

And sounds good, I'll let this sit for now since CI is green and continue to monitor the best way forward

tpantelis: Discussed with @skitt - after bumping the operator framework version and adjusting our operator code (planned epic for this release) and requiring go 1.17, there should be no obstacle in bumping to K8s 0.23 across the projects.

tpantelis (approved):

dfarrell07 (approved):
LGTM! @sridhargaddam / @vthapar / @aswinsuryan does anyone else want to review or can we merge this?

sridhargaddam:
> LGTM! @sridhargaddam / @vthapar / @aswinsuryan does anyone else want to review or can we merge this?

@dfarrell07 I could not look into this as I was occupied with other stuff, but @vthapar wanted to try this out on an OCP cluster. I'll let him comment. I'm fine if we want to go ahead and merge it.

vthapar:
@dfarrell07 @sridhargaddam While testing I ran into a issue with OCP 4.10. This fix requires OCP 4.11 or later to work. I've been unable to get a working setup with OCP 4.11 I'd like to confirm it working with 4.11 at least once before we merge it. I should have results in a day at most.

vthapar (approved):
I didn't review the code but did test it out and confirmed E2E to be passing in both kind and AWS+OCP.

PR #1691, opened 2022-02-08
Clean up Endpoints and Clusters on uninstall

tpantelis:
Delete all from the local datastore and delete local objects from the broker.

### DIFF
--- a/pkg/controllers/datastoresyncer/datastore_cluster_sync_test.go
+++ b/pkg/controllers/datastoresyncer/datastore_cluster_sync_test.go
@@ -30,7 +30,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var _ = Describe("Cluster syncing", testClusterSyncing)
+var (
+	_ = Describe("Cluster syncing", testClusterSyncing)
+	_ = Describe("Cluster cleanup", testClusterCleanup)
+)
 
 func testClusterSyncing() {
 	t := newTestDriver()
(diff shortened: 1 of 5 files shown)


pkg/controllers/datastoresyncer/datastore_cluster_sync_test.go
```diff
@@ -112,3 +115,37 @@ func testClusterSyncing() {
...
+		}))
+	})
+
+	It("should remove local Clusters from the remote datastore", func() {
+		Expect(t.syncer.Cleanup()).To(Succeed())
+
+		test.AwaitNoResource(t.brokerClusters, clusterID)
+
+		time.Sleep(500 * time.Millisecond)
+		test.AwaitResource(t.brokerClusters, otherClusterID)
```
sridhargaddam, line 142: Usually every cluster has only a single Cluster Object right, any specific reason for validating two cluster objects? 
To be more precise, if we are not the owner for otherClusterID, why are we expecting that to be deleted on the Broker?

pkg/controllers/datastoresyncer/datastore_endpoint_sync_test.go
```diff
@@ -213,3 +214,48 @@ func testEndpointExclusivity() {
...
+		test.CreateResource(t.brokerEndpoints, test.SetClusterIDLabel(existingRemoteEndpoint, existingRemoteEndpoint.Spec.ClusterID))
+	})
+
+	It("should remove local Endpoints from the remote datastore", func() {
+		Expect(t.syncer.Cleanup()).To(Succeed())
+
+		test.AwaitNoResource(t.brokerEndpoints, existingLocalEndpoint.GetName())
+
+		time.Sleep(500 * time.Millisecond)
+		test.AwaitResource(t.brokerEndpoints, existingRemoteEndpoint.GetName())
```
sridhargaddam, line 252: Same as above

pkg/controllers/datastoresyncer/datastore_cluster_sync_test.go
tpantelis: I'm ensuring it didn't delete the `Cluster` object for other cluster from the broker. It's calling  `AwaitResource` on L142 instead of `AwaitNoResource` - perhaps you overlooked that?

sridhargaddam: Oh yeah. Indeed i overlooked it :P

sridhargaddam (approved):

dfarrell07 (approved):

Issue #1693, opened 2022-02-11
Globalnet does not work for services without selector


anfredette:
**What happened**:
Created a service without selector and it was not reachable from another cluster in the clusterset.

**What you expected to happen**:
That the service be reachable from anywhere in the clusterset

**How to reproduce it (as minimally and precisely as possible)**:
- Create a service w/o selector
- Manually create a pod endpoint for the service
- Confirm that it is accessible on the local cluster
- Export the service
- Attempt to access the global IP address from nettest on another cluster in the clusterset
- Note that it is not accessible

Note: The actual global IP address must be used instead of the domain name until https://github.com/submariner-io/lighthouse/issues/603 is resolved.

**Anything else we need to know?**:
It is believed that this issues was introduced by https://github.com/submariner-io/submariner/pull/1643

A work-around is to manually create an endpoint for the internal service created by the export. For example, execute the following script (assuming service and pod names are as defined below)
```
SVC_NAME=nginx-svc
SVC_POD_NAME=nginx-pod
SVC_PORT=80
INTERNAL_SVC_NAME=$(kubectl get services -l submariner.io/exportedServiceRef=$SVC_NAME | grep submariner | awk '{print $1}')
SVC_POD_IP=$(kubectl get pods -n default $SVC_POD_NAME -o wide | grep $SVC_POD_NAME | awk '{print $6}')

cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Endpoints
metadata:
  name: $INTERNAL_SVC_NAME
subsets:
  - addresses:
      - ip: $SVC_POD_IP
    ports:
      - port: $SVC_PORT
EOF
```

**Environment**:
- Diagnose information (use `subctl diagnose all`):
- Gather information (use `subctl gather`):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

sridhargaddam:
This is still relevant. The PR from @anfredette addresses the LH/ServiceDiscovery use-cases.

anfredette:
I didn't intend to close this issue per comment from @sridhargaddam.  I am reopening it again.

anfredette:
This was fixed by the following two PRs:
https://github.com/submariner-io/submariner/pull/1695
https://github.com/submariner-io/submariner-operator/pull/1845

PR #1582, opened 2021-10-27
Add more E2E tests for external connectivity use case

mkimuram:
This PR adds more e2e tests for external connectivity use case. Tests are refactored to align with E2E tests for normal cross-cluster use cases, before adding tests.

Note that most codes are copied from https://github.com/submariner-io/submariner/pull/1558 , so that #1558 can focus on adding a new feature.

Dependency:
- [x] https://github.com/submariner-io/submariner/pull/1695

Fixes: #1581

PR #1695, opened 2022-02-12
Implement endpoints controller for service without selector

### LINKED ISSUE #1694: Service without selector needs endpoints for internal service
Support globalnet for service without selector.

**Why is this needed**:
With https://github.com/submariner-io/submariner/pull/1643, iptables rule to forward to kube-proxy chain for service is removed, instead forwarding to globalIngressIP to endpoints is offloaded to externalIP feature of the Service, by creating internal service. Endpoints of the service __with__ selector can be synced between original and internal services by sharing labels, but Endpoints of the service __without__ selector needs to be handled in a different way.

mkimuram:
This PR implement endpoints controller for service without selector.

Dependency:
- [x] https://github.com/submariner-io/submariner-operator/pull/1845

Fixes: https://github.com/submariner-io/submariner/issues/1694

### DIFF
--- a/pkg/globalnet/constants/constants.go
+++ b/pkg/globalnet/constants/constants.go
@@ -20,6 +20,7 @@ package constants
 
 const (
 	ClusterGlobalEgressIPName = "cluster-egress.submariner.io"
+	EndpointClonedFrom        = "endpoints.submariner.io/cloned-from"
 
 	SmGlobalnetIngressChain = "SUBMARINER-GN-INGRESS"
 	SmGlobalnetEgressChain  = "SUBMARINER-GN-EGRESS"
--- a/pkg/globalnet/controllers/endpoints_controller.go
+++ b/pkg/globalnet/controllers/endpoints_controller.go
@@ -1,0 +1,146 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"reflect"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/util"
+	"github.com/submariner-io/submariner/pkg/globalnet/constants"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+func startEndpointsController(name, namespace string, config *syncer.ResourceSyncerConfig) (*endpointsController, error) {
+	// We'll panic if config is nil, this is intentional
+	var err error
+
+	klog.Infof("Creating Endpoints controller for %s/%s", namespace, name)
+
+	_, gvr, err := util.ToUnstructuredResource(&corev1.Endpoints{}, config.RestMapper)
+	if err != nil {
+		return nil, errors.Wrap(err, "error converting resource")
+	}
+
+	controller := &endpointsController{
+		baseSyncerController: newBaseSyncerController(),
+		name:                 name,
+		namespace:            namespace,
+		endpoints:            config.SourceClient.Resource(*gvr),
+	}
+
+	fieldSelector := fields.Set(map[string]string{"metadata.name": name}).AsSelector().String()
+
+	controller.resourceSyncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:                "Endpoints syncer",
+		ResourceType:        &corev1.Endpoints{},
+		SourceClient:        config.SourceClient,
+		SourceNamespace:     namespace,
+		RestMapper:          config.RestMapper,
+		Federator:           federate.NewCreateOrUpdateFederator(config.SourceClient, config.RestMapper, namespace, "" /* localClusterID */),
+		Transform:           controller.process,
+		SourceFieldSelector: fieldSelector,
+		ResourcesEquivalent: areEndpointsEqual,
+	})
+
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating the syncer")
+	}
+
+	if err := controller.Start(); err != nil {
+		return nil, err
+	}
+
+	return controller, nil
+}
+
+func (c *endpointsController) Stop() {
+	c.baseController.Stop()
+}
+
+func (c *endpointsController) Start() error {
+	err := c.baseSyncerController.Start()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *endpointsController) process(from runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	ep := from.(*corev1.Endpoints)
+
+	switch op {
+	case syncer.Create, syncer.Update:
+		return c.onCreateOrUpdate(ep)
+	case syncer.Delete:
+		return c.onDelete(ep)
+	}
+
+	return nil, false
+}
+
+func (c *endpointsController) onCreateOrUpdate(ep *corev1.Endpoints) (runtime.Object, bool) {
+	key, _ := cache.MetaNamespaceKeyFunc(ep)
+
+	klog.Infof("Processing Endpoints %q", key)
+
+	// Skip cloning already cloned Endpoints resource
+	if _, ok := ep.Labels[constants.EndpointClonedFrom]; ok {
+		klog.Infof("Skip cloning already cloned Endpoints %q", key)
+
+		return ep, false
+	}
+
+	clonedEp := &corev1.Endpoints{}
+	clonedEp.Namespace = ep.Namespace
+	clonedEp.Name = GetInternalSvcName(ep.Name)
+	clonedEp.Labels = map[string]string{constants.EndpointClonedFrom: ep.Name}
+	clonedEp.Subsets = ep.Subsets
+
+	return clonedEp, false
+}
+
+func (c *endpointsController) onDelete(ep *corev1.Endpoints) (runtime.Object, bool) {
+	key, _ := cache.MetaNamespaceKeyFunc(ep)
+
+	klog.Infof("Endpoints %q deleted", key)
+
+	return &corev1.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GetInternalSvcName(ep.Name),
+			Namespace: ep.Namespace,
+		},
+	}, false
+}
+
+func areEndpointsEqual(obj1, obj2 *unstructured.Unstructured) bool {
+	subsets1, _, _ := unstructured.NestedSlice(obj1.Object, "subsets")
+	subsets2, _, _ := unstructured.NestedSlice(obj2.Object, "subsets")
+
+	return reflect.DeepEqual(subsets1, subsets2)
+}
--- a/pkg/globalnet/controllers/base_controllers.go
+++ b/pkg/globalnet/controllers/base_controllers.go
@@ -258,3 +258,14 @@ func GetInternalSvcName(name string) string {
 
 	return strings.ToLower(svcName)
 }
+
+func deleteEndpoints(namespace, name string,
+	client dynamic.NamespaceableResourceInterface) error {
+	err := client.Namespace(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if apierrors.IsNotFound(err) {
+		klog.Warningf("Could not find Endpoints %s/%s to delete", namespace, name)
+		return nil
+	}
+
+	return errors.Wrapf(err, "error deleting Endpoints %s/%s", namespace, name)
+}
--- a/pkg/globalnet/controllers/controllers_suite_test.go
+++ b/pkg/globalnet/controllers/controllers_suite_test.go
@@ -93,6 +93,7 @@ type testDriverBase struct {
 	globalIngressIPs       dynamic.ResourceInterface
 	services               dynamic.ResourceInterface
 	serviceExports         dynamic.ResourceInterface
+	endpoints              dynamic.ResourceInterface
 	pods                   dynamic.NamespaceableResourceInterface
 	nodes                  dynamic.ResourceInterface
 	watches                *fakeDynClient.WatchReactor
@@ -100,7 +101,7 @@ type testDriverBase struct {
 
 func newTestDriverBase() *testDriverBase {
 	t := &testDriverBase{
-		restMapper: test.GetRESTMapperFor(&submarinerv1.Endpoint{}, &corev1.Service{}, &corev1.Node{}, &corev1.Pod{},
+		restMapper: test.GetRESTMapperFor(&submarinerv1.Endpoint{}, &corev1.Service{}, &corev1.Node{}, &corev1.Pod{}, &corev1.Endpoints{},
 			&submarinerv1.GlobalEgressIP{}, &submarinerv1.ClusterGlobalEgressIP{}, &submarinerv1.GlobalIngressIP{}, &mcsv1a1.ServiceExport{}),
 		scheme:       runtime.NewScheme(),
 		ipt:          fakeIPT.New(),
@@ -127,6 +128,8 @@ func newTestDriverBase() *testDriverBase {
 	t.clusterGlobalEgressIPs = t.dynClient.Resource(*test.GetGroupVersionResourceFor(t.restMapper, &submarinerv1.ClusterGlobalEgressIP{}))
 
 	t.pods = t.dynClient.Resource(*test.GetGroupVersionResourceFor(t.restMapper, &corev1.Pod{}))
+
+	t.endpoints = t.dynClient.Resource(*test.GetGroupVersionResourceFor(t.restMapper, &corev1.Endpoints{})).Namespace(namespace)
 
 	t.services = t.dynClient.Resource(*test.GetGroupVersionResourceFor(t.restMapper, &corev1.Service{})).Namespace(namespace)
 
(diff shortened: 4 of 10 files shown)


mkimuram:
@sridhargaddam @nerdalert @anfredette 

We might need to discuss how it should be implemented, first. However, to make it easier to discuss further, I've implemented it to just sync original endpoints to cloned endpoints if the original endpoints has "endpoints.submariner.io/exported: true" label. By doing so, the sync logic can be completely separated from other globalnet's behaviors, like if the endpoints are for service without selector or exported or if the service for the endpoints  even exists.

Note that we would need to add RBAC rule to allow globalnet controller to create endpoints.

pkg/globalnet/controllers/endpoints_controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+var _ = Describe("Endpoints controller", func() {
+	t := newEndpointsControllerTestDriver()
+
+	var (
+		epName string
+		ep     *corev1.Endpoints
+	)
+
+	When("an endpoints without label is created", func() {
```
tpantelis:
```suggestion
	When("an Endpoints resource without the exported label is created", func() {
```

pkg/globalnet/controllers/endpoints_controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+			t.createEndpoints(newEndpoints(epName, map[string]string{}))
+			t.awaitEndpoints(epName)
+		})
+
+		It("should not create a cloned endpoints", func() {
+			t.awaitNoEndpoints(controllers.GetInternalSvcName(epName))
+		})
+	})
+
+	When("an endpoints with label is created", func() {
```
tpantelis:
```suggestion
	When("an Endpoints resource with the exported label is created", func() {
```

pkg/globalnet/constants/constants.go
```diff
@@ -31,5 +31,7 @@ const (
 	SmGlobalnetEgressChainForNamespace       = "SM-GN-EGRESS-NS"
 	SmGlobalnetEgressChainForCluster         = "SM-GN-EGRESS-CLUSTER"
 
-	SmGlobalIP = "submariner.io/globalIp"
+	SmGlobalIP         = "submariner.io/globalIp"
```
tpantelis: We can drop the _Sm_ prefix.

pkg/globalnet/controllers/endpoints_controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+		BeforeEach(func() {
+			epName = "epWithLabel"
+			ep = t.createEndpoints(newEndpoints(epName, map[string]string{"endpoints.submariner.io/exported": "true"}))
```
tpantelis:
```suggestion
			ep = t.createEndpoints(newEndpoints(epName, map[string]string{constants.ExportedEndpoint: "true"}))
```

pkg/globalnet/constants/constants.go
```diff
@@ -31,5 +31,7 @@ const (
...
+	SmExportedEndpoint = "endpoints.submariner.io/exported"
+	SmLabelTrue        = "true"
```
tpantelis: A bit of a nit but I think it's overkill to create a constant for "_true_".  But if we do I think we should be consistent and define it in admiral and use it everywhere.

pkg/globalnet/controllers/endpoints_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	controller.resourceSyncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:            "Endpoints syncer",
+		ResourceType:    &corev1.Endpoints{},
+		SourceClient:    config.SourceClient,
+		SourceNamespace: corev1.NamespaceAll,
+		RestMapper:      config.RestMapper,
+		Federator:       federate.NewCreateFederator(config.SourceClient, config.RestMapper, corev1.NamespaceAll),
+		Transform:       controller.process,
+	})
+
```
tpantelis, line 69: To be complete and consistent with other controllers, we probably should reconcile possible internal Endpoints whose exported Endpoints were deleted during downtime.

tpantelis (changes requested):

pkg/globalnet/controllers/endpoints_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	if val, ok := ep.Labels[constants.SmExportedEndpoint]; !ok || val != constants.SmLabelTrue {
+		return nil, false
+	}
+
+	switch op {
+	case syncer.Create:
+		return c.onCreate(ep)
+	case syncer.Delete:
+		return c.onDelete(ep)
+	case syncer.Update:
```
sridhargaddam: Currently, we are skipping the Update event. What if the user modifies the ip/port of the original endpoint?

pkg/globalnet/controllers/endpoints_controller.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (c *endpointsController) process(from runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	ep := from.(*corev1.Endpoints)
+
+	// Skip endpoints without "endpoints.submariner.io/exported: true" label
```
sridhargaddam: This logic works with an assumption that user will create original endpoints with a specific label `endpoints.submariner.io/exported`. But its very likely that user might miss adding this label to the original endpoint as they will just create regular service without selectors and an associated endpoint. IMHO we should make that use-case work out of the box without expecting this label from the user.

sridhargaddam (changes requested):
Thanks for the PR @mkimuram

pkg/globalnet/controllers/endpoints_controller.go
vthapar: We need some means to know if endpoint is created by user for Exported Service or not. If no label, then we have to check if ServiceExport is created or not. Also, we have to handle case where ServiceExport is created after Endpoint is created.

Requirement to have label, though minor, is not without precedent in kubernetes. We can handle Update where one may add label after creation when they decide to export service.

Still, we can't add controller for *all* endpoints as there will be way too many endpoints in a cluster. Note lighthouse ads label for endpointslices created by it. We can easily add code in `subctl` export service to add label for external services.

pkg/globalnet/controllers/endpoints_controller.go
vthapar: We should also handle the case where user may add the label later. Possible for scenarios where external service already existed but now chose to export it.

pkg/globalnet/controllers/endpoints_controller.go
sridhargaddam: > Still, we can't add controller for _all_ endpoints as there will be way too many endpoints in a cluster. Note lighthouse ads label for endpointslices created by it. We can easily add code in `subctl` export service to add label for external services.

We don't need a controller for all endpoints. We can create a controller/listener with a label selector (just like how we are creating the pod listeners) where the selector will match the name of the original exported service.

mkimuram:
@tpantelis @sridhargaddam @vthapar

Thank you for your comments. I will update the PR as suggested.

pkg/globalnet/controllers/endpoints_controller.go
mkimuram: @sridhargaddam 

I've changed the codes to handle only Endpoints for exported Services. Current results of unit tests shows that it works well regardless of the order of exporting Services and creating Endpoints, however, it doesn't seems to work if original Endpoints resource is updated after cloned Endpoints are created by the controller (in this case, the cloned Endpoints is also expected to be updated to have the same Subsets, but it doesn't seems to be updated).

Could you check if I'm doing something wrong?
(Are there any requirements for admiral to actually update the resource? I assume that just returning the expected resource in `Transform` function makes the resource updated).

anfredette:
@mkimuram, as an FYI, I tested your original PR on some KIND clusters today, and I couldn't get it to work, though it's possible I was doing something wrong.  I'll follow up with you on it tomorrow.  I can give your new one a try tomorrow as well.

mkimuram:
@anfredette

Thank you for your testing.
You can test this PR like below way (although, it still has an [issue](https://github.com/submariner-io/submariner/pull/1695#discussion_r807450882)).

<details>

1. Run e2e test for deploy the test environment (test itself should fail)
```
make e2e using=external-net,globalnet
```

2. Update submariner-globalnet clusterrole and recreate globalnet pod (https://github.com/submariner-io/submariner-operator/pull/1845 will fix this)

```
export KUBECONFIG=output/kubeconfigs/kind-config-cluster1
kubectl get clusterrole submariner-globalnet -o json \
  | jq '.rules += [{"apiGroups":[""],"resources":["endpoints"],"verbs":["create","get","list","watch","update","delete"]}]' \
  | kubectl apply -f -
kubectl delete pod -l "app=submariner-globalnet" -n submariner-operator

export KUBECONFIG=output/kubeconfigs/kind-config-cluster2
kubectl get clusterrole submariner-globalnet -o json \
  | jq '.rules += [{"apiGroups":[""],"resources":["endpoints"],"verbs":["create","get","list","watch","update","delete"]}]' \
  | kubectl apply -f -
kubectl delete pod -l "app=submariner-globalnet" -n submariner-operator
```

3. Create Service without selector and Endpoints (`Endpoints.subsets.addresses.ip` should point to test-vm docker container's IP)

```
export KUBECONFIG=output/kubeconfigs/kind-config-cluster1
cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Service
metadata:
  name: test-vm
spec:
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
EOF

cat << EOF | kubectl apply -f -
apiVersion: v1
kind: Endpoints
metadata:
  name: test-vm
subsets:
  - addresses:
      - ip: 172.117.0.4
    ports:
      - port: 80
EOF

subctl export service -n default test-vm
```

4. Confirm that internal Endpoints is created by controller and globalingressip is allocated

```
kubectl get endpoints
NAME                                          ENDPOINTS         AGE
kubernetes                                    172.18.0.7:6443   29m
submariner-4cl54k5nvyvsg3p65n5la4lbtg5scllc   172.117.0.4:80    6m51s
test-vm                                       172.117.0.4:80    6m56s

kubectl get globalingressip test-vm -o yaml | grep allocatedIP
  allocatedIP: 242.254.1.253
```

5. Confirm that test-vm can be accessed from pod in cluster2

```
export KUBECONFIG=output/kubeconfigs/kind-config-cluster2
kubectl -n default run tmp-shell --rm -i --tty --image quay.io/submariner/nettest -- bash
curl 242.254.1.253
```

</details>

anfredette:
@mkimuram, FYI, today I just checked out your pr, and ran the following:
`make deploy using=lighthouse DEPLOY_ARGS='--globalnet --cable_driver vxlan'`

Without doing anything else, it failed the basic make deploy check.  [Details here](https://gist.github.com/anfredette/d8be2639ba2ec8dc8e1359d0474401e3).

Without your pr, the above make deploy ran fine and I was able to do my manual testing for service without selector.  I'll give your test approach a try also.

anfredette (changes requested):
See comments made on the "Conversation" tab.

mkimuram:
@anfredette 

Could you try `make deploy using=external-net,globalnet,lighthouse,vxlan` instead of passing similar args to `DEPLOY_ARGS` directly? In Makefile, at least `external-net` updates other than `DEPLOY_ARGS`, so it may fail due to this difference (I didn't test with `lighthouse` and `vxlan`, but I hope that it should work).

anfredette:
Will do. I'm trying your approach first. Then I'll come back and give this
a try.

anfredette:
> kubectl get clusterrole submariner-globalnet -o json \
>   | jq '.rules += [{"apiGroups":[""],"resources":["endpoints"],"verbs":["create","get","list","watch","update","delete"]}]' \
>   | kubectl apply -f -

If I checkout and build with https://github.com/submariner-io/submariner-operator/pull/1845, do I still need to run the above command?

mkimuram:
Please check if clusterrole submariner-globalnet has ["create","get","list","watch","update","delete"] right for apiGroups [""]'s "endpoints" resource. If not, you need to. Just for testing, it would be safe to run the command any time, because it just adds above, even if it already exists.

mkimuram:
> If I checkout and build with https://github.com/submariner-io/submariner-operator/pull/1845, do I still need to run the above command?

The deploy script for e2e uses the subctl command inside the dappper, so you would need to do it even if you build subctl by yourself (unless you replace the subctl inside dapper). If you manually deploy your environment by using the subctl command, it would be work well without the command.

anfredette:
Got it.

pkg/globalnet/controllers/endpoints_controller.go
mkimuram: @tpantelis @sridhargaddam 

I noticed there is `NewCreateOrUpdateFederator`, not `NewCreateFederator`. Is my understanding right that we need to use `NewCreateOrUpdateFederator` if updating resource is required?
If so, what needs to be passed as `localClusterID` and `keepMetadataField`?
(At a glance, just passing "" and nothing seems working, but I need to ensure that it is the right way).

tpantelis: > @tpantelis @sridhargaddam
> 
> I noticed there is `NewCreateOrUpdateFederator`, not `NewCreateFederator`. Is my understanding right that we need to use `NewCreateOrUpdateFederator` if updating resource is required? 

Yes

If so, what needs to be passed as `localClusterID` and `keepMetadataField`? (At a glance, just passing "" and nothing seems working, but I need to ensure that it is the right way).

`localClusterID` is used for 2-way syncing with the broker - that's not relevant here so pass "". You also don't need any `keepMetadataField`.

mkimuram: @tpantelis Thank you for your quick answer. I will update the code as so.

anfredette (commented):
I re-ran the test following Masaki's instructions above, and it does work. 

I then tested with lighthouse by running `make e2e using=lighthouse,external-net,globalnet` to set up the clusters. I also used [this script](https://github.com/anfredette/kube-scripts/blob/main/create-service-wo-selector.sh) that I used for testing https://github.com/submariner-io/lighthouse/pull/684 to set up the service without selector. This script uses a regular pod as an endpoint instead of an external VM, but I think the Submariner flows are the same in either case. That test worked as well and I confirmed that the services names could be resolved from cluster2. I.e., `curl nginx-svc.default.svc.clusterset.local` worked in this example.

One observation I have is that the code creates an internal service, called submariner-uvhtheozeyskii6xhvh5dxgtbdktupw3, as well as an internal endpoint with the same name, and assigns the external IP address to the internal service.

```
$ kubectl get services
NAME                                          TYPE        CLUSTER-IP     EXTERNAL-IP     PORT(S)   AGE
kubernetes                                    ClusterIP   100.0.0.1      <none>          443/TCP   176m
nginx-svc                                     ClusterIP   100.0.138.96   <none>          80/TCP    161m
submariner-uvhtheozeyskii6xhvh5dxgtbdktupw3   ClusterIP   100.0.11.170   242.254.1.253   80/TCP    161m
$ kubectl get endpoints
NAME                                          ENDPOINTS         AGE
kubernetes                                    172.18.0.4:6443   176m
nginx-svc                                     10.0.128.5:80     161m
submariner-uvhtheozeyskii6xhvh5dxgtbdktupw3   10.0.128.5:80     161m
```

This may be a newbie misunderstanding, but is the internal service necessary? For my manual workaround documented in https://github.com/submariner-io/submariner/issues/1693, I only needed to manually create the internal endpoint to get the datapath to work. If we had 10 endpoints, would we need to create 10 extra services? Would they all get different external IP addresses or the same? Furthermore, why does globalnet need to care about all the endpoints? Wouldn't the local service loadbalancer be responsible for load balancing the sessions across the endpoints?

We can discuss this more during the non-cluster connectivity meeting tomorrow morning. Thanks!

mkimuram:
@anfredette In short, after #1643,
- For non-exported Service, globalnet controller do nothing (as it was),
- For exported Service with selector, globalnet controller creates internal Service, and Endpoints for it is created by k8s,
- For exported Service without selector, globalnet controller creates internal Service, but Endpoints for it isn't created by k8s.

So, this PR is to add codes to automatically create Endpoints for internal Service, if Service is Service without selector (as you documented to manually do in #1693). I know that this explanation won't be enough, so let's discuss in the meeting.

pkg/globalnet/controllers/endpoints_controller.go
mkimuram: @sridhargaddam  I've updated the PR and it now can handle update. PTAL
 
The last commit is needed for deleting cloned Endpoints if original Endpoints is deleted while controller stops, but it avoids controller become running unless RBAC isn't granted (as handled in submariner-io/submariner-operator#1845 , so it needs to be merged, first).

pkg/globalnet/controllers/endpoints_controller_test.go
mkimuram: Test codes are moved to pkg/globalnet/controllers/service_export_controller_test.go and changed in this way.

pkg/globalnet/constants/constants.go
mkimuram: Removed prefix.

pkg/globalnet/controllers/endpoints_controller.go
mkimuram: Add codes at `ensureClonedHasOriginal` in pkg/globalnet/controllers/service_export_endpoints_controllers.go

pkg/globalnet/controllers/endpoints_controller.go
mkimuram: Add codes to handle update and it can handle exporting service before/after endpoints creation.

mkimuram:
@tpantelis @sridhargaddam @vthapar

I've squashed the commit and it is now ready to be reviewed. PTAL

Note that the second commit is needed because ExternalIP feature of k8s doesn't seem to forward packets from pod on the same cluster when accessed via ExternalIP, therefore this access can't be reachable. However, it won't need to be covered, because there is no use case for it (This cluster is just to connect cluster and external, so no user pod will be deployed). 
Cleanup for E2E tests will be done in https://github.com/submariner-io/submariner/pull/1582.

@dfarrell07

Now that submariner-operator#1845 is merged, this PR works well in my environment (without any manual interaction).
However, it seems that CI is skipping the test. Could you check what is happening (is it just taking time)?

dfarrell07:
@mkimuram It just needed the `ready-to-test` label, which I've added. It's added automatically by the bot when there are two +1 reviews, or it's fine to add it manually before then if you need to see a result like in this case.

tpantelis (approved):

anfredette (approved):
LGTM

I've also tested this code successfully using various versions of the following scripts:
https://github.com/anfredette/kube-scripts/blob/main/create-service-wo-selector.sh
https://github.com/anfredette/kube-scripts/blob/main/create-multi-service-wo-selector.sh

sridhargaddam (approved):
Awesome work @mkimuram

anfredette:
FYI, I rebased this pr since @mkimuram is out this week.

sridhargaddam:
Thank you @anfredette

PR #1696, opened 2022-02-16
Track framework.ExecWithOptions changes in shipyard

elevran:
Shipyard's definition changed to accept pointer, not struct (large struct lint error),
reflecting this change into submariner

Signed-off-by: Etai Lev Ran <etail@il.ibm.com>


elevran:
This is broken on tip of devel but not in CI (using tagged versions?).
Closing

PR #1708, opened 2022-02-24
Fix new linter errors from golangci-lint 1.44.2

dfarrell07:
Fixes gofumpt and revive errors that show up when we upgrade from
golangci-lint 1.43.0 to 1.44.2.

### DIFF
--- a/pkg/cable/driver.go
+++ b/pkg/cable/driver.go
@@ -30,7 +30,6 @@ import (
 
 // Driver is used by the ipsec engine to actually connect the tunnels.
 type Driver interface {
-
 	// Init initializes the driver with any state it needs.
 	Init() error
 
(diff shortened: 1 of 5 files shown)


tpantelis (approved):

dfarrell07:
There's actually an even newer version of gofumpt that flags quite a bit more. I can run it instead?

https://pastebin.com/VhiUnPp5
https://github.com/mvdan/gofumpt/releases/tag/v0.3.0

dfarrell07:
After thinking about it more, I think it makes sense to only make the changes the current/latest linter flags. We should keep the standards consistent with the automation, driven by it. We can make the changes gofumpt 0.3 flags when/if golangci-lint pulls it in.

dfarrell07:
This does fix the issues flagged on https://github.com/submariner-io/shipyard/pull/734 (https://github.com/submariner-io/shipyard/runs/5324947094?check_suite_focus=true) and there are not any omitted there, I checked locally with golangci-lint 1.44.2.

skitt (approved):

PR #1710, opened 2022-02-27
Implement Gateway datapath cleanup

### LINKED ISSUE #1636: Implement gateway dataplane cleanup
Pass an UNINSTALL environment variable to initiate cleanup of dataplane components (IP tables etc).

Task for submariner-io/enhancements#50

sridhargaddam:
This PR implements the cleanup for the following cable drivers.
1. Libreswan
2. Wireguard
3. VxLAN

Fixes: https://github.com/submariner-io/submariner/issues/1636

### DIFF
--- a/pkg/netlink/netlink.go
+++ b/pkg/netlink/netlink.go
@@ -29,7 +29,9 @@ import (
 	"syscall"
 
 	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/vishvananda/netlink"
+	"k8s.io/klog"
 )
 
 type Interface interface {
@@ -55,6 +57,10 @@ type Interface interface {
 }
 
 var NewFunc func() Interface
+
+const (
+	allZeroAddress = "0.0.0.0/0"
+)
 
 type netlinkType struct{}
 
@@ -178,7 +184,7 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
 	}
 
 	for i := range routes {
-		if routes[i].Dst == nil || routes[i].Dst.String() == "0.0.0.0/0" {
+		if routes[i].Dst == nil || routes[i].Dst.String() == allZeroAddress {
 			if routes[i].LinkIndex == 0 {
 				return nil, fmt.Errorf("default gateway interface could not be determined")
 			}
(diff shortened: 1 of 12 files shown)


aswinsuryan (approved):

pkg/util/util.go
```diff
@@ -147,3 +152,66 @@ func EnsureValidName(name string) string {
...
+
+		if err = netLink.XfrmPolicyDel(&currentXfrmPolicyList[i]); err != nil {
+			return errors.Wrapf(err, "error deleting XFRM policy %s", currentXfrmPolicyList[i])
+		}
+	}
+
+	return nil
+}
+
+func DeleteVxLANIfaceAlongWithRoutes(iface string, tableID int) error {
```
maryamtahhan: probably better to call this function DeleteIfaceAlongWithRoutes... or I can do it when I push IP in IP...

maryamtahhan (commented):
Looks good to me. I would just genericize the names of functions in pkg/util/util.go to not include any specific cable types as other/new cable types will probably also use these functions.

if a function is truely tied to a cable type then perhaps utils.go is not where it should go?

pkg/util/util.go
sridhargaddam: Yeah, makes sense @maryamtahhan. I'll update the PR. Thanks for reviewing :)

maryamtahhan (approved):

nyechiel:
@skitt can you please review?

pkg/util/util.go
```diff
@@ -147,3 +152,66 @@ func EnsureValidName(name string) string {
 		return c
 	}, name)
 }
+
+func DeleteXfrmRules() error {
```
tpantelis: I suggest we put this in the `netlink` package which is more meaningful than `util`.

pkg/util/util.go
```diff
@@ -147,3 +152,66 @@ func EnsureValidName(name string) string {
...
+
+		if err = netLink.XfrmPolicyDel(&currentXfrmPolicyList[i]); err != nil {
+			return errors.Wrapf(err, "error deleting XFRM policy %s", currentXfrmPolicyList[i])
+		}
+	}
+
+	return nil
+}
+
+func DeleteIfaceAlongWithRoutes(iface string, tableID int) error {
```
tpantelis: Same here. 

```suggestion
func DeleteIfaceAndAssociatedRoutes(iface string, tableID int) error {
```

pkg/util/util.go
```diff
@@ -147,3 +152,66 @@ func EnsureValidName(name string) string {
...
+	link, err := nl.LinkByName(iface)
```
tpantelis: Instead of using the `vishvananda` interface directly, we can use our `netlink.Interface` abstraction here and in the function calls below.

pkg/util/util.go
sridhargaddam: Done

pkg/util/util.go
sridhargaddam: Done

sridhargaddam (commented):

pkg/netlink/netlink.go
```diff
@@ -194,3 +202,64 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
 
 	return nil, fmt.Errorf("unable to find default route")
 }
+
+func (n *netlinkType) DeleteIfaceAndAssociatedRoutes(iface string, tableID int) error {
+	link, err := netlink.LinkByName(iface)
```
tpantelis:
```suggestion
	link, err := n.LinkByName(iface)
```

pkg/netlink/netlink.go
```diff
@@ -194,3 +202,64 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
...
+	if err != nil {
+		if !errors.Is(err, netlink.LinkNotFoundError{}) {
+			klog.Warningf("Failed to retrieve the vxlan-tunnel interface: %v", err)
+		}
+
+		return nil
+	}
+
+	currentRouteList, err := netlink.RouteList(link, syscall.AF_INET)
```
tpantelis:
```suggestion
	currentRouteList, err := n.RouteList(link, syscall.AF_INET)
```

pkg/netlink/netlink.go
```diff
@@ -194,3 +202,64 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
...
+
+	if err != nil {
+		klog.Warningf("Unable to cleanup routes, error retrieving routes on the link %s: %v", iface, err)
+	} else {
+		for i := range currentRouteList {
+			klog.V(log.DEBUG).Infof("Processing route %v", currentRouteList[i])
+			if currentRouteList[i].Table == tableID {
+				if err = netlink.RouteDel(&currentRouteList[i]); err != nil {
```
tpantelis:
```suggestion
				if err = n.RouteDel(&currentRouteList[i]); err != nil {
```

pkg/netlink/netlink.go
```diff
@@ -194,3 +202,64 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
...
+					klog.Errorf("Error removing route %s: %v", currentRouteList[i], err)
+				}
+			}
+		}
+	}
+
+	err = netlink.LinkDel(link)
```
tpantelis:
```suggestion
	err = n.LinkDel(link)
```

pkg/netlink/netlink.go
```diff
@@ -194,3 +202,64 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
 
 	return nil, fmt.Errorf("unable to find default route")
 }
+
+func (n *netlinkType) DeleteIfaceAndAssociatedRoutes(iface string, tableID int) error {
```
tpantelis: Since this function uses existing interface methods, I suggest we put it in the package but not part of the interface so it doesn't have to be specifically implemented by the fake.
 
```suggestion
func DeleteIfaceAndAssociatedRoutes(iface string, tableID int) error {
    n := New()
```

pkg/netlink/netlink.go
```diff
@@ -194,3 +202,64 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
...
+
+	err = netlink.LinkDel(link)
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete the vxlan interface")
+	}
+
+	return nil
+}
+
+func (n *netlinkType) DeleteXfrmRules() error {
```
tpantelis: Same here.

```suggestion
func DeleteXfrmRules() error {
    n := New()
```

tpantelis (approved):

PR #1717, opened 2022-03-03
Update vxlan cable to use common cable functions

maryamtahhan:
Update the vxlan cable driver to use common functions shared with the IP in IP cable driver

Depends on https://github.com/submariner-io/submariner/pull/1716

### DIFF
--- a/pkg/cable/vxlan/vxlan.go
+++ b/pkg/cable/vxlan/vxlan.go
@@ -22,8 +22,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"strconv"
-	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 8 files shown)


pkg/cable/vxlan/vxlan.go
```diff
@@ -218,55 +217,9 @@ func createvxlanIface(iface *vxlanIface) error {
 	return nil
 }
 
-func isVxlanConfigTheSame(newLink, currentLink netlink.Link) bool {
-	required := newLink.(*netlink.Vxlan)
-	existing := currentLink.(*netlink.Vxlan)
-
-	if required.VxlanId != existing.VxlanId {
```
maryamtahhan: Add this back in as it has an additional check

maryamtahhan: in fact this function has significantly deviated from IP in IP - so remove the common function from cableutils and add back in the vxlan function check

maryamtahhan:
Will close this till the IP in IP changes are approved/merged - so as to avoid reviewing the same code twice

PR #1721, opened 2022-03-08
Delete iptable rules when remote ep is deleted

### LINKED ISSUE #1720: Relevant postrouting iptables rules on cluster1 not being deleted after subctl uninstall cluster2
**What happened**:
Postrouting iptables rules on cluster1 allowing traffic to cluster2 subnets not removed 
 after running  'subctl uninstall Cluster2'

Relevant iptables rules should be deleted

1. Run Kind deployment (https://submariner.io/getting-started/quickstart/kind/), where cluster1 is also the broker
2. Run 'subctl uninstall cluster2'
3. Verify that cluster2 EP removed from cluster1

>> $ kubectl get endpoints.submariner.io -A 
>>NAMESPACE               NAME                                            AGE
submariner-k8s-broker   cluster1-submariner-cable-cluster1-172-20-0-4   20m
submariner-operator     cluster1-submariner-cable-cluster1-172-20-0-4   20m

4. check iptables (nat  table) on cluster1 GW node, we can still see rules that accept traffic from cluster2 subnets (100.2.0.0/16 and 10.2.0.0/16)

>> $iptables -t nat -L SUBMARINER-POSTROUTING
>> Warning: iptables-legacy tables present, use iptables-legacy to see them
SNAT       all  --  240.0.0.0/8          anywhere             to:10.1.128.0
ACCEPT     all  --  10.1.0.0/16          100.2.0.0/16        
ACCEPT     all  --  100.2.0.0/16         10.1.0.0/16         
ACCEPT     all  --  10.1.0.0/16          10.2.0.0/16         
ACCEPT     all  --  10.2.0.0/16          10.1.0.0/16         

[subcatl_gather.gz](https://github.com/submariner-io/submariner/files/8198844/subcatl_gather.gz)

sridhargaddam:
When a remote endpoint is deleted, route agent gets notification
from the event handler but the current code was not handling
this event. This PR fixes it.

Fixes: https://github.com/submariner-io/submariner/issues/1720

### DIFF
--- a/pkg/routeagent_driver/handlers/kubeproxy/iptables_iface.go
+++ b/pkg/routeagent_driver/handlers/kubeproxy/iptables_iface.go
@@ -83,38 +83,47 @@ func (kp *SyncHandler) createIPTableChains() error {
 
 func (kp *SyncHandler) updateIptableRulesForInterClusterTraffic(inputCidrBlocks []string, operation Operation) {
 	for _, inputCidrBlock := range inputCidrBlocks {
-		if operation == Add {
-			err := kp.programIptableRulesForInterClusterTraffic(inputCidrBlock)
-			if err != nil {
-				klog.Errorf("Failed to program iptable rule. %v", err)
-			}
-		} else if operation == Delete {
-			// TODO: Handle this use-case
-			klog.Warning("Handle the delete use-case")
+		err := kp.programIptableRulesForInterClusterTraffic(inputCidrBlock, operation)
+		if err != nil {
+			klog.Errorf("Failed to program iptable rules. %v", err)
 		}
 	}
 }
 
-func (kp *SyncHandler) programIptableRulesForInterClusterTraffic(remoteCidrBlock string) error {
+func (kp *SyncHandler) programIptableRulesForInterClusterTraffic(remoteCidrBlock string, operation Operation) error {
 	ipt, err := iptables.New()
 	if err != nil {
 		return errors.Wrap(err, "error initializing iptables")
 	}
 
 	for _, localClusterCidr := range kp.localClusterCidr {
-		ruleSpec := []string{"-s", localClusterCidr, "-d", remoteCidrBlock, "-j", "ACCEPT"}
-		klog.V(log.DEBUG).Infof("Installing iptables rule for outgoing traffic: %s", strings.Join(ruleSpec, " "))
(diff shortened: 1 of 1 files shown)


pkg/routeagent_driver/handlers/kubeproxy/iptables_iface.go
```diff
@@ -83,38 +83,47 @@ func (kp *SyncHandler) createIPTableChains() error {
...
 	for _, inputCidrBlock := range inputCidrBlocks {
-		if operation == Add {
-			err := kp.programIptableRulesForInterClusterTraffic(inputCidrBlock)
-			if err != nil {
-				klog.Errorf("Failed to program iptable rule. %v", err)
-			}
-		} else if operation == Delete {
-			// TODO: Handle this use-case
-			klog.Warning("Handle the delete use-case")
+		err := kp.programIptableRulesForInterClusterTraffic(inputCidrBlock, operation)
```
vthapar, line 86: Why not retain the `if operation` and write a new method for `cleanupIpTableRulesforInterClusterTraffic`. It will simplify code and make it more readable. Other than ipt.New() nothing else is common between Add/Delete cases.

sridhargaddam: No, its not just ipt.New(), the ruleSpec and the flow will be more or less the same. 
In fact, I tried that initially, but linter was complaining that code is duplicated. So modified it to make linter happy.

yboaron: I checked it locally and it fixed the issue
Maybe we can keep the if operation and just set function pointer to ipt.AppendUnique or ipt.Delete

yboaron: In second thought, if we don't plan to add new logic to intercluster ipt rules,  I think the suggested change is OK

yboaron (approved):

skitt (approved):

pkg/routeagent_driver/handlers/kubeproxy/iptables_iface.go
vthapar: Good to know that you did try that way initially.

vthapar (approved):

sridhargaddam:
Backport PR: https://github.com/submariner-io/submariner/pull/1723

Issue #1740, opened 2022-03-17
Use minimum GITHUB_TOKEN permissions for each job


dfarrell07:
The `GITHUB_TOKEN` used by some GitHub Actions can be configured to [`permissive` or `restricted`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token). We currently use `permissive` as a GitHub Organizational default. We should instead use `restricted` as the GitHub Organizational default and explicitly grant the minimum required permissions to each job/workflow.

> You can use the permissions key in your workflow file to modify permissions for the GITHUB_TOKEN for an entire workflow or for individual jobs. This allows you to configure the minimum required permissions for a workflow or job. When the permissions key is used, all unspecified permissions are set to no access, with the exception of the metadata scope, which always gets read access.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

Example per-workflow: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input

Example per-job: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api

Depends on https://github.com/submariner-io/submariner/pull/2042
Depends on https://github.com/submariner-io/submariner/pull/2054
Depends on https://github.com/submariner-io/submariner/issues/1799
Depends on https://github.com/submariner-io/admiral/pull/438
Depends on https://github.com/submariner-io/cloud-prepare/pull/424
Depends on https://github.com/submariner-io/lighthouse/pull/913
Depends on https://github.com/submariner-io/releases/pull/515
Depends on https://github.com/submariner-io/shipyard/pull/994
Depends on https://github.com/submariner-io/subctl/pull/310
Depends on https://github.com/submariner-io/submariner-charts/pull/284
Depends on https://github.com/submariner-io/submariner-operator/pull/2292
Depends on https://github.com/submariner-io/submariner-website/pull/839
Depends on https://github.com/submariner-io/enhancements/pull/140

dfarrell07:
Better docs: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

dfarrell07:
It's weird this doesn't fail: https://github.com/submariner-io/lighthouse/pull/821

Maybe the permissions changes don't take effect on PRs to prevent some attacks? Maybe they have to be merged?

dfarrell07:
[...]

PR #1751, opened 2022-03-23
Use Dependabot to update GHAs, now supports SHAs

dfarrell07:
This is a bit of a test of Dependabot's support for managing SHA-based
versions of GitHub Actions.

In the original press release about Dependabot's support for GHAs,
they say:

> Dependabot creates pull requests that update the action to the latest
released tag (e.g., v2), regardless of if you’re currently on a release
tag, a pre-release tag, or a specific hash.

github.blog/2020-06-25-dependabot-now-updates-your-actions-workflows/

But I think SHA-based updates were added as a part of:

https://github.com/dependabot/dependabot-core/issues/2835

I see examples where Dependabot apparently works with SHAs:

https://github.com/systemd/systemd/pull/22638/files
https://github.com/google/go-github/pull/2049/files

And those projects are using config like is proposed here:

https://github.com/systemd/systemd/blob/main/.github/dependabot.yml#L5
https://github.com/google/go-github/blob/master/.github/dependabot.yml#L15

As always, we should be careful about taking updates from third parties
and review theses PRs to update GHAs carefully. Especially until #1740
reduces the permissions of each job to the minimum required.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,10 @@
 ---
 version: 2
 updates:
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: daily
   - package-ecosystem: gomod
     directory: "/"
     schedule:


skitt (approved):

dfarrell07:
None of the checks are starting :eyes:

dfarrell07:
> None of the checks are starting eyes

Yeah GitHub is having issues. https://www.githubstatus.com/

tpantelis (approved):

PR #1752, opened 2022-03-23
multi-active-gw-POC

astoycos:
Initial round of commits for MAG 

- Setup tunnel infrastructure between local worker nodes and multiple GW nodes 
- Setup tunnel infrastructure between multiple remote GWs 
- Setups/reconciles multipath routes between local worker nodes and multiple GW nodes 
- Setups/reconciles multipath routing between multiple remote GWs 
- Removes some checks that ensure a single GW per cluster 
- Billy's changes to add subctl `--multi-active-gateway` option


pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -195,16 +195,18 @@ func (g *gatewayMonitor) handleCreatedOrUpdatedEndpoint(obj runtime.Object, numR
...
-		klog.V(log.DEBUG).Infof("Transitioned to non-gateway node with endpoint private IP %s", endpoint.Spec.PrivateIP)
-
-		g.syncMutex.Lock()
-		if g.isGatewayNode {
-			g.stopControllers()
-			g.isGatewayNode = false
-		}
-		g.syncMutex.Unlock()
 	}
+	// Only stop controllers on endpoint deletion since we can have multiple gateways
```
astoycos, line 199: @Billy99 This is the only bit I changed in the globalnet controller

pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go
```diff
@@ -153,161 +154,123 @@ func (kp *SyncHandler) configureRoute(remoteSubnet string, operation Operation,
...
-	return nil
+	return desiredRoutes, nil
 }
 
-func (kp *SyncHandler) removeUnknownRoutes(vxlanGw net.IP, currentRouteList []netlink.Route) {
-	for i := range currentRouteList {
-		// Contains(endpoint destinations, route destination string, and the route gateway is our actual destination.
-		klog.V(log.DEBUG).Infof("Processing route %v", currentRouteList[i])
+// Reconcile the routes based on the GWIPs installed on this device using rtnetlink.
+func (kp *SyncHandler) reconcileIntraClusterRoutes() error {
```
anfredette, line 220: With this pr, all changes result in a call to reconcileIntraClusterRoutes(), which compares the existing routes with the desired routes, and then adds/deletes as needed.  This is as opposed to just adding and deleting individual routes based on specific events when possible.  There are pros and cons to this approach.  The big pro is that it is simpler and more resilient.  We maintain the desired state.  Then, when anything changes, we just make sure that the desired state matches the actual state. If anything was missed in the interim, the routes recover.  The main con is that all the routes need to be checked every time anything changes, so there will be more processing overhead.

This works, and I think the simplicity/resiliency benefits make it worth trying.  If this proves to be a scaling bottleneck, then it can be optimized later.

@sridhargaddam, @aswinsuryan, and others, I'd like to get your feedback on this approach.

pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go
```diff
@@ -153,161 +154,123 @@ func (kp *SyncHandler) configureRoute(remoteSubnet string, operation Operation,
...
 	}
 
-	// Update internal Loadbalancing Groups
-	for _, gwIp := range kp.gwIPs.Elements() {
-		vxlanGwIP, err := getVxlanVtepIPAddress(gwIp)
-		if err != nil {
-			return errors.Wrap(err, "failed to derive the remoteVtepIP")
+	for key, route := range currentRouteList {
+		// route exists and is up to date keep it
+		if _, ok := desiredRouteList[key]; ok {
```
anfredette, line 246: The way this lookup works, if a route exists, but we're adding or removing a nexthop gateway to the Multipath field, the route won't match, and we'll end up deleting and recreating the route.  It's probably more efficient and less disruptive to do a RouteReplace instead.

anfredette, line 246: This code is not doing a full route comparison between the existing route and the desired route, so it's possible that the values in the key match, but something else doesn't.  It may be reasonable to assume that the route agent is the only one creating these routes, and they are only created one way, so if the key values match, the other info must match as well.  If so, we might not want to go through the effort to build the full desired route list until we've determined which routes are missing.  Otherwise, we should do a full comparison and replace the route if necessary.

anfredette (commented):
I've tested this code and it works for me. I still need to complete reviewing this pr, but I wanted to post a few comments to solicit feedback.

So far I support merging it as-is to facilitate progress on the POC while we consider possible optimizations, but I will hold off on approving it until I look through the rest of it.  Thanks!

Issue #1749, opened 2022-03-23
Error while appending rules to iptable chains


sridhargaddam:
In an OCP deployment with Submariner version 0.12, it was seen that after we create a chain, its taking few milliseconds for the chain to be actually programmed on the node. From the Globalnet pod logs...

```
+ trap 'exit 1' SIGTERM SIGINT
+ SUBMARINER_VERBOSITY=2
+ '[' '' == true ']'
+ DEBUG=-v=2
+ exec submariner-globalnet -v=2 -alsologtostderr
W0322 08:45:15.847390       1 client_config.go:608] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0322 08:45:15.852666       1 main.go:76] Starting submariner-globalnet{acm-nmanos-devcluster-a2-aws submariner-operator [] false}
I0322 08:45:23.452615       1 request.go:645] Throttling request took 1.148508792s, request: GET:https://172.30.0.1:443/apis/whereabouts.cni.cncf.io/v1alpha1?timeout=32s
I0322 08:45:26.457138       1 gateway_monitor.go:117] Starting GatewayMonitor to monitor the active Gateway node in the cluster.
[...]

PR #1765, opened 2022-03-30
Optimize code that looks for iptable chain presence

sridhargaddam:
Instead of listing all the existing chains and checking
if the required chain is present or not, we can directly
use the iptables library API, chainExists, which is more
optimal way to verify for the chain presence.

Related to: https://github.com/submariner-io/submariner/issues/1749

### DIFF
--- a/pkg/iptables/fake/iptables.go
+++ b/pkg/iptables/fake/iptables.go
@@ -156,6 +156,18 @@ func (i *IPTables) ClearChain(table, chain string) error {
 	return nil
 }
 
+func (i *IPTables) ChainExists(table, chain string) (bool, error) {
+	i.mutex.Lock()
+	defer i.mutex.Unlock()
+
+	chainSet := i.tableChains[table]
+	if chainSet != nil {
+		return chainSet.Contains(chain), nil
+	}
+
+	return false, nil
+}
+
 func (i *IPTables) AddChainsFor(table string, chains ...string) {
 	i.mutex.Lock()
 	defer i.mutex.Unlock()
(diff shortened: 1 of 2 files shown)


skitt:
This is great, it just needs a `fake` implementation as you’ve no doubt noticed ;-).

tpantelis (approved):

pkg/iptables/fake/iptables.go
```diff
@@ -156,6 +156,18 @@ func (i *IPTables) ClearChain(table, chain string) error {
 	return nil
 }
 
+func (i *IPTables) ChainExists(table, chain string) (bool, error) {
```
yboaron, line 159: Not sure if the error return value is needed for  ChainExists function, cause ChainExists always returns error ==nil, but maybe we should keep this way for consistency

pkg/iptables/iptables.go
```diff
@@ -73,16 +74,13 @@ func (i *iptablesWrapper) Delete(table, chain string, rulespec ...string) error
...
+		return nil
 	}
 
-	for _, val := range existingChains {
-		if val == chain {
-			// Chain already exists
-			return nil
-		}
+	if err != nil {
+		return errors.Wrap(err, "error listing IP table chains")
```
yboaron: maybe "error finding IP table chain" ? we don't list anymore

pkg/iptables/fake/iptables.go
sridhargaddam: This is a fake implementation, so we have to keep the API in sync for consistency.

yboaron (approved):

sridhargaddam:
Backport PR: https://github.com/submariner-io/submariner/pull/1785

PR #1768, opened 2022-04-01
Move external network E2E to main E2E matrix, add to flake finder

dfarrell07:

### DIFF
--- a/.github/workflows/e2e-full.yml
+++ b/.github/workflows/e2e-full.yml
@@ -15,6 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         deploytool: ['operator']
+        external_net: ['']
         globalnet: ['', 'globalnet']
         cable_driver: ['libreswan', 'wireguard', 'vxlan']
         ovn: ['', 'ovn']
@@ -23,35 +24,27 @@ jobs:
             globalnet: 'globalnet'
           - ovn: 'ovn'
             cable_driver: 'wireguard'
+        include:
+          - external_net: 'external-net'
+          - external_net: 'external-net'
+            globalnet: 'globalnet'
     steps:
       - name: Check out the repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
 
       - name: Run E2E deployment and tests
+        if: matrix.external_net != 'external-net'
         uses: submariner-io/shipyard/gh-actions/e2e@devel
         with:
           using: ${{ matrix.cable_driver }} ${{ matrix.deploytool }} ${{ matrix.globalnet }} ${{ matrix.ovn }}
 
-      - name: Post mortem
-        if: failure()
-        uses: submariner-io/shipyard/gh-actions/post-mortem@devel
-
-  e2e-external:
-    name: External Network E2E
-    if: contains(github.event.pull_request.labels.*.name, 'ready-to-test')
-    timeout-minutes: 30
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        external_net: ['external-net']
-        globalnet: ['', 'globalnet']
-    steps:
-      - name: Check out the repository
-        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-
-      - name: Run E2E deployment and tests
+      - name: Run External Network E2E deployment and tests
+        if: matrix.external_net == 'external-net'
         uses: submariner-io/shipyard/gh-actions/e2e@devel
         with:
           using: ${{ matrix.external_net }} ${{ matrix.globalnet }}
-          skip: ""
+          skip: ""  # Override skipping external network tests
+
+      - name: Post mortem
+        if: failure()
+        uses: submariner-io/shipyard/gh-actions/post-mortem@devel
--- a/.github/workflows/flake_finder.yml
+++ b/.github/workflows/flake_finder.yml
@@ -15,16 +15,29 @@ jobs:
       fail-fast: false
       matrix:
         deploytool: ['operator']
+        external_net: ['']
         globalnet: ['', 'globalnet']
         cable_driver: ['libreswan', 'wireguard']
+        include:
+          - external_net: 'external-net'
+          - external_net: 'external-net'
+            globalnet: 'globalnet'
     steps:
       - name: Check out the repository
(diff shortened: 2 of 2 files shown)


.github/workflows/e2e-full.yml
```diff
@@ -15,6 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         deploytool: ['operator']
+        external_net: ['']
```
dfarrell07, line 18: Before anyone asks, yes it is required to include this default `external_net: ['']`. I first sent exactly this patch without those lines and no external net E2E jobs were defined.

dfarrell07:
This will require an update to the required jobs config in GH.

.github/workflows/e2e-full.yml
```diff
@@ -23,35 +24,27 @@ jobs:
             globalnet: 'globalnet'
           - ovn: 'ovn'
             cable_driver: 'wireguard'
+        include:
+          - external_net: 'external-net'
+          - external_net: 'external-net'
+            globalnet: 'globalnet'
```
mkimuram, line 30: As I read https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs#example-including-new-combinations , I understand that only the combinations of ('external-net') and ('external-net', 'globalnet') is added by this include lines. And these are currently defined in e2e-external job, and are merged to this e2e job. Looks good to me.

mkimuram:
Also confirmed that proper external network tests passed in "End to End Full". And the flake finder is defined in the same way, so it should work well (I guess that we need to merge to check if it actually works well in https://github.com/submariner-io/submariner/actions/workflows/flake_finder.yml).

- `End to End Full / E2E (external-net) (pull_request) Successful in 13m`
https://github.com/submariner-io/submariner/runs/5792701103?check_suite_focus=true

```
Ran 6 of 26 Specs in 72.015 seconds
2821
SUCCESS! -- 6 Passed | 0 Failed | 7 Pending | 13 Skipped
```

- `End to End Full / E2E (external-net, globalnet) (pull_request) Successful in 14m`
https://github.com/submariner-io/submariner/runs/5792701164?check_suite_focus=true

```
Ran 13 of 26 Specs in 259.893 seconds
3094
SUCCESS! -- 13 Passed | 0 Failed | 7 Pending | 6 Skipped
```

@dfarrell07 It seems that below job is still shown. Is it something that will be disappear once this PR is merged?

- `External Network E2E (external-net, globalnet) Expected — Waiting for status to be reported`

Once it becomes clear, I will approve. Anyway, thank you very much for working on this to improve!

.github/workflows/e2e-full.yml
mkolesnik: This seems to be a quirk in GHA though on their documentation it seems like it should be working.

Perhaps open a bug for this and link to the two runs (or maybe such a bug is already open)?

mkolesnik (commented):
> This will require an update to the required jobs config in GH.

I remember leaving this comment but can't find it (maybe another such patch?) so:

I think the best course of action is to place all experimental features (this, wireguard, etc) in a separate `e2e-experimental.yaml` or such file.
They could still trigger from `ready-to-test` label but it would be much easier to differentiate (especially for branch protections) which jobs are required for merging and which are experimental features (sometimes more flaky) which aren't mandatory for merging a patch.

dfarrell07:
> I think the best course of action is to place all experimental features (this, wireguard, etc) in a separate e2e-experimental.yaml or such file.

I don't really want to get in to the business of defining what's stable and what's not by moving jobs between E2E workflows. I think that would result in lots more required jobs updates. Both External E2E and Wireguard were required jobs. It's not clear to me that Wireguard is more experimental than VXLAN. Let's leave discussions of stability for explicit docs, not kinda-arbitrary workflow splits.

That's also basically the inverse of the goal of this patch - to avoid bugs and reduce maintenance by de-duplicating. This adds a post-mortem and makes the timeout constant, but in the future the two workflows might diverge in different ways. If we go back to splitting the on-PR workflows, we'd also probably want to keep the on-timer workflows split for kinda-consistent maintenance. 

So if you want that change @mkolesnik you can send a PR and we can discuss, but that's not what I propose.

dfarrell07:
> It seems that below job is still shown. Is it something that will be disappear once this PR is merged?

@mkimuram That's GitHub configuration for the jobs required to pass. Since this job was renamed it's still looking for the old job.
An Owner will need to update that configuration when this is merged.

mkolesnik:
Fine by me to better discuss and define it, but we should put a pin in this PR until it's decided since this change will make external E2E required for any PR

dfarrell07:
> this change will make external E2E required for any PR

I'm not sure I understand what you mean. As you can see, it's not currently required and that shouldn't change after merge. It will run on every PR yes, but I think that's what we want and currently have.

Maybe you're looking at the old job, which was required?

.github/workflows/e2e-full.yml
```diff
@@ -15,6 +15,7 @@ jobs:
       fail-fast: false
```
sridhargaddam, line 15: Do we have to increase the timeout-minutes as we are now including this new external_net jobs?

sridhargaddam:
> > This will require an update to the required jobs config in GH.
> 
> I remember leaving this comment but can't find it (maybe another such patch?) so:
> 
> I think the best course of action is to place all experimental features (this, wireguard, etc) in a separate `e2e-experimental.yaml` or such file. They could still trigger from `ready-to-test` label but it would be much easier to differentiate (especially for branch protections) which jobs are required for merging and which are experimental features (sometimes more flaky) which aren't mandatory for merging a patch.

I'm fine either way, as long as we get some feedback about stability of external_net jobs using the flake finder. Please note that for supporting external_network use-cases there are no explicit changes in Submariner. The changes are mostly outside of Submariner, but validating the use-case regularly will allow us to know early if the external-net use-case is broken due to internal changes in Submariner.

.github/workflows/e2e-full.yml
dfarrell07: The External Net job was running with 30m timeout before, this changes it to use the same 45m timeout the other E2E jobs use.

dfarrell07:
@mkolesnik Please re-review

mkolesnik (approved):

skitt (approved):

PR #1767, opened 2022-04-01
remove redundant parentheses

jinglina:
remove redundant parentheses

PR #1770, opened 2022-04-04
Remove unnecessary parentheses

skitt:
This is an automatic change produced with

     gofmt -r '(a) -> a' -w $(find pkg -name '*.go')

Depends on #1767 


tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
Do we want to add a "linter" to run this check? It could just be a make target with that command in Shipyard. We could just raise an issue for now, I can add it if we decide we want it.

skitt:
> Do we want to add a "linter" to run this check? It could just be a make target with that command in Shipyard. We could just raise an issue for now, I can add it if we decide we want it.

Yup, that would make sense. Perhaps there’s a linter in golangci-lint which can take care of this...

PR #1772, opened 2022-04-05
Build x86 and ARM images, using crossarch

skitt:

Depends on https://github.com/submariner-io/shipyard/pull/829

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -1,0 +1,19 @@
+---
+name: Multi-arch Builds
+
+on:
+  pull_request:
+
+jobs:
+  check-multiarch:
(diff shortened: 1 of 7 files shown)


skitt:
The UPX logs in the multi-arch job show that this is building amd64 and arm64 binaries and images.

.github/workflows/multiarch.yml
```diff
@@ -0,0 +1,19 @@
+---
+name: Multi-arch builds
```
dfarrell07:
```suggestion
name: Multi-arch Builds
```

Hate to be so nitty, but the consistent UI is nice.

dfarrell07 (approved):

tpantelis (approved):

Issue #1780, opened 2022-04-08
Remove globalIP annotation from node as part of uninstall operation


sridhargaddam:
What happened:
Globalnet adds a globalIP annotation to the Gateway node which is used as healthcheck-ip.
Ideally this annotation should be removed as part of `subctl uninstall ....` operation, otherwise it could create issues during re-deployment.

How to reproduce it (as minimally and precisely as possible):
Check nodes annotations on cluster2
Run Submariner deployment

bin/subctl deploy-broker --kubeconfig output/kubeconfigs/kind-config-cluster1 --globalnet
bin/subctl join --kubeconfig output/kubeconfigs/kind-config-cluster1 broker-info.subm --clusterid cluster1 --natt=false
bin/subctl join --kubeconfig output/kubeconfigs/kind-config-cluster2 broker-info.subm --clusterid cluster2 --natt=false
Uninstall Submariner from cluster2

bin/subctl uninstall --kubeconfig output/kubeconfigs/kind-config-cluster2
Check node annotations on cluster2 gateway node

Environment:
Kind

sridhargaddam:
Leaving the annotation on the node could create issues when submariner is re-installed on the cluster.

sridhargaddam:
Logs from one of the clusters which initially had a globalCIDR of 242.2.0.0/16 and after re-deployment of Submariner a different globalCIDR of 242.1.0.0/16 was allocated.

```
I0408 06:15:14.464203       1 cableengine.go:102] CableEngine controller started, driver: "libreswan"
I0408 06:15:14.464368       1 datastoresyncer.go:68] Starting the datastore syncer
I0408 06:15:14.464614       1 tunnel.go:38] Starting the tunnel controller
I0408 06:15:14.565274       1 healthchecker.go:108] CableEngine HealthChecker started with PingInterval: 1, MaxPacketLossCount: 5
I0408 06:15:14.732686       1 datastoresyncer.go:214] Ensuring we are the only endpoint active for this cluster
I0408 06:15:14.733322       1 datastoresyncer.go:293] Creating local submariner Cluster: types.SubmarinerCluster{ID:"cluster3", Spec:v1.ClusterSpec{ClusterID:"cluster3", ColorCodes:[]string{"blue"}, ServiceCIDR:[]string{"10.43.0.0/16"}, ClusterCIDR:[]string{"10.42.0.0/24"}, GlobalCIDR:[]string{"242.1.0.0/16"}}} 
[...]

PR #1781, opened 2022-04-08
Validate globalIP on node with GlobalCIDR of cluster

sridhargaddam:
This PR validates the globalIP present on the gateway node with
the GlobalCIDR allocated to the Cluster before updating the local
endpoint health-check ip.

Related to: https://github.com/submariner-io/submariner/issues/1780

### DIFF
--- a/pkg/controllers/datastoresyncer/node_handler.go
+++ b/pkg/controllers/datastoresyncer/node_handler.go
@@ -19,6 +19,8 @@ limitations under the License.
 package datastoresyncer
 
 import (
+	"net"
+
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/submariner/pkg/globalnet/constants"
 	k8sv1 "k8s.io/api/core/v1"
@@ -35,7 +37,21 @@ func (d *DatastoreSyncer) handleCreateOrUpdateNode(obj runtime.Object, numRequeu
 
 	globalIPOfNode := node.GetAnnotations()[constants.SmGlobalIP]
 
-	return d.updateLocalEndpointIfNecessary(globalIPOfNode)
+	// Validate that globalIPOfNode falls in the globalCIDR allocated to the cluster.
+	if globalIPOfNode != "" {
+		_, ipnet, err := net.ParseCIDR(d.localCluster.Spec.GlobalCIDR[0])
+		if err != nil {
+			// Ideally this will not happen as globalCIDR is expected to be a valid CIDR.
+			klog.Errorf("Error parsing the GlobalCIDR %q: %v", d.localCluster.Spec.GlobalCIDR, err)
+			return false
+		}
+
+		if ipnet.Contains(net.ParseIP(globalIPOfNode)) {
+			return d.updateLocalEndpointIfNecessary(globalIPOfNode)
+		}
+	}
+
+	return false
 }
 
 func (d *DatastoreSyncer) areNodesEquivalent(obj1, obj2 *unstructured.Unstructured) bool {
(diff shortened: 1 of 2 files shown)


pkg/controllers/datastoresyncer/node_handler.go
```diff
@@ -35,7 +37,21 @@ func (d *DatastoreSyncer) handleCreateOrUpdateNode(obj runtime.Object, numRequeu
...
+		_, ipnet, err := net.ParseCIDR(d.localCluster.Spec.GlobalCIDR[0])
+		if err != nil {
+			// Ideally this will not happen as globalCIDR is expected to be a valid CIDR.
+			klog.Errorf("Error parsing the GlobalCIDR %q: %v", d.localCluster.Spec.GlobalCIDR, err)
+			return false
+		}
+
+		if ipnet.Contains(net.ParseIP(globalIPOfNode)) {
+			return d.updateLocalEndpointIfNecessary(globalIPOfNode)
+		}
```
tpantelis, line 51: What would be the scenario where the global IP wouldn't be valid wrt to the CIDR? Perhaps we should log an error or warning?

pkg/controllers/datastoresyncer/node_handler.go
```diff
@@ -35,7 +37,21 @@ func (d *DatastoreSyncer) handleCreateOrUpdateNode(obj runtime.Object, numRequeu
 
 	globalIPOfNode := node.GetAnnotations()[constants.SmGlobalIP]
 
-	return d.updateLocalEndpointIfNecessary(globalIPOfNode)
+	// Validate that globalIPOfNode falls in the globalCIDR allocated to the cluster.
+	if globalIPOfNode != "" {
+		_, ipnet, err := net.ParseCIDR(d.localCluster.Spec.GlobalCIDR[0])
```
tpantelis, line 42: Should we assume the array is  non-empty? Also it seems to assume there's only one entry -  what if there is more than one?

tpantelis (changes requested):
Can you explain the problem/scenario that necessitated this validation?

pkg/controllers/datastoresyncer/node_handler.go
sridhargaddam: >Should we assume the array is non-empty? 

This code runs only when Globalnet is enabled. See here - https://github.com/submariner-io/submariner/blob/54fcb0a46d4acfc9d34b1c1f022b5b83fe6316e3/pkg/controllers/datastoresyncer/datastoresyncer.go#L94

>Also it seems to assume there's only one entry - what if there is more than one?

Currently we don't support multiple GlobalCIDRs and the implementation considers only one GlobalCIDR. You can look at the IPAM code for reference.

pkg/controllers/datastoresyncer/node_handler.go
sridhargaddam: Basically I noticed this issue while debugging an issue where the same cluster was used for multiple re-deployments. Globalnet adds a globalIP annotation to the Gateway node which is used as healthcheck-ip.
Ideally this annotation should be removed as part of `subctl uninstall ....` operation, but it was not (will push a separate PR to fix it), so its creating issue during re-deployment when a different GlobalCIDR was allocated to the same cluster. 

Please see - https://github.com/submariner-io/submariner/issues/1780#issuecomment-1093293544

sridhargaddam:
> Can you explain the problem/scenario that necessitated this validation?

Please see the comments and the associated bug report.

tpantelis (approved):

aswinsuryan (approved):

sridhargaddam:
Backport PR: https://github.com/submariner-io/submariner/pull/1810

PR #1805, opened 2022-04-24
Verify local cidr ipset is updated with local PodCIDRs

### LINKED ISSUE #1774: MTU issues seen between an OnPrem cluster vs Public cloud
While running e2e tests in an OCP Cluster where one of the cluster is on OpenStack (OnPrem) and the other is on Public cloud (AWS), it was seen that the initial connection goes through, but there are errors during data transmission. 

```
      Expected
          <string>: listening on 0.0.0.0:1234 ...
          connect to 10.231.0.98:1234 from 242.0.0.4:46183 (242.0.0.4:46183)
          
      to contain substring
          <string>: b5539269-607a-4893-bd05-3981b1a7d985
```
As it can be seen above, the client was able to connect to the service but the data transmission failed. Submariner e2e tests validate if the connection is successfully established and it also verifies if the data transmission includes the unique UUID from the client, hence the e2e tests are failing.

Setup details:
-------------------
OpenShift version: 4.10.2
Cloud platform: OnPrem (OSP) vs Public Cloud (AWS)
CNI: OpenShift SDN
Submariner Globalnet Enabled
Submariner version: 0.12.0

Additional notes:
-----------------------
So it is clear that its an MTU related issue and on debugging it further, it turns out that its because of iptables vs nft differences.
Submariner MTU driver in route-agent programs iptable rules to clamp tcp mss to PMTU in the Mangle table. Since the underlying platform is running nftables, the automatic translation layer should ideally translate them to corresponding nftable rules, but it appears like this conversion is not happening properly.

IPTable rules programmed on the Gateway node (mangle table):
```
TCPMSS     tcp  --  0.0.0.0/0            0.0.0.0/0            match-set SUBMARINER-LOCALCIDRS src match-set SUBMARINER-REMOTECIDRS dst tcp flags:0x06/0x02 TCPMSS clamp to PMTU
TCPMSS     tcp  --  0.0.0.0/0            0.0.0.0/0            match-set SUBMARINER-REMOTECIDRS src match-set SUBMARINER-LOCALCIDRS dst tcp flags:0x06/0x02 TCPMSS clamp to PMTU
```

Translated nft rules dumped using the nft binary (`nft -a list table mangle`) on the same node:
```
chain SUBMARINER-POSTROUTING {
        meta l4proto tcp # match-set SUBMARINER-LOCALCIDRS src # match-set SUBMARINER-REMOTECIDRS dst tcp flags & (syn|rst) == syn counter packets 114 bytes 6840 tcp option maxseg size set rt mtu
        meta l4proto tcp # match-set SUBMARINER-REMOTECIDRS src # match-set SUBMARINER-LOCALCIDRS dst tcp flags & (syn|rst) == syn counter packets 42 bytes 2520 tcp option maxseg size set rt mtu
}
```

If we look at the nft rules carefully, the `....# match-set SUBMARINER-REMOTECIDRS...` is commented out as the iptables mtu target/expressions are not supported by the automatic translation. The problem is described further in this issue - https://unix.stackexchange.com/questions/672742/why-mss-clamping-in-iptables-nft-seems-to-take-no-effect-in-nftables

yboaron:
Verify that the SUBMARINER_LOCALCIDRS ipset is updated with the local
PodCIDRs because SNAT (with GN IP) is performed in NAT Postrouting table
which is hit after the MANGLE POSTROUTING table.

Partially Fixes: https://github.com/submariner-io/submariner/issues/1774


skitt (approved):

tpantelis (approved):

sridhargaddam:
@yboaron Can you please link this PR to the issue in the commit message - https://github.com/submariner-io/submariner/issues/1774#issuecomment-1091584458

i.e., Partially Fixes https://github.com/submariner-io/submariner/issues/1774#issuecomment-1091584458

mkolesnik:
Backported in  #1821

PR #1809, opened 2022-04-28

yboaron:
Submariner MTU driver in route-agent programs iptable rules to
clamp tcp mss to PMTU in the Mangle table. When the underlying
platform is running nftables, the auto-translate (iptables-translate)
 fails to convert these rules (most probably because
iptables-translate currently doesn't support converting mss-clamping
rules targeting ipsets).

This PR configures mss-clamping rules that don't target ipsets when
the underlying platform is running nftables.

Once auto-translate for mss-clamping rules targeting ipsets is
supported, this fix could be reverted.

### DIFF
--- a/pkg/iptables/iptables.go
+++ b/pkg/iptables/iptables.go
@@ -19,6 +19,8 @@ limitations under the License.
 package iptables
 
 import (
+	"bytes"
+	"os/exec"
 	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
@@ -26,6 +28,12 @@ import (
 	level "github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/stringset"
 	"k8s.io/klog"
+)
+
+const (
+	iptablesCmd = "iptables"
+	versionCmd  = "--version"
+	nftSubstr   = "nf_tables"
 )
 
 type Interface interface {
@@ -187,4 +195,26 @@ func UpdateChainRules(ipt Interface, table, chain string, rules [][]string) erro
 	}
 
 	return nil
+}
+
+// iptables mode should be retrieved from iptables library (when supported )
+
+func IsNftMode(ipt Interface) bool {
+	path, err := exec.LookPath(iptablesCmd)
+	if err != nil {
+		klog.Warningf("Unable to find %s path: %q", iptablesCmd, err)
+		return false
+	}
+
+	cmd := exec.Command(path, versionCmd)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+
+	err = cmd.Run()
+	if err != nil {
+		klog.Warningf("Unable to run %s %s: %q", iptablesCmd, versionCmd, err)
+		return false
+	}
+
+	return strings.Contains(out.String(), nftSubstr)
 }
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
@@ -33,9 +33,10 @@ import (
 
 type mtuHandler struct {
 	event.HandlerBase
-	ipt         iptables.Interface
-	remoteIPSet ipset.Named
-	localIPSet  ipset.Named
+	ipt             iptables.Interface
+	remoteIPSet     ipset.Named
+	localIPSet      ipset.Named
+	iptablesNftMode bool
 }
 
 func NewMTUHandler() event.Handler {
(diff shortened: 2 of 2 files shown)


pkg/iptables/iptables.go
```diff
@@ -28,6 +30,12 @@ import (
 	"k8s.io/klog"
 )
 
+const (
+	IptablesCmd = "iptables"
+	VersionCmd  = "--version"
+	NftSubstr   = "nf_tables"
```
tpantelis: These don't need to be publicly scoped. That said they're only used in one place and thus don't need to be constants.

pkg/iptables/iptables.go
```diff
@@ -188,3 +196,29 @@ func UpdateChainRules(ipt Interface, table, chain string, rules [][]string) erro
 
 	return nil
 }
+
+// iptables mode should be retrieved from iptables library (when supported )
+
+func IsNftMode(ipt Interface) bool {
```
tpantelis, line 202: This function doesn't use `ipt`.  It seems this function should be part of the `Interface`?

pkg/iptables/iptables.go
```diff
@@ -188,3 +196,29 @@ func UpdateChainRules(ipt Interface, table, chain string, rules [][]string) erro
...
+	if err != nil {
+		klog.Warningf("Unable to run iptables --version: %q", err)
+		return false
+	}
+
+	if !strings.Contains(out.String(), NftSubstr) {
+		return false
+	}
+
+	return true
```
tpantelis:
```suggestion
	return strings.Contains(out.String(), NftSubstr)
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -58,53 +59,62 @@ func (h *mtuHandler) Init() error {
...
+		if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecDest...); err != nil {
+			return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecSource, " "))
+		}
 	}
 
 	return nil
 }
 
 func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
+	if h.iptablesNftMode {
```
sridhargaddam, line 114: Before we merge this PR, let us confirm that its indeed working on a real setup.

Here we are avoiding adding rules with localCIDR, so I'd like to hear from @aswinsuryan if this is fine and if there were any other reasons why we used to add iptable rules even with localCIDRs.

pkg/iptables/iptables.go
```diff
@@ -188,3 +196,29 @@ func UpdateChainRules(ipt Interface, table, chain string, rules [][]string) erro
...
+		return false
+	}
+
+	cmd := exec.Command(path, VersionCmd)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+
+	err = cmd.Run()
+	if err != nil {
+		klog.Warningf("Unable to run iptables --version: %q", err)
```
sridhargaddam:
```suggestion
		klog.Warningf("Unable to run %s %s: %q", IptablesCmd, VersionCmd, err)
```

pkg/iptables/iptables.go
sridhargaddam: +1 to "dont need to be publicly scoped"

The info messages below seem to use hardcoded names, so we can instead use these macros.

pkg/iptables/iptables.go
```diff
@@ -188,3 +196,29 @@ func UpdateChainRules(ipt Interface, table, chain string, rules [][]string) erro
 
 	return nil
 }
+
+// iptables mode should be retrieved from iptables library (when supported )
+
+func IsNftMode(ipt Interface) bool {
+	path, err := exec.LookPath(IptablesCmd)
+	if err != nil {
+		klog.Warningf("Unable to find iptables path: %q", err)
```
sridhargaddam:
```suggestion
		klog.Warningf("Unable to find %s path: %q", IptablesCmd, err)
```

sridhargaddam (changes requested):
Thanks for the PR @yboaron. Some minor comments. Adding "Requested changes" mainly to hold this PR from getting merged until we validate the fix on an OSP/AWS setup.

pkg/iptables/iptables.go
yboaron: Yes you are right, I thought to include it in the interface to make it easier to retrieve mode from iptables library when supported.

Both options seem fine to me, would you rather exclude it from the interface?

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: Yep, agree, we need to verify it solves the mtu issue with nftables platforms, before merging it.

aswinsuryan: It was added since some tests were failing, IIRC some of the host networking tests. If e2e passes it should be safe to avoid these rules.

pkg/iptables/iptables.go
tpantelis: I'd prefer on the interface so we can mock it in unit tests.

tpantelis (changes requested):

yboaron:
I'm going to close this PR , and will open new PR to address MTU issue

PR #1817, opened 2022-05-05
Use Dependabot to update GHAs on release-0.12

dfarrell07:
Add Dependabot configuration to automatically update GitHub Actions on
the release-0.12 branch.

This will facilitate quick updates while using SHA-based versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,11 @@ updates:
     directory: '/'
     schedule:
       interval: daily
+  - package-ecosystem: github-actions
+    directory: '/'
+    target-branch: "release-0.12"
+    schedule:
+      interval: daily
   - package-ecosystem: gomod
     directory: "/"
     schedule:


dfarrell07:
This seems much easier than backporting every GHA update we need, but it would mean that not every commit added to the release branch is strictly a cherry-pick.

skitt:
> This seems much easier than backporting every GHA update we need, but it would mean that not every commit added to the release branch is strictly a cherry-pick.

I’m of the opinion that this is fine; we do want GHA updates on release branches, but we don’t want to be tied to backporting whatever changes we make in the `devel` GHAs. Perhaps we can require that any GHA dependency bump is merged and tested on `devel` before being merged on the release branches.

skitt:
I’ve added this to https://docs.google.com/document/d/1IhTMQMWkDj_V9ZfHI9szy3xZ3dZb0yMEHhP2y_ivzRU/edit

skitt (approved):

.github/dependabot.yml
```diff
@@ -5,6 +5,11 @@ updates:
     directory: '/'
     schedule:
       interval: daily
+  - package-ecosystem: github-actions
+    directory: '/'
+    target-branch: "release-0.12"
```
skitt, line 10: This is something we should take care of when cutting release branches in future (since this section is present in `devel` now).

dfarrell07: Yeah at the same time we cut release branches we should update this config on devel to point to the new branch. It's only read from devel, so whatever ends up on the release branch itself doesn't matter.

dfarrell07:
> I’ve added this to https://docs.google.com/document/d/1IhTMQMWkDj_V9ZfHI9szy3xZ3dZb0yMEHhP2y_ivzRU/edit

We ran out of time to talk about it today, but I requested feedback in Slack.

https://kubernetes.slack.com/archives/C010RJV694M/p1652108705916989

dfarrell07:
> I’ve added this to https://docs.google.com/document/d/1IhTMQMWkDj_V9ZfHI9szy3xZ3dZb0yMEHhP2y_ivzRU/edit

We talked about this on the dev call today and there were no objections. :heavy_check_mark:

tpantelis (approved):

PR #1830, opened 2022-05-11

### LINKED ISSUE #1774: MTU issues seen between an OnPrem cluster vs Public cloud

yboaron:
Submariner MTU handler in route-agent programs iptable rules to
    clamp tcp mss to PMTU in the Mangle table. We noticed that tcp
    mss not being clamped to PMTU when the platform is running nftables
    and Globalnet is enabled.
    
  This PR adds iptables rules that force tcp mss clamping in case
  Globalnet is configured, the mss value is calculated based on
  the MTU of the default interface and the cable driver overhead.

Fixes: https://github.com/submariner-io/submariner/issues/1774

### DIFF
--- a/pkg/routeagent_driver/environment/env.go
+++ b/pkg/routeagent_driver/environment/env.go
@@ -24,4 +24,5 @@ type Specification struct {
 	ClusterCidr []string
 	ServiceCidr []string
 	Uninstall   bool
+	GlobalCidr  []string
 }
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
@@ -19,16 +19,32 @@ limitations under the License.
 package mtu
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
 	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cable/vxlan"
 	"github.com/submariner-io/submariner/pkg/event"
 	"github.com/submariner-io/submariner/pkg/ipset"
 	"github.com/submariner-io/submariner/pkg/iptables"
+	netlinkAPI "github.com/submariner-io/submariner/pkg/netlink"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	"k8s.io/klog"
 	utilexec "k8s.io/utils/exec"
+)
+
+type forceMssSts int
+
+const (
+	notNeeded forceMssSts = iota
+	needed
+	configured
+)
+
+const (
+	// TCP MSS = Default_Iface_MTU - TCP_H(20)-IP_H(20)-max_IpsecOverhed(80).
+	maxIpsecOverhead = 120
 )
 
 type mtuHandler struct {
@@ -37,11 +53,18 @@ type mtuHandler struct {
 	ipt              iptables.Interface
 	remoteIPSet      ipset.Named
 	localIPSet       ipset.Named
+	forceMss         forceMssSts
 }
 
-func NewMTUHandler(localClusterCidr []string) event.Handler {
+func NewMTUHandler(localClusterCidr []string, isGlobalnet bool) event.Handler {
+	forceMss := notNeeded
+	if isGlobalnet {
+		forceMss = needed
+	}
+
 	return &mtuHandler{
 		localClusterCidr: localClusterCidr,
+		forceMss:         forceMss,
 	}
 }
 
@@ -79,6 +102,19 @@ func (h *mtuHandler) Init() error {
 		return errors.Wrapf(err, "error creating ipset %q", constants.LocalCIDRIPSet)
 	}
 
+	if err := iptables.PrependUnique(h.ipt, constants.MangleTable, constants.PostRoutingChain,
+		forwardToSubMarinerPostRoutingChain); err != nil {
+		return errors.Wrapf(err, "error inserting iptables rule %q",
+			strings.Join(forwardToSubMarinerPostRoutingChain, " "))
+	}
+
+	// iptable rules to clamp TCP MSS to a fixed value will be programmed when the local endpoint is created
+	if h.forceMss == needed {
+		return nil
+	}
+
+	klog.Info("Creating iptables clamp-mss-to-pmtu rules")
+
 	ruleSpecSource := []string{
 		"-m", "set", "--match-set", constants.LocalCIDRIPSet, "src", "-m", "set", "--match-set",
 		constants.RemoteCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
@@ -88,12 +124,6 @@ func (h *mtuHandler) Init() error {
 		"-m", "set", "--match-set", constants.RemoteCIDRIPSet, "src", "-m", "set", "--match-set",
 		constants.LocalCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
 		"--clamp-mss-to-pmtu",
-	}
-
-	if err := iptables.PrependUnique(h.ipt, constants.MangleTable, constants.PostRoutingChain,
-		forwardToSubMarinerPostRoutingChain); err != nil {
-		return errors.Wrapf(err, "error inserting iptables rule %q",
-			strings.Join(forwardToSubMarinerPostRoutingChain, " "))
 	}
 
 	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecSource...); err != nil {
@@ -121,6 +151,17 @@ func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
 		if err != nil {
 			return errors.Wrap(err, "error adding localClusterCidr IP set entry")
 		}
+	}
+
+	if h.forceMss == needed {
+		klog.Info("Creating iptables set-mss rules")
+
+		err := h.forceMssClamping(endpoint)
+		if err != nil {
+			return errors.Wrap(err, "error forcing TCP MSS clamping")
+		}
+
+		h.forceMss = configured
 	}
 
 	return nil
@@ -228,6 +269,42 @@ func (h *mtuHandler) Stop(uninstall bool) error {
 
 	if err := h.remoteIPSet.Destroy(); err != nil {
 		klog.Errorf("Error deleting ipset %q: %v", constants.RemoteCIDRIPSet, err)
+	}
+
+	return nil
+}
+
+func (h *mtuHandler) forceMssClamping(endpoint *submV1.Endpoint) error {
+	defaultHostIface, err := netlinkAPI.GetDefaultGatewayInterface()
+	if err != nil {
+		return errors.Wrapf(err, "Unable to find the default interface on host")
+	}
+
+	overHeadSize := maxIpsecOverhead
+	if endpoint.Spec.Backend == vxlan.CableDriverName {
+		overHeadSize = vxlan.VxlanOverhead
+	}
+
+	klog.Infof("forceMssClamping to: IF_MTU(%d)-Overhead(%d)=%d",
+		defaultHostIface.MTU, overHeadSize, (defaultHostIface.MTU - overHeadSize))
+
+	ruleSpecSource := []string{
+		"-m", "set", "--match-set", constants.LocalCIDRIPSet, "src", "-m", "set", "--match-set",
+		constants.RemoteCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
+		"--set-mss", strconv.Itoa(defaultHostIface.MTU - overHeadSize),
+	}
+	ruleSpecDest := []string{
+		"-m", "set", "--match-set", constants.RemoteCIDRIPSet, "src", "-m", "set", "--match-set",
+		constants.LocalCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
+		"--set-mss", strconv.Itoa(defaultHostIface.MTU - overHeadSize),
+	}
+
+	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecSource...); err != nil {
+		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecSource, " "))
+	}
+
+	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecDest...); err != nil {
+		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecDest, " "))
 	}
 
 	return nil
--- a/pkg/routeagent_driver/main.go
+++ b/pkg/routeagent_driver/main.go
@@ -86,7 +86,7 @@ func main() {
 		ovn.NewHandler(&env, smClientset),
 		cabledriver.NewXRFMCleanupHandler(),
 		cabledriver.NewVXLANCleanup(),
-		mtu.NewMTUHandler(env.ClusterCidr),
+		mtu.NewMTUHandler(env.ClusterCidr, (len(env.GlobalCidr) != 0)),
 	); err != nil {
 		klog.Fatalf("Error registering the handlers: %s", err.Error())
 	}
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler_test.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler_test.go
@@ -46,7 +46,7 @@ var _ = Describe("MTUHandler", func() {
 		ipset.NewFunc = func() ipset.Interface {
 			return ipSet
 		}
-		handler = mtu.NewMTUHandler([]string{"10.1.0.0/24"})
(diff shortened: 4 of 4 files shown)


pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -159,10 +110,15 @@ func (h *mtuHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
...
 	for _, subnet := range subnets {
-		err := h.remoteIPSet.DelEntry(subnet)
-		if err != nil {
-			klog.Errorf("Error deleting the subnet %q from the remote IPSet: %v", subnet, err)
+		ruleSpecDest := []string{
+			"-d", subnet, "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
+			"--set-mss", h.mssClampValue,
+		}
+
+		if err := h.ipt.Delete(constants.MangleTable, constants.PostRoutingChain, ruleSpecDest...); err != nil {
```
```suggestion
		if err := h.ipt.Delete(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecDest...); err != nil {
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -159,10 +110,15 @@ func (h *mtuHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
...
+			klog.Errorf("Error deleting iptables rule from %q chain: %v", constants.PostRoutingChain, err)
```
```suggestion
			klog.Errorf("Error deleting iptables rule %q: %v", strings.Join(ruleSpecDest, " "), err)
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -61,96 +61,47 @@ func (h *mtuHandler) Init() error {
...
-		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecSource, " "))
+	defaultHostIface, err := netlinkAPI.GetDefaultGatewayInterface()
+	if err != nil {
+		return errors.Wrapf(err, "Unable to find the default interface on host")
 	}
 
-	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecDest...); err != nil {
-		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecSource, " "))
-	}
+	// TODO: set maxOverlayHeaderSize based on cable driver type
```
sridhargaddam: As you know, VxLAN cable driver would have a lower overhead and since this fix cannot be immediately consumed, shall we include the cable driver checks as well? I'm fine if you want to use a separate PR for that.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -19,30 +19,30 @@ limitations under the License.
...
-	"github.com/submariner-io/submariner/pkg/ipset"
 	"github.com/submariner-io/submariner/pkg/iptables"
+	netlinkAPI "github.com/submariner-io/submariner/pkg/netlink"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	"k8s.io/klog"
-	utilexec "k8s.io/utils/exec"
+)
+
+const (
+	maxOverlayHeaderSize = 120
```
sridhargaddam: Please add some notes on why/how we derived at the value of 120 for future reference.

sridhargaddam (changes requested):
Thanks @yboaron for the PR, nice work.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: Yes, I prefer to have it in a separate PR.

sridhargaddam (changes requested):
Thanks for fixing the review comments. I'm checking something, will update the vote shortly.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -19,30 +19,31 @@ limitations under the License.
...
-	"github.com/submariner-io/submariner/pkg/ipset"
 	"github.com/submariner-io/submariner/pkg/iptables"
+	netlinkAPI "github.com/submariner-io/submariner/pkg/netlink"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	"k8s.io/klog"
-	utilexec "k8s.io/utils/exec"
+)
+
+const (
+	// TCP MSS = MTU - TCP_H(20)-IP_H(20)-max_IpsecOverhed(80)
```
```suggestion
	// TCP MSS = Default_Iface_MTU - TCP_H(20)-IP_H(20)-max_IpsecOverhed(80)
```

Just in case if you have to respin the PR for any additional comments, otherwise its fine in the current state.

skitt:
Thanks for this, including the explanation in the commit message!
It seems the unit test needs to be updated (or is it identifying a real issue in the implementation?).

pkg/routeagent_driver/main.go
```diff
@@ -73,6 +77,28 @@ func main() {
 		klog.Fatalf("Error building submariner clientset: %s", err.Error())
 	}
 
+	var localCluster *submarinerv1.Cluster
+
+	isGlobalnet := false
+	// During installation, sometimes creation of clusterCRD by submariner-gateway-pod would take few secs.
+	for i := 0; i < 100; i++ {
```
mangelajo: I'd recommend the use of a retry lib instead of a loop:
[example](https://github.com/submariner-io/submariner/blob/devel/pkg/endpoint/public_ip.go#L124)

mangelajo (changes requested):
I was passing by, don't take me too seriously :D

pkg/routeagent_driver/main.go
yboaron: OK, 10x.
Done

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -123,6 +146,15 @@ func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
 		}
 	}
 
+	if h.forceMss == needed {
```
sridhargaddam, line 156: @yboaron so here we are programming both the rules right? 
i.e., existing rules to clamp tcp mss to PMTUD and
if globalnet deployment, new rules to clamp tcp mss to a fixed value. 

IMHO, I feel we should just program one of them depending on the deployment.

i.e., move this check to Init() method and in Line 105/110 if Vanilla Submariner, we will continue to use the same rules (aka PMTUD) but if its Globalnet we will program rules to clamp the MSS to a fixed value.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -19,29 +19,52 @@ limitations under the License.
...
+type forceMssSts int
+
+const (
+	notNeeded forceMssSts = iota
+	needed
+	configured
+)
+
+const (
+	// TCP MSS = MTU - TCP_H(20)-IP_H(20)-max_IpsecOverhed(80).
```
```suggestion
	// TCP MSS = Default_Iface_MTU - TCP_H(20)-IP_H(20)-max_IpsecOverhed(80).
```

sridhargaddam (changes requested):
Thanks @yboaron

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -123,6 +151,15 @@ func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
 		}
 	}
 
+	if h.forceMss == needed {
+		err := h.forceMssClamping(endpoint)
```
sridhargaddam, line 159: Any specific reason why we are programming these rules in the LocalEndpointCreated API. Can we program in the Init() method itself. i.e., Just the ruleSpecSource/ruleSpecDest will change depending on Globalnet is enabled or not.

sridhargaddam (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: I also thought to add the rules in init() function, but since we want to calculate the MSS value based on the cable driver , I moved it to LocalEndpointCreated function where we have handy the cable driver.
But, yeah agree that it's more straightforward to have the rules in init() function.

Any idea how we can retrieve the cable driver in case we would like to add the rules in init() function? 

One more thing, I'm not sure how critical it is, but current PR doesn't cover runtime config change of cable driver,  I mean  something like the following use case:

SM deployed with Globalnet and VxLAN --> clamp iptables rules will be configured with value based on VxLAN header
Cable driver config changed from VxLAN to IPsec --> I don't think that mtuhandler is being called upon endpoint update

So, we'll endup with force clamp rules to based on VxLAN instead of ipsec.

sridhargaddam: > I also thought to add the rules in init() function, but since we want to calculate the MSS value based on the cable driver , I moved it to LocalEndpointCreated function where we have handy the cable driver. But, yeah agree that it's more straightforward to have the rules in init() function.

> Any idea how we can retrieve the cable driver in case we would like to add the rules in init() function?

In that case, we can read the submariner CR object which has both the CableDriver info as well as GlobalCIDR info in the retry loop. The Submariner CR would be read by the submariner-operator while deploying Submariner components, so we can expect it to be present when the init() method is invoked.
CC: @skitt do you see any issue if we use information from Submariner CR in route-agent?


> One more thing, I'm not sure how critical it is, but current PR doesn't cover runtime config change of cable driver, I mean something like the following use case:
> 
> SM deployed with Globalnet and VxLAN --> clamp iptables rules will be configured with value based on VxLAN header Cable driver config changed from VxLAN to IPsec --> I don't think that mtuhandler is being called upon endpoint update
> 
> So, we'll endup with force clamp rules to based on VxLAN instead of ipsec.

IMO if the user is interested in updating the cableDriver, we expect the user to run "subctl uninstall" before deploying with VxLAN cable driver. As part of uninstall the rules will be flushed and fresh rules will be programmed, so it should be fine.

sridhargaddam (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
tpantelis: > In that case, we can read the submariner CR object which has both the CableDriver info as well as GlobalCIDR info in the retry loop. The Submariner CR would be read by the submariner-operator while deploying Submariner components, so we can expect it to be present when the init() method is invoked. CC: @skitt do you see any issue if we use information from Submariner CR in route-agent?
> 

I don't think we want the route agent reading the Submariner CR. The operator already passes the GlobalCIDR via env var, why not pass the cable driver as well? Or read it from the Backend of the local Endpoint as it does now.

However, looking ahead, assuming a single, static cable driver will be further problematic for supporting multiple cable drivers, should we ever revisit that.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -79,6 +102,17 @@ func (h *mtuHandler) Init() error {
 		return errors.Wrapf(err, "error creating ipset %q", constants.LocalCIDRIPSet)
 	}
 
+	if err := iptables.PrependUnique(h.ipt, constants.MangleTable, constants.PostRoutingChain,
+		forwardToSubMarinerPostRoutingChain); err != nil {
+		return errors.Wrapf(err, "error inserting iptables rule %q",
+			strings.Join(forwardToSubMarinerPostRoutingChain, " "))
+	}
+
+	// if it is needed to force MSS clamping we don't need to define the clamp-mss-to-pmtu rules
```
sridhargaddam: Please update this comment with a note saying that iptable rules to clamp TCP MSS to a fixed value will be programmed when the local endpoint is created.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: In that case maybe it is worth retrieving the cable driver from endpoint CR and configuring iptables rules in LocalEndpointCreated as implemented in current PR version.

In case multiple cable drivers will be supported   I assume we'll need to configure the iptables rules in LocalEndpointCreated function 

@sridhargaddam what do you think ?

sridhargaddam: In order to support multiple cable drivers when clamping the TCP MSS to a fixed value, we will have to program individual iptable rules for each destination and cannot use the IPSets (like localIPset and remoteIPset) with a single IPtable rule. Such rules have to be programmed in the RemoteEndpointCreated methods AFAICT.

sridhargaddam (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: As decided, we'll continue with the current approach and when multiple cable drivers will be supported we'll to revisit this handler.

pkg/routeagent_driver/environment/env.go
```diff
@@ -24,4 +24,5 @@ type Specification struct {
 	ClusterCidr []string
 	ServiceCidr []string
 	Uninstall   bool
+	GlobalCidr  string
```
sridhargaddam: Please check if this should be an array of strings like the other Cluster/Service CIDRs. 

In submariner-gateway env' its an array and similarly in globalnet as well.
https://github.com/submariner-io/submariner/blob/8c0d04e3fff7d7da5396d625b09ba91e7e3aa688/pkg/types/types.go#L36

https://github.com/submariner-io/submariner/blob/8c0d04e3fff7d7da5396d625b09ba91e7e3aa688/pkg/globalnet/controllers/types.go#L80

sridhargaddam (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -232,3 +273,39 @@ func (h *mtuHandler) Stop(uninstall bool) error {
...
+		constants.LocalCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
+		"--set-mss", strconv.Itoa(defaultHostIface.MTU - overHeadSize),
+	}
+
+	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecSource...); err != nil {
+		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecSource, " "))
+	}
+
+	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecDest...); err != nil {
+		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecSource, " "))
```
```suggestion
		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecDest, " "))
```

sridhargaddam (changes requested):

sridhargaddam (approved):
Thanks for your patience Yossi. LGTM.

tpantelis (approved):

sridhargaddam:
@yboaron can you please handle the backport?

dfarrell07:
This was backported in https://github.com/submariner-io/submariner/pull/1846.

PR #1846, opened 2022-05-24

yboaron:
Submariner MTU driver in route-agent programs iptable rules to
clamp tcp mss to PMTU in the Mangle table. We noticed that tcp
mss not being clamped to PMTU when platform is running nftables
and Globalnet is enabled.

This PR adds iptables rules that force tcp mss clamping in case
Globalnet is configured, the mss value is calculated based on
the MTU of the default interface and the cable driver overhead.

Fixes: https://github.com/submariner-io/submariner/issues/1774

### DIFF
--- a/pkg/routeagent_driver/environment/env.go
+++ b/pkg/routeagent_driver/environment/env.go
@@ -24,4 +24,5 @@ type Specification struct {
 	ClusterCidr []string
 	ServiceCidr []string
 	Uninstall   bool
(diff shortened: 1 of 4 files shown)


tpantelis (approved):

sridhargaddam (approved):

dfarrell07:
This is a backport of https://github.com/submariner-io/submariner/pull/1830.

PR #1851, opened 2022-05-30
Fix connection failure with LB mode + VxLAN Cabledriver

### LINKED ISSUE #1822: Loadbalancer mode not working with VxLAN cable driver on AWS
While installing Submariner with VxLAN cable driver and Loadbalancer mode, it was seen that on AWS cluster, the vxlan-tunnel interface disappears and there are couple of errors in the submariner-gateway pod.

**Environment**:
- Cloud provider or hardware configuration: AWS
- Install tools: Submariner version: 0.12 on OCP 4.10.2

sridhargaddam:
When using Loadbalancer mode with VxLAN cable driver, it was seen that
connections are sometimes not setup. The reason for this is the missing
vxlan-tunnel interface which was deleted by route-agent driver assuming
that the Gateway node migrated.

The reason why route-agent considers it as Gateway migration is because
the local-endpoint is deleted and then re-created. Ideally, the local
endpoint should not be deleted when the Gateway pod comes up, but the
current checks in the code tries to match the entire backendConfig and
when it finds that something does not match with the endpoint stored in
the datastore it deletes and re-creates the endpoint. This PR fixes this
issue by ensuring that local-endpoint is simply updated and not deleted.

Fixes: https://github.com/submariner-io/submariner/issues/1822

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -132,7 +132,19 @@ func CompareEndpointSpec(left, right *subv1.EndpointSpec) bool {
 
 	// maybe we have to use just reflect.DeepEqual(left, right), but in this case the subnets order will influence.
 	return left.ClusterID == right.ClusterID && left.CableName == right.CableName && left.Hostname == right.Hostname &&
-		left.Backend == right.Backend && equality.Semantic.DeepEqual(left.BackendConfig, right.BackendConfig)
+		left.Backend == right.Backend && isBackendConfigSame(left, right)
+}
+
+func isBackendConfigSame(left, right *subv1.EndpointSpec) bool {
+	if left.BackendConfig[subv1.UsingLoadBalancer] == "true" &&
+		right.BackendConfig[subv1.UsingLoadBalancer] == "true" {
+		// When Gateway pod comes up with loadbalancer mode enabled, it inserts a preferred-server-timestamp in
+		// the BackendConfig when the Gateway pod comes up. So, in loadbalancer mode, we just have to compare
+		// the load-balancer status.
+		return true
+	}
+
+	return equality.Semantic.DeepEqual(left.BackendConfig, right.BackendConfig)
 }
 
 func EnsureValidName(name string) string {


skitt (approved):

pkg/util/util.go
```diff
@@ -132,9 +132,22 @@ func CompareEndpointSpec(left, right *subv1.EndpointSpec) bool {
...
+func isBackendConfigSame(left, right *subv1.EndpointSpec) bool {
+	if left.BackendConfig[subv1.UsingLoadBalancer] == "true" &&
+		right.BackendConfig[subv1.UsingLoadBalancer] == "true" {
+		// When Gateway pod comes up with loadbalancer mode enabled, it inserts a preferred-server-timestamp in
+		// the BackendConfig when the Gateway pod comes up. So, in loadbalancer mode, we just have to compare
+		// the load-balancer status.
+		return true
+	} else {
+		return equality.Semantic.DeepEqual(left.BackendConfig, right.BackendConfig)
+	}
```
skitt:
```suggestion
	}
	
	return equality.Semantic.DeepEqual(left.BackendConfig, right.BackendConfig)
```

yboaron:
Overall, looks good to me, but I have 
Q: will we hit a similar issue if one of the fields in of local endpoint will be updated?  or it's just init race

sridhargaddam:
> Overall, looks good to me, but I have Q: will we hit a similar issue if one of the fields in of local endpoint will be updated? or it's just init race

The specific problem fixed in this PR is not a race. The local endpoint is always updated with the latest timestamp as [preferred-time-stamp](https://github.com/submariner-io/submariner/blob/43219bcc143712bc4aba16b0e7f2614bc74f147c/pkg/endpoint/local_endpoint.go#L155) when the gateway pod comes up, to indicate to the remote clusters that the local Gateway wants to act as a preferred server.

yboaron (approved):

PR #1860, opened 2022-06-06
add format judge about the cableName

pengbinbin1:

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -90,6 +90,10 @@ func GetClusterIDFromCableName(cableName string) string {
 	// 0           1    2   3    4    5       6   7  8 9  10
 	// submariner-cable-my-super-long_cluster-id-172-16-32-5
 	cableSplit := strings.Split(cableName, "-")
+	if len(cableSplit) < 3 {
+		klog.Error(fmt.Errorf("cableName:%s format is invalid", cableName))
+		return ""
+	}
 	clusterID := cableSplit[2]
 
 	for i := 3; i < len(cableSplit)-4; i++ {


pkg/util/util.go
```diff
@@ -90,6 +90,10 @@ func GetClusterIDFromCableName(cableName string) string {
 	// 0           1    2   3    4    5       6   7  8 9  10
 	// submariner-cable-my-super-long_cluster-id-172-16-32-5
 	cableSplit := strings.Split(cableName, "-")
+	if len(cableSplit) != 11{
+		klog.Error(fmt.Errorf("cableName:%s format is invalid",cableName))
+		return ""
+	}
```
tpantelis, line 96: What is the motivation for this change? It is actually incorrect - the token length does not need to be 11 - the comment above is just an example of one that is long. The unit tests for this function fail with this change.

tpantelis (changes requested):

pkg/util/util.go
pengbinbin1: Maybe judge whether the length is less than 3 is appropriate ,because if the cableSplit's length is less than 3,then the cableSplit[2] will be panic

tpantelis: Sure - we expect the length to be at least 3. 

This really shouldn't be an issue unless someone purposely tampers with it. Are you seeing an issue or how did this code come to your attention?

pengbinbin1: I'am reading the code and  found it inadvertently

dfarrell07: Looks like a good catch to me, of course we want to avoid the possibility of purposely tampering/crashing. Thanks @pengbinbin1!

dfarrell07 (approved):

skitt (approved):

pkg/util/util.go
tpantelis: The check keeps it from panicking but what caller functionality will be broken by returning an empty cluster ID....perhaps we're better off crashing. That said, this function isn't actually used except in unit tests so we should just remove it 😄

tpantelis (changes requested):

pkg/util/util.go
dfarrell07: Thanks for the context @tpantelis, that makes sense.

tpantelis:
The `GetClusterIDFromCableName` function was removed as part of https://github.com/submariner-io/submariner/pull/1879.

PR #1863, opened 2022-06-06
Update K8s dependabot config

tpantelis:
Ignore all `k8s.io` and `sigs.k8s.io` dependencies and versions.


dfarrell07 (approved):
Makes sense

skitt:
By specifying an upper bound on versions, I was hoping to get dependabot to track minor updates for us; but I’m not sure that actually works, so let’s forget it.

skitt (approved):

PR #1870, opened 2022-06-13
Use NettestImageURL var from shipyard

tpantelis:
...to make the `nettest` image configurable for air-gapped deployments.


dfarrell07:
Are we thinking about backporting this? If so, I guess the version bump should be a separate commit.

tpantelis:
> Are we thinking about backporting this? 

Not that I'm aware of. If so, we'd have to backport several more PRs.

dfarrell07 (approved):

skitt (approved):

PR #1872, opened 2022-06-14
Support override MTU value used in TCP MSS clamping

### LINKED ISSUE #1858: Support option to override the MTU value used in TCP MSS clamping
Currently, when Globalnet is enabled in the cluster, to address the MTU issues, Submariner clamps the TCP MSS to a fixed value derived from the default-interface-mtu minus the tunnel-overhead. In certain environments where the MTU between the Gateway nodes is even lower, we should provide an option for the users to override this MTU value.

This can be done by adding annotations on the Gateway node or by configuring them as env' variables on the submariner-gateway pod.

yboaron:
Currently, when Globalnet is enabled in the cluster,
to address the MTU issues, Submariner clamps the TCP
MSS to a value derived from the default-interface-mtu minus
the tunnel-overhead.

This PR allows users to set MSS clamping value using
node annotation.

To force MSS clamping to a specific value it needs
to annotate the relevant node:
kubectl annotate node <node_name> submariner.io/tcp-mss-clamping=<value>

Fixes: https://github.com/submariner-io/submariner/issues/1858

### DIFF
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -95,6 +95,7 @@ const (
 	PreferredServerConfig   = "preferred-server"
 	PublicIP                = "public-ip"
 	UsingLoadBalancer       = "using-loadbalancer"
+	TCPMssValue             = "submariner.io/tcp-clamp-mss"
 )
 
 // Valid PublicIP resolvers.
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
@@ -54,9 +54,10 @@ type mtuHandler struct {
 	remoteIPSet      ipset.Named
 	localIPSet       ipset.Named
 	forceMss         forceMssSts
+	tcpMssValue      int
 }
 
-func NewMTUHandler(localClusterCidr []string, isGlobalnet bool) event.Handler {
+func NewMTUHandler(localClusterCidr []string, isGlobalnet bool, tcpMssValue int) event.Handler {
 	forceMss := notNeeded
 	if isGlobalnet {
 		forceMss = needed
@@ -65,6 +66,7 @@ func NewMTUHandler(localClusterCidr []string, isGlobalnet bool) event.Handler {
 	return &mtuHandler{
 		localClusterCidr: localClusterCidr,
 		forceMss:         forceMss,
+		tcpMssValue:      tcpMssValue,
 	}
 }
 
@@ -275,28 +277,55 @@ func (h *mtuHandler) Stop(uninstall bool) error {
 }
 
 func (h *mtuHandler) forceMssClamping(endpoint *submV1.Endpoint) error {
-	defaultHostIface, err := netlinkAPI.GetDefaultGatewayInterface()
-	if err != nil {
-		return errors.Wrapf(err, "Unable to find the default interface on host")
+	tcpMssSrc := "user"
+	tcpMssValue := h.tcpMssValue
+
+	if tcpMssValue == 0 {
+		defaultHostIface, err := netlinkAPI.GetDefaultGatewayInterface()
+		if err != nil {
+			return errors.Wrapf(err, "Unable to find the default interface on host")
+		}
+
+		overHeadSize := maxIpsecOverhead
+		if endpoint.Spec.Backend == vxlan.CableDriverName {
+			overHeadSize = vxlan.VxlanOverhead
+		}
+
+		tcpMssValue = defaultHostIface.MTU - overHeadSize
+		tcpMssSrc = "default"
 	}
 
-	overHeadSize := maxIpsecOverhead
-	if endpoint.Spec.Backend == vxlan.CableDriverName {
-		overHeadSize = vxlan.VxlanOverhead
-	}
-
-	klog.Infof("forceMssClamping to: IF_MTU(%d)-Overhead(%d)=%d",
-		defaultHostIface.MTU, overHeadSize, (defaultHostIface.MTU - overHeadSize))
-
+	klog.Infof("forceMssClamping to: %d (%s) ", tcpMssValue, tcpMssSrc)
 	ruleSpecSource := []string{
 		"-m", "set", "--match-set", constants.LocalCIDRIPSet, "src", "-m", "set", "--match-set",
 		constants.RemoteCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
-		"--set-mss", strconv.Itoa(defaultHostIface.MTU - overHeadSize),
+		"--set-mss", strconv.Itoa(tcpMssValue),
 	}
 	ruleSpecDest := []string{
 		"-m", "set", "--match-set", constants.RemoteCIDRIPSet, "src", "-m", "set", "--match-set",
 		constants.LocalCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
-		"--set-mss", strconv.Itoa(defaultHostIface.MTU - overHeadSize),
+		"--set-mss", strconv.Itoa(tcpMssValue),
+	}
+
+	rules, err := h.ipt.List(constants.MangleTable, constants.SmPostRoutingChain)
+	if err != nil {
+		return errors.Wrapf(err, "error listing the rules in %s chain", constants.SmPostRoutingChain)
+	}
+
+	isPresent := false
+
+	for _, rule := range rules {
+		if strings.Contains(rule, strings.Join(ruleSpecSource, " ")) || strings.Contains(rule, strings.Join(ruleSpecDest, " ")) {
+			isPresent = true
+			break
+		}
+	}
+
+	if len(rules) > 0 && !isPresent {
+		if err := h.ipt.ClearChain(constants.MangleTable, constants.SmPostRoutingChain); err != nil {
+			klog.Warningf("Error flushing iptables chain %q of %q table: %v", constants.SmPostRoutingChain,
+				constants.MangleTable, err)
+		}
 	}
 
 	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecSource...); err != nil {
--- a/pkg/routeagent_driver/main.go
+++ b/pkg/routeagent_driver/main.go
@@ -22,14 +22,17 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/kelseyhightower/envconfig"
 	"github.com/pkg/errors"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	cni "github.com/submariner-io/submariner/pkg/cni"
 	"github.com/submariner-io/submariner/pkg/event"
 	"github.com/submariner-io/submariner/pkg/event/controller"
 	"github.com/submariner-io/submariner/pkg/event/logger"
+	"github.com/submariner-io/submariner/pkg/node"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/cabledriver"
 	cniapi "github.com/submariner-io/submariner/pkg/routeagent_driver/cni"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
@@ -38,7 +41,6 @@ import (
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/mtu"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/ovn"
 	"k8s.io/client-go/kubernetes"
-	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
@@ -69,6 +71,11 @@ func main() {
 		klog.Fatalf("Error building kubeconfig: %s", err.Error())
 	}
 
+	k8sClientSet, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		klog.Fatalf("Error building clientset: %s", err.Error())
+	}
+
 	smClientset, err := submarinerClientset.NewForConfig(cfg)
 	if err != nil {
 		klog.Fatalf("Error building submariner clientset: %s", err.Error())
@@ -87,7 +94,7 @@ func main() {
 		ovn.NewHandler(&env, smClientset),
 		cabledriver.NewXRFMCleanupHandler(),
 		cabledriver.NewVXLANCleanup(),
-		mtu.NewMTUHandler(env.ClusterCidr, len(env.GlobalCidr) != 0),
+		mtu.NewMTUHandler(env.ClusterCidr, len(env.GlobalCidr) != 0, getTCPMssValue(k8sClientSet)),
 	); err != nil {
 		klog.Fatalf("Error registering the handlers: %s", err.Error())
 	}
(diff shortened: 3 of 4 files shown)


pkg/routeagent_driver/constants/constants.go
```diff
@@ -50,6 +50,7 @@ const (
 	// the egress traffic to the corresponding CNIInterfaceIP on that host.
 	RouteAgentHostNetworkTableID = 150
 
-	NATTable    = "nat"
-	FilterTable = "filter"
+	NATTable         = "nat"
+	FilterTable      = "filter"
+	ForceMssClamping = "submariner.io/mssClamping"
```
sridhargaddam: IMO its better to declare this constant at the same place where we are declaring other constants - i.e., https://github.com/submariner-io/submariner/blob/8e8728427df87ef146fa1243008dc6e72f078b04/pkg/apis/submariner.io/v1/types.go#L92

Also, what is your opinion on the following name?
submariner.io/tcp-mss-clamping

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -54,9 +54,10 @@ type mtuHandler struct {
 	remoteIPSet      ipset.Named
 	localIPSet       ipset.Named
 	forceMss         forceMssSts
+	usrForceMssVal   int
 }
 
-func NewMTUHandler(localClusterCidr []string, isGlobalnet bool) event.Handler {
+func NewMTUHandler(localClusterCidr []string, isGlobalnet bool, usrForceMssVal int) event.Handler {
```
```suggestion
func NewMTUHandler(localClusterCidr []string, isGlobalnet bool, usrForcedMssVal int) event.Handler {
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -274,29 +276,31 @@ func (h *mtuHandler) Stop(uninstall bool) error {
...
 		"-m", "set", "--match-set", constants.LocalCIDRIPSet, "src", "-m", "set", "--match-set",
 		constants.RemoteCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
-		"--set-mss", strconv.Itoa(defaultHostIface.MTU - overHeadSize),
+		"--set-mss", strconv.Itoa(usrForceMssVal),
 	}
 	ruleSpecDest := []string{
 		"-m", "set", "--match-set", constants.RemoteCIDRIPSet, "src", "-m", "set", "--match-set",
 		constants.LocalCIDRIPSet, "dst", "-p", "tcp", "-m", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
-		"--set-mss", strconv.Itoa(defaultHostIface.MTU - overHeadSize),
+		"--set-mss", strconv.Itoa(usrForceMssVal),
```
sridhargaddam: It is very likely that user might initially deploy Submariner, notice the problem and then want to update the tcpMss value by adding the necessary annotation on the GW node. So the code here should first delete the previous iptable rule (if present) and create a new one with the usrForcedMssValue

sridhargaddam (changes requested):
Thanks for the PR @yboaron

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: Not sure that I fully understand your comment, I assume that in case the user didn't annotate node we still want to clamp MSS to default value .
I think that latest version of PR should address your concern.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -274,29 +276,52 @@ func (h *mtuHandler) Stop(uninstall bool) error {
...
+		if endpoint.Spec.Backend == vxlan.CableDriverName {
+			overHeadSize = vxlan.VxlanOverhead
+		}
 
-	klog.Infof("forceMssClamping to: IF_MTU(%d)-Overhead(%d)=%d",
-		defaultHostIface.MTU, overHeadSize, (defaultHostIface.MTU - overHeadSize))
+		usrForceMssVal = defaultHostIface.MTU - overHeadSize
+	}
 
+	klog.Infof("forceMssClamping to: %d", usrForceMssVal)
```
sridhargaddam: Please add a log for us to know if the MSSClamp value is the default one or something that was forced by user. As we will mostly be relying on the logs while debugging any issue.

yboaron: Yeah, that's a good idea.
Done

pkg/apis/submariner.io/v1/types.go
```diff
@@ -95,6 +95,7 @@ const (
 	PreferredServerConfig   = "preferred-server"
 	PublicIP                = "public-ip"
 	UsingLoadBalancer       = "using-loadbalancer"
+	ForceTCPMssClamping     = "submariner.io/tcp-mss-clamping"
```
```suggestion
	TCPMssValue  = "submariner.io/tcp-mss-value"
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -54,9 +54,10 @@ type mtuHandler struct {
 	remoteIPSet      ipset.Named
 	localIPSet       ipset.Named
 	forceMss         forceMssSts
+	usrForcedMssVal  int
```
tpantelis:
```suggestion
	tcpMssValue  int
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -54,9 +54,10 @@ type mtuHandler struct {
...
 }
 
-func NewMTUHandler(localClusterCidr []string, isGlobalnet bool) event.Handler {
+func NewMTUHandler(localClusterCidr []string, isGlobalnet bool, usrForcedMssVal int) event.Handler {
```
tpantelis:
```suggestion
func NewMTUHandler(localClusterCidr []string, isGlobalnet bool, tcpMssValue int) event.Handler {
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -274,29 +276,54 @@ func (h *mtuHandler) Stop(uninstall bool) error {
 	return nil
 }
 
-func (h *mtuHandler) forceMssClamping(endpoint *submV1.Endpoint) error {
-	defaultHostIface, err := netlinkAPI.GetDefaultGatewayInterface()
-	if err != nil {
-		return errors.Wrapf(err, "Unable to find the default interface on host")
-	}
+func (h *mtuHandler) forceMssClamping(endpoint *submV1.Endpoint, usrForceMssVal int) error {
```
tpantelis: The mss value is stored in `mtuHandler` so we don't need to pass it here.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -274,29 +276,54 @@ func (h *mtuHandler) Stop(uninstall bool) error {
...
+	mssValueSrc := "user"
+	if usrForceMssVal == 0 {
```
tpantelis:
```suggestion
        tcpMssValue := h.tcpMssValue
	if tcpMssValue == 0 {
```

pkg/apis/submariner.io/v1/types.go
sridhargaddam:
```suggestion
	ForceTCPMssClamping     = "submariner.io/tcp-clamp-mss"
```

On a second thought, "tcp-mss-clamping" seems to convey a boolean value like true/false. Shall we use "tcp-clamp-mss" as it can convey an assignment.

pkg/routeagent_driver/main.go
```diff
@@ -152,3 +155,23 @@ func annotateNode(clusterCidr []string, cfg *restclient.Config) error {
...
+		return usrForceMssVal
+	}
+
+	localNode, err := node.GetLocalNode(k8sClientSet)
+	if err != nil {
+		klog.Errorf("Error  getting information on the local node: %s", err.Error())
+		return usrForceMssVal
+	}
+
+	usrForceMssVal, _ = strconv.Atoi(localNode.GetAnnotations()[v1.ForceTCPMssClamping])
```
tpantelis: We should check and log the return error instead of ignoring it. Also `Atoi` returns an error for an empty string so handle that.

```suggestion
        tcpMssStr := localNode.GetAnnotations()[v1.TCPMssValue]
        if tcpMssStr == "" {
            return 0
        }
        
	tcpMssValue, err := strconv.Atoi(tcpMssStr)
	if err != nil {
	    klog.Errorf("Error parsing %q annotation", v1.TCPMssValue)
	}
```

pkg/routeagent_driver/main.go
```diff
@@ -152,3 +155,23 @@ func annotateNode(clusterCidr []string, cfg *restclient.Config) error {
 
 	return nil
 }
+
+func getUsrForceMssVal(cfg *restclient.Config) int {
```
```suggestion
func getTCPMssValue(cfg *restclient.Config) int {
```

pkg/routeagent_driver/main.go
```diff
@@ -152,3 +155,23 @@ func annotateNode(clusterCidr []string, cfg *restclient.Config) error {
...
+	if err != nil {
+		klog.Errorf("Error building clientset: %s", err.Error())
+		return usrForceMssVal
+	}
+
+	localNode, err := node.GetLocalNode(k8sClientSet)
+	if err != nil {
+		klog.Errorf("Error  getting information on the local node: %s", err.Error())
+		return usrForceMssVal
+	}
```
```suggestion
	k8sClientSet, err := kubernetes.NewForConfig(cfg)
	if err != nil {
		klog.Errorf("Error building clientset: %s", err.Error())
		return 0
	}

	localNode, err := node.GetLocalNode(k8sClientSet)
	if err != nil {
		klog.Errorf("Error  getting information on the local node: %s", err.Error())
		return 0
	}
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -274,29 +276,54 @@ func (h *mtuHandler) Stop(uninstall bool) error {
...
+		if strings.Contains(rule, strings.Join(ruleSpecSource, " ")) || strings.Contains(rule, strings.Join(ruleSpecDest, " ")) {
+			isPresent = true
+			break
+		}
+	}
+
+	if !isPresent {
+		if err := h.ipt.ClearChain(constants.MangleTable, constants.SmPostRoutingChain); err != nil {
+			klog.Errorf("Error flushing iptables chain %q of %q table: %v", constants.SmPostRoutingChain,
+				constants.MangleTable, err)
```
sridhargaddam, line 327: Should we return an error here?

pkg/routeagent_driver/main.go
```diff
@@ -152,3 +155,23 @@ func annotateNode(clusterCidr []string, cfg *restclient.Config) error {
 
 	return nil
 }
+
+func getUsrForceMssVal(cfg *restclient.Config) int {
+	usrForceMssVal := 0
+
+	k8sClientSet, err := kubernetes.NewForConfig(cfg)
```
sridhargaddam: I think we can directly pass the k8sClientSet as an argument to this API instead of constructing it again here. 
i.e., in the caller of this API, please create the k8sClientSet and use it both for this API as well as for annotateNode() methods.

pkg/routeagent_driver/main.go
sridhargaddam: Since this is an user input, we should validate if there is any error in this conversion.

pkg/routeagent_driver/main.go
```diff
@@ -152,3 +155,23 @@ func annotateNode(clusterCidr []string, cfg *restclient.Config) error {
...
+	if err != nil {
+		klog.Errorf("Error building clientset: %s", err.Error())
+		return usrForceMssVal
+	}
+
+	localNode, err := node.GetLocalNode(k8sClientSet)
+	if err != nil {
+		klog.Errorf("Error  getting information on the local node: %s", err.Error())
```
sridhargaddam:
```suggestion
		klog.Errorf("Error getting information on the local node: %s", err.Error())
```

sridhargaddam (changes requested):

pkg/apis/submariner.io/v1/types.go
sridhargaddam: I'm fine with this. We can even use "submariner.io/tcp-mss-clamp-value"

tpantelis: yeah - "TCP mss clamping" describes the action  - this represents the value.

pkg/routeagent_driver/main.go
yboaron: 10x

pkg/routeagent_driver/main.go
yboaron: Yeah, good point

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: Hmm, maybe we can log warning or something like that and create the new iptables rules

pkg/apis/submariner.io/v1/types.go
yboaron: I'm not good at choosing names :-( , let's use 'tcp-clamp-mss' as u proposed

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: Good point, I'll take care of that.
10x

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
sridhargaddam: Ack, warning should be fine. We should also add a check to see that there are indeed some rules in the chain otherwise this warning will be displayed on a fresh setup as well. 

i.e., if !isPresent && len(rules) > 0

pkg/routeagent_driver/main.go
```diff
@@ -134,21 +141,39 @@ func init() {
...
+	err := cniapi.AnnotateNodeWithCNIInterfaceIP(nodeName, k8sClientSet, clusterCidr)
 	if err != nil {
 		return errors.Wrap(err, "error annotating node with CNI interface IP")
 	}
 
 	return nil
 }
+
+func getTCPMssValue(k8sClientSet *kubernetes.Clientset) int {
+	usrForceMssVal := 0
```
sridhargaddam: This is used only in line 164 but not in Line 170. We can either remove it all-together or keep it consistent.

pkg/routeagent_driver/main.go
```diff
@@ -134,21 +141,39 @@ func init() {
...
+
+	tcpMssStr := localNode.GetAnnotations()[v1.TCPMssValue]
+
+	if tcpMssStr == "" {
+		return 0
+	}
+
+	tcpMssValue, err := strconv.Atoi(tcpMssStr)
+	if err != nil {
+		klog.Errorf("Error parsing %q annotation", v1.TCPMssValue)
```
sridhargaddam, line 173: If there is an error, we are still returning the tcpMssValue in Line178. I think strconv.Atoi in such cases might return 0 for tcpMssValue, but it could be misleading for anyone looking at the code. Please fix it.

sridhargaddam (changes requested):

pkg/routeagent_driver/main.go
tpantelis: We should remove it.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
yboaron: Done

tpantelis (approved):

sridhargaddam (approved):

sridhargaddam:
@yboaron @skitt IMO we should allow user to override the MTU if required. I feel we should backport this PR to both release-0.13 and release-0.12 branches. What are your thoughts?

yboaron:
> @yboaron @skitt IMO we should allow user to override the MTU if required. I feel we should backport this PR to both release-0.13 and release-0.12 branches. What are your thoughts?

We have this PR in 0.13 , we didn't backport it to 0.12.
Hmm, we can backport it to 0.12 though I'm not sure how critical is this fix .
If we plan to release 0.12.3 we can backport it.

sridhargaddam:
> We have this PR in 0.13 , we didn't backport it to 0.12. Hmm, we can backport it to 0.12 though I'm not sure how critical is this fix . If we plan to release 0.12.3 we can backport it.

Since we are hardcoding the MTU, having an option to override (if necessary) is always helpful if users run into issues (or want to use a more optimal MTU). We can backport it to release-0.12 branch and it will be included in the next 0.12.3 whenever we make a release.

yboaron:
> > We have this PR in 0.13 , we didn't backport it to 0.12. Hmm, we can backport it to 0.12 though I'm not sure how critical is this fix . If we plan to release 0.12.3 we can backport it.
> 
> Since we are hardcoding the MTU, having an option to override (if necessary) is always helpful if users run into issues (or want to use a more optimal MTU). We can backport it to release-0.12 branch and it will be included in the next 0.12.3 whenever we make a release.

OK then , let's backport it for 0.12

PR #1885, opened 2022-06-23
Add grep to Globalnet and RouteAgent images

yboaron:
In order for iptables-wrapper to work properly we
 need to make sure that grep is included in the image.

  Fixes: https://github.com/submariner-io/submariner/issues/1883

PR #1887, opened 2022-06-23
Bump to Fedora 36

### LINKED ISSUE #1886: Bump submariner images to F35 (or F36)
Currently the images are on F34 which is EOL since 2022-06-07 - https://docs.fedoraproject.org/en-US/releases/eol/

We should bump to F35 (or F36) to keep images up to date and secure

skitt:
See individual commits.

Depends on #1885.
Fixes #1886.

### DIFF
--- a/package/Dockerfile.submariner-gateway
+++ b/package/Dockerfile.submariner-gateway
@@ -23,7 +23,7 @@ COPY package/dnf_install /
 # kmod is required so that libreswan can load modules
 RUN /dnf_install -a ${TARGETPLATFORM} -v ${FEDORA_VERSION} -r /output/gateway \
     glibc bash glibc-minimal-langpack coreutils-single \
-    libcurl-minimal iproute iptables iptables-nft libreswan kmod
+    libcurl-minimal iproute libreswan kmod
 
 FROM --platform=${TARGETPLATFORM} scratch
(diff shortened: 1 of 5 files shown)


skitt:
This fails because of `iptables` changes.

nyechiel:
@skitt are we planning to get this done for 0.13? The PR is up for review in the 0.13 board but the linked issue is in the future backlog.

skitt:
> @skitt are we planning to get this done for 0.13? The PR is up for review in the 0.13 board but the linked issue is in the future backlog.

We can discuss it in the calls next week, but I’d like to get this into 0.13 to avoid a release with already-obsolete base images.

mkolesnik (approved):
LGTM

nyechiel:
@sridhargaddam can you review/approve please?

dfarrell07 (commented):
In a commit message, you say

> We can do the same parsing using bash's builting getopt, so do that
instead.

I think you mean `getopts` here. As we're actually moving from `getopt`, I think the difference between the two is significant enough to fix.

> We can do the same parsing using bash's built-in getopts, so do that
instead.

dfarrell07 (approved):

sridhargaddam (approved):
LGTM

PR #1903, opened 2022-06-30
Restart health check pinger in case it failed

### LINKED ISSUE #1857: Sometimes remote endpoint RTT is reported as '0'
I have deployed SM clustersset with GN on aws, gcp and azure, and sometimes RTT is reported as '0 'even though the endpoint is successfully connected and round trip is ~70mSec.

The following error is reported in GW crd [1] while I can ping manually the remote endpoint [2].

Restarting the gw pod solved the problem.
Attached is GW's pod log [gwpod.txt](https://github.com/submariner-io/submariner/files/8807027/gwpod.txt)

[1]
status:
  connections:
  - endpoint:
      backend: libreswan
      backend_config:
        natt-discovery-port: "4490"
        preferred-server: "true"
        preferred-server-timestamp: "1654000314"
        udp-port: "4500"
        using-loadbalancer: "true"
      cable_name: submariner-cable-gcp-a-10-0-128-2
      cluster_id: gcp-a
      healthCheckIP: 242.2.255.254
      hostname: yblb-7sv57-worker-b-7h5tq.c.gc-acm-dev.internal
      nat_enabled: true
      private_ip: 10.0.128.2
      public_ip: 34.148.193.38
      subnets:
      - 242.2.0.0/16
    latencyRTT:
      average: 0s
      last: 0s
      max: 0s
      min: 0s
      stdDev: 0s
    status: connected
    statusMessage: 'Failed to run the pinger for the remote endpoint IP "242.2.255.254":
      write ip4 0.0.0.0->242.2.255.254: sendto: object is remote'
    usingIP: 34.148.193.38
    usingNAT: true

[2]

sh-4.4# ping 242.2.255.254
PING 242.2.255.254 (242.2.255.254) 56(84) bytes of data.
64 bytes from 242.2.255.254: icmp_seq=1 ttl=64 time=71.5 ms
64 bytes from 242.2.255.254: icmp_seq=2 ttl=64 time=71.3 ms
64 bytes from 242.2.255.254: icmp_seq=3 ttl=64 time=71.3 ms

yboaron:
When a health check pinger starts running before the
relevant routes configured on GW node, it fails with
[1] error and health check functionality does not work for run remote EP.

This PR restarts the pinger after timeout in case it fails to run.

[1]
E0629 13:34:50.788674       1 pinger.go:114] Unable to start pinger
for IP "242.0.255.254": error running the pinger: write ip4
0.0.0.0->242.0.255.254: sendto: object is remote

Fixes: https://github.com/submariner-io/submariner/issues/1857

### DIFF
--- a/pkg/cableengine/healthchecker/pinger.go
+++ b/pkg/cableengine/healthchecker/pinger.go
@@ -29,7 +29,10 @@ import (
 	"k8s.io/klog"
 )
 
-const Privileged = true
+const (
+	Privileged         = true
+	doPingRetryTimeout = 5
+)
 
 var (
 	defaultMaxPacketLossCount uint = 5
@@ -111,8 +114,8 @@ func (p *pingerInfo) Start() {
 			default:
 				err := p.doPing()
 				if err != nil {
-					klog.Errorf("Unable to start pinger for IP %q: %v", p.ip, err)
-					return
+					klog.Errorf("Unable to start pinger for IP %q: %v, retry in %d seconds", p.ip, err, doPingRetryTimeout)
+					time.Sleep(doPingRetryTimeout * time.Second)
 				}
 			}
 		}


nyechiel:
@yboaron how confident are you with this fix? Should we get it into 0.13?

tpantelis (approved):

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -109,10 +112,12 @@ func (p *pingerInfo) Start() {
 			case <-p.stopCh:
 				return
 			default:
-				err := p.doPing()
-				if err != nil {
-					klog.Errorf("Unable to start pinger for IP %q: %v", p.ip, err)
-					return
+				for {
```
sridhargaddam: @yboaron @tpantelis 
I'm probably missing something here. We are now using nested for loops. 
Even if we do not have a for loop in L115, because of the outer for loop at L110, the API (p.doPing) would be executed isn't it? Main difference would be the sleep in L119.

Can you please explain the need for introducing the for loop in Line 115?

sridhargaddam (commented):
Please clarify

pkg/cableengine/healthchecker/pinger.go
sridhargaddam: and we are not returning now when there is an error. Do we still need the for loop?

tpantelis: you're right - we really don't need the inner loop.

yboaron: Yep, you are right.
Deleted the inner loop.
10x

yboaron:
> @yboaron how confident are you with this fix? Should we get it into 0.13

I'm pretty confident about this PR altough it's hard to reproduce the problem.
I think it would be nice to get it in 0.13 though we've a simple workaround for this fix (restart GW pod)

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -111,8 +114,11 @@ func (p *pingerInfo) Start() {
 			default:
 				err := p.doPing()
 				if err != nil {
-					klog.Errorf("Unable to start pinger for IP %q: %v", p.ip, err)
-					return
+					klog.Errorf("Unable to start pinger for IP %q: %v, retry in %d seconds", p.ip, err, doPingRetryTimeout)
+					time.Sleep(doPingRetryTimeout * time.Second)
+				} else {
+					klog.Infof("Successfully started pinger for IP %q ", p.ip)
+					break
```
tpantelis: I think we should make the minimal change here and just retry (ie not return) on failure. A nil return from `doPing` means it ran to completion so the log message here isn't meaningful. Also let L112 above decide to exit the loop instead of breaking here.

tpantelis (changes requested):

pkg/cableengine/healthchecker/pinger.go
yboaron: Sounds reasonable, do we report the pinger status(working/not working) somewhere ? cr.status?

tpantelis (approved):

sridhargaddam (approved):

nyechiel:
@yboaron @sridhargaddam my understanding is that this one will need to be backported to both `release-0.13` and `release-0.13`. Yossi, can you please handle the backports once the PR is merged?

sridhargaddam:
> @yboaron @sridhargaddam my understanding is that this one will need to be backported to both `release-0.13` and `release-0.13`. Yossi, can you please handle the backports once the PR is merged?

@yboaron a small typo. I'm sure Nir meant, `release-0.13` and `release-0.12` branches.

nyechiel:
> > @yboaron @sridhargaddam my understanding is that this one will need to be backported to both `release-0.13` and `release-0.13`. Yossi, can you please handle the backports once the PR is merged?
> 
> @yboaron a small typo. I'm sure Nir meant, `release-0.13` and `release-0.12` branches.

Yes, sorry!

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -111,8 +114,8 @@ func (p *pingerInfo) Start() {
 			default:
 				err := p.doPing()
 				if err != nil {
-					klog.Errorf("Unable to start pinger for IP %q: %v", p.ip, err)
-					return
+					klog.Errorf("Unable to start pinger for IP %q: %v, retry in %d seconds", p.ip, err, doPingRetryTimeout)
+					time.Sleep(doPingRetryTimeout * time.Second)
```
mkolesnik, line 118: Coming late, but shouldnt it die after a few attempts?

This way a rogue pod can run forever if it constantly fails, which is quite a different behavior that what we had before (unless I'm missing something)

sridhargaddam: The health-checker (aka pinger) is performed from the submariner-gateway pod itself and its objective is to run forever. Even during a success scenario pinger runs at regular interval and reports the RTT which can be seen in `subctl show ...` command.

yboaron: It runs as go func ( ~ separate thread) as part of GW pod , so k8s scheduler won't help us here if we'll exit this function sm-gw pod will continue running.

We'll be stuck without pinger running and no component will restart it (I think that was the bug root cause). 


Another direction could be to expose the pinger health somehow and let k8s scheduler monitor and restart the pod in case of failure (using liveness check)

nyechiel:
@yboaron can you please backport this to `release-0.13` and link the PR as "Depends on" to https://github.com/submariner-io/releases/issues/418?

pkg/cableengine/healthchecker/pinger.go
mkolesnik: OK thanks this is clear now

PR #1948, opened 2022-08-03
Set ReadHeaderTimeout to 60s to mitigate Slowloris

dfarrell07:
Go 1.18 added this new timeout and gosec now fails if it's unset.

Leaving this timeout unset exposes the server to a Slowloris attack,
where an attacker keeps many connections open by periodic updates to
HTTP request headers. This can eventually result in being unable to open
new connections because of too many open files.

The 60s value is copied from nginx:
nginx.org/en/docs/http/ngx_http_core_module.html#client_header_timeout

Relates-to: submariner-io/shipyard#900

### DIFF
--- a/main.go
+++ b/main.go
@@ -318,7 +318,7 @@ func submarinerClusterFrom(submSpec *types.SubmarinerSpecification) *types.Subma
 }
 
 func startHTTPServer() *http.Server {
-	srv := &http.Server{Addr: ":8080"}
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

tpantelis:
I verified this fixes the linting error.

dfarrell07:
And E2E is passing, so I'll go ahead and send the other similar PRs to other repos.

sridhargaddam (approved):

skitt (approved):

PR #1958, opened 2022-08-10
Add gateway state restore to redundancy tests

MaxBab:
The redundancy tests should restore the gateways to the initial state
based on the state gathered during "BeforeEach" execution.

Depends on https://github.com/submariner-io/shipyard/pull/914

Signed-off-by: Maxim Babushkin <mbabushk@redhat.com>


MaxBab:
After discussion on slack channel, a different approach of test flow will be implemented.
Closing this PR.

PR #1980, opened 2022-08-24
Minimize GHA permissions

dfarrell07:
Set the GitHub Actions token permission to null in most workflows.

This results in:

GITHUB_TOKEN Permissions
  Metadata: read

The default permissions, used without the null override, are:

GITHUB_TOKEN Permissions
  Actions: read
  Checks: read
  Contents: read
  Deployments: read
  Discussions: read
  Issues: read
  Metadata: read
  Packages: read
  Pages: read
  PullRequests: read
  RepositoryProjects: read
  SecurityEvents: read
  Statuses: read

Two jobs require non-null permissions to function.

The Anchore vulnerability scan GHA that reports results on merges needs
security-events write permission to publish SARIF reports using
github/codeql-action/upload-sarif. Fails with HTTP 403 otherwise.

The dependent issues GHA needs PR/issues write permissions to add/remove
`dependent` labels. It needs status write permission to block/unblock
PRs when dependencies are missing/met. Fails with HttpError otherwise.

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -4,6 +4,8 @@ name: Branch Checks
 on:
   pull_request:
 
+permissions: {}
+
 jobs:
   target_branch:
     name: PR targets branch
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -7,6 +7,8 @@ on:
       - 'CODEOWNERS'
       - 'CODEOWNERS.in'
 
+permissions: {}
+
 jobs:
   updated:
     name: Up-to-date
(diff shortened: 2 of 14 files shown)


tpantelis (approved):

dfarrell07:
FYI I verified the dependent issues PR write permission here:

https://github.com/dfarrell07/lighthouse/runs/7997670733?check_suite_focus=true#step:2:19

and the statues permission here:

https://github.com/dfarrell07/lighthouse/runs/7998606414?check_suite_focus=true#step:2:22

skitt (approved):

dfarrell07:
The next version of this PR (https://github.com/submariner-io/admiral/pull/405) has a better commit message and answers a question that had be bothering me about why we need write permissions when we seem to see read as the default.

dfarrell07:
> The next version of this PR ([submariner-io/admiral#405](https://github.com/submariner-io/admiral/pull/405)) has a better commit message and answers a question that had be bothering me about why we need write permissions when we seem to see read as the default.

tl;dr we're reducing the permissions even more than this commit message makes it seem

PR #1989, opened 2022-08-27
Use append unique instead of update

### LINKED ISSUE #1981: OVN+Globalnet: Missing rule in SUBMARINER-POSTROUTING chain
**What happened**:

When Globalnet is deployed with OVN, the `SUBMARINER-POSTROUTING` chain sometimes does not have the rule to forward the traffic to `SUBMARINER-GN-EGRESS` chain. It was seen that this rule is present in one of the clusters but missing on the second cluster in a KIND+OVN+Globalnet deployment.

This rule is important as its an entry point to the Globalnet pipeline where the SNAT operation is performed on the Gateway node.

```
iptables -I SUBMARINER-POSTROUTING 1 -s 0.0.0.0/0 -d 0.0.0.0/0 -j SUBMARINER-GN-EGRESS -t nat 
```

aswinsuryan:
When the route-agent after the globalnet controller programs the flow,
it replaces the flows added by globalnet controller. Now append unique adds
the flow and the flows if added by the globalnet controller will get a
higher priority.
Signed-off-by: Aswin Suryanarayanan <aswinsuryan@gmail.com>

Fixes : #1981


skitt (approved):

tpantelis (approved):

sridhargaddam (changes requested):
Can you check if we are deleting the rules when a remote endpoint is removed?

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -155,22 +155,16 @@ func (ovn *Handler) setupForwardingIptables() error {
...
-		"error updating %q rules", constants.SmPostRoutingChain)
-}
-
-func (ovn *Handler) getNoMasqueradRuleSpecs() [][]string {
-	var rules [][]string
-
 	for _, endpoint := range ovn.remoteEndpoints {
 		for _, subnet := range endpoint.Spec.Subnets {
-			rules = append(rules, []string{"-d", subnet, "-j", "ACCEPT"})
+			err := ovn.ipt.AppendUnique("nat", constants.SmPostRoutingChain, []string{"-d", subnet, "-j", "ACCEPT"}...)
```
sridhargaddam: It looks like we are not deleting the iptable rules when the remote endpoint is deleted.

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -144,6 +144,12 @@ func (ovn *Handler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
 	}
 
 	if ovn.isGateway {
+		for _, subnet := range endpoint.Spec.Subnets {
+			if err = ovn.addNoMasqueradeIPTables(subnet); err != nil {
+				return errors.Wrap(err, "error adding no-masquerade rules")
```
sridhargaddam: Please include the subnet (or the rule) for which the error occurred.

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -186,6 +192,12 @@ func (ovn *Handler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 	}
 
 	if ovn.isGateway {
+		for _, subnet := range endpoint.Spec.Subnets {
+			if err = ovn.removeNoMasqueradeIPTables(subnet); err != nil {
+				return errors.Wrap(err, "error removing no-masquerade rules")
```
sridhargaddam: same

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -197,6 +209,13 @@ func (ovn *Handler) TransitionToNonGateway() error {
 	defer ovn.mutex.Unlock()
 
 	ovn.isGateway = false
+	for _, endpoint := range ovn.remoteEndpoints {
+		for _, subnet := range endpoint.Spec.Subnets {
+			if err := ovn.removeNoMasqueradeIPTables(subnet); err != nil {
+				return errors.Wrap(err, "error removing no-masquerade rules")
```
sridhargaddam: same

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -207,5 +226,13 @@ func (ovn *Handler) TransitionToGateway() error {
 
 	ovn.isGateway = true
 
+	for _, endpoint := range ovn.remoteEndpoints {
+		for _, subnet := range endpoint.Spec.Subnets {
+			if err := ovn.addNoMasqueradeIPTables(subnet); err != nil {
+				return errors.Wrap(err, "error adding no-masquerade rules")
```
sridhargaddam: same

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -144,6 +144,12 @@ func (ovn *Handler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
 	}
 
 	if ovn.isGateway {
+		for _, subnet := range endpoint.Spec.Subnets {
```
sridhargaddam, line 147: We will still have to handle the out of order events which can trigger race conditions. Basically, we have to look at the lastProcessedTime and process/ignore the event similar to how we are doing in kubeproxy handler - https://github.com/submariner-io/submariner/blob/0c3e043a374c6a793c564a9d98fa8fb8990b310e/pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go#L108

But anyway, we can use a separate PR for that.

sridhargaddam (changes requested):

pkg/routeagent_driver/handlers/ovn/handler.go
aswinsuryan: sure will do as a follow up pr

pkg/routeagent_driver/handlers/ovn/handler.go
aswinsuryan: Done

pkg/routeagent_driver/handlers/ovn/handler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/handler.go
aswinsuryan: Done.

sridhargaddam (approved):
Thanks for addressing the comments @aswinsuryan

PR #2008, opened 2022-09-12
Cleanup iptable rules programmed by OVN RA plugin

### LINKED ISSUE #1868: Cleanup OVN resources during uninstall
Cleanup all OVN resources during uninstall if OVNKubernetes is the CNI

**Why is this needed**:
During install, networkplugin-syncer and route-agent add some resources to OVN. Need to make sure these are all cleaned up properly during an uninstall.

sridhargaddam:
After running uninstall on an OVN Cluster, it was seen that iptable chains in NAT as well as FILTER tables are still present on the nodes. This PR fixes it.

Fixes: https://github.com/submariner-io/submariner/issues/1868

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/uninstall.go
+++ b/pkg/routeagent_driver/handlers/ovn/uninstall.go
@@ -60,10 +60,8 @@ func (ovn *Handler) Stop(uninstall bool) error {
 			constants.RouteAgentHostNetworkTableID, err)
 	}
 
-	err = ovn.cleanupForwardingIptables()
-	if err != nil {
-		klog.Errorf("Error deleting iptable rules %v", err)
-	}
+	ovn.flushAndDeleteIPTableChains(constants.FilterTable, constants.ForwardChain, forwardingSubmarinerFWDChain)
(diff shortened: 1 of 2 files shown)


pkg/routeagent_driver/handlers/ovn/uninstall.go
```diff
@@ -60,10 +60,8 @@ func (ovn *Handler) Stop(uninstall bool) error {
 			constants.RouteAgentHostNetworkTableID, err)
 	}
 
-	err = ovn.cleanupForwardingIptables()
```
tpantelis, line 63: So we don't need to run this?

sridhargaddam: No. It only does partial cleanup which is necessary when there is a Gateway failover.

tpantelis (approved):

nyechiel (approved):

aswinsuryan (approved):

sridhargaddam:
Backport PR to release-0.13 branch: https://github.com/submariner-io/submariner/pull/2026

Issue #2013, opened 2022-09-13
Refactor gateways (redundancy) tests


MaxBab:
The GatewayFailOver and GatewayPodRestart tests are part of the "redundancy" tests group.
The "gateways" tests implementation allows execution of the tests only on kind environment and could not be executed on real OCP environments.

As discussed in the following threads [1] [2]:
Existing "gateway-failover" test is not a real scenario.
In the existing scenario, the gateway node that should be failed, only removes the "submariner.io/gateway: "true"" label from the node and then starts a new gateway node.
This will not work in the real environment because of the following:
  - The remove of the label will not work, because the `MachineSet` of the cluster will set the label back at the same moment.
    The `MachineSet` is responsible for keeping the gateway node in the same condition state.
  - Creation of the pure new gateway node will not work as well, because the ports for the new gateway node will not be opened and as a result the gateway node will remain unreachable.

Due to the above, the following changes should be done as part of the refactoring:
- The "gateway-failover" test will require 2 gateway nodes existing in the cluster to be executed.
  This is a more customer related scenario.
- The kind testing environment should reflect the changes and be able to continue "gateway-failover" test after the refactoring.
- While in the kind environment, the failover process will remain the same, in the real k8s/OCP env, we need to crash the gateway node in order to perform a real failover. Crashing the gateway node is another way to properly test the failover scenario.
- Adapt if required other redundancy tests to meet the refactoring.

[1] - https://coreos.slack.com/archives/C0134E73VH6/p1654177190977899
[2] - https://coreos.slack.com/archives/C0134E73VH6/p1660056801840349

PR #2014, opened 2022-09-14
Refactor GatewayFailOver test

MaxBab:
Current implementation of the test allows execution only on "kind" environment.
In order to execute the test on any real environment, the following changed required:

- The test will run only when 2 GW node will be available
- The test will search for the cluster with 2 GW nodes and set it as
      primary cluster for test execution
- If found, will execute, otherwise will skip
- The failover step will be differ between "kind" and real environment. In k8s/ocp environment, the GW node will be crashed to simulate a real failover, while in "kind" it will use current step (change the gateway label to false), as crashing the gw node will crash the whole machine which holds the "kind" env.

Details of refactoring:
https://github.com/submariner-io/submariner/issues/2013

Depends On https://github.com/submariner-io/shipyard/pull/945 Depends On https://github.com/submariner-io/shipyard/pull/946 Depends On https://github.com/submariner-io/shipyard/pull/947

Signed-off-by: Maxim Babushkin <mbabushk@redhat.com>

PR #2018, opened 2022-09-15
Update RouteAgentRestart test

MaxBab:
The following test relates to the e2e in "kind" environment only. Due to refactor of GatewayFailOver test, "kind" environment will have all working nodes serving as gateway nodes.
This is done to save the resources.

As a result, the test will fail because it's looking for a non gateway worker node and it will missing.

Since the RouteAgent pod is running on every cluster node, this change will search for the RouteAgent pod on any cluster node in case the non GW node is not found.

Refactoring details:
https://github.com/submariner-io/submariner/issues/2013

Depends On https://github.com/submariner-io/shipyard/pull/948

Signed-off-by: Maxim Babushkin <mbabushk@redhat.com>

### DIFF
--- a/test/e2e/redundancy/route_agent_restart.go
+++ b/test/e2e/redundancy/route_agent_restart.go
@@ -48,9 +48,14 @@ func testRouteAgentRestart(f *subFramework.Framework, onGateway bool) {
 	clusterBName := framework.TestContext.ClusterIDs[framework.ClusterB]
 
 	nodes := f.FindNodesByGatewayLabel(framework.ClusterA, onGateway)
+	// Testing GatewayFailover test in "kind" env will require
+	// all nodes to serve as GW nodes due to resources limitation.
+	// It means that current test may fail as will not find non GW worker.
+	// In case required node will not be found, look for any non GW/worker
+	// node. And since the "route agent" pod runs on every node, it will work.
 	if len(nodes) == 0 && !onGateway {
-		framework.Skipf("Skipping the test as cluster %q doesn't have any suitable non-gateway nodes...", clusterAName)
-		return
+		By(fmt.Sprintf("No Non Gateway worker nodes found in %q. Looking for non worker nodes", clusterAName))
+		nodes = f.FindAnyNonGatewayRouteAgentPodNodes(framework.ClusterA)
 	}
 
 	By(fmt.Sprintf("Found node %q on %q", nodes[0].Name, clusterAName))


test/e2e/redundancy/route_agent_restart.go
```diff
@@ -48,9 +48,11 @@ func testRouteAgentRestart(f *subFramework.Framework, onGateway bool) {
 	clusterBName := framework.TestContext.ClusterIDs[framework.ClusterB]
 
 	nodes := f.FindNodesByGatewayLabel(framework.ClusterA, onGateway)
+	// This will happen only in "kind" environment because of the flow of
+	// testing gateway failover test
```
sridhargaddam: Not sure i understand this statement. Can you please eloborate, why this will happen only in KIND and not in other deployments? 

Also, gateway failover test even if its executed prior to this, should reset the state of the clusters back to the same state as how it was prior to the test-case, if this is not happening, it should be fixed.
Similarly, IMO these tests can be executed in any order, so this test-case can be executed prior to gateway-failover tests.

MaxBab: Hi @sridhargaddam 

As part of the GatewayFailover test refactoring, I'm changing the flow of the test to the following:
One of the clusters will have 2 gateway nodes in active/passive state.
During the failover step, the active gateway node will be crashed and passive gateway become active.

In order to replicate the same test flow in "kind" environment, I'm creating the following infrastructure:
First cluster will have 2 nodes - master and worker.
The worker will serve as GW node.

Second cluster will have 3 nodes - master and 2 worker nodes.
Both worker nodes will serve as GW nodes to replicate the test flow change I described above.

With this infrastructure, the RouteAgentRestart test will not be able to find a non GW node as it request.
This scenatio will never happen in non "kind" environment because there will always be a non GW worker node.

That's why I need this workaround for the "kind" env only.

Hope my explanation made it clear.

Maybe I need to make a note in a test more understandable.

MaxBab: @sridhargaddam 

I have updated the comment to make it more clear.

test/e2e/redundancy/route_agent_restart.go
```diff
@@ -48,9 +48,14 @@ func testRouteAgentRestart(f *subFramework.Framework, onGateway bool) {
...
+	// Testing GatewayFailover test in "kind" env will require
+	// all nodes to serve as GW nodes due to resources limitation.
+	// It means that current test may fail as will not find non GW worker.
+	// In case required node will not be found, look for any non GW/worker
+	// node. And since the "route agent" pod runs on every node, it will work.
 	if len(nodes) == 0 && !onGateway {
-		framework.Skipf("Skipping the test as cluster %q doesn't have any suitable non-gateway nodes...", clusterAName)
-		return
+		By(fmt.Sprintf("No Non Gateway worker nodes found in %q. Looking for non worker nodes", clusterAName))
+		nodes = f.FindAnyNonGatewayRouteAgentPodNodes(framework.ClusterA)
```
tpantelis, line 58: Since the route agent runs on every node, `FindAnyNonGatewayRouteAgentPodNodes` is effectively the same as `FindNodesByGatewayLabel` where the GW label isn't set to true.  It seems the changes here are due to the recent change to `FindNodesByGatewayLabel` where it no longer treats the absence of the label as _false_. I'm still a bit wary of that change. I think we should restore the previous behavior. With the new behavior, not returning the node if the label is absent essentially implies it is a GW node which is incorrect. I know that `GatewayCleanup` relies on the specific _false_ value but I think that's a bit fragile as it's possible a node's GW label could've been set to _false_  external to the test. I think a safer approach is to have `DoFailover` simply save off the GW node name in the `Framework` for `GatewayCleanup` to use. So with `FindNodesByGatewayLabel` restored to its previous behavior, we don't need `FindAnyNonGatewayRouteAgentPodNodes`.

With that, we don't really need any changes here but perhaps we shouldn't assume `ClusterA` has the desired GW node criteria. Instead, similar to the GW failover test, we can search for the desired cluster:

```Go
var nodes []*v1.Node
var clusterA ClusterIndex

for idx := range TestContext.ClusterIDs {
    nodes = f.FindNodesByGatewayLabel(idx, onGateway)
    if len(nodes) >= 0 {
        clusterA = idx
        break
    }
}

if len(nodes) == 0  {
    s := ""
    if !onGateway {
        s = "non-"
    }

    framework.Skipf("Skipping the test as no cluster has any suitable %sgateway nodes...", s)
    return
}

clusterB := framework.FindOtherClusterIndex(clusterA)
```

tpantelis (changes requested):

test/e2e/redundancy/route_agent_restart.go
MaxBab: Hi @tpantelis 

I already proposed a patch to save and restore a state of the gateway nodes, but you said that it's not a way we should go.
https://github.com/submariner-io/shipyard/pull/914

In addition, this way of saving/restoring a state relevant only for "kind" environment.
Any regular k8s/ocp env will not need it.

Maybe it's better to add additional parameter `FindNodesByGatewayLabel` function?
Currently is has the following parameters: `FindNodesByGatewayLabel(cluster ClusterIndex, isGateway bool)`.
So, we could add another one to control if we would like to search for the absent label nodes or not.
Like: `FindNodesByGatewayLabel(cluster ClusterIndex, isGateway, nonLabelNode bool)`.

It's a bit clumsy than creating a way of saving/restoring the state of the GW node only for "kind".

What do you think?

tpantelis: The discussion in https://github.com/submariner-io/shipyard/pull/914 is around not modifying the GW label and using other approaches. It was closed referencing a slack discussion but I apologize that I don't recall the details of that tho I do remember talking about saving state internally. 

So the issue is not with `FindNodesByGatewayLabel` but how to restore the GW label at test cleanup. I think searching for nodes with the label exactly equal to _false_ is fragile and possibly problematic as I mentioned. We could use another temporary label, eg `submariner.io/gwFailoverTest`, and use that to locate the node(s) to restore. Or save the state internally in the `Framework`. I think the latter is easiest.

MaxBab: By "another temporary label", you mean that test will set a label on the specific gw node in "kind" env and at the end will restore that gw state as per the label, and then remove the label from the node?

If that's the case, and you're ok with that, I would like to go with that approach.

tpantelis: yes that is what I meant.

MaxBab: Hi @tpantelis 

I just dig through the patches once again and remember what was the actual purpose of this change and adding the `FindAnyNonGatewayRouteAgentPodNodes` function.

Please, keep up with me.
First of all, to remind to myself as well, all these patches going around the `GatewayFailover` test refactoring to allow execution of redundancy tests on ocp environment.
The change of this test `RouteAgentRestart` and new function `FindAnyNonGatewayRouteAgentPodNodes` will serve only "kind" environment.
The reason behind it, is because of the `GatewayFailover` test refactoring, "kind" environment will have the following structure:
```
cluster1:
- control node
- worker (serve as GW node)
cluster2:
- control node
- worker (serve as GW node)
- worker (serve as GW node)
```

In order to simulate real scenario failover one of the clusters will have 2 GW nodes in active/passive state.
For "kind" environment is means that all existing worker nodes will act as GW nodes and there will be no non GW nodes available.
This is done to save the resource and not create additional worker node for "kind" environment.

With all the above, `RouteAgentRestart` test will run as it is now, without any refactoring, with no issues, because any real ocp based environment will have a non GW node.
But "kind" environment does not (resource saving).
So in order to make the `RouteAgentRestart` test being executable of "kind" environment, it needs to get modified.

Now, regarding the last change of `FindNodesByGatewayLabel` function:
https://github.com/submariner-io/shipyard/pull/945
With that change, the function will return only the nodes with the GW label exist.
It could be a "true" or "false" value of the label, but the label must exist.
It is used by the `GatewayCleanup` function.

Finally, if I will restore the functionality of the `FindNodesByGatewayLabel` function to return a non GW nodes as well, it will not help to execute the `RouteAgentRestart` test and it will fail in "kind" environment in any case.

The reason for it is the fact that `FindNodesByGatewayLabel` returns only "worker" nodes and filters out the "control" nodes.
https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/nodes.go#L41

But in "kind" environment, only "control plane" node could be tested with `RouteAgentRestart` with condition of GW node false.
Restoring the function and changing the `GatewayCleanup` function flow will not take any effect.

This is why I proposed a patch with the function that return any node with the "routeagent" pod, including the "control-plane" nodes.
https://github.com/submariner-io/shipyard/pull/948

Sorry for the long reading.
I wanted to make myself clear.

tpantelis: OK thanks for the explanation - that clarifies things. So the relevant code in `FindNodesByGatewayLabel` is:

```Go
// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
return KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
    LabelSelector: "!node-role.kubernetes.io/master",
})
```

So according to the comment the route agent won't be scheduled/run on a master node. If that's the case then the test won't be able to run on `kind` even with `FindAnyNonGatewayRouteAgentPodNodes`. But I'm assuming that's not the case since I assume you've been able to successfully run the test on `kind` where it finds a master node.

Given that, why don't we change `FindNodesByGatewayLabel` to include (ie not ignore) master nodes (perhaps prioritizing non-master nodes, ie put master nodes at the end of the returned array)?

MaxBab: Doesn't it look to you that adding "master" nodes to the returned array of `FindNodesByGatewayLabel` function, makes this function not reflecting the meaning of the function by its name.
When I'm looking at the `FindNodesByGatewayLabel` function, to me it looks like it should find to me the GW node, with label "true" or "false".

So maybe in this case, a difference function could suite better?
https://github.com/submariner-io/shipyard/pull/948/files

MaxBab: And by the way, a comment within the function relates to the GW functionality only and not the route-agent pod.

tpantelis: > Doesn't it look to you that adding "master" nodes to the returned array of `FindNodesByGatewayLabel` function, makes this function not reflecting the meaning of the function by its name. 

The function name indicates to find GW nodes - it doesn't say anything about "master"  or "worker" nodes. To me, the fact that it filters out "master" nodes is a hidden side effect not reflected in the name. In fact I wasn't aware of that (or had forgotten) until you pointed it out.

> When I'm looking at the `FindNodesByGatewayLabel` function, to me it looks like it should find to me the GW node, with label "true" or "false".

It's semantics and the way the function is named. The semantics we use for determining that a node is a GW is that the label has to be present and equal to "true", otherwise it isn't a GW. So using exact label match with "false" and not returning a node if the label is absent essentially implies it is a GW node which actually violates those semantics. In fact we could further simplify it by defining the criteria for a GW such that the label only needs to be present, regardless of the value. So perhaps interpret the function in terms of those semantics rather than the mechanics of searching by a label value. Perhaps refactoring it to two functions, `FindGatewayNodes` and `FindNonGatewayNodes`, which don't say anything about a label, would clarify the semantics.

MaxBab: Yes, I think two functions will make it more cleaner.
I like the idea.

Now, when thinking about the `FindNonGatewayNodes` function, I wonder, should we add a parameter to control if we would like to return only worker (non gw) nodes or both worker (non gw) and master nodes?
Generally I think we should because usually, master does prohibit creation of the pods on the master itself.
It does not count for the `route-agent` pod and this pod runs on every node in the cluster.

So, maybe we could name it like `func FindNonGatewayWorkerNodes(workersOnly bool)`.
If "false" it will return all non gw nodes from the cluster.
What do you think?

tpantelis: > Now, when thinking about the `FindNonGatewayNodes` function, I wonder, should we add a parameter to control if we would like to return only worker (non gw) nodes or both worker (non gw) and master nodes? Generally I think we should because usually, master does prohibit creation of the pods on the master itself. It does not count for the `route-agent` pod and this pod runs on every node in the cluster.
> 
> So, maybe we could name it like `func FindNonGatewayWorkerNodes(workersOnly bool)`. If "false" it will return all non gw nodes from the cluster. What do you think?

I think we should keep it simple and return all nodes. If we have a caller that needs to restrict it to only worker nodes then they can post-filter the returned nodes.

MaxBab: Ok, makes sense.
Will work on that.

Thanks.

MaxBab:
Closing as all necessary changes will be handled by the following patch.
https://github.com/submariner-io/submariner/pull/2014

PR #2053, opened 2022-10-11
Skip public-ip resolution in air-gapped clusters

sridhargaddam:
When AIR_GAPPED_DEPLOYMENT is set to true as an env' variable on the Gateway pod, this PR does the following.
1. It looks for gateway.submariner.io/public-ip=ipv4:\<address\> annotation on the gateway node. If the annotation is present, its value is used as public-ip in the local endpoint.
2. If the annotation is missing on the gateway node, an empty public-ip is configured in the local endpoint.

In both the above scenarios the gateway pod will avoid making any API/DNS/LB queries to resolve the public-ip associated with the gateway node. Also, the periodic public-ip watcher is disabled for air-gapped deployments.

### DIFF
--- a/main.go
+++ b/main.go
@@ -127,7 +127,10 @@ func main() {
 
 	submSpec.CableDriver = strings.ToLower(submSpec.CableDriver)
 
-	localEndpoint, err := endpoint.GetLocal(&submSpec, k8sClient)
+	airGapped := isAirGappedDeployment()
+	logger.Infof("AIR_GAPPED_DEPLOYMENT is set to %t", airGapped)
+
+	localEndpoint, err := endpoint.GetLocal(&submSpec, k8sClient, airGapped)
 	fatalOnErr(err, "Error creating local endpoint object")
 
 	cableEngine := cableengine.NewEngine(localCluster, localEndpoint)
@@ -170,7 +173,7 @@ func main() {
 
 	cableEngineSyncer.Run(stopCh)
 
-	publicIPWatcher := getPublicIPWatcher(&submSpec, k8sClient, submarinerClient, localEndpoint)
+	publicIPWatcher := getPublicIPWatcher(&submSpec, k8sClient, submarinerClient, localEndpoint, airGapped)
 
 	becameLeader := func(context.Context) {
 		if err = cableEngine.StartEngine(); err != nil {
@@ -264,6 +267,10 @@ func main() {
 	}
 }
 
+func isAirGappedDeployment() bool {
+	return os.Getenv("AIR_GAPPED_DEPLOYMENT") == "true"
+}
+
 func getCableHealthChecker(cfg *rest.Config, submSpec *types.SubmarinerSpecification) healthchecker.Interface {
 	var cableHealthchecker healthchecker.Interface
 	var err error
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -170,7 +173,7 @@ func main() {
 
 	cableEngineSyncer.Run(stopCh)
 
-	publicIPWatcher := getPublicIPWatcher(&submSpec, k8sClient, submarinerClient, localEndpoint)
+	publicIPWatcher := getPublicIPWatcher(&submSpec, k8sClient, submarinerClient, localEndpoint, airGapped)
```
tpantelis, line 176: Minor thing but instead of passing in the flag I think it makes more sense to just check it outside:

```suggestion
	var publicIPWatcher *endpoint.PublicIPWatcher
	if !airGapped {
	    publicIPWatcher = getPublicIPWatcher(&submSpec, k8sClient, submarinerClient, localEndpoint)
	 }
```

pkg/endpoint/public_ip.go
```diff
@@ -88,6 +96,47 @@ func getPublicIP(submSpec *types.SubmarinerSpecification, k8sClient kubernetes.I
 	return "", nil
 }
 
+func resolveIPInAirGappedDeployment(k8sClient kubernetes.Interface, namespace, config string, airGapped bool) (string, error) {
+	if !airGapped {
+		return "", errors.Errorf("this function should be called only for air-gapped deployments")
+	}
```
tpantelis: Since the function name indicates air gapped, we don't really need to pass in a flag. Plus the lone call site already checks for air gapped prior to calling.

```suggestion
func resolveIPInAirGappedDeployment(k8sClient kubernetes.Interface, namespace, config string) (string, error) {
```

pkg/endpoint/public_ip.go
```diff
@@ -88,6 +96,47 @@ func getPublicIP(submSpec *types.SubmarinerSpecification, k8sClient kubernetes.I
...
+			return "", errors.Errorf("invalid format for %q annotation: %q", v1.GatewayConfigPrefix+v1.PublicIP, config)
+		}
+
+		if parts[0] != v1.IPv4 {
+			continue
+		}
+
+		ip, err := resolvePublicIP(k8sClient, namespace, parts)
+
+		return ip, err
```
```suggestion
		return resolvePublicIP(k8sClient, namespace, parts)
```

pkg/endpoint/public_ip.go
```diff
@@ -88,6 +96,47 @@ func getPublicIP(submSpec *types.SubmarinerSpecification, k8sClient kubernetes.I
...
+	if !ok {
+		return "", errors.Errorf("unknown resolver %q in %q annotation", parts[0], v1.GatewayConfigPrefix+v1.PublicIP)
+	}
+
+	ip, err := method(k8sClient, namespace, parts[1])
+	if err == nil {
+		return ip, nil
+	}
+
+	return "", err
```
```suggestion
	return method(k8sClient, namespace, parts[1])
```

tpantelis (changes requested):

main.go
sridhargaddam: This is leading to the following error from the linter.
```
main.go:92:1: cognitive complexity 31 of func `main` is high (> 30) (gocognit)
```

In order to address the error, I will have to re-factor the main() method and since its not part of this feature, it might be worth doing it in a separate PR. Hope you are okay.

tpantelis: ok - I can refactor afterwards.

tpantelis (approved):

dfarrell07 (approved):

sridhargaddam:
I'll add the release-note as part of subctl PR change which includes the new flag.

PR #2087, opened 2022-10-27
Allow setting TCP MSS clamping value also for non GN

yboaron:
We noticed in some cases MTU issues even when
GN was disabled.

This PR allows the user to force a particular MSS clamping value 
also for non GN.

To force a particular MSS clamping value use the submariner.io/tcp-clamp-mss 
node annotation on Gateway nodes.

e.g. kubectl annotate node <node_name> submariner.io/tcp-clamp-mss=\<value>

Signed-off-by: Yossi Boaron <yboaron@redhat.com>

### DIFF
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
@@ -62,7 +62,7 @@ var logger = log.Logger{Logger: logf.Log.WithName("MTU")}
 
 func NewMTUHandler(localClusterCidr []string, isGlobalnet bool, tcpMssValue int) event.Handler {
 	forceMss := notNeeded
-	if isGlobalnet {
+	if isGlobalnet || tcpMssValue != 0 {
 		forceMss = needed
 	}
 


pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -62,7 +62,7 @@ var logger = log.Logger{Logger: logf.Log.WithName("MTU")}
 
 func NewMTUHandler(localClusterCidr []string, isGlobalnet bool, tcpMssValue int) event.Handler {
 	forceMss := notNeeded
-	if isGlobalnet {
+	if isGlobalnet || tcpMssValue != 0 {
```
sridhargaddam, line 65: Not in the scope of this PR, but we should probably add some validation on the MTU value that is configured by the user (i.e., to ensure that its in the valid reasonable range)

sridhargaddam (approved):

nyechiel (approved):

nyechiel:
This would need a backport to both `release-0.14` and `release-0.13`.

sridhargaddam:
Backport PRs:
Release-0.14 branch: https://github.com/submariner-io/submariner/pull/2091
Release-0.13 branch: https://github.com/submariner-io/submariner/pull/2092

PR #2122, opened 2022-11-13
Avoid XfrmPolicyList return nil data.

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

Avoid xfrmPolicyList interface return nil data to crash the golang.

### DIFF
--- a/pkg/netlink/netlink.go
+++ b/pkg/netlink/netlink.go
@@ -261,6 +261,12 @@ func DeleteXfrmRules() error {
 	}
 
 	for i := range currentXfrmPolicyList {
+		// Maybe some data is dirty,  skip them for avoiding golang panic.
+		if currentXfrmPolicyList[i].Dst == nil || currentXfrmPolicyList[i].Src == nil {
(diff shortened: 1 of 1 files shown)


pkg/netlink/netlink.go
```diff
@@ -261,6 +261,12 @@ func DeleteXfrmRules() error {
 	}
 
 	for i := range currentXfrmPolicyList {
+		// Maybe some data is dirty,  skip them for avoiding golang panic.
+		if currentXfrmPolicyList[i].Dst == nil || currentXfrmPolicyList[i].Src == nil {
```
sridhargaddam, line 265: Did you notice any PANIC in a real setup?

yanggangtony: @sridhargaddam 
No, just the code “dst or src pointer” may be trigger panic.

skitt (changes requested):
This isn’t necessary, `(*IPNet).String()` handles nil references correctly. `Dst` or `Src` being `nil` won’t cause a panic.

yanggangtony:
> This isn’t necessary, `(*IPNet).String()` handles nil references correctly. `Dst` or `Src` being `nil` won’t cause a panic.

@skitt 
Thanks for review and give advice..
I just test the code .Its fine..
So just close the pr..

PR #2163, opened 2022-12-01
Run dependabot for go weekly, all updates at once

dfarrell07:
To minimize the code review work required to merge go update PRs, cause
dependabot to give us all the updates at the same time, once a week.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,8 +23,7 @@ updates:
   - package-ecosystem: gomod
     directory: "/"
     schedule:
-      interval: daily
-    open-pull-requests-limit: 10
+      interval: weekly
     ignore:
       # protobuf must be handled manually
       - dependency-name: google.golang.org/protobuf


tpantelis (approved):

.github/dependabot.yml
```diff
@@ -28,5 +28,3 @@ updates:
     ignore:
       # protobuf must be handled manually
       - dependency-name: google.golang.org/protobuf
-      # Our own dependencies are handled during releases
-      - dependency-name: github.com/submariner-io/*
```
skitt: Does this actually work on devel commits? What we want is bumps to the latest commit on the devel branches; I wonder if this won’t cause dependabot to submit PRs updating to the latest tag, either on devel or on the last release branch once it exists (which will do the wrong thing).

skitt (changes requested):

.github/dependabot.yml
dfarrell07: I'll remove that change for now, send a different PR with a different attempt.

skitt (approved):

PR #2164, opened 2022-12-05
Fix issues with gateway public-ip resolver

sridhargaddam:
This PR implements changes in the public-ip resolution code

1. Recently we started to notice consistent errors while resolving
   public-ip when using the api.my-ip.io server. This PR removes this
   server entry from the list of resolvers and includes some new set 
   of servers as resolvers.
2. In the current code, we use a fixed sequence for resolving the public-ip.
   Due to this, the same resolver gets used all the time.
   It is only when the resolver fails, the next one from the list will
   be used as the resolver. This PR now randomizes the resolverList
   and we pick a random server everytime the periodic public-ip watcher
   is triggered.
3. The periodic public IP watcher interval is now increased to 60 secs.

### DIFF
--- a/pkg/endpoint/local_endpoint.go
+++ b/pkg/endpoint/local_endpoint.go
@@ -20,6 +20,7 @@ package endpoint
 
 import (
 	"fmt"
+	"math/rand"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


sridhargaddam:
The PR is in good shape, however, I would like to test it once with load-balancer mode on AWS/GCP and once I find it working, will remove the draft status.

skitt (approved):

tpantelis (approved):

sridhargaddam:
Backport PRs:
release-0.14: https://github.com/submariner-io/submariner/pull/2183
release-0.13: https://github.com/submariner-io/submariner/pull/2352

Issue #2125, opened 2022-11-14
Add periodic `go get <subm-internal>@devel` integration automation


dfarrell07:
**What would you like to be added**:

Add automation to run `go get *@devel` for all Submariner-internal repos on a weekly basis. This may be done with a manual GHA, but better if we can use Dependabot (or the new/similar bot from M$ft).

**Why is this needed**:

To avoid integration issues just before or during a release, to let us know when there are internal incompatibilities so they can be fixed.

This idea came out of a sprint retrospective.

dfarrell07:
This is blocked on getting control of submariner-bot from Miguel. I emailed him a few weeks ago, will follow-up.

dfarrell07:
> This is blocked on getting control of submariner-bot from Miguel. I emailed him a few weeks ago, will follow-up.

I pinged him again, will update.

dfarrell07:
The current issue is CI isn't starting on the PR's opened by this automation. Details: https://github.com/submariner-io/submariner/pull/2334#issuecomment-1467745545

dfarrell07:
Discussing with the team, we have a better idea. New idea is to do this internal integration (and maybe other go.mod updates) at the start of the automated release process. That solves a number of issues, including allowing vuln scans to run on non-devel branches (not possible to use cron triggers there) as well as catching code-level integration issues like this issue was meant to catch.

PR #2173, opened 2022-12-05
Add GHA to periodically integrate submariner-io/*

dfarrell07:
Add a weekly GitHub action to update all internal Submariner repo dependencies to the latest on devel.

Avoids integration issues just before or during a release and lets us know when there are internal incompatibilities so they can be fixed.

Relates-to: submariner-io/submariner#2125

### DIFF
--- a/.github/workflows/periodic.yml
+++ b/.github/workflows/periodic.yml
@@ -36,8 +36,9 @@ jobs:
           commit-message: |
             Update submariner-io/* dependencies to latest
 
-            This checks the current status of this repository against the latest
-            version of all the Submariner projects it depends on.
+            This upgrades all our dependencies on other Submariner projects to their
+            latest development snapshots, ensuring the code in the projects remains
+            coherent and that tests of development images verify the latest code.
           signoff: true
           labels: automated, dependencies
 
--- a/.github/workflows/periodic.yml
+++ b/.github/workflows/periodic.yml
@@ -8,6 +8,39 @@ on:
 permissions: {}
 
 jobs:
+  internal-integration:
+    name: Internal Integration
+    if: github.repository_owner == 'submariner-io'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+
+      - name: Update internal submariner-io/* dependencies to latest
+        run: |
+          for dep in $(awk '!/module/ && /github.com.submariner-io/ { print $1 }' go.mod)
+            do go get ${dep}@devel
+          done
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
+        with:
+          title: Update submariner-io/* dependencies to latest
+          body: |
+            This checks the current status of this repository against the latest version of all the Submariner projects it depends on.
+            If something fails, the failure should be investigated and at least tracked as an issue blocking the next release.
+            Since some CI only runs periodically, if on-PR CI passes it's still good to merge this update for full integration coverage.
+          commit-message: |
+            Update submariner-io/* dependencies to latest
+
+            This checks the current status of this repository against the latest
+            version of all the Submariner projects it depends on.
+          signoff: true
+          labels: automated, dependencies
+
   markdown-link-check-periodic:
     name: Markdown Links (all files)
     if: github.repository_owner == 'submariner-io'


.github/workflows/periodic.yml
```diff
@@ -8,6 +8,30 @@ on:
...
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+      - name: Update internal submariner-io/* dependencies to latest
+        run: |
+          go get github.com/submariner-io/admiral@devel
+          go get github.com/submariner-io/shipyard@devel
```
skitt: We can make this generic:

```suggestion
          for dep in $(awk '!/module/ && /github.com.submariner-io/ { print $1 }' go.mod); do go get ${dep}@devel; done
```

.github/workflows/periodic.yml
```diff
@@ -8,6 +8,30 @@ on:
...
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
+        with:
+          title: Update submariner-io/* dependencies to latest
+          body: To catch cross-repo integration issues before releases.
```
skitt: The pull-request should work as a commit if merged, the body here doesn’t seem like a great commit message — it describes the purpose of the job, not the real purpose of the commit.

skitt (changes requested):

.github/workflows/periodic.yml
dfarrell07: What do you mean by "real purpose of the commit" vs "the purpose of the job"? All the commit does is add the job, so they have the same purpose. Happy to make better commit messages, but I thought about it a lot and I don't understand what you're looking for.

skitt: I’m not talking about the commit message for this PR, I’m talking about the commit message generated by the job in the PRs it will create. If the goal is only to check that everything is OK, and the PRs aren’t supposed to be merged, something like

> This checks the current status of this repository against the latest version of all the Submariner project it depends on. If all goes well, this PR shouldn’t be merged. If something fails, the failure should be investigated and at least tracked as an issue (if appropriate) blocking the next release.

If the goal is to merge them,

> This pulls in the latest versions of all Submariner projects this project depends on. This is an automated update.

dfarrell07: Oh I see! Okay, thanks that's good feedback!

mkolesnik:
QQ why not let dependabot do it?

skitt:
> QQ why not let dependabot do it?

Do you know how to get dependabot to push PRs using the latest commit of a given branch, ignoring any tags?

mkolesnik:
> Do you know how to get dependabot to push PRs using the latest commit of a given branch, ignoring any tags?

Seems its rather limited, but this might be possible with Renovate - https://github.com/marketplace/renovate

skitt:
> > Do you know how to get dependabot to push PRs using the latest commit of a given branch, ignoring any tags?
> 
> Seems its rather limited, but this might be possible with Renovate - https://github.com/marketplace/renovate

That’s not what I was asking. If the alternative is adding yet another third-party tool, I’d rather have a 25-line GHA which does exactly what we want and only what we want.

mkolesnik:
> That’s not what I was asking. If the alternative is adding yet another third-party tool, I’d rather have a 25-line GHA which does exactly what we want and only what we want.

That's the only answer I got for you since dependabot is quite limited,
I'm not saying having another GHA is necessarily bad, but if the tool we have for these use cases isn't good enough (dependabot) then we should consider replacing it.
This is not the only limitation dependabot had which affects us, but its another annoying thing we need to creatively work around instead of replacing it with a better tool.

skitt:
I’m about confused by your questions, but never mind. Yes, we can consider replacing dependabot with something else; if that’s important to you, please open an issue. I don’t think that should block this though.

skitt:
This would have saved us some hassle on the 0.15.0-m1 release, see the additional Go changes needed in https://github.com/submariner-io/lighthouse/pull/1024 and the dependency in https://github.com/submariner-io/subctl/pull/475.

.github/workflows/periodic.yml
skitt: You mentioned that you’d updated this, but I don’t see a change in the PR.

.github/workflows/periodic.yml
```diff
@@ -8,6 +8,31 @@ on:
...
+        run: |
+          for dep in $(awk '!/module/ && /github.com.submariner-io/ { print $1 }' go.mod)
+            do go get ${dep}@devel
+          done
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
+        with:
+          title: Update submariner-io/* dependencies to latest
+          body: To catch cross-repo integration issues before releases.
```
skitt: This still needs fixing.

skitt (changes requested):

.github/workflows/periodic.yml
```diff
@@ -8,6 +8,34 @@ on:
...
+          done
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
+        with:
+          title: Update submariner-io/* dependencies to latest
+          body: |
+            This checks the current status of this repository against the latest version of all the Submariner projects it depends on.
+            If something fails, the failure should be investigated and at least tracked as an issue blocking the next release.
+            Since some CI only runs periodically, if on-PR CI passes it's still good to merge this update for full integration coverage.
```
skitt, line 35: Thanks, this is much better.

Looking at [the `create-pull-request` docs](https://github.com/peter-evans/create-pull-request#action-inputs), I see that the default commit message isn’t great (“[create-pull-request] automated change”) — apologies for not noticing this earlier, I thought the PR title and body would be used (and looked into the details because, while the updated PR title and body work for a PR, the body isn’t useful as a commit message).

Could you add a commit message too?

skitt (changes requested):

.github/workflows/periodic.yml
```diff
@@ -8,6 +8,39 @@ on:
...
+          commit-message: |
+            Update submariner-io/* dependencies to latest
+
+            This checks the current status of this repository against the latest
+            version of all the Submariner projects it depends on.
```
skitt: Sorry, I’m obviously not communicating clearly enough.

The commit message generated by the job needs to make sense as a commit message in the history of the project it goes into. Its role in the context of the job is irrelevant; the commit itself doesn’t “check the current status …”, that’s what the job does.

```suggestion
            Update submariner-io/* dependencies to latest

            This upgrades all our dependencies on other Submariner projects to their
            latest development snapshots, ensuring the code in the projects remains
            coherent and that tests of development images verify the latest code.
```

skitt (changes requested):

.github/workflows/periodic.yml
dfarrell07: I see what you're saying, but to me it seems the commit does "check the status of this repo against the latest...". It causes tests to be run against those versions, both on-PR and periodic. I can change to the suggestion of course, that's more clear/verbose.

skitt: A commit message describes the changes that the commit makes. The only thing that the commit changes is a bunch of dependencies; there are no changes _in the commit_ which check the status of anything. Our CI means that because a PR is created with the commit, the status will be checked; that’s why it makes sense to say that in the _PR_ description.

skitt (approved):

skitt (approved):

tpantelis (approved):

PR #2175, opened 2022-12-06
Add osx trash to gitignore

wuyingjun-lucky:
Signed-off-by: wuyingjun <wuyingjun@cmss.chinamobile.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ scripts/broker-info.subm*
 /Dockerfile.*
 .shflags
 Makefile.dapper
+# OSX trash
+.DS_Store


skitt (changes requested):
I know we have similar entries in our `.gitignore` files already, but please don’t add more; instead, set up [a global `.gitignore` on your system](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) to match the specifics of your local setup.

wuyingjun-lucky:
All right. 
It feels a bit weird without adding this entry.  [k8s ignore file](https://github.com/kubernetes/kubernetes/blob/master/.gitignore)
/close

PR #2250, opened 2023-01-16
Skip processing IPv6 addresses in mtu_handler

### LINKED ISSUE #1739: [Epic] Ensure Submariner works with dual-stack clusters with IPv4 connections
**What happened**:
Starting from K8s 1.23 version, it supports dual-stack K8s clusters where pods will get both IPv4 and IPv6 addresses. Submariner currently does not support IPv6, but it should atleast allow IPv4 communication to happen between the clusters.

- [x] [Skip IPv6 node addresses while adding fdb entries](https://github.com/submariner-io/submariner/pull/2245)
- [x] [Skip processing IPv6 addresses in mtu_handler](https://github.com/submariner-io/submariner/pull/2250)
- [x] [Submariner support for Dual Stack networks](https://github.com/submariner-io/submariner/pull/2294) Includes fixes in submariner-gateway pods, route-agent pods and mtu_handler.
- [x] [Support in SM Globalnet for dual-stack clusters](https://github.com/submariner-io/submariner/pull/2323)
- [x] [Support for deploying dual-stack k8s KIND clusters](https://github.com/submariner-io/shipyard/pull/1173) 
- [x] [Document dual-stack option in make command](https://github.com/submariner-io/submariner-website/pull/898)
- [x] [Dual-stack job in CI](https://github.com/submariner-io/shipyard/pull/1174)
- [x] [Release notes](https://github.com/submariner-io/submariner/pull/2327/files)

sridhargaddam:
Submariner mtu_handler creates IPSet entries which are subsequently used in the iptable rules to clamp TCP Mss to PMTU/fixed value. The IPsets created by the mtu_handler are of IPV4 protocolFamily. In a
dual-stack env' the the handler was trying to add an IPv6 address which results in a bunch of errors.
Until Submariner supports IPv6, lets skip the IPv6 addresses in the mtu_handler.

Partially Fixes: https://github.com/submariner-io/submariner/issues/1739

### DIFF
--- a/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
+++ b/pkg/routeagent_driver/handlers/mtu/mtuhandler.go
@@ -32,6 +32,7 @@ import (
 	netlinkAPI "github.com/submariner-io/submariner/pkg/netlink"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	utilexec "k8s.io/utils/exec"
+	k8snet "k8s.io/utils/net"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
@@ -143,7 +144,7 @@ func (h *mtuHandler) Init() error {
 }
 
 func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
-	subnets := extractSubnets(&endpoint.Spec)
+	subnets := extractIPv4Subnets(&endpoint.Spec)
 	for _, subnet := range subnets {
 		err := h.localIPSet.AddEntry(subnet, true)
 		if err != nil {
@@ -173,7 +174,7 @@ func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
 }
 
 func (h *mtuHandler) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
-	subnets := extractSubnets(&endpoint.Spec)
+	subnets := extractIPv4Subnets(&endpoint.Spec)
 	for _, subnet := range subnets {
 		err := h.localIPSet.DelEntry(subnet)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -215,11 +216,12 @@ func (h *mtuHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
 	return nil
 }
 
-func extractSubnets(endpoint *submV1.EndpointSpec) []string {
+func extractIPv4Subnets(endpoint *submV1.EndpointSpec) []string {
 	subnets := make([]string, 0, len(endpoint.Subnets))
 
 	for _, subnet := range endpoint.Subnets {
-		if !strings.HasPrefix(subnet, endpoint.PrivateIP+"/") {
```
tpantelis, line 222: So we no longer need this check?

sridhargaddam: No. I don't think we require this check. Prior to pushing the PR, I discussed with @aswinsuryan (original author of the code) to see if there is a use-case behind this check.

skitt (approved):

tpantelis (approved):

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
aswinsuryan: I think  reused a method from here at that time and it may not be valid here  given the IPs that an endpoint can take,

[1] https://github.com/submariner-io/submariner/blob/4eef0296c913b8d4bb9f7be33edbcdcb63891ecb/pkg/cable/libreswan/libreswan.go#L288

sridhargaddam: Thanks for confirming @aswinsuryan

PR #2263, opened 2023-01-25
Skip endpoints from Broker with overlapping CIDRs

sridhargaddam:
Currently, we have validations in route-agent,
network-plugin-syncer which will check if CIDRs
overlap when processing an endpoint. However, 
there is no such check in the cableDriver code. 
Because of this the IPsec tunnels are configured
even for cluster when the CIDR overlaps. This will
not affect the pods on other nodes, but has an effect
on the Gateway node pods.

This PR fixes this issue by ensuring that we do not
sync any endpoint from the Broker cluster if its CIDR
overlaps with the local cluster.

### DIFF
--- a/pkg/controllers/datastoresyncer/datastoresyncer.go
+++ b/pkg/controllers/datastoresyncer/datastoresyncer.go
@@ -30,6 +30,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
 	"github.com/submariner-io/admiral/pkg/watcher"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cidr"
 	"github.com/submariner-io/submariner/pkg/types"
 	k8sv1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


skitt (approved):

pkg/controllers/datastoresyncer/datastoresyncer.go
```diff
@@ -214,6 +216,27 @@ func (d *DatastoreSyncer) shouldSyncCluster(obj runtime.Object, numRequeues int,
...
 
+func (d *DatastoreSyncer) shouldSyncRemoteEndpoint(obj runtime.Object, numRequeues int,
+	op resourceSyncer.Operation,
+) (runtime.Object, bool) {
+	endpoint := obj.(*submarinerv1.Endpoint)
+
+	for _, localSubnet := range d.localEndpoint.Spec.Subnets {
+		overlap, err := cidr.IsOverlapping(endpoint.Spec.Subnets, localSubnet)
+		if err != nil {
+			logger.Errorf(err, "unable to validate if remote CIDR overlaps with local CIDR")
```
tpantelis:
```suggestion
			logger.Errorf(err, "Unable to validate if remote CIDR overlaps with local CIDR")
```

pkg/controllers/datastoresyncer/datastoresyncer.go
```diff
@@ -214,6 +216,27 @@ func (d *DatastoreSyncer) shouldSyncCluster(obj runtime.Object, numRequeues int,
...
+			return nil, false
+		}
+
+		if overlap {
+			logger.Errorf(err, "skip processing the remote endpoint %#v as subnets are overlapping", endpoint)
```
tpantelis:
```suggestion
			logger.Errorf(nil, "Skip processing the remote endpoint %#v as subnets are overlapping", endpoint)
```

tpantelis (commented):
A couple minor things.

tpantelis (approved):

dfarrell07:
This went into 0.13 here: https://github.com/submariner-io/submariner/pull/2643

PR #2275, opened 2023-02-01
Fix stale Globalnet internal services

sridhargaddam:
Normally, a user is not expected to delete the internal service created by the Globalnet controller. However, when it's accidentally done while the globalnet controller is down, the internal service remains until the finalizer is removed. We have seen that this intermediate state of the service can potentially create issues in some deployments. In this PR, the Globalnet controller while booting up will identify such internal services and delete them.

### DIFF
--- a/pkg/globalnet/controllers/gateway_monitor.go
+++ b/pkg/globalnet/controllers/gateway_monitor.go
@@ -272,6 +272,17 @@ func (g *gatewayMonitor) startControllers() error {
 
 	g.controllers = append(g.controllers, c)
 
+	// A user is not normally expected to delete the internal service created by the Globalnet controller.
+	// However, when it's accidentally done while the globalnet controller is down, the internal service
+	// remains until the finalizer is removed. We have seen that this intermediate state of
+	// the service can potentially create issues in some deployments. Hence, we identify such internal
+	// services and delete them when the Globalnet controller pod comes up.
+	err = RemoveStaleInternalServices(g.syncerConfig)
+	if err != nil {
+		// Just log an error message as it's non-fatal
+		logger.Errorf(err, "Error removing stale internal services created by Globalnet controller")
+	}
+
 	// The GlobalIngressIP controller needs to be started before the ServiceExport and Service controllers to ensure
 	// reconciliation works properly.
 	c, err = NewGlobalIngressIPController(g.syncerConfig, pool)
--- a/pkg/globalnet/controllers/uninstall.go
+++ b/pkg/globalnet/controllers/uninstall.go
@@ -27,6 +27,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/finalizer"
 	"github.com/submariner-io/admiral/pkg/resource"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/util"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	versioned "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	"github.com/submariner-io/submariner/pkg/globalnet/constants"
@@ -215,4 +217,36 @@ func RemoveGlobalIPAnnotationOnNode(cfg *rest.Config) {
 	}
 
 	logger.Infof("Successfully removed globalIP annotation from node %q", nodeName)
+}
+
+func RemoveStaleInternalServices(config *syncer.ResourceSyncerConfig) error {
+	_, gvr, err := util.ToUnstructuredResource(&corev1.Service{}, config.RestMapper)
+	if err != nil {
+		return errors.Wrap(err, "error converting resource")
+	}
+
+	services := config.SourceClient.Resource(*gvr)
+
+	svcList, err := services.Namespace(corev1.NamespaceAll).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: InternalServiceLabel,
+	})
+	if err != nil {
+		return errors.Wrap(err, "error listing the services")
+	}
+
+	for i := range svcList.Items {
+		svc := &svcList.Items[i]
+		if !svc.GetDeletionTimestamp().IsZero() {
+			logger.Warningf("Globalnet internal service %s/%s has deletionTimestamp set", svc.GetNamespace(), svc.GetName())
+
+			err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.GetNamespace())), svc,
+				InternalServiceFinalizer)
+			if err != nil {
+				return errors.Wrapf(err, "error while removing the finalizer from globalnet internal service %q/%q",
+					svc.GetNamespace(), svc.GetName())
+			}
+		}
+	}
+
+	return nil
 }


pkg/globalnet/controllers/uninstall.go
```diff
@@ -216,3 +218,38 @@ func RemoveGlobalIPAnnotationOnNode(cfg *rest.Config) {
...
+			if err != nil {
+				return errors.Wrapf(err, "error converting %#v to Service", svcList.Items[i])
+			}
+
+			if finalizer.IsPresent(&svc.ObjectMeta, InternalServiceFinalizer) {
+				if err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.Namespace)), svc,
+					InternalServiceFinalizer); err != nil {
+					return fmt.Errorf("error while removing the finalizer from globalnet internal service %q/%q",
+						svc.Namespace, svc.Name)
+				}
```
skitt: This loses the details of the error:

```suggestion
				if err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.Namespace)), svc,
					InternalServiceFinalizer); err != nil {
					return errors.Wrapf(err, "error while removing the finalizer from globalnet internal service %q/%q",
						svc.Namespace, svc.Name)
				}
```

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -272,6 +272,17 @@ func (g *gatewayMonitor) startControllers() error {
...
 	g.controllers = append(g.controllers, c)
 
+	// A user is not normally expected to delete the internal service created by the Globalnet controller.
+	// However, when it's accidentally done while the globalnet controller is down, the internal service
+	// remains until the finalizer is removed. We have seen that this intermediate state of
+	// the service can potentially create issues in some deployments. Hence, we identify such internal
+	// services and delete them when the Globalnet controller pod comes up.
+	err = RemoveStaleInternalServices(g.syncerConfig)
+	if err != nil {
+		// Just log an error message as its non-fatal
```
skitt: Nit:

```suggestion
		// Just log an error message as it's non-fatal
```

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -272,6 +272,17 @@ func (g *gatewayMonitor) startControllers() error {
...
+		logger.Errorf(err, "Error removing stale internal services created by Globalnet controller")
```
skitt, line 283: I know it’s not as convenient, but if this isn’t an actual error, we shouldn’t log it as an error:

```suggestion
		logger.Warningf("Error removing stale internal services created by Globalnet controller: %v", err)
```

skitt (changes requested):

pkg/globalnet/controllers/gateway_monitor.go
sridhargaddam: Some CNIs using kube-proxy continue to support the service (by load-balancing the requests to the backend pods) even if the deletionTimestamp is set on the service. However, others like OVNK do not support it. So, I'd say that it would be better to log an error message as it would be difficult to spot the issue otherwise.

pkg/globalnet/controllers/uninstall.go
```diff
@@ -216,3 +218,38 @@ func RemoveGlobalIPAnnotationOnNode(cfg *rest.Config) {
...
+			if err != nil {
+				return errors.Wrapf(err, "error converting %#v to Service", svcList.Items[i])
+			}
+
+			if finalizer.IsPresent(&svc.ObjectMeta, InternalServiceFinalizer) {
+				if err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.Namespace)), svc,
+					InternalServiceFinalizer); err != nil {
+					return errors.Wrapf(err, "error while removing the finalizer from globalnet internal service %q/%q",
+						svc.Namespace, svc.Name)
+				}
```
tpantelis:
```suggestion
				err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.Namespace)), svc,
					InternalServiceFinalizer)
				return errors.Wrapf(err, "error while removing the finalizer from globalnet internal service %q/%q",
						svc.Namespace, svc.Name)
				
```

pkg/globalnet/controllers/uninstall.go
```diff
@@ -216,3 +218,38 @@ func RemoveGlobalIPAnnotationOnNode(cfg *rest.Config) {
...
+
+	for i := range svcList.Items {
+		if !svcList.Items[i].GetDeletionTimestamp().IsZero() {
+			err = config.Scheme.Convert(&svcList.Items[i], svc, nil)
+			if err != nil {
+				return errors.Wrapf(err, "error converting %#v to Service", svcList.Items[i])
+			}
+
+			if finalizer.IsPresent(&svc.ObjectMeta, InternalServiceFinalizer) {
+				if err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.Namespace)), svc,
```
tpantelis: There's no need to convert to `Service` obj - `Unstructured` is a `runtime.Object`.
 
```suggestion
	for i := range svcList.Items {
	        svc := &svcList.Items[i]
		if !svc.GetDeletionTimestamp().IsZero() {
			if finalizer.IsPresent(svc, InternalServiceFinalizer) {
				if err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.GetNamespace())), svc,
```

tpantelis (changes requested):

tpantelis:
Is the scenario described in the summary just theory or actually seen in a real deployment? If the latter, I'm curious how a user accidentally deleted an internal service and why was the globalnet controller down at the time.  Couldn't a user also accidentally delete a service while the controller is running? Perhaps we should have a controller that handles both edge cases.

sridhargaddam:
> Is the scenario described in the summary just theory or actually seen in a real deployment? If the latter, I'm curious how a user accidentally deleted an internal service and why was the globalnet controller down at the time. Couldn't a user also accidentally delete a service while the controller is running? Perhaps we should have a controller that handles both edge cases.

Yes it was reported by our QE

pkg/globalnet/controllers/uninstall.go
```diff
@@ -216,3 +218,32 @@ func RemoveGlobalIPAnnotationOnNode(cfg *rest.Config) {
...
+
+func RemoveStaleInternalServices(config *syncer.ResourceSyncerConfig) error {
+	_, gvr, err := util.ToUnstructuredResource(&corev1.Service{}, config.RestMapper)
+	if err != nil {
+		return errors.Wrap(err, "error converting resource")
+	}
+
+	services := config.SourceClient.Resource(*gvr)
+
+	svcList, err := services.Namespace(corev1.NamespaceAll).List(context.TODO(), metav1.ListOptions{})
```
tpantelis: Since we're only interested in globalnet internal services, we can optimize the `List` call by using a label selector for the `InternalServiceLabel`:

```suggestion
	svcList, err := services.Namespace(corev1.NamespaceAll).List(context.TODO(), metav1.ListOptions{
		LabelSelector: InternalServiceLabel,
	})
```

pkg/globalnet/controllers/uninstall.go
```diff
@@ -216,3 +218,32 @@ func RemoveGlobalIPAnnotationOnNode(cfg *rest.Config) {
...
+
+	for i := range svcList.Items {
+		svc := &svcList.Items[i]
+		if !svc.GetDeletionTimestamp().IsZero() {
+			if finalizer.IsPresent(svc, InternalServiceFinalizer) {
+				err := finalizer.Remove(context.TODO(), resource.ForDynamic(services.Namespace(svc.GetNamespace())), svc,
+					InternalServiceFinalizer)
+
+				return errors.Wrapf(err, "error while removing the finalizer from globalnet internal service %q/%q",
+					svc.GetNamespace(), svc.GetName())
```
tpantelis, line 246: My prior suggestion was incorrect - I guess I misread the code - sorry about that. It should be the way you had it.

```suggestion
				if err != nil {
				    return errors.Wrapf(err, "error while removing the finalizer from globalnet internal service %q/%q",
					    svc.GetNamespace(), svc.GetName())
				}
```

pkg/globalnet/controllers/uninstall.go
```diff
@@ -216,3 +218,32 @@ func RemoveGlobalIPAnnotationOnNode(cfg *rest.Config) {
...
+
+	svcList, err := services.Namespace(corev1.NamespaceAll).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error listing the services")
+	}
+
+	for i := range svcList.Items {
+		svc := &svcList.Items[i]
+		if !svc.GetDeletionTimestamp().IsZero() {
+			if finalizer.IsPresent(svc, InternalServiceFinalizer) {
```
tpantelis: If we search by the label as I suggested above then we don't really need to check for the `InternalServiceFinalizer`, ie it should be present.

tpantelis (changes requested):

pkg/globalnet/controllers/gateway_monitor.go
sridhargaddam: I've added a warning message if a internal service with deletionTimestamp exists. Hope you are fine with having this as an error message.

pkg/globalnet/controllers/uninstall.go
sridhargaddam: This is a good optimization.

skitt (approved):

tpantelis (approved):

PR #2301, opened 2023-02-21
Set automated submariner-dependencies PR author

dfarrell07:
The commit author defaults to the user that triggered the workflow, which resolves to the person that added the periodic workflow (me).

While this is correct in a way, I did cause these commits to be created, it's not very clear. It may be less confusing set it to a non-account.


tpantelis (approved):

maayanf24 (approved):

skitt (approved):

dfarrell07:
Keep getting sporadic unrelated failures, blocking merge. Still trying, latest rerun is [Latest Release to Latest Version](https://github.com/submariner-io/submariner/actions/runs/4365336864/jobs/7633816402#logs).

PR #2324, opened 2023-03-02
Use k8s.io generic sets instead of Admiral stringset

skitt:

@tpantelis I’d particularly appreciate a careful review of the locking in the second commit. Go locks aren’t re-entrant, see https://groups.google.com/g/golang-nuts/c/XqW1qcuZgKg/m/Ui3nQkeLV80J for details.

### DIFF
--- a/pkg/globalnet/controllers/gateway_monitor.go
+++ b/pkg/globalnet/controllers/gateway_monitor.go
@@ -39,6 +39,7 @@ import (
 	routeAgent "github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes/scheme"
 	k8snet "k8s.io/utils/net"
@@ -50,7 +51,7 @@ func NewGatewayMonitor(spec Specification, localCIDRs []string, config *watcher.
 		baseController: newBaseController(),
 		spec:           spec,
 		isGatewayNode:  atomic.Bool{},
-		localSubnets:   stringset.New(localCIDRs...).Elements(),
+		localSubnets:   sets.New(localCIDRs...).UnsortedList(),
 		remoteSubnets:  stringset.NewSynchronized(),
 	}
 
--- a/pkg/globalnet/controllers/gateway_monitor.go
+++ b/pkg/globalnet/controllers/gateway_monitor.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
-	"github.com/submariner-io/admiral/pkg/stringset"
 	"github.com/submariner-io/admiral/pkg/syncer"
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
 	admUtil "github.com/submariner-io/admiral/pkg/util"
@@ -52,7 +51,7 @@ func NewGatewayMonitor(spec Specification, localCIDRs []string, config *watcher.
 		spec:           spec,
 		isGatewayNode:  atomic.Bool{},
 		localSubnets:   sets.New(localCIDRs...).UnsortedList(),
-		remoteSubnets:  stringset.NewSynchronized(),
+		remoteSubnets:  sets.New[string](),
 	}
 
 	var err error
@@ -164,7 +163,7 @@ func (g *gatewayMonitor) handleCreatedOrUpdatedEndpoint(obj runtime.Object, numR
 
 		for _, remoteSubnet := range endpoint.Spec.Subnets {
 			if k8snet.IsIPv4CIDRString(remoteSubnet) {
-				g.remoteSubnets.Add(remoteSubnet)
+				g.remoteSubnets.Insert(remoteSubnet)
 				g.markRemoteClusterTraffic(remoteSubnet, AddRules)
 			}
 		}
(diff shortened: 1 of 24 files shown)


pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -161,10 +161,12 @@ func (g *gatewayMonitor) handleCreatedOrUpdatedEndpoint(obj runtime.Object, numR
 			return false
 		}
 
+		g.syncMutex.Lock()
```
tpantelis: This lock is intended to protect the `isGatewayNode` flag so I'm wary of also using it in a disparate manner. This is a case where the standalone [synchronized set](https://github.com/submariner-io/admiral/pull/557/commits/05d6c279d17e1818f0d861803567d36c4a1ae667) is useful.

skitt: How about https://github.com/submariner-io/submariner/pull/2325 instead, for `isGatewayNode` and `controllers`?

tpantelis: We really don't even need to synchronize `remoteSubnets` since it's only accessed via the work queue which is single-threaded. If we had multiple worker threads then we'd need synchronization. Same for `ingressIPMap` in _ingress_pod_controller.go_.

skitt: OK, I’ll drop the synchronization around those.

pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go
```diff
@@ -72,20 +73,23 @@ func (kp *SyncHandler) updateRoutingRulesForCIDRBlock(inputCidrBlock string, ope
...
 
 	case Delete:
-		if kp.routeCacheGWNode.Remove(inputCidrBlock) {
+		if kp.routeCacheGWNode.Has(inputCidrBlock) {
 			if err := kp.configureRoute(inputCidrBlock, operation, viaGW); err != nil {
 				logger.Errorf(err, "Failed to delete route %q for HostNetwork support on the Gateway node",
 					inputCidrBlock)
 			}
+
+			kp.routeCacheGWNode.Delete(inputCidrBlock)
```
tpantelis: The prior behavior was to remove `inputCidrBlock` from the set regardless of an error on L87 so, to preserve that, this line should be before L87.

skitt: Right, same when adding it, I’ll fix both.

skitt:
The OVN errors are significant, no need to retry the jobs.

tpantelis:
It looks like some changes got lost in the last push.  There's still references to `stringset` in

_./pkg/globalnet/controllers/types.go
./pkg/globalnet/controllers/ingress_pod_controller.go
./pkg/globalnet/controllers/gateway_monitor.go
./pkg/routeagent_driver/handlers/kubeproxy/kp_iptables.go_

skitt:
> It looks like some changes got lost in the last push. There's still references to `stringset` in
> 
> _./pkg/globalnet/controllers/types.go ./pkg/globalnet/controllers/ingress_pod_controller.go ./pkg/globalnet/controllers/gateway_monitor.go ./pkg/routeagent_driver/handlers/kubeproxy/kp_iptables.go_

Yes, I pulled out the commit touching synchronized sets, I’ll open a separate PR for that one.

skitt:
> > It looks like some changes got lost in the last push. There's still references to `stringset` in
> > _./pkg/globalnet/controllers/types.go ./pkg/globalnet/controllers/ingress_pod_controller.go ./pkg/globalnet/controllers/gateway_monitor.go ./pkg/routeagent_driver/handlers/kubeproxy/kp_iptables.go_
> 
> Yes, I pulled out the commit touching synchronized sets, I’ll open a separate PR for that one.

Scratch that, I did that thinking that the OVN breakage might be related to the synchronized removals, but that’s not it; I’ve restored the second commit and fixed the first one.

dfarrell07 (approved):

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
```diff
@@ -141,12 +141,12 @@ func (ovn *SyncHandler) ensureSubmarinerInfra() error {
...
 	remoteSubnets := ovn.remoteEndpointSubnetSet()
 
-	if remoteSubnets.Size() == 0 {
+	if len(remoteSubnets) == 0 {
 		logger.V(log.DEBUG).Info("No Remote Subnets to Process")
 		return nil
 	}
 
-	logger.V(log.DEBUG).Infof("Reconciling these raw remote subnets %v to logical router policies", remoteSubnets.Elements())
+	logger.V(log.DEBUG).Infof("Reconciling these raw remote subnets %v to logical router policies", remoteSubnets)
```
tpantelis:
```suggestion
	logger.V(log.DEBUG).Infof("Reconciling these raw remote subnets %v to logical router policies", remoteSubnets.UnsortedList())
```

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
```diff
@@ -141,12 +141,12 @@ func (ovn *SyncHandler) ensureSubmarinerInfra() error {
 func (ovn *SyncHandler) setupOvnClusterRouterLRPs() error {
 	remoteSubnets := ovn.remoteEndpointSubnetSet()
 
-	if remoteSubnets.Size() == 0 {
+	if len(remoteSubnets) == 0 {
```
tpantelis: Why not use `remoteSubnets.Len()` here?

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
skitt: Oops, not sure what I was thinking there...

pkg/networkplugin-syncer/handlers/ovn/ovn_submariner_infrastructure.go
skitt: They’re equivalent (currently), but I see from https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/sets/set_generic_test.go that the API does appear to be `s.Len()` rather than `len(s)` (which makes sense since the former is controlled by the implementation, and the latter is a happy coincidence on the current implementation).

tpantelis (approved):

PR #2325, opened 2023-03-03
Use an atomic Bool for isGatewayNode

skitt:
This allows the manual synchronization to be removed.

syncMutex is repurposed to protect controllers; we want to ensure that
startControllers() finishes before stopControllers() can be called,
and that stopControllers() is never called twice in parallel.

### DIFF
--- a/pkg/globalnet/controllers/gateway_monitor.go
+++ b/pkg/globalnet/controllers/gateway_monitor.go
@@ -21,6 +21,7 @@ package controllers
 import (
 	"os"
 	"strings"
+	"sync/atomic"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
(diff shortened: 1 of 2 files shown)


pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -184,25 +185,18 @@ func (g *gatewayMonitor) handleCreatedOrUpdatedEndpoint(obj runtime.Object, numR
 
 		configureTCPMTUProbe()
 
-		g.syncMutex.Lock()
-		if !g.isGatewayNode {
-			g.isGatewayNode = true
-
+		if g.isGatewayNode.CompareAndSwap(false, true) {
 			err := g.startControllers()
```
skitt, line 187: I think this is really what needs to be protected — `stopControllers()` mustn’t be called while `startControllers()` is in progress (_i.e._ `controllers` needs to be synced).

tpantelis: yeah I believe that's correct. `Stop` is called outside the informer thread otherwise we really wouldn't need any synchronization.

tpantelis (approved):

dfarrell07 (approved):

PR #2342, opened 2023-03-16
Use the new sets.Clear() function

skitt:
This depends on the release of K8s 1.27.

Depends on https://github.com/submariner-io/submariner/pull/2426

### DIFF
--- a/pkg/ipset/fake/ipset.go
+++ b/pkg/ipset/fake/ipset.go
@@ -78,7 +78,7 @@ func (i *IPSet) FlushSet(set string) error {
 		return nil
 	}
 
-	entries.Delete(entries.UnsortedList()...)
+	entries.Clear()
 
 	return nil
 }
--- a/pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go
+++ b/pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go
@@ -33,7 +33,7 @@ import (
 
 func (kp *SyncHandler) updateRoutingRulesForHostNetworkSupport(inputCidrBlocks []string, operation Operation) {
 	if operation == Flush {
-		kp.routeCacheGWNode.Delete(kp.routeCacheGWNode.UnsortedList()...)
+		kp.routeCacheGWNode.Clear()
 
 		err := kp.netLink.FlushRouteTable(constants.RouteAgentHostNetworkTableID)
 		if err != nil {


dfarrell07 (approved):

skitt:
@tpantelis we can bump apimachinery without touching client-go, see https://github.com/skitt/submariner-operator/pull/118; in fact because of the required code changes for the apimachinery bump, we need this in first before we can upgrade client-go.

tpantelis:
> @tpantelis we can bump apimachinery without touching client-go, see [skitt/submariner-operator#118](https://github.com/skitt/submariner-operator/pull/118); in fact because of the required code changes for the apimachinery bump, we need this in first before we can upgrade client-go.

we probably can but shouldn't we try to keep all K8s dependency versions consistent?  If there's a compelling reason where we need to bump one w/o the others but using `sets.Clear` doesn't fit that criteria.

skitt:
> we probably can but shouldn't we try to keep all K8s dependency versions consistent? If there's a compelling reason where we need to bump one w/o the others but using `sets.Clear` doesn't fit that criteria.

I agree that `sets.Clear` isn’t important, but the first commit in this PR fixes code which trips the build if we try to bump apimachinery; we need that before any of the other K8s bumps can go in.

pkg/globalnet/controllers/controllers_suite_test.go
```diff
@@ -506,6 +505,7 @@ func awaitStatusConditions(client dynamic.ResourceInterface, name string, atInde
 	var conditions []metav1.Condition
 	var notFound error
 
+	//nolint:staticcheck // Replacement unclear, see https://github.com/kubernetes/kubernetes/issues/116712
```
tpantelis: I encountered this too.  The replacement is [PollUntilContextTimeout](https://github.com/submariner-io/lighthouse/blob/65fa08c9fd02335a44b3e9ee4b02f5c682ee9e2a/pkg/agent/controller/controller_suite_test.go#L562).

tpantelis:
> > we probably can but shouldn't we try to keep all K8s dependency versions consistent? If there's a compelling reason where we need to bump one w/o the others but using `sets.Clear` doesn't fit that criteria.
> 
> I agree that `sets.Clear` isn’t important, but the first commit in this PR fixes code which trips the build if we try to bump apimachinery; we need that before any of the other K8s bumps can go in.

I addressed the API adjustments with https://github.com/submariner-io/submariner/pull/2426/files#diff-ce3b5bd8d2e2441b2c065d6b9337368965304cc15e784e014abc08cae1c8aa4e.

tpantelis (approved):

PR #2346, opened 2023-03-20
log warning: VxLan cable-driver is supported only with no NAT deployments

### LINKED ISSUE #1928: Log Warning when VxLAN cable-driver is used with NAT
**What happened**:
When using VxLAN cable-driver with NAT enabled, we can run into issues in certain environments. Currently Submariner supports VxLAN cable-driver in a no-NAT deployments. 

So, as discussed in the [original issue](https://github.com/submariner-io/submariner/issues/1417#issuecomment-862928685), lets log a warning in the gateway pod (and diagnose?)

cyclinder:
When using VxLAN cable-driver with NAT enabled, we can run into issues in certain environments. Currently Submariner supports VxLAN cable-driver in a no-NAT deployments.

when in this case, we add a warning message.

Fix https://github.com/submariner-io/submariner/issues/1928


main.go
```diff
@@ -141,6 +142,9 @@ func main() {
 	}
 
 	components.submSpec.CableDriver = strings.ToLower(components.submSpec.CableDriver)
+	if components.submSpec.CableDriver == vxlan.CableDriverName && components.submSpec.NATEnabled {
+		logger.Warning("VxLan cable-driver is supported only with no NAT deployments")
+	}
```
tpantelis: I'd prefer to do this check in the [vxlan cable driver](https://github.com/submariner-io/submariner/blob/eafa40319d27327460351f9e107a351d0d9d6225/pkg/cable/vxlan/vxlan.go#L99).

tpantelis (changes requested):

main.go
cyclinder: Thanks for reviewing!  Updated.

dfarrell07:
Thanks for the PR @cyclinder!

dfarrell07:
FYI @tpantelis is in crunch time trying to finish some large changes before 0.15.0, that's why his re-review is slower than usual.

pkg/cable/vxlan/vxlan.go
```diff
@@ -97,6 +97,10 @@ func NewDriver(localEndpoint *types.SubmarinerEndpoint, localCluster *types.Subm
 		localCluster:  *localCluster,
 	}
 
+	if strings.ToLower(localEndpoint.Spec.CableName) == CableDriverName && localEndpoint.Spec.NATEnabled {
```
dfarrell07:
```
pkg/cable/vxlan/vxlan.go:100:5: equalFold: consider replacing with strings.EqualFold(localEndpoint.Spec.CableName, CableDriverName) (gocritic)
	if strings.ToLower(localEndpoint.Spec.CableName) == CableDriverName && localEndpoint.Spec.NATEnabled {
	   ^
```

This is flagged by Go linting

https://github.com/submariner-io/submariner/actions/runs/4485520060/jobs/7926833046?pr=2346

cyclinder: Thanks for the reply, Updated!  No push, I'm feel ok :)

tpantelis (approved):

sridhargaddam (approved):

skitt (approved):

PR #2354, opened 2023-03-24
Improve public IP resolver logging

### LINKED ISSUE #2350: Improve public IP resolver logging
Currently, when public IP resolvers cause errors, we log those errors, but we don’t log the overall state — so just looking at the logs is misleading.

We should either log success at the same severity, or only log errors if we can’t get a public IP address _at all_.

sridhargaddam:
Fixes: https://github.com/submariner-io/submariner/issues/2350

### DIFF
--- a/pkg/endpoint/public_ip.go
+++ b/pkg/endpoint/public_ip.go
@@ -31,7 +31,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/submariner-io/admiral/pkg/log"
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/types"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


sridhargaddam:
Sample output with simulated errors.

pkg/endpoint/public_ip.go
```diff
@@ -104,11 +105,15 @@ func getPublicIP(submSpec *types.SubmarinerSpecification, k8sClient kubernetes.I
 		}
 
 		// If this resolver failed, we log it, but we fall back to the next one
-		logger.V(log.DEBUG).Infof("Error resolving public IP with resolver %s, config: %s: %v", resolver, config, err)
+		if wrappedErr == nil {
+			wrappedErr = errors.Wrapf(err, "\nResolver[%q]", resolver)
+		} else {
+			wrappedErr = errors.Wrapf(wrappedErr, "\nResolver[%q]: %v", resolver, err.Error())
+		}
```
tpantelis: Rather than wrapping each error in a nested manner, perhaps use [NewAggregate](https://github.com/kubernetes/apimachinery/blob/8d1258da8f386b809d312cdda316366d5612f54e/pkg/util/errors/errors.go#L46) instead - a composite/aggregate error seems more suitable here..

tpantelis (changes requested):

skitt (approved):

dfarrell07 (approved):

tpantelis (approved):

PR #2357, opened 2023-03-27
Add profiling endpoints to the metrics servers

skitt:
This provides simple access to profiling information in the gateway and globalnet binaries. Forwarding the port locally provides access to a web-based interface, http://localhost:6060/debug/pprof/ (if the port is forwarded to :6060), which can be used directly or with tools such as https://github.com/google/pprof.

Screenshot from 2023-03-27 15-31-31

Screenshot from 2023-03-27 15-34-38

Screenshot from 2023-03-27 15-30-13

If this is deemed useful we can generalise it to other binaries which don't already provide metrics.

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"strings"
 	"sync"
@@ -346,6 +347,7 @@ func startHTTPServer(spec *types.SubmarinerSpecification) *http.Server {
 	srv := &http.Server{Addr: ":" + spec.MetricsPort, ReadHeaderTimeout: 60 * time.Second}
 
 	http.Handle("/metrics", promhttp.Handler())
+	http.HandleFunc("/debug", pprof.Profile)
 
 	go func() {
 		if err := srv.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
--- a/pkg/globalnet/main.go
+++ b/pkg/globalnet/main.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"flag"
 	"net/http"
+	"net/http/pprof"
 	"time"
 
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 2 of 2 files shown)


dfarrell07 (approved):
Very cool! Just to be clear, this doesn't expose a new port (same port as metrics server) that we might need to document, just a new endpoint for the existing port.

main.go
```diff
@@ -334,6 +335,7 @@ func startHTTPServer(spec *types.SubmarinerSpecification) *http.Server {
 	srv := &http.Server{Addr: ":" + spec.MetricsPort, ReadHeaderTimeout: 60 * time.Second}
 
 	http.Handle("/metrics", promhttp.Handler())
+	http.HandleFunc("/debug", pprof.Profile)
```
sridhargaddam, line 350: Does this enable profiling unconditionally even in production (ie.., non debug builds)? Generally profiling is enabled to investigate any performance issues and is turned off once we are done. 

Any idea about the overhead it introduces in the system?

sridhargaddam (changes requested):

sridhargaddam:
> Very cool! Just to be clear, this doesn't expose a new port (same port as metrics server) that we might need to document, just a new endpoint for the existing port.

If I understand correctly, the way to use this is as follows.
1. Submariner Gateway (and globalnet) pods already support metrics via http interface and listen on the TCP Port which is configured as env variable SUBMARINER_METRICSPORT (default value 32780) on the pod.
2. So, to test this, you can export your KUBECONFIG to point to the cluster and run `kubectl port-forward -n submariner-operator pod/<submariner-gateway-pod-name> 32780:32780` on your host
3. You can now access the profiling info in your browser at the following URL - http://localhost:32780/debug/pprof/

main.go
skitt: I’ve updated the commit message:

> The performance impact is minimal, except when streaming CPU profiles; that only happens on demand, e.g. with a query on .../debug/pprof/profile?seconds=30 (for a 30-second profile). See https://medium.com/google-cloud/continuous-profiling-of-go-programs-96d4416af77b which explains how Google run with permanent profiling; the 5% cost (amortised to 0.5%) is incurred by streaming profiles periodically, and doesn't apply to this change.

The idea here is to provide permanent access to profiling features. This will allow us (with appropriate permissions) to examine heap problems, blocked threads etc. in production setups without having to reproduce the trigger.

skitt:
> If I understand correctly, the way to use this is as follows.

Exactly! If this is merged for 0.15 I will give a demo in the demo session next week.

main.go
sridhargaddam: Cool, thanks.

sridhargaddam (approved):

tpantelis (approved):

Issue #2345, opened 2023-03-20
Gateway connections sometimes not getting established


tpantelis:
We've been seeing flakiness with E2E tests where the common theme is that one of the gateway connections doesn’t get fully established. In one run in which _cluster3_ is attempting to install the `Endpoint` from _cluster1_, we see the following in the log:

```
2023-03-20T04:07:25.648Z INF ..gine/cableengine.go:195 CableEngine          Installing Endpoint cable "submariner-cable-cluster1-172-18-0-4"
  002 listening for IKE messages
  003 IP interfaces vetha57b6c2c and veth43c47b84 share address 10.132.0.1!
…
  2023-03-20T04:07:25.661Z INF ..reswan/libreswan.go:347 libreswan            Creating connection(s) for {"metadata":{"name":"cluster1-submariner-cable-cluster1-172-18-0-4","namespace":"submariner-operator","uid":"91194d54-b2a8-44ab-a0af-b5d3dfe6528f","resourceVersion":"1279","generation":1,"creationTimestamp":"2023-03-20T04:07:24Z","labels":{"submariner-io/clusterID":"cluster1"},"managedFields":[{"manager":"submariner-gateway","operation":"Update","apiVersion":"submariner.io/v1","time":"2023-03-20T04:07:24Z","fieldsType":"FieldsV1","fieldsV1":{"f:metadata":{"f:labels":{".":{},"f:submariner-io/clusterID":{}}},"f:spec":{".":{},"f:backend":{},"f:backend_config":{".":{},"f:natt-discovery-port":{},"f:preferred-server":{},"f:udp-port":{}},"f:cable_name":{},"f:cluster_id":{},"f:healthCheckIP":{},"f:hostname":{},"f:nat_enabled":{},"f:private_ip":{},"f:public_ip":{},"f:subnets":{}}}}]},"spec":{"cluster_id":"cluster1","cable_name":"submariner-cable-cluster1-172-18-0-4","healthCheckIP":"10.130.0.1","hostname":"cluster1-control-plane","subnets":["100.66.0.0/16","10.130.0.0/16"],"private_ip":"172.18.0.4","public_ip":"13.64.10.200","nat_enabled":false,"backend":"libreswan","backend_config":{"natt-discovery-port":"4490","preferred-server":"false","udp-port":"4500"}}} in bi-directional mode
[...]

PR #2363, opened 2023-03-28
Cancel whack calls after 5s

skitt:
See individual commits.

This should help with #2345.


pkg/cable/libreswan/libreswan.go
```diff
@@ -169,8 +170,11 @@ func (i *libreswan) Init() error {
 var trafficStatusRE = regexp.MustCompile(`.* "([^"]+)"[^,]*, .*inBytes=(\d+), outBytes=(\d+).*`)
 
 func retrieveActiveConnectionStats() (map[string]int, map[string]int, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
tpantelis: Minor thing... Although both `TODO` and `Background` behave the same, ie do nothing, I think the semantics of `Background` apply here, ie as a top-level `Context`.

```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```

pkg/cable/libreswan/libreswan.go
```diff
@@ -301,13 +305,20 @@ func whack(args ...string) error {
 	var err error
 
 	for i := 0; i < 3; i++ {
-		cmd := exec.Command("/usr/libexec/ipsec/whack", args...)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+		err := func() error {
+			ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
tpantelis:
```suggestion
			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```

pkg/cable/libreswan/libreswan.go
tpantelis: Can you make `5*time.Second` a constant, eg `whackTimeout`?

tpantelis (changes requested):

pkg/cable/libreswan/libreswan.go
skitt: I thought `TODO()` was a better fit, one of its documented use-cases is “because the surrounding function has not yet been extended to accept a Context parameter” — ideally these functions should be extended to take a parent context as input.

tpantelis: yeah perhaps.  `TODO` is fine.

tpantelis (approved):

sridhargaddam (approved):

sridhargaddam:
@skitt @tpantelis should we backport this PR?

tpantelis:
> @skitt @tpantelis should we backport this PR?

Probably but let's let it bake in **devel** first to see how it works.

dfarrell07:
Should 0.15.0-rc0 depend on this?

dfarrell07:
Four E2E jobs failed, re-running.

dfarrell07:
This went in before release-0.15 was cut, so it doesn't need to be backported there before RC0.

PR #2374, opened 2023-04-01
Define status subresource in globalnet CRDs

tpantelis:
`GlobalEgressIP`, `ClusterGlobalEgressIP` and `GlobalEgressIP` follow the spec-status separation paradigm so it makes sense to define status as an actual subresource. The controllers for each already use the `StatusUpdateFederator` to specifically only update the status field.

This will need to be backported to 0.15 to unblock rc0 as globalnet E2E is failing. The main reason is that **admiral** `CreateOrUpdate`  doesn't properly handle CRDs that have the status field but don't define it as a subresource. While we need to fix that, it made it apparent that the globalnet CRDs really should define the subresource.

### DIFF
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -196,6 +196,7 @@ func (c *Connection) SetStatus(status ConnectionStatus, messageFormat string, a 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:shortName="geip"
+// +kubebuilder:subresource:status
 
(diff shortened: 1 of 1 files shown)


sridhargaddam (approved):

vthapar (approved):

sridhargaddam (dismissed):
@tpantelis thanks for working on the fix. I'd like to collect some feedback from skitt, so requesting changes.

@skitt, we have an equivalent PR on the devel (under review), but I guess @tpantelis wanted to try if the changes are working on 0.15 so pushed this PR directly on release-0.15 branch. I'd like to hear from you if we can merge this in the current state or if you prefer the devel PR to be merged and then backported to release-0.15 branch.

skitt (approved):

PR #2400, opened 2023-04-06
DO_NOT_MERGE: rp_filter setting not applied on vx-submariner

sridhargaddam:


skitt:
@sridhargaddam I think this can be closed now, right?

PR #2402, opened 2023-04-06
Enable jobs on merge queues

skitt:
See
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue for details.

The target_branch and apply-suggestions-commits jobs are required, so they need to run on merge group events, but their results are only valid outside the merge queue, so they are ignored if the event isn't a pull_request event.

The DCO app is required by the CNCF, but doesn't support merge queues yet. Until it does, this restores the old DCO check for PRs, ignoring it in merge queues; that way, we can require that, ensuring that PR commits have a DCO, without blocking the merge queue.

vulnerability-scanning isn't relevant in the merge queue, so it is skipped there.

(The general reasoning is that PR checks should verify everything related to the commits themselves; merge queue checks should verify everything that changes as a result of the position of the PR in the queue, which doesn't include commit texts, so anything related to that can be ignored in the merge queue if it's been checked in the PR. GitHub doesn't support different requirements for PRs and merge queues yet, so this needs to be handled by skipping irrelevant tests in merge queues in the workflow configuration itself.)

Depends on https://github.com/z0al/dependent-issues/issues/538

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -3,12 +3,14 @@ name: Branch Checks
 
 on:
   pull_request:
+  merge_group:
 
 permissions: {}
 
 jobs:
   target_branch:
     name: PR targets branch
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
       - name: Check that the PR targets devel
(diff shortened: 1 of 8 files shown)


dfarrell07:
We think Dependabot's new feature to group PRs might be a better solution for us currently, given the lack of support for Merge Queues in our dependencies.

PR #2423, opened 2023-04-14
Fix connection issues with RHEL9 GW nodes

### LINKED ISSUE #2422: Connection failures seen with RHEL-9 Gateway nodes
**What happened**:

On RHEL-9/CoreOS 413.92.xx nodes when OCP is installed with OpenShiftSDN, it was seen that connectivity between non-Gateway to non-Gateway nodes is broken. Upon debugging the issue, the problem turns out to be wrong configuration related to reverse path filtering (i.e., /proc/sys/net/ipv4/conf/vx-submariner/rp_filter) for vx-submariner interface on the Gateway nodes.

Surprisingly, when the route-agent, after creating the vx-submariner, configures the LooseMode Reverse Path filtering on the iface, there is no error from from the netlink library, but the LooseMode (aka "2") configuration is not applied on the interface.

```
[90m2023-03-31T18:43:27.319Z[0m [33mDBG[0m ..oxy/gw_transition.go:47 KubeProxy            The current node has become a Gateway
[90m2023-03-31T18:43:27.319Z[0m [1m[31mERR[0m[0m ..oxy/routes_iface.go:159 KubeProxy   Error retrieving link by name "vx-submariner" [36merror=[0m[31m"Link not found"[0m
[90m2023-03-31T18:43:27.319Z[0m [32mINF[0m ..oxy/gw_transition.go:55 KubeProxy     Creating the vxlan interface: vx-submariner on the gateway node
[90m2023-03-31T18:43:27.320Z[0m [33mDBG[0m ../kubeproxy/vxlan.go:279 KubeProxy  Successfully configured reverse path filter to loose mode on "vx-submariner"
[90m2023-03-31T18:43:27.392Z[0m [32mINF[0m ..oller/controller.go:150 EventController  Event controller started
``` 

Note: This issue is applicable only when using OpenShiftSDN and not with OVN-K CNI.

non-Gateway to non-Gateway communication should work fine.

A quick work-around is to restart the route-agent pods using the following command and this will fix the issue.
```
kubectl delete pod -n submariner-operator -l app=submariner-routeagent
```

**Environment**:
Submariner version 0.15.0-m3, 
Kubernetes Server version: v1.26.2+06e8c46, 
OCP 4.13.0-rc0
RHEL-9/CentOS Stream CoreOS 413.92.202303190222-0

sridhargaddam:
When a RHEL-9 node is used as a Submariner Gateway node, it was seen that immediately after creating the interface if we configure the rp_filter setting its getting overwritten. However, if we add a small delay for the interface to come up and then configure the rp_filter it works fine. This is still to be considred as a temporary work-around until a proper root cause is identified.

Fixes: https://github.com/submariner-io/submariner/issues/2422

### DIFF
--- a/pkg/netlink/netlink.go
+++ b/pkg/netlink/netlink.go
@@ -53,6 +53,7 @@ type Basic interface {
 	XfrmPolicyDel(policy *netlink.XfrmPolicy) error
 	XfrmPolicyList(family int) ([]netlink.XfrmPolicy, error)
 	EnableLooseModeReversePathFilter(interfaceName string) error
+	GetReversePathFilter(interfaceName string) ([]byte, error)
 	ConfigureTCPMTUProbe(mtuProbe, baseMss string) error
 }
 
@@ -152,6 +153,20 @@ func (n *netlinkType) EnableLooseModeReversePathFilter(interfaceName string) err
 	// Enable loose mode (rp_filter=2) reverse path filtering on the vxlan interface.
 	err := setSysctl("/proc/sys/net/ipv4/conf/"+interfaceName+"/rp_filter", []byte("2"))
 	return errors.Wrapf(err, "unable to update rp_filter proc entry for interface %q", interfaceName)
+}
+
+func (n *netlinkType) GetReversePathFilter(interfaceName string) ([]byte, error) {
+	path := "/proc/sys/net/ipv4/conf/" + interfaceName + "/rp_filter"
+
+	existing, err := os.ReadFile(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to read proc entry for interface %q", interfaceName)
+	}
+
+	// Ignore leading and terminating newlines
+	existing = bytes.Trim(existing, "\n")
+
+	return existing, nil
 }
 
 func (n *netlinkType) FlushRouteTable(tableID int) error {
--- a/pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
+++ b/pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
@@ -19,11 +19,14 @@ limitations under the License.
 package kubeproxy
 
 import (
+	"bytes"
 	goerrors "errors"
+	"fmt"
 	"net"
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
@@ -271,12 +274,12 @@ func (kp *SyncHandler) createVxLANInterface(activeEndPoint string, ifaceType int
 			}
 		}
 
-		err = kp.netLink.EnableLooseModeReversePathFilter(VxLANIface)
+		err = kp.ensureLooseModeIsConfigured(VxLANIface)
 		if err != nil {
-			return errors.Wrap(err, "error enabling loose mode")
+			return errors.Wrap(err, "error while validating loose mode")
 		}
 
-		logger.V(log.DEBUG).Infof("Successfully configured reverse path filter to loose mode on %q", VxLANIface)
+		logger.Infof("Successfully configured reverse path filter to loose mode on %q", VxLANIface)
 	} else if ifaceType == VxInterfaceWorker {
 		// non-Gateway/Worker Node
 		attrs := &vxLanAttributes{
@@ -300,4 +303,28 @@ func (kp *SyncHandler) createVxLANInterface(activeEndPoint string, ifaceType int
 	}
 
 	return nil
+}
+
+func (kp *SyncHandler) ensureLooseModeIsConfigured(iface string) error {
+	for i := 0; i < 10; i++ {
+		// Revisit: This is a temporary work-around to fix https://github.com/submariner-io/submariner/issues/2422
+		// Allow the vx-submariner interface to get initialized.
+		time.Sleep(100 * time.Millisecond)
+
+		rpFilterSetting, err := kp.netLink.GetReversePathFilter(iface)
+		if err == nil {
+			if bytes.Equal(rpFilterSetting, []byte("2")) {
+				return nil
+			}
+		} else {
+			logger.Warningf("Error retrieving reverse path filter for %q: %v", iface, err)
+		}
+
+		err = kp.netLink.EnableLooseModeReversePathFilter(iface)
+		if err != nil {
+			return errors.Wrapf(err, "error enabling loose mode on iface %q", iface)
+		}
+	}
+
+	return fmt.Errorf("loose mode not configured on iface %q", iface)
 }
--- a/pkg/netlink/fake/netlink.go
+++ b/pkg/netlink/fake/netlink.go
@@ -257,6 +257,10 @@ func (n *basicType) EnableLooseModeReversePathFilter(_ string) error {
 	return nil
 }
 
+func (n *basicType) GetReversePathFilter(_ string) ([]byte, error) {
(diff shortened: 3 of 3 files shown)


pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
```diff
@@ -271,12 +274,21 @@ func (kp *SyncHandler) createVxLANInterface(activeEndPoint string, ifaceType int
...
+
 		err = kp.netLink.EnableLooseModeReversePathFilter(VxLANIface)
 		if err != nil {
-			return errors.Wrap(err, "error enabling loose mode")
+			return errors.Wrapf(err, "error enabling loose mode on iface %q", VxLANIface)
+		}
+
+		_, err = kp.ensureLooseModeIsConfigured(VxLANIface)
+		if err != nil {
+			return errors.Wrap(err, "error while validating loose mode")
```
skitt: Let’s just call `ensureLooseModeIsConfigured`, it ends up doing the same as the initial wait and `EnableLooseMode`.

skitt (changes requested):

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
sridhargaddam: I thought it would be easier to read with the comment that we are waiting for the interface to get initialised, but anyways, I can change it. Thanks for reviewing.

pkg/netlink/netlink.go
```diff
@@ -154,6 +155,20 @@ func (n *netlinkType) EnableLooseModeReversePathFilter(interfaceName string) err
...
+
+	existing, err := os.ReadFile(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to read proc entry for interface %q", interfaceName)
+	}
+
+	// Ignore leading and terminating newlines
+	existing = bytes.Trim(existing, "\n")
+
+	return existing, nil
```
tpantelis, line 169: How about return `string` for convenience?

```suggestion
	return string(bytes.Trim(existing, "\n")), nil
```

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
```diff
@@ -301,3 +304,25 @@ func (kp *SyncHandler) createVxLANInterface(activeEndPoint string, ifaceType int
 
 	return nil
 }
+
+func (kp *SyncHandler) ensureLooseModeIsConfigured(iface string) (bool, error) {
```
tpantelis: The returned `bool` isn't used - it's redundant with the returned error.

```suggestion
func (kp *SyncHandler) ensureLooseModeIsConfigured(iface string) error {
```

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
```diff
@@ -301,3 +304,25 @@ func (kp *SyncHandler) createVxLANInterface(activeEndPoint string, ifaceType int
...
+	for i := 0; i < 10; i++ {
+		// Revisit: This is a temporary work-around to fix https://github.com/submariner-io/submariner/issues/2422
+		// Allow the vx-submariner interface to get initialized.
+		time.Sleep(100 * time.Millisecond)
```
tpantelis, line 312: Perhaps we should move this after L324 below to avoid a 100ms wait in case it's initially set.

tpantelis (changes requested):

skitt (approved):

pkg/netlink/netlink.go
sridhargaddam: It is aligned with the setSysctl APIs. 

https://github.com/submariner-io/submariner/blob/ba52b8b2bb3494cb4e9c781d12f1ff68bd7b2097/pkg/netlink/netlink.go#L174

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
sridhargaddam: I've seen that immediately after creating the interface, if we set the rp_filter to 2 and read it, its value sometimes is 2 and later gets updated. Adding this delay seems to allow the interface to get initialised and then updating the rp_filter to 2 was working in the setup I was looking at.

As this is a one time operation (only for GW node) and is not periodically invoked, this small delay should be okay IMO for other platforms.

tpantelis: If we do as suggested [here](https://github.com/submariner-io/submariner/issues/2422#issuecomment-1511351483) then it would/should be set to 2 and we'd avoid the delay. But not a big deal - it's negligible anyway.

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
```diff
@@ -301,3 +304,25 @@ func (kp *SyncHandler) createVxLANInterface(activeEndPoint string, ifaceType int
...
+
+		rpFilterSetting, err := kp.netLink.GetReversePathFilter(iface)
+		if err == nil {
+			if bytes.Equal(rpFilterSetting, []byte("2")) {
+				return nil
+			}
+		}
```
tpantelis: We should at least log the error here.

```suggestion
		if err == nil {
	            if bytes.Equal(rpFilterSetting, []byte("2")) {
		        return nil
		    }
		} else {
		    logger.Warningf("Error retrieving reverse path filter: %v", err)
		}
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
sridhargaddam: TBH, it does not really help us in anyway even if we have the log message. But if you insist, I can add the log message.

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
sridhargaddam: See the later messages...

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
```diff
@@ -301,3 +304,25 @@ func (kp *SyncHandler) createVxLANInterface(activeEndPoint string, ifaceType int
...
+
+		err = kp.netLink.EnableLooseModeReversePathFilter(iface)
+		if err != nil {
```
yboaron, line 324: Shouldn't we continue with the for loop instead of exiting in case of error here ?

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
tpantelis: I wouldn't say I'm insisting - I think it's better to handle the error in some way rather than ignore it, especially since we're depending on successful retrieval to proceed.  Who knows - maybe something strange could cause retrieval to fail...

pkg/routeagent_driver/handlers/kubeproxy/vxlan.go
sridhargaddam: Failure to update the rp_filter setting on an interface was always logged and treated as an error. This PR preserves the existing behaviour.

Coming to the issue on RHEL-9, we do not get any error while updating the rp_filter, its just that it gets overwritten with some default config when a new interface is created.

tpantelis (approved):

yboaron (approved):

PR #2447, opened 2023-04-25
Use ipset-go instead of calling the ipset binary

skitt:
This reduces our requirements on the environment in which Submariner components run.

The ipset binaries are kept in the container images, for debugging purposes and because subctl gather relies on ipset in the route agent.


skitt:
The ipset-go library hasn’t produced any releases (yet), but it’s small, has tests, and responds to pull requests.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -177,14 +203,24 @@ func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
 func (h *mtuHandler) LocalEndpointRemoved(endpoint *submV1.Endpoint) error {
 	subnets := extractIPv4Subnets(&endpoint.Spec)
 	for _, subnet := range subnets {
-		err := h.localIPSet.DelEntry(subnet)
+		entry, err := newIPv4SubnetEntry(subnet)
+		if err != nil {
+			return errors.Wrapf(err, "error deleting local IP set entry %s", subnet)
```
tpantelis: Elsewhere in this function we just log errors, it seems we should be consistent here, especially since this isn't a recoverable/transient error.

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -177,14 +203,24 @@ func (h *mtuHandler) LocalEndpointCreated(endpoint *submV1.Endpoint) error {
...
 		if err != nil {
 			logger.Errorf(err, "Error deleting the subnet %q from the local IPSet", subnet)
 		}
 	}
 
 	for _, subnet := range h.localClusterCidr {
-		err := h.localIPSet.DelEntry(subnet)
+		entry, err := newIPv4SubnetEntry(subnet)
+		if err != nil {
+			return errors.Wrapf(err, "error deleting local IP set entry %s", subnet)
```
tpantelis: Same here...

dfarrell07:
> The ipset-go library hasn’t produced any releases (yet), but it’s small, has tests, and responds to pull requests.

Not necessary voting -1, but it also hasn't had any contributors other than you, doesn't have a great [git history](https://github.com/lrh3321/ipset-go/commits/master), and doesn't seem to be used by any consumers. :sweat_smile: 

Still looking..

skitt:
There’s also https://github.com/nadoo/ipset, which has no dependencies and has produced releases, but it’s GPL-3-licensed (which I don’t mind, but probably won’t fly with the CNCF), and it has no tests.

dfarrell07:
I was able to ask Stephen a bunch of questions about this on a recent call. Overall I think we don't have a great path either way and there isn't one choice that seems obviously best/worst.

* Moving to only-nftables likely isn't going to happen as OpenShiftSDN is moving to focus on OVN.
* The current solution adds a few os.exec calls that this would avoid.
* It's not clear if this path would reduce the likelihood of concurrency issues we see with the current wrapped binary, as those seem to come from timing between userspace and the kernel (which both solutions could have).
* The current library we're using was abandoned by Tencent and forked in to K8s. We're not using the latest version of that, so if we stick with the current design we should likely look into keeping that up-to-date.
* We may want to move subctl gather to use this lib too, to drop the binary there, but that would be a more major change.

If we look through the code of this lib and feel safe about it, I think it's fine to move to it.

nyechiel:
@skitt this one needs a rebase

nyechiel:
> @skitt this one needs a rebase

ping :)

nyechiel:
bump

tpantelis:
Do we still need or want to do this?

PR #2478, opened 2023-05-04
Bump ovn-kubernetes to the current tip

skitt:
This brings us up-to-date with changes made in the last year.


tpantelis:
`CreateOrUpdateLogicalRouterPorts` was changed to `CreateOrUpdateLogicalRouterPort` by https://github.com/ovn-org/ovn-kubernetes/commit/ec8a85faac32336fe64cf3dcf90061f0144799ba#diff-5d62dde13a2856ea3d4b324968dbaeed261fcb74cb3b81909edf40a6634fa1f4 to pass an optional `*nbdb.GatewayChassis`.  It doesn't seem that's relevant in our case so we can pass nil.

tpantelis:
[This commit](https://github.com/ovn-org/ovn-kubernetes/commit/20f1c848e3635fa04453242c983ca993b4a306b4#diff-5d62dde13a2856ea3d4b324968dbaeed261fcb74cb3b81909edf40a6634fa1f4) essentially replaces `CreateOrUpdateLogicalRouterStaticRoutesWithPredicate` with `CreateOrReplaceLogicalRouterStaticRouteWithPredicate`.  However it's not a drop-in replacement as the latter also removes additional routes. I'm not sure if that's what we want or if that's even relevant. Otherwise we can simply call `CreateOrUpdateLogicalRouterStaticRoutesWithPredicateOps` with nil `[]libovsdb.Operation`.

pkg/networkplugin-syncer/handlers/ovn/ovn_logical_router_static_route.go
```diff
@@ -42,7 +42,7 @@ func (ovn *SyncHandler) reconcileSubOvnLogicalRouterStaticRoutes(port, nextHop s
 			return item.OutputPort != nil && *item.OutputPort == port && item.IPPrefix == lrsr.IPPrefix
 		}
 
-		err = libovsdbops.CreateOrUpdateLogicalRouterStaticRoutesWithPredicate(ovn.nbdb, submarinerLogicalRouter, lrsr, LRSRPred)
+		err = libovsdbops.CreateOrReplaceLogicalRouterStaticRouteWithPredicate(ovn.nbdb, submarinerLogicalRouter, lrsr, LRSRPred)
```
tpantelis: `CreateOrReplaceLogicalRouterStaticRouteWithPredicate` also removes additional routes. Is  that wanted or relevant in our case? Otherwise we can simply call `CreateOrUpdateLogicalRouterStaticRoutesWithPredicateOps` with nil `[]libovsdb.Operation`.

skitt: I was following the commit you noted too, but yes, it might not be appropriate.

nyechiel:
@aswinsuryan @sridhargaddam  is this something we want to merge before 0.16.0-rc0?

aswinsuryan:
We can revisit this once https://github.com/submariner-io/submariner/pull/2621 is merged as some of the API is not now used.

sridhargaddam:
> @aswinsuryan @sridhargaddam is this something we want to merge before 0.16.0-rc0?

We discused this on slack and decided to keep this on hold as rc0 is on monday and we want to avoid any breakages.
https://kubernetes.slack.com/archives/C010RJV694M/p1694177898896719

dfarrell07:
To clarify - there were some functions removed that we used, and there are not clear replacements. Stephen tried pinging an OVN dev a while back and didn't get a response. I guess we should keep asking.

PR #2499, opened 2023-05-26
Avoid cableEngine cleanup when the gateway pod is restarted

### LINKED ISSUE #2498: Gateway pod restart is causing datapath disruption
**What happened**:

In the current code, it was seen that when the submariner-gateway pod gets restarted, there is a brief datapath disruption until the new pod comes online. It seems to be introduced as part of the [following PR](https://github.com/submariner-io/submariner/pull/1689/commits/205a12bfa73bcce8d0f5c85a3ecfdbd4ee52b7f5) while addressing OVN use-cases. 

The cleanup code in submariner-gateway pod is not really required as its handled by the [route-agent code](https://github.com/submariner-io/submariner/tree/devel/pkg/routeagent_driver/cabledriver) when there is any transition in the Gateway nodes. 

When the submariner-gateway pod is restarted, ideally it would be good if we can avoid any disruptions to the datapath.

1. Deploy a KIND cluster.
2. Deploy two pods one on non-GW nodes of cluster-1 and cluster-2
3. Ping from pod in cluster1 to the pod in cluster2 
4. While the ping is going on, delete the submariner-gateway pod in cluster2
5. You will notice a brief interruption in the ping traffic until the submariner-gateway pod comes online.

sridhargaddam:
In the current code, it was seen that when the submariner-gateway pod gets restarted, there is a brief datapath disruption until the new pod comes online. The cleanup code in submariner-gateway pod is not required as its handled by the route-agent code when there is any transition in the active Gateway of the cluster.

This PR removes the cleanup code from submariner-gateway pod which fixes the issue with VxLAN Cable Driver, but Libreswan Cable Driver seems to have some other issue[s] which needs to be addressed separately.

Screenshot from 2023-05-26 17-01-59

Partially fixes: https://github.com/submariner-io/submariner/issues/2498

### DIFF
--- a/main.go
+++ b/main.go
@@ -207,10 +207,6 @@ func main() {
 
 	<-components.stopCh
 
-	if err := components.cableEngine.Cleanup(); err != nil {
-		logger.Error(nil, "Error cleaning up cableEngine resources before removing Gateway")
-	}
-
 	logger.Info("All controllers stopped or exited. Stopping main loop")
 
 	if err := httpServer.Shutdown(context.TODO()); err != nil {


tpantelis (approved):

sridhargaddam:
@tpantelis @skitt @aswinsuryan @yboaron I verified this PR couple of times in a KIND deployment and captured my observations in the commit message. I haven't got a chance to verify it on an OCP deployment. Do you think we should validate this once (although the fix looks valid and is reasonable) even with OCP before merging. Any thoughts?

skitt:
> Do you think we should validate this once (although the fix looks valid and is reasonable) even with OCP before merging.

Given that the change was introduced for OVNK, I think we should verify it on OCP with OVNK before merging.

skitt:
Ah, I was basing my comment above only on the notification email, which didn’t include the nice table in the description. I see you’ve already tested this with OVNK on kind, so I don’t think testing on OCP is absolutely necessary.

dfarrell07 (approved):

sridhargaddam:
Backport PRs:
release-0.15: https://github.com/submariner-io/submariner/pull/2504
release-0.14: https://github.com/submariner-io/submariner/pull/2518
release-0.13: https://github.com/submariner-io/submariner/pull/2519

PR #2526, opened 2023-06-04
Retry reading local node details from k8s-api

### LINKED ISSUE #2525: GW pod restarts when it fails to read local node information once
The gateway pod reads local node information from k8s-api to retrieve 
node details required for local endpoint creation.

In the current code, in case of failure [1]  to read the local node information from k8s-api, the code will exit with an error and the pod will be restarted by k8s with the default back-off timing.

we have seen the GW restart counter to be ~15K on some deployments recently over a period of 3-4 months, although it seems that root cause was almost always some temporary failure in the k8s-api or some other k8s-infra and not something directly related to GW functionality.

Also, we noticed that in some cases restarting the GW pod could lead to DP failure.

Env:
Submariner 0.15

[1] 
+ trap 'exit 1' SIGTERM SIGINT
+ export CHARON_PID_FILE=/var/run/charon.pid
+ CHARON_PID_FILE=/var/run/charon.pid
+ rm -f /var/run/charon.pid
+ SUBMARINER_VERBOSITY=2
+ '[' false == true ']'
+ DEBUG=-v=2
++ cat /proc/sys/net/ipv4/conf/all/send_redirects
+ [[ 1 = 0 ]]
+ echo 0
+ exec submariner-gateway -v=2 -alsologtostderr
[90m2023-05-29T09:33:31.760Z[0m [32mINF[0m ..-source/app/main.go:109 main                 Starting the submariner gateway engine
[90m2023-05-29T09:33:31.767Z[0m [32mINF[0m ..-source/app/main.go:117 main                 Parsed env variables
W0529 09:33:31.769761       1 client_config.go:617] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
[90m2023-05-29T09:33:31.782Z[0m [32mINF[0m ..-source/app/main.go:133 main                 Creating the cable engine
[90m2023-05-29T09:33:31.782Z[0m [32mINF[0m ..-source/app/main.go:144 main                 AIR_GAPPED_DEPLOYMENT is set to false
[90m2023-05-29T09:33:35.895Z[0m [1m[31mFTL[0m[0m ..nt/local_endpoint.go:51 Endpoint             Error getting information on the local node [36merror=[0m[31m"unable to find local node \"m2prdocs-tjb8n-worker-nk4xm\": Get \"https://10.207.0.1:443/api/v1/nodes/m2prdocs-tjb8n-worker-nk4xm\": dial tcp 10.207.0.1:443: connect: no route to host"[0m

yboaron:
Retry reading local node information from k8s-api
in case of error.

Fixes: https://github.com/submariner-io/submariner/issues/2525


pkg/node/node.go
```diff
@@ -21,23 +21,48 @@ package node
...
+	}, func() error {
+		var err error
+
+		node, err = clientset.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "unable to find local node %q", nodeName)
+		}
+
+		return nil
+	})
```
tpantelis, line 65: Would we expect a subsequent retry to succeed? If not, this will try 5 times - after that we'll still return an error and the pod will exit and restart so I'm not sure what this actually solves. Plus there are other places where we access the API server on startup where we don't retry and could similarly cause an exit and restart.

From the summary in the associated issue, it seems the motivation for this is to avoid a possible DP failure due to pod restart. I'm not sure of the details there but it seems we should try to address that (if possible) since restarts could still occur (unless we try to have the whole process retry indefinitely on startup).

tpantelis (changes requested):

pkg/node/node.go
yboaron: Well, there are two points here:

1. To avoid unnecessary traffic hit, I agree that it's better if GW pod restarts don't affect DP at all. But in the current implementation, there is a brief DP failure when GW pod is restarted (for libreswan cable driver).

2. We would like to reduce the restart counter for GW pod as much as possible, we have seen the GW restart counter to be ~15K on some deployments recently over a period of 3-4 months, although it seems that root cause was almost always some temporary failure in the k8s-api or some other k8s-infra and not something directly related to GW functionality.

I think it's worth retrying to call the k8s-api again in places where it can reduce the number of restarts caused by a temporary malfunction in the k8s-api even just to address point 2. 

> Plus there are other places where we access the API server on startup where we don't retry and could similarly cause an exit and restart

I think we should address also the other places. could you please share the relevant places?

sridhargaddam: Agree with Yossi. Not all API failures could trigger a restart, at some places in the code we already have the retry mechanism, but this PR addresses and important issue what was seen in some customer deployments. Lets get this merged and look for potential places where we have to retry.

sridhargaddam (approved):

pkg/node/node.go
tpantelis: OK. It looks like this is the first place in the GW where we access the API server so if this eventually succeeds then subsequent accesses should succeed as well.

skitt (approved):

pkg/node/node.go
```diff
@@ -21,23 +21,48 @@ package node
...
 	}
 
-	node, err := clientset.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
-	if err != nil {
-		return nil, errors.Wrapf(err, "unable to find local node %q", nodeName)
-	}
+	var node *v1.Node
+
+	err := retry.OnError(nodeRetry, func(err error) bool {
+		logger.Infof("retrying to read local node details: %s", err)
```
tpantelis:
```suggestion
		logger.Warningf("Error reading the local node - retrying: %v", err)
```

tpantelis (changes requested):

tpantelis (approved):

PR #2542, opened 2023-06-07
Provide detailed logs for certificate errors

skitt:
This adds an error handler which logs detailed information about certificates causing errors. This should help us debug certificate issues seen in production.

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ package main
 
 import (
 	"context"
+	"crypto/x509"
 	"errors"
 	"flag"
 	"fmt"
@@ -28,6 +29,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/kelseyhightower/envconfig"
@@ -35,6 +37,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/log/kzerolog"
+	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
 	"github.com/submariner-io/admiral/pkg/watcher"
 	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
@@ -51,6 +54,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/pod"
 	"github.com/submariner-io/submariner/pkg/types"
 	corev1 "k8s.io/api/core/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
@@ -100,8 +104,9 @@ const (
 )
 
 var (
-	VERSION = "not-compiled-properly"
-	logger  = log.Logger{Logger: logf.Log.WithName("main")}
+	VERSION            = "not-compiled-properly"
+	logger             = log.Logger{Logger: logf.Log.WithName("main")}
+	lastBadCertificate atomic.Value
 )
 
 func main() {
@@ -122,6 +127,19 @@ func main() {
 	httpServer := startHTTPServer(&components.submSpec)
 
 	var err error
+
+	//nolint:reassign // We need to reassign ErrorHandlers to register our handler
+	utilruntime.ErrorHandlers = append(utilruntime.ErrorHandlers, func(err error) {
+		var unknownAuthorityError x509.UnknownAuthorityError
+		if errors.As(err, &unknownAuthorityError) && lastBadCertificate.Swap(unknownAuthorityError.Cert) != unknownAuthorityError.Cert {
+			logger.Errorf(err, "Certificate error: %s", resource.ToJSON(err))
+		}
+		var certificateInvalidError x509.CertificateInvalidError
+		if errors.As(err, &certificateInvalidError) && lastBadCertificate.Swap(certificateInvalidError.Cert) != certificateInvalidError.Cert {
+			logger.Errorf(err, "Certificate error: %s", resource.ToJSON(err))
+		}
+		// The generic handler has already logged the error, no need to repeat if we don't want extra detail
+	})
 
 	components.restConfig, err = clientcmd.BuildConfigFromFlags(localMasterURL, localKubeconfig)
 	logger.FatalOnError(err, "Error building kubeconfig")


main.go
```diff
@@ -123,6 +125,19 @@ func main() {
...
+	//nolint:reassign // We need to reassign ErrorHandlers to register our handler
+	utilruntime.ErrorHandlers = append(utilruntime.ErrorHandlers, func(err error) {
+		var unknownAuthorityError x509.UnknownAuthorityError
+		if errors.As(err, &unknownAuthorityError) {
+			logger.Errorf(err, "Unknown authority error: %v", unknownAuthorityError)
+		}
+
+		var invalidCertificateError x509.CertificateInvalidError
+		if errors.As(err, &invalidCertificateError) {
+			logger.Errorf(err, "Invalid certificate error: %v", invalidCertificateError)
```
tpantelis: Using `%v`, we just get the error string printed:

`Invalid certificate error: x509: certificate has expired or is not yet valid: more detail`

which will be printed anyway.

I assume you want the detailed fields printed. With `%#v`, we get:

`Invalid certificate error: x509.CertificateInvalidError{Cert:(*x509.Certificate)(0xc0003f0580), Reason:1, Detail:"more detail"}`

However this does not follow/expand pointers. 

Converting to JSON does (using `resource.ToJSON(err)`):

```
Invalid certificate error: {
  "Cert": {
    "Raw": "AQIDBA==",
    "RawTBSCertificate": null,
    "RawSubjectPublicKeyInfo": null,
    "RawSubject": "BQY=",
    "RawIssuer": null,
    "Signature": null,
    "SignatureAlgorithm": 7,
    "PublicKeyAlgorithm": 2,
    "PublicKey": "PublicKey",
    "Version": 1,
    "SerialNumber": 123,
    "Issuer": {
      "Country": [
        "US"
      ],
      "Organization": [
        "RH"
      ],
      "OrganizationalUnit": null,
      "Locality": null,
      "Province": null,
      "StreetAddress": null,
      "PostalCode": null,
      "SerialNumber": "",
      "CommonName": "",
      "Names": null,
      "ExtraNames": null
    },
    "Subject": {
      "Country": null,
      "Organization": null,
      "OrganizationalUnit": null,
      "Locality": null,
      "Province": null,
      "StreetAddress": null,
      "PostalCode": null,
      "SerialNumber": "",
      "CommonName": "",
      "Names": null,
      "ExtraNames": null
    },
    "NotBefore": "0001-01-01T00:00:00Z",
    "NotAfter": "0001-01-01T00:00:00Z",
    "KeyUsage": 0,
    "Extensions": null,
    "ExtraExtensions": null,
    "UnhandledCriticalExtensions": null,
    "ExtKeyUsage": null,
    "UnknownExtKeyUsage": null,
    "BasicConstraintsValid": false,
    "IsCA": false,
    "MaxPathLen": 0,
    "MaxPathLenZero": false,
    "SubjectKeyId": null,
    "AuthorityKeyId": null,
    "OCSPServer": null,
    "IssuingCertificateURL": null,
    "DNSNames": null,
    "EmailAddresses": null,
    "IPAddresses": null,
    "URIs": null,
    "PermittedDNSDomainsCritical": false,
    "PermittedDNSDomains": null,
    "ExcludedDNSDomains": null,
    "PermittedIPRanges": null,
    "ExcludedIPRanges": null,
    "PermittedEmailAddresses": null,
    "ExcludedEmailAddresses": null,
    "PermittedURIDomains": null,
    "ExcludedURIDomains": null,
    "CRLDistributionPoints": null,
    "PolicyIdentifiers": null
  },
  "Reason": 1,
  "Detail": "more detail"
}
```

```suggestion
			logger.Errorf(err, "Invalid certificate error: %s", resource.ToJSON(invalidCertificateError))
```

tpantelis (changes requested):

main.go
skitt: Right, that’s exactly what I was after, I was waiting to bring my test cluster back up today (with an expired certificate on the broker) to check. I’ve simplified the code since the error type is shown explicitly in the JSON.

main.go
```diff
@@ -123,6 +126,13 @@ func main() {
 
 	var err error
 
+	//nolint:reassign // We need to reassign ErrorHandlers to register our handler
+	utilruntime.ErrorHandlers = append(utilruntime.ErrorHandlers, func(err error) {
+		if errors.Is(err, x509.UnknownAuthorityError{}) || errors.Is(err, x509.CertificateInvalidError{}) {
+			logger.Errorf(err, "Certificate error: %s", resource.ToJSON(err))
```
skitt, line 132: I’m concerned this will result in huge amounts of logs, I’ll change this so that certificates are only logged once.

sridhargaddam (approved):

dfarrell07 (approved):

main.go
```diff
@@ -123,6 +128,26 @@ func main() {
...
+			if errors.As(err, &unknownAuthorityError) && unknownAuthorityError.Cert != lastBadCertificate {
+				logger.Errorf(err, "Certificate error: %s", resource.ToJSON(err))
+				lastBadCertificate = unknownAuthorityError.Cert
+			}
+			if errors.As(err, &certificateInvalidError) && certificateInvalidError.Cert != lastBadCertificate {
+				logger.Errorf(err, "Certificate error: %s", resource.ToJSON(err))
+				lastBadCertificate = certificateInvalidError.Cert
+			}
+			// The generic handler has already logged the error, no need to repeat if we don't want extra detail
+		}
```
tpantelis, line 140: I've found `errors.Is` to be problematic - it only works for error types that define an `Is` method which these 2 errors don't. So we need to use `errors.As`. Also we can simplify a bit by using an `atomic.Value` for `lastBadCertificate`:

```
var lastBadCertificate atomic.Value
...

//nolint:reassign // We need to reassign ErrorHandlers to register our handler
utilruntime.ErrorHandlers = append(utilruntime.ErrorHandlers, func(err error) {
    var unknownAuthorityError x509.UnknownAuthorityError
    if errors.As(err, &unknownAuthorityError) && unknownAuthorityError.Cert != lastBadCertificate.Load() {
        logger.Errorf(err, "Certificate error: %s", resource.ToJSON(err))
	lastBadCertificate.Store(unknownAuthorityError.Cert)
    }

    var certificateInvalidError x509.CertificateInvalidError
    if errors.As(err, &certificateInvalidError) && certificateInvalidError.Cert != lastBadCertificate.Load() {
        logger.Errorf(err, "Certificate error: %s", resource.ToJSON(err))
	lastBadCertificate.Store(certificateInvalidError.Cert)
    }
})

```

tpantelis (changes requested):

main.go
skitt: Ah yes, they don’t implement `Is()`. Thanks for the `atomic.Value` tip, but using separate `Load()` and `Store()` doesn’t provide the same semantics as a mutex; however `Swap()` does.

tpantelis: It doesn't - it depends if you want strong or eventual consistency. Either way would be fine here.

tpantelis (approved):

main.go
skitt: I meant that in this scenario, using `Swap()` ensures that there isn’t a race between comparing and storing the value, in the same way as protecting the whole function with a mutex.

Issue #2529, opened 2023-06-05
Restart of GW pod cause to short DP failure (libreswan cable driver)


yboaron:
- I deployed Submariner (Submariner 0.15, cable driver - libreswan) on two Kind clusters 
- run  pod with nettest image on cluster1 (p1, on non_GW node) and cluster2 (p2, on non_GW node)
- Retrieve the IP address of pod p2
- Kubectl exec into p1 pod on cluster1
-  run 'ping -i .1 <  p2 IP address >'
- Restart GW pod on cluster1 by running the next steps:
   kubectl exec -it  -n submariner-operator  <gw pod name> /bin/sh
   run 'ps -aux ' , the output should be something like [1]
   Kill the pluto process
- Checking the ping output, I can see that connectivity to remote cluster wad down for ~2 seconds , check [2]

[1]
$  ps -aux 
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.4  0.1 742112 49836 ?        Ssl  Jun04   3:31 submariner-gateway -v=2 -alsologtostderr
root          31  0.0  0.0 843036 17304 ?        Sl   Jun04   0:01 /usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --nofork
root       10090  0.0  0.0   4464  3604 pts/0    Ss   11:44   0:00 /bin/sh
root       10099  0.0  0.0   7012  1496 pts/0    R+   11:44   0:00 ps -aux

[2]
64 bytes from 10.131.1.16: icmp_seq=147 ttl=60 time=0.231 ms
64 bytes from 10.131.1.16: icmp_seq=148 ttl=60 time=0.229 ms
64 bytes from 10.131.1.16: icmp_seq=149 ttl=60 time=0.264 ms
64 bytes from 10.131.1.16: icmp_seq=150 ttl=60 time=0.221 ms
64 bytes from 10.131.1.16: icmp_seq=151 ttl=60 time=0.293 ms
64 bytes from 10.131.1.16: icmp_seq=171 ttl=60 time=0.238 ms
64 bytes from 10.131.1.16: icmp_seq=172 ttl=60 time=0.244 ms
64 bytes from 10.131.1.16: icmp_seq=173 ttl=60 time=0.327 ms
64 bytes from 10.131.1.16: icmp_seq=174 ttl=60 time=0.234 ms

maayanf24:
@yboaron - can you link the PR here?

[...]

PR #2543, opened 2023-06-11
Shortening datapath downtime for libreswan cable

yboaron:
Currently there is about 2 seconds of downtime in the datapath for libreswan cable when the gateway pod restarts.

When Pluto starts running [1] the IPsec datapath rules created by the previous GW pod are cleared, and traffic is only recovered after ConnectToEndpoint events are handled by the libreswan cable driver.

This PR reduces datapath downtime by:
A. Checking Pluto control socket status at faster rate
B. Starting Pluto right before the first ConnectToEndpoint event
   is handled by libreswan cable driver.

With this PR, the datapath downtime is reduced to ~0.5 seconds.

Related to https://github.com/submariner-io/submariner/issues/2529

[1]
https://github.com/submariner-io/submariner/blob/devel/pkg/cable/libreswan/libreswan.go#L158


pkg/cable/libreswan/libreswan.go
```diff
@@ -337,6 +334,15 @@ func whack(args ...string) error {
 // ConnectToEndpoint establishes a connection to the given endpoint and returns a string
 // representation of the IP address of the target endpoint.
 func (i *libreswan) ConnectToEndpoint(endpointInfo *natdiscovery.NATEndpointInfo) (string, error) {
+	if !i.plutoStarted {
+		// Ensure Pluto is started
+		if err := i.runPluto(); err != nil {
+			return "", errors.Wrap(err, "error starting Pluto")
```
sridhargaddam: This is going to change the behaviour. In the current code if `runPluto` returns error, we treat that as a FATAL error as part of Init() method. However, in the updated code its treated as an normal error while connecting to an endpoint. Is this intentional?

yboaron: Good point, I assume that if Pluto fails we should consider it as FATAL.

It looks like  we have the following options:
 
1.  update only Pluto control socket status check , in my Kind environment downtime was ~1.1 Sec with this fix.
2. Seperate Engine '**StartEngine**' API into two functions, to something like **InitEngine** and **StartEngineOperation** ,  **StartEngineOperation** should be called right before  Endpoints handler is started and we should treat failure here as fatal.  for libreswan Pluto should be started at **StartEngineOperation** phase.
3. hack the current code to treat Pluto error as fatal.


I think that since we expect that with **--dpdaction=hold** fix the number of gw pod restarts will decrease significantly, maybe we should go first with '1'

WDYT ?

sridhargaddam: This issue seems to be very specific to libreswan, so Option-2 may not be a good idea IMO. 

That means if we go with Option-1, the downtime will be ~1.1 seconds instead of 0.5 secs? 
Did you get a chance to verify this behaviour on an OCP deployment?

Ideally when downtime cannot be avoided due to some limitations in Libreswan driver, the lower the duration the better as applications can sometimes tolerate that downtime. Can you please try once on an OCP deployment and let us know your findings?

Also, I'd like to hear from @skitt if he has any feedback on this.

skitt: I agree we shouldn’t change the API for this. Looking at this purely from the perspective of the Libreswan cable driver, not being able to start Pluto is a fatal error, but the behaviour of the agent should be decided based on what the remediation should be.

It seems to me that the safest option is to stick to the current behaviour, and abort the gateway pod if Pluto can’t be started (moving this to a just-in-time start doesn’t change much).

If we want to reconsider things:

* if we abort the gateway pod, we’ll retry starting Pluto later on, after some delay (while the pod restarts)
* if we return an error, we’ll retry starting Pluto too, the next time the connections are synced (5 sec?)

So crashing the pod might actually result in a quicker retry, and possibly fix issues preventing Pluto from starting. Put another way, it seems to me that restarting later on from the same pod is unlikely to actually find different conditions, whereas restarting the entire pod could fix things.

In summary, I vote for keeping the current behaviour: move the code as suggested in the PR, but make failure to run Pluto fatal as it was before.

yboaron: Delaying this PR  till we get  feedback on https://bugzilla.redhat.com/show_bug.cgi?id=2216154

yboaron: So, following the discussion on ^^ BZ, we identified two issues related to Libreswan datapath downtime when GW pod was restarted.

1. Datapath is down from the moment Pluto is started until endpoints are reconfigured  
2. When Libreswan is configured to use UDP encapsulation, traffic is down as long as Pluto not running.


This PR handles '1' by running Pluto just before configuring the endpoints.

cc @skitt @sridhargaddam

skitt (approved):

nyechiel:
Nice work @yboaron! This was tested with OCP, correct?

We probably want to get this merged and backported to 0.15, 0.14 and 0.13. 0.13.6 was already released (so we can target 0.13.7), but we are planning to release both 0.15.2 and 0.14.6 this week so we should make sure they include this fix.

sridhargaddam (approved):

PR #2569, opened 2023-06-20
Add handlers to create GatewayRoute and NonGatewayRoute CRs

### LINKED ISSUE #203: Add handler to create GatewayRoute and Non GatewayRouteCRs

aswinsuryan:
These handlers will create the GatewayRoute and NonGatewayRoute CRs in response to endpoint create event

It will respond to gateway transition event, by creating the CRs in the new gateway node and deleting them in the old nodes.

Fixes: https://github.com/submariner-io/enhancements/issues/203


pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+	// TODO enable when we switch to new implementation
+	// return []string{cni.OVNKubernetes}
+	return []string{}
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) LocalEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	gatewayRouteHandler.mutex.Lock()
+	defer gatewayRouteHandler.mutex.Unlock()
+
+	nextHopIP, err := gatewayRouteHandler.getNextHopOnK8sMgmtIntf(endpoint)
```
yboaron: Do you use the endpoint parameter in getNextHopOnK8sMgmtIntf function ?

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if err != nil || nextHopIP == nil {
```
yboaron: IIUC, if we fail to retrieve ovn interface IP  address  routes to remote cluster are broken.
Should we retry in case of failure? 
Is there any point in continuing if we fail here or we should fatal exit here?

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+	gatewayRouteHandler.nextHopIP = nextHopIP.String()
+
+	return nil
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	gatewayRouteHandler.mutex.Lock()
+	defer gatewayRouteHandler.mutex.Unlock()
+
+	if gatewayRouteHandler.isGateway {
```
yboaron: Maybe you should consider adding [1] as first line in this function  to avoid unneeded mutex.lock and unlock  in non GW nodes 

[1]
`if !gatewayRouteHandler.isGateway{
return nil 
}     `

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			_, err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Create(context.TODO(),
+				gatewayRouteHandler.getGatewayRoute(endpoint), metav1.CreateOptions{})
+			if !apierrors.IsAlreadyExists(err) {
+				return errors.Wrapf(err, "error processing the remote endpoint creation for %q", endpoint.Name)
+			}
+			return nil
+		})
+
+		return errors.Wrapf(retryErr, "processing the remote endpoint event failed even after retry")
```
yboaron: Shouldn't we return error only if retryErr != nil ? 

Also log that GWRoute CR for remoteendpoint successfully created could be useful

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return nil
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) RemoteEndpointRemoved(endpoint *submarinerv1.Endpoint) error {
+	gatewayRouteHandler.mutex.Lock()
```
yboaron: Can you elaborate why mutex is needed here? 
we only create K8S CR

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+		if err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Delete(context.TODO(),
+			endpoint.Name, metav1.DeleteOptions{}); err != nil {
+			return errors.Wrapf(err, "error deleting gatewayRoute %q", endpoint.Name)
+		}
+	}
+
+	return nil
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) getNextHopOnK8sMgmtIntf(*submarinerv1.Endpoint) (*net.IP, error) {
```
yboaron: Can we move this function to some util/other ovn file?
It just takes local cluster cidrs, I can see that we already have similar function in https://github.com/submariner-io/submariner/blob/devel/pkg/routeagent_driver/handlers/ovn/host_networking.go#L113C21-L113C44

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (gatewayRouteHandler *GatewayRouteHandler) TransitionToGateway() error {
+	gatewayRouteHandler.mutex.Lock()
+	defer gatewayRouteHandler.mutex.Unlock()
+
+	gatewayRouteHandler.isGateway = true
+
+	for _, endpoint := range gatewayRouteHandler.remoteEndpoints {
+		err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).DeleteCollection(context.TODO(),
+			metav1.DeleteOptions{}, metav1.ListOptions{
```
yboaron: Can't we just update the nextHop fileld in existing CR instead of delete and create ?

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+			APIVersion: "submariner.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: endpoint.Name,
+		},
+		RoutePolicySpec: submarinerv1.RoutePolicySpec{
+			RemoteCIDRs: endpoint.Spec.Subnets,
+			NextHops:    []string{gatewayRouteHandler.nextHopIP},
+		},
+	}
```
yboaron: Should we handle also 
LocalEndpointUpdated and RemoteEndpointUpdated events here?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil
+	}
+
+	nonGatewayRouteHandler.transitSwitchIP = transitSwitchIP
+
+	return nil
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) GetName() string {
+	return "submariner-gw-route-handler"
```
yboaron: submariner-nongw-route-handler

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) Init() error {
+	logger.Infof("Starting NonGatewayRouteHandler")
+
+	nodeName, ok := os.LookupEnv("NODE_NAME")
+	if !ok {
+		return errors.New("error getting the Node name")
+	}
+
+	node, err := nonGatewayRouteHandler.k8sClientSet.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
yboaron: Maybe you should use https://github.com/submariner-io/submariner/blob/devel/pkg/node/node.go#L45 (it support also retry)  ?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,189 @@
...
+	nonGatewayRouteHandler.mutex.Lock()
+	defer nonGatewayRouteHandler.mutex.Unlock()
+
+	nonGatewayRouteHandler.localEndpoint = endpoint
+
+	return nil
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	nonGatewayRouteHandler.mutex.Lock()
```
yboaron: Maybe you should check if !nonGatewayRouteHandler.isGateway and return nil as first line in this function?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,189 @@
...
+
+	annotations := node.GetAnnotations()
+
+	transitSwitchIP, ok := annotations["k8s.ovn.org/node-transit-switch-port-ifaddr"]
+	if !ok {
+		logger.Infof("No transit switch IP configured")
+		return nil
+	}
+
+	nonGatewayRouteHandler.transitSwitchIP = transitSwitchIP
```
yboaron: What if transitSwitchIP value changed?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) GetNetworkPlugins() []string {
+	// TODO enable when we switch to new implementation
+	// return []string{cni.OVNKubernetes}
+	return []string{}
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) LocalEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	nonGatewayRouteHandler.mutex.Lock()
```
yboaron: Can you explain why we need the mutex ?

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
tpantelis: The `isGateway` flag also needs to be protected by the mutex. Unless we use an atomic bool.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+
+	return nil
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	gatewayRouteHandler.mutex.Lock()
+	defer gatewayRouteHandler.mutex.Unlock()
+
+	if gatewayRouteHandler.isGateway {
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
tpantelis: Since we're just creating here, we won't get a conflict.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+			if !apierrors.IsNotFound(err) {
+				return errors.Wrapf(err, "error deleting gatewayRoute %q", endpoint.Name)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) getGatewayRoute(endpoint *submarinerv1.Endpoint) *submarinerv1.GatewayRoute {
```
tpantelis:
```suggestion
func (gatewayRouteHandler *GatewayRouteHandler) newGatewayRoute(endpoint *submarinerv1.Endpoint) *submarinerv1.GatewayRoute {
```

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,206 @@
...
+	return &submarinerv1.GatewayRoute{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "GatewayRoute",
+			APIVersion: "submariner.io/v1",
+		},
```
tpantelis: Why are you adding the type data here? This shouldn't be necessary.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
tpantelis: To protect fields like `localEndpoint` that could be accessed by multiple threads.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,189 @@
...
+		if err := nonGatewayRouteHandler.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace).Delete(context.TODO(),
+			endpoint.Name, metav1.DeleteOptions{}); err != nil {
+			return errors.Wrapf(err, "error deleting nonGatewayRoute %q", endpoint.Name)
+		}
+	}
+
+	return nil
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) TransitionToNonGateway() error {
```
yboaron: What if prev active GW pod is not running for some reason (node down)? 
we end up with two CRs that can point to two different next hop IPs , shouldn't it affect the traffic?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,189 @@
...
+	nonGatewayRouteHandler.mutex.Lock()
+	defer nonGatewayRouteHandler.mutex.Unlock()
+
+	nonGatewayRouteHandler.isGateway = false
+	if nonGatewayRouteHandler.transitSwitchIP != "" {
+		for _, endpoint := range nonGatewayRouteHandler.remoteEndpoints {
+			if err := nonGatewayRouteHandler.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace).Delete(context.TODO(),
+				endpoint.Name, metav1.DeleteOptions{}); err != nil {
+				return errors.Wrapf(err, "error deleting nonGatewayRoute %q", endpoint.Name)
```
yboaron: Don't we want to continue to next remote endpoint instead of returning here ?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
yboaron: I thought that handler framework serialize the events, (call event x+1 only after event x handling was completed),
that isn't the case , right ?

tpantelis: Looking at the handler framework code, it looks like that **should** be the case as long as the fields are only accessed via handler framework callbacks. In fact the framework maintains an `isGateway` flag internally and could make that available to handlers - I can look into that later.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Removed it.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: it will be retried if we return an error by the watcher.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: errors.Wrapf takes care of that , it returns nil if there is no error .
 
We can check the successful creation with a kubectl command, shouldn't that be sufficient.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: we want to lock the isGatewayFlag too.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Changed to update.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: I think we generally don't update endpoints in submariner now.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: There will be a retry on failure wouldn't that be better?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: changed the logic to do an update.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: @tpantelis Without the type date , I am getting error.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: it is a topology change right ? we don't support it right now I think, like if ovn-k8s-mp0 interface IP change.

dfarrell07:
@aswinsuryan Needs a manual rebase

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return nil
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	gatewayRouteHandler.mutex.Lock()
+	defer gatewayRouteHandler.mutex.Unlock()
+
+	if gatewayRouteHandler.isGateway {
+		_, err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Create(context.TODO(),
```
yboaron: I have Q about events order, is it guaranteed that RemoteEndpointCreated event is received always after LocalEndpointCreated was received?

If RemoteEndpointCreated is received before LocalEndpointCreated GatewayRoute.nextHopIP will be nil/empty  , right ?

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil || nextHopIP == nil {
+		return errors.Wrapf(err, "error getting the ovn kubernetes management interface IP")
+	}
+
+	gatewayRouteHandler.nextHopIP = nextHopIP.String()
+
+	return nil
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
```
yboaron: Don't you need to store endpoint in gatewayRouteHandler.remoteEndpoints  in this function?

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
sridhargaddam: We should not assume that RemoteEndpoint notification will be received after localEndpoint. Specifically it can happen when the pod gets restarted and I remember seeing it multiple times.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
yboaron: If we plan to support it in phase-2, maybe it is worth adding a TODO note?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,181 @@
...
+	logger.Infof("Starting NonGatewayRouteHandler")
+
+	node, err := nodeutil.GetLocalNode(nonGatewayRouteHandler.k8sClientSet)
+	if err != nil {
+		return errors.Wrap(err, "error getting the g/w node")
+	}
+
+	annotations := node.GetAnnotations()
+
+	transitSwitchIP, ok := annotations["k8s.ovn.org/node-transit-switch-port-ifaddr"]
```
yboaron: Define "k8s.ovn.org/node-transit-switch-port-ifaddr" as const ?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,181 @@
...
+	return "submariner-nongw-route-handler"
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) GetNetworkPlugins() []string {
+	// TODO enable when we switch to new implementation
+	// return []string{cni.OVNKubernetes}
+	return []string{}
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) LocalEndpointCreated(endpoint *submarinerv1.Endpoint) error {
```
yboaron: Do you use localEndpoint in this handler ?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,181 @@
...
+
+	nonGatewayRouteHandler.localEndpoint = endpoint
+
+	return nil
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	nonGatewayRouteHandler.mutex.Lock()
+	defer nonGatewayRouteHandler.mutex.Unlock()
+
```
yboaron: Where do you add  the remote endpoint to remoteEndpoints ?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,181 @@
...
+	transitSwitchIP, ok := annotations["k8s.ovn.org/node-transit-switch-port-ifaddr"]
+	if !ok {
+		logger.Infof("No transit switch IP configured")
+		return nil
+	}
+
+	nonGatewayRouteHandler.transitSwitchIP = transitSwitchIP
+
+	return nil
+}
```
yboaron: IIUC, in case transitSwitchIP is empty this controller shouldn't do anything , right ?

Can we handle that in NewNonGatewayRouteHandler ?  I mean if  transitSwitchIP is empty then return nil as handler ?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,181 @@
...
+	})
+
+	return errors.Wrapf(retryErr, "processing the remote endpoint event failed even after retry")
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) RemoteEndpointRemoved(endpoint *submarinerv1.Endpoint) error {
+	nonGatewayRouteHandler.mutex.Lock()
+	defer nonGatewayRouteHandler.mutex.Unlock()
+
+	delete(nonGatewayRouteHandler.remoteEndpoints, endpoint.Name)
```
yboaron: I have a generic Q about EP storing in nonGatewayRouteHandler.remoteEndpoints or other handlers, 
we store a pointer to EP in this array, I can see that the pointer is sent from controller/watcher to all handlers, how do we make sure that the EP object isn't deleted/freed while we store a pointer to this object?

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
yboaron: Then maybe it makes sense to return error in RemoteEndpointCreated if gatewayRouteHandler.nextHopIP is nil (and watcher retry mechanism should handle this race)

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
tpantelis: strange... what error and where?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Added TODO

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: it was solved when we added the api to scheme.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: With the updated logic , we are not relying on localendpoint and init will throw error if nextHopIp is missing.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Removed.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Oh missed it , added.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: When I try that it, it adds increases the complexity in main. I think that affects readability.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Added.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Shouldn't that be taken care by golang memory management.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,148 @@
...
+
+	gatewayRouteHandler.isGateway = true
+
+	for _, endpoint := range gatewayRouteHandler.remoteEndpoints {
+		if _, err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Update(context.TODO(),
+			gatewayRouteHandler.newGatewayRoute(endpoint), metav1.UpdateOptions{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				_, err = gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Create(context.TODO(),
+					gatewayRouteHandler.newGatewayRoute(endpoint), metav1.CreateOptions{})
+				return errors.Wrapf(err, "error deleting gatewayRoute %q", endpoint.Name)
```
yboaron: error updating ...

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,177 @@
...
+	nonGatewayRouteHandler.mutex.Lock()
+	defer nonGatewayRouteHandler.mutex.Unlock()
+
+	if !nonGatewayRouteHandler.isGateway || nonGatewayRouteHandler.transitSwitchIP == "" {
+		return nil
+	}
+
+	nonGatewayRouteHandler.remoteEndpoints[endpoint.Name] = endpoint
+
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
yboaron: Q about RemoteEndpointCreated event handling, don't we want to run retry on conflict also for gwroute?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	nonGatewayRouteHandler.mutex.Lock()
+	defer nonGatewayRouteHandler.mutex.Unlock()
+
+	if !nonGatewayRouteHandler.isGateway || nonGatewayRouteHandler.transitSwitchIP == "" {
+		return nil
+	}
+
+	nonGatewayRouteHandler.remoteEndpoints[endpoint.Name] = endpoint
```
yboaron: Don't you want to store ep details also for nonactive GW node?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,177 @@
...
+	nonGatewayRouteHandler.isGateway = true
+	if nonGatewayRouteHandler.transitSwitchIP != "" {
+		for _, endpoint := range nonGatewayRouteHandler.remoteEndpoints {
+			if _, err := nonGatewayRouteHandler.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace).Update(context.TODO(),
+				nonGatewayRouteHandler.newNonGatewayRoute(endpoint), metav1.UpdateOptions{}); err != nil {
+				if !apierrors.IsNotFound(err) {
+					_, err = nonGatewayRouteHandler.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace).Create(context.TODO(),
+						nonGatewayRouteHandler.newNonGatewayRoute(endpoint), metav1.CreateOptions{})
+
+					return errors.Wrapf(err, "error deleting nonGatewayRoute %q", endpoint.Name)
```
yboaron: error updating ..

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
skitt: Yes, an object won’t be GCed if there’s a live pointer to it.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
yboaron: Ack

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
yboaron: Nice, 10x

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: it is not required here as create will not cause a conflict , removed.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: yes we need to, changed.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,148 @@
...
+func NewGatewayRouteHandler(env *environment.Specification, smClientSet submarinerClientset.Interface) *GatewayRouteHandler {
+	// We'll panic if env is nil, this is intentional
+	return &GatewayRouteHandler{
+		config:          env,
+		smClient:        smClientSet,
+		remoteEndpoints: map[string]*submarinerv1.Endpoint{},
+	}
+}
+
+func (gatewayRouteHandler *GatewayRouteHandler) Init() error {
```
tpantelis: [Receiver names](https://github.com/golang/go/wiki/CodeReviewComments#receiver-names) should be short (I thought we had a linter that flags this...).

```suggestion
func (h *GatewayRouteHandler) Init() error {
```

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,148 @@
...
+	defer gatewayRouteHandler.mutex.Unlock()
+
+	gatewayRouteHandler.remoteEndpoints[endpoint.Name] = endpoint
+
+	if gatewayRouteHandler.isGateway {
+		_, err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Create(context.TODO(),
+			gatewayRouteHandler.newGatewayRoute(endpoint), metav1.CreateOptions{})
+		if !apierrors.IsAlreadyExists(err) {
+			return errors.Wrapf(err, "error processing the remote endpoint creation for %q", endpoint.Name)
+		}
```
tpantelis: Shouldn't we ensure err is non-nil?

```suggestion
		if err != nil && !apierrors.IsAlreadyExists(err) {
			return errors.Wrapf(err, "error processing the remote endpoint creation for %q", endpoint.Name)
		}
```

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,148 @@
...
+	gatewayRouteHandler.mutex.Lock()
+	defer gatewayRouteHandler.mutex.Unlock()
+
+	delete(gatewayRouteHandler.remoteEndpoints, endpoint.Name)
+
+	if gatewayRouteHandler.isGateway {
+		if err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Delete(context.TODO(),
+			endpoint.Name, metav1.DeleteOptions{}); err != nil {
+			return errors.Wrapf(err, "error deleting gatewayRoute %q", endpoint.Name)
+		}
```
tpantelis: Perhaps we should also check for `IsNotFound` just in case.

```suggestion
		err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Delete(context.TODO(),
			endpoint.Name, metav1.DeleteOptions{})
		if err != nil && !apierrors.IsNotFound(err) {
			return errors.Wrapf(err, "error deleting gatewayRoute %q", endpoint.Name)
		}
```

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,148 @@
...
+
+	for _, endpoint := range gatewayRouteHandler.remoteEndpoints {
+		if _, err := gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Update(context.TODO(),
+			gatewayRouteHandler.newGatewayRoute(endpoint), metav1.UpdateOptions{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				_, err = gatewayRouteHandler.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Create(context.TODO(),
+					gatewayRouteHandler.newGatewayRoute(endpoint), metav1.CreateOptions{})
+				return errors.Wrapf(err, "error updating gatewayRoute %q", endpoint.Name)
+			}
+		}
```
tpantelis: It seems you want admiral's `util.CreateOrUpdate` here.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+func NewNonGatewayRouteHandler(smClientSet submarinerClientset.Interface, k8sClientset *clientset.Clientset) *NonGatewayRouteHandler {
+	// We'll panic if env is nil, this is intentional
+	return &NonGatewayRouteHandler{
+		smClient:        smClientSet,
+		remoteEndpoints: map[string]*submarinerv1.Endpoint{},
+		k8sClientSet:    k8sClientset,
+	}
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) Init() error {
```
tpantelis: Same as above re: receiver names.

```suggestion
func (h *NonGatewayRouteHandler) Init() error {
```

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+		logger.Infof("No transit switch IP configured")
+		return nil
+	}
+
+	nonGatewayRouteHandler.transitSwitchIP, err = jsonToIP(transitSwitchIP)
+	if !ok {
+		return errors.Wrapf(err, "error parsing the transit switch IP")
+	}
+
+	return nil
```
tpantelis: This should check `err != nil` but can be simplified to:

```suggestion
	return errors.Wrapf(err, "error parsing the transit switch IP")
```

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+		return errors.Wrap(err, "error getting the g/w node")
+	}
+
+	annotations := node.GetAnnotations()
+
+	// TODO transitSwitchIP changes support needs to be added.
+	transitSwitchIP, ok := annotations[ovnTransitSwitchIPAnnotation]
+	if !ok {
+		logger.Infof("No transit switch IP configured")
+		return nil
```
tpantelis: Is this an error condition, ie return an error?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+	nonGatewayRouteHandler.remoteEndpoints[endpoint.Name] = endpoint
+
+	if !nonGatewayRouteHandler.isGateway || nonGatewayRouteHandler.transitSwitchIP == "" {
+		return nil
+	}
+
+	_, err := nonGatewayRouteHandler.smClient.SubmarinerV1().
+		NonGatewayRoutes(endpoint.Namespace).Create(context.TODO(),
+		nonGatewayRouteHandler.newNonGatewayRoute(endpoint), metav1.CreateOptions{})
+	if !apierrors.IsAlreadyExists(err) {
```
tpantelis:
```suggestion
	if err != nil && !apierrors.IsAlreadyExists(err) {
```

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+	delete(nonGatewayRouteHandler.remoteEndpoints, endpoint.Name)
+
+	if !nonGatewayRouteHandler.isGateway || nonGatewayRouteHandler.transitSwitchIP == "" {
+		return nil
+	}
+
+	if err := nonGatewayRouteHandler.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace).Delete(context.TODO(),
+		endpoint.Name, metav1.DeleteOptions{}); err != nil {
+		return errors.Wrapf(err, "error deleting nonGatewayRoute %q", endpoint.Name)
+	}
```
tpantelis: Same as above re: checking `IsNotFound`.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return nil
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) TransitionToGateway() error {
+	nonGatewayRouteHandler.mutex.Lock()
+	defer nonGatewayRouteHandler.mutex.Unlock()
+
+	nonGatewayRouteHandler.isGateway = true
+	if nonGatewayRouteHandler.transitSwitchIP != "" {
```
tpantelis: Shortcut here to reduce indentation.

```suggestion
	if nonGatewayRouteHandler.transitSwitchIP == "" {
	    return nil
	}
```

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+		for _, endpoint := range nonGatewayRouteHandler.remoteEndpoints {
+			if _, err := nonGatewayRouteHandler.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace).Update(context.TODO(),
+				nonGatewayRouteHandler.newNonGatewayRoute(endpoint), metav1.UpdateOptions{}); err != nil {
+				if !apierrors.IsNotFound(err) {
+					_, err = nonGatewayRouteHandler.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace).Create(context.TODO(),
+						nonGatewayRouteHandler.newNonGatewayRoute(endpoint), metav1.CreateOptions{})
+
+					return errors.Wrapf(err, "error updating nonGatewayRoute %q", endpoint.Name)
+				}
+			}
```
tpantelis: Same as above re: `util.CreateOrUpdate`.

pkg/routeagent_driver/main.go
```diff
@@ -99,6 +103,7 @@ func main() {
 		eventlogger.NewHandler(),
 		kubeproxy.NewSyncHandler(env.ClusterCidr, env.ServiceCidr),
 		ovn.NewHandler(&env, smClientset),
+		ovn.NewGatewayRouteHandler(&env, smClientset),
```
tpantelis, line 107: Shouldn't we also add `ovn.NewNonGatewayRouteHandler`?

pkg/routeagent_driver/handlers/ovn/utils.go
```diff
@@ -0,0 +1,85 @@
...
+	"encoding/json"
+	"fmt"
+	"net"
+
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+func getNextHopOnK8sMgmtIntf() (string, error) {
+	link, err := netlink.LinkByName(OVNK8sMgmntIntfName)
```
tpantelis: Use the netlink wrapper (here and below) to facilitate unit testing.

```suggestion
        netLink := netlink.New()
	link, err := netLink.LinkByName(OVNK8sMgmntIntfName)
```

tpantelis (approved):

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
sridhargaddam: I think we use the receiver name consistency (i.e., if its h, it has to be h throughout the file)  linter, but not the length of the name itself. But I agree, a shorter name is a convention we are following in our code.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
+	"github.com/submariner-io/submariner/pkg/event"
+	nodeutil "github.com/submariner-io/submariner/pkg/node"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+const (
+	ovnTransitSwitchIPAnnotation = "k8s.ovn.org/node-transit-switch-port-ifaddr"
```
sridhargaddam: Please move this to the pkg/routeagent_driver/handlers/ovn/constants.go file

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+	// We'll panic if env is nil, this is intentional
+	return &NonGatewayRouteHandler{
+		smClient:        smClientSet,
+		remoteEndpoints: map[string]*submarinerv1.Endpoint{},
+		k8sClientSet:    k8sClientset,
+	}
+}
+
+func (nonGatewayRouteHandler *NonGatewayRouteHandler) Init() error {
+	logger.Infof("Starting NonGatewayRouteHandler")
```
sridhargaddam: We can use "logger.Info" here and at other places too.

```suggestion
	logger.Info("Starting NonGatewayRouteHandler")
```

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
sridhargaddam: Does this `ovnTransitSwitchIPAnnotation` IP point to the ovn-k8s-mp0 iface IP on the transit GW?

Also, assuming that the annotation was missing and we simply return, how does the NonGatewayRouteHandler handler recover from the situation?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error getting the g/w node")
+	}
+
+	annotations := node.GetAnnotations()
+
+	// TODO transitSwitchIP changes support needs to be added.
+	transitSwitchIP, ok := annotations[ovnTransitSwitchIPAnnotation]
+	if !ok {
+		logger.Infof("No transit switch IP configured")
```
sridhargaddam: Is this annotation seen even with non OVN IC deployments too?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
sridhargaddam: +1

pkg/routeagent_driver/handlers/ovn/utils.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to retrieve link by name")
+	}
+
+	addrs, err := netlink.AddrList(link, 0)
```
sridhargaddam: We can optimize it by querying only IPv4 addresses as follows and remove the check in Line43

addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)

pkg/routeagent_driver/handlers/ovn/utils.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return ip != nil && ip.To4() != nil, nil
+}
+
+func jsonToIP(jsonData string) (string, error) {
+	var data map[string]string
+
+	err := json.Unmarshal([]byte(jsonData), &data)
+	if err != nil {
+		return "", errors.Wrapf(err, "error marshalling the json ip")
```
sridhargaddam:
```suggestion
		return "", errors.Wrapf(err, "error unmarshalling the json ip")
```

pkg/routeagent_driver/main.go
sridhargaddam: Currently the new Handlers are not registered (as they return an empty name in GetNetworkPlugins), but to be consistent, we should add it.

tpantelis (changes requested):
I meant to request changes before.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/main.go
aswinsuryan: added it , sorry it was missed while I created a selected a subset of changes for this commit.

pkg/routeagent_driver/handlers/ovn/utils.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: This will not be seen in non-IC deployments, and non-gateway routes will not be created in such cases.

pkg/routeagent_driver/handlers/ovn/utils.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/utils.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: when I try to pass  `h.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace)` util.CreateOrUpdate it is giving me error updateStatus is missing.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Modified the code for the comment given for next line. If this annothation is missing we will not start this controller as it will be a non-ic deployment.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: when I try to pass h.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace) util.CreateOrUpdate it is giving me error updateStatus is missing.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
tpantelis: Can you provide the code you used and the exact message received?

aswinsuryan: I tried this one

		ngwr := h.newNonGatewayRoute(endpoint)
		result, err := util.CreateOrUpdate(context.TODO(), h.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace), ngwr,
			func(existing runtime.Object) (runtime.Object, error) {
				existingNgwr := ngwr
				return existingNgwr, nil
			})

aswinsuryan: and the error was `Cannot use 'h.smClient.SubmarinerV1().NonGatewayRoutes(endpoint.Namespace)' (type NonGatewayRouteInterface) as the type resource.Interface Type does not implement 'resource.Interface' as some methods are missing: UpdateStatus(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error)`

tpantelis: ok that's a compile error. You need to pass a `resource.Interface` wrapper. You can either create one for the generated client interfaces, as is done for [Gateways here](https://github.com/submariner-io/submariner/blob/d48ecae9a34ee3227d0c449357857a43431b2bf4/pkg/cableengine/syncer/syncer.go#L107), or just use the general K8s controller-runtime API and [associated resource.Interface](https://github.com/submariner-io/admiral/blob/79a20667691f873c34a509935699aa5114bd59d7/pkg/resource/kubernetes.go#L241).

aswinsuryan: Thanks, changed it .

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: done

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,179 @@
...
+	h.isGateway = true
+
+	for _, endpoint := range h.remoteEndpoints {
+		gwr := h.newGatewayRoute(endpoint)
+
+		result, err := util.CreateOrUpdate(context.TODO(), h.gatewayResourceInterface(endpoint.Namespace), gwr,
+			func(existing runtime.Object) (runtime.Object, error) {
+				existingGwr := gwr
+				return existingGwr, nil
+			})
```
tpantelis: The assignment to `existingGwr` is unnecessary but you can use `util.Replace` instead.

```suggestion
		result, err := util.CreateOrUpdate(context.TODO(), h.gatewayResourceInterface(endpoint.Namespace), gwr, util.Replace(gwr))
```

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
```diff
@@ -0,0 +1,179 @@
...
+			return errors.Wrapf(err, "error creating/updating GatewayRoute")
+		}
+
+		if result == util.OperationResultCreated {
+			logger.V(log.TRACE).Infof("GatewayRoute does not exist - created: %+v", gwr)
+		} else if result == util.OperationResultUpdated {
+			logger.V(log.TRACE).Infof("GatewayRoute already exists - updated %+v", gwr)
+		} else {
+			logger.V(log.TRACE).Info("GatewayRoute already exists but doesn't need updating")
+		}
```
tpantelis: You can simplify:

```suggestion
                 logger.V(log.TRACE).Infof("GatewayRoute %s: %#v", result, gwr)
```

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,199 @@
...
+			return errors.Wrapf(err, "error creating/updating NonGatewayRoute")
+		}
+
+		if result == util.OperationResultCreated {
+			logger.V(log.TRACE).Infof("NonGatewayRoute does not exist - created: %+v", ngwr)
+		} else if result == util.OperationResultUpdated {
+			logger.V(log.TRACE).Infof("NonGatewayRoute already exists - updated %+v", ngwr)
+		} else {
+			logger.V(log.TRACE).Info("NonGatewayRoute already exists but doesn't need updating")
+		}
```
tpantelis: Same comments apply as in the other handler.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done

tpantelis (approved):

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,191 @@
...
+	mutex           sync.Mutex
+	smClient        submarinerClientset.Interface
+	remoteEndpoints map[string]*submarinerv1.Endpoint
+	k8sClientSet    clientset.Interface
+	isGateway       bool
+	transitSwitchIP string
+}
+
+func NewNonGatewayRouteHandler(smClientSet submarinerClientset.Interface, k8sClientset *clientset.Clientset) *NonGatewayRouteHandler {
+	// We'll panic if env is nil, this is intentional
```
yboaron: env?

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, endpoint := range h.remoteEndpoints {
+		ngwr := h.newNonGatewayRoute(endpoint)
+
+		result, err := util.CreateOrUpdate(context.TODO(), h.nonGatewayResourceInterface(endpoint.Namespace),
+			ngwr, util.Replace(ngwr))
+		if err != nil {
+			return errors.Wrapf(err, "error creating/updating NonGatewayRoute")
+		}
+
+		logger.V(log.TRACE).Infof("GatewayRoute %s: %#v", result, ngwr)
```
yboaron: NonGatewayRoute

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: It was a comment for initial implementation, removed.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
tpantelis, line 1: Sorry I didn't notice this before. Our file naming convention is to use underscores and not camel case so, also "submariner" is redundant with the project name so I propose _gateway_route_handler.go_, which also matches the struct name.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
tpantelis, line 1: Similarly _non_gateway_route_handler.go_

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/submarinerGWRouteHandler.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/submarinerNonGWRouteHandler.go
aswinsuryan: Done.

tpantelis (approved):

yboaron (approved):

tpantelis:
I'll give @sridhargaddam a chance to review.

sridhargaddam (approved):

PR #2576, opened 2023-06-28
Automate calico ippools creation by routeagent

yboaron:
To make calico compatible with Submariner, it needs to create calico IPPools for the subnets associated with the Pod and Service CIDRs of the remote clusters.

This PR adds a new handler (CalicoIPPool) for the routeagent, the CalicoIPPool  handler will be responsible for creating calico IPPools for remote endpoints, running Calico API server is a 
prerequisite for automating Calico IPPools creation.

Depends-On : https://github.com/submariner-io/submariner-operator/pull/2713

### DIFF
--- a/pkg/routeagent_driver/handlers/calico/ippool_handler.go
+++ b/pkg/routeagent_driver/handlers/calico/ippool_handler.go
@@ -1,0 +1,211 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package calico
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync/atomic"
+
+	"github.com/pkg/errors"
+	calicoapi "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+	calicocs "github.com/projectcalico/api/pkg/client/clientset_generated/clientset"
+	"github.com/submariner-io/admiral/pkg/log"
+	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cidr"
+	"github.com/submariner-io/submariner/pkg/cni"
+	"github.com/submariner-io/submariner/pkg/event"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	errorutils "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/rest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	submarinerIPPool = "submariner.io/ippool"
+)
+
+type calicoIPPoolHandler struct {
+	event.HandlerBase
+	restConfig      *rest.Config
+	client          *calicocs.Clientset
+	remoteEndpoints map[string]*submV1.Endpoint
+	isGateway       atomic.Bool
+}
+
+var logger = log.Logger{Logger: logf.Log.WithName("CalicoIPPool")}
+
+func NewCalicoIPPoolHandler(restConfig *rest.Config) event.Handler {
+	return &calicoIPPoolHandler{
+		restConfig:      restConfig,
+		remoteEndpoints: map[string]*submV1.Endpoint{},
+	}
+}
+
+func (h *calicoIPPoolHandler) GetNetworkPlugins() []string {
+	return []string{cni.Calico}
+}
+
+func (h *calicoIPPoolHandler) GetName() string {
+	return "Calico IPPool handler"
+}
+
+func (h *calicoIPPoolHandler) Init() error {
+	var err error
+
+	h.client, err = calicocs.NewForConfig(h.restConfig)
+
+	return errors.Wrap(err, "error initializing Calico clientset")
+}
+
+func (h *calicoIPPoolHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	h.remoteEndpoints[endpoint.Name] = endpoint
+	if !h.isGateway.Load() {
+		logger.V(log.TRACE).Info("Ignore RemoteEndpointCreated event (node isn't Gateway)")
+		return nil
+	}
+
+	err := h.createIPPool(endpoint)
+
+	return errors.Wrap(err, "failed to handle RemoteEndpointCreated event")
+}
+
+func (h *calicoIPPoolHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	delete(h.remoteEndpoints, endpoint.Name)
+
+	if !h.isGateway.Load() {
+		logger.V(log.TRACE).Info("Ignore RemoteEndpointRemoved event (node isn't Gateway)")
+		return nil
+	}
+
+	err := h.deleteIPPool(endpoint)
+
+	return errors.Wrap(err, "failed to handle RemoteEndpointRemoved event")
+}
+
+func (h *calicoIPPoolHandler) TransitionToNonGateway() error {
+	logger.Info("TransitionToNonGateway")
+
+	h.isGateway.Swap(false)
+
+	return nil
+}
+
+func (h *calicoIPPoolHandler) TransitionToGateway() error {
+	var retErrors []error
+	logger.Info("TransitionToGateway")
+
+	h.isGateway.Swap(true)
+
+	for _, endpoint := range h.remoteEndpoints {
+		err := h.createIPPool(endpoint)
+		if err != nil {
+			logger.Warningf("Failed to create ippool %s", endpoint.GetName())
+			retErrors = append(retErrors,
+				errors.Wrapf(err, "error creating Calico IPPool for endpoint %q ", endpoint.GetName()))
+		}
+	}
+
+	return errorutils.NewAggregate(retErrors)
+}
+
+func (h *calicoIPPoolHandler) Stop(uninstall bool) error {
+	if !uninstall || !h.isGateway.Load() {
+		return nil
+	}
+
+	logger.Info("Uninstalling Calico IPPools used for Submariner")
+
+	labelSelector := labels.SelectorFromSet(map[string]string{submarinerIPPool: "true"}).String()
+	err := h.client.ProjectcalicoV3().IPPools().DeleteCollection(context.TODO(), metav1.DeleteOptions{},
+		metav1.ListOptions{LabelSelector: labelSelector})
+
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "Failed to delete Calico IPPools using labelSelector %q", labelSelector)
+	}
+
+	logger.Infof("Successfully delete Calico IPPools using labelSelector %q", labelSelector)
+
+	return nil
+}
+
+func (h *calicoIPPoolHandler) createIPPool(endpoint *submV1.Endpoint) error {
+	subnets := cidr.ExtractIPv4Subnets(endpoint.Spec.Subnets)
+	var retErrors []error
+
+	for _, subnet := range subnets {
+		iPPoolObj := &calicoapi.IPPool{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   getEndpointSubnetIPPoolName(endpoint, subnet),
+				Labels: map[string]string{submarinerIPPool: "true"},
+			},
+			Spec: calicoapi.IPPoolSpec{
+				CIDR:        subnet,
+				NATOutgoing: false,
+				Disabled:    true,
+			},
+		}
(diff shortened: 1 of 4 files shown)


pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	submarinerIPPool = "submariner.io/ippool"
+)
+
+type createUpdateFunc func(ipPoolObj *calicoapi.IPPool) error
+
+type calicoIppoolHandler struct {
```
tpantelis: For consistency

```suggestion
type calicoIPPoolHandler struct {
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+	event.HandlerBase
+	restConfig      *rest.Config
+	cs              *calicocs.Clientset
```
tpantelis:
```suggestion
	client              *calicocs.Clientset
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+		remoteEndpoints: map[string]*submV1.Endpoint{},
+	}
+}
+
+func (h *calicoIppoolHandler) GetNetworkPlugins() []string {
+	return []string{cni.Calico}
+}
+
+func (h *calicoIppoolHandler) GetName() string {
+	return "Calico_Ippool handler"
```
tpantelis:
```suggestion
	return "Calico IPPool handler"
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (h *calicoIppoolHandler) Init() error {
+	var err error
+
+	h.cs, err = calicocs.NewForConfig(h.restConfig)
+	if err != nil {
+		return errors.Wrap(err, "error initializing Calico clientset")
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return errors.Wrap(err, "error initializing Calico clientset")
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (h *calicoIppoolHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	err := h.handleRemoteEndpointCreateOrUpdate(endpoint, true)
+	if err != nil {
+		return errors.Wrapf(err, "failed to handle RemoteEndpointCreated event")
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "failed to handle RemoteEndpointCreated event")
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (h *calicoIppoolHandler) RemoteEndpointUpdated(endpoint *submV1.Endpoint) error {
+	err := h.handleRemoteEndpointCreateOrUpdate(endpoint, false)
+	if err != nil {
+		return errors.Wrapf(err, "failed to handle RemoteEndpointUpdated event")
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "failed to handle RemoteEndpointUpdated event")
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (h *calicoIppoolHandler) RemoteEndpointRemoved(endpoint *submV1.Endpoint) error {
+	delete(h.remoteEndpoints, endpoint.Name)
+
+	if !h.isGateway {
+		logger.Info("Ignore RemoteEndpointRemoved event (node isn't Gateway)")
```
tpantelis: I don't think we need this logged by deault.

```suggestion
		logger.V(log.TRACE).Info("Ignore RemoteEndpointRemoved event (node isn't Gateway)")
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+type createUpdateFunc func(ipPoolObj *calicoapi.IPPool) error
+
+type calicoIppoolHandler struct {
+	event.HandlerBase
+	restConfig      *rest.Config
+	cs              *calicocs.Clientset
+	remoteEndpoints map[string]*submV1.Endpoint
+	isGateway       bool
```
tpantelis: We should protect these with a mutex as is done in other handlers (eg ).https://github.com/submariner-io/submariner/blob/c2cae00ace9c6c50cb1294a94cd36d75f19f207a/pkg/routeagent_driver/handlers/kubeproxy/gw_transition.go#L29

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,256 @@
...
+				Labels: map[string]string{submarinerIPPool: "true"},
+			},
+			Spec: calicoapi.IPPoolSpec{
+				CIDR:        subnet,
+				NATOutgoing: false,
+				Disabled:    true,
+			},
+		}
+
+		err := funcToRun(iPPoolObj)
```
tpantelis: I don't think the update operation is appropriate here b/c no field can actually change. The only thing that's variable is the `subnet` but that is also part of the name so it's unique for every instance and can't change. I think what you want to do here is basically "diff" the new subnets with existing `IPPools`, ie 

- obtain the previous `IPPools`
- for each subnet, if there's no previous `IPPool` create it
- delete any previous `IPPools` w/o a corresponding subnet

tpantelis (changes requested):

pkg/routeagent_driver/handlers/calico/ippool_handler.go
yboaron: Do you still think we should use mutex here ?

pkg/routeagent_driver/handlers/calico/ippool_handler.go
yboaron: Well, since changing cluster name and CIDRs post submariner deployment isn't supported, we don't need to handle endpoint update event here. I removed the relevant code.

yboaron:
Few observations after checking this PR on OCP with Calico clusters on IBM ROKS, 

- According to Calico recommendation we should use  projectcalico.org/v3 to create IPPools and not crd.projectcalico.org/v1 (see [1] )

-  To enable projectcalico.org/v3  API the Calico API server should be installed in the cluster  [2]
- Calico API server not installed on OCP clusters on ROKS

- After installing Calico API server by following the instructions from [2], Calico IPPools needed for Submariner created by this PR as expected.


I filed [3] issue on Calico project, let's wait for their reply to decide how to proceed with this PR

[1] 
https://github.com/projectcalico/calico/issues/6412
[2] 
https://docs.tigera.io/calico/latest/operations/install-apiserver#how-to
[3] 
https://github.com/projectcalico/calico/issues/7845

pkg/routeagent_driver/handlers/calico/ippool_handler.go
tpantelis: As discussed elsewhere it shouldn't be needed unless mutable fields can be accessed outside the parent's event loop. This appears to be the case with `isGateway` in the `Stop` method. To be completely safe, you can use an [atomic bool](https://pkg.go.dev/sync/atomic#Bool) for `isGateway`.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		if _, err := h.client.ProjectcalicoV3().IPPools().Create(context.TODO(), iPPoolObj,
+			metav1.CreateOptions{}); err != nil && !apierrors.IsAlreadyExists(err) {
+			logger.Errorf(err, "Error creating Calico IPPool for ClusterID %q subnet %q", endpoint.Spec.ClusterID, subnet)
+			retErr = err
+		} else {
+			logger.Infof("Successfully create Calico IPPool %q", iPPoolObj.GetName())
+		}
+	}
```
tpantelis: Minor thing but we should only log if it was actually created. Also we don't need to log the error here since we're returning it:

```suggestion
		err := h.client.ProjectcalicoV3().IPPools().Create(context.TODO(), iPPoolObj,
			metav1.CreateOptions{}))
	         if err == nil {
	             logger.Infof("Successfully created Calico IPPool %q", iPPoolObj.GetName())
	             continue
	         }
	         
	         if !apierrors.IsAlreadyExists(err) {
			retErrors = append(retErrors, errors.Wrapf(err, "error creating Calico IPPool for ClusterID %q subnet %q", endpoint.Spec.ClusterID, subnet)))
			
		}
	}
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,198 @@
...
+
+	logger.Infof("Successfully delete Calico IPPools using labelSelector %q", labelSelector)
+
+	return nil
+}
+
+func (h *calicoIPPoolHandler) createIPPool(endpoint *submV1.Endpoint) error {
+	subnets := cidr.ExtractIPv4Subnets(endpoint.Spec.Subnets)
+	retErr := error(nil)
```
tpantelis: I think for completeness we should return an [aggregate error](https://github.com/kubernetes/apimachinery/blob/c9b3b3a37189129d0d80b894dc5d393ade9d3b79/pkg/util/errors/errors.go#L46) instead of just returning the last one, ie:

```
import errorutils "k8s.io/apimachinery/pkg/util/errors"
...
var retErrors []error

...
retErrors = append(retErrors, ...)
...

return errorutils.NewAggregate(retErrors)
```

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,198 @@
...
+		} else {
+			logger.Infof("Successfully create Calico IPPool %q", iPPoolObj.GetName())
+		}
+	}
+
+	return retErr
+}
+
+func (h *calicoIPPoolHandler) deleteIPPool(endpoint *submV1.Endpoint) error {
+	retErr := error(nil)
```
tpantelis: Same as above re: aggregate error

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,198 @@
...
+	for _, subnet := range subnets {
+		poolName := getEndpointSubnetIPPoolName(endpoint, subnet)
+		if err := h.client.ProjectcalicoV3().IPPools().Delete(context.TODO(),
+			poolName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+			logger.Errorf(err, "Error deleting Calico IPPool for ClusterID %q subnet %q", endpoint.Spec.ClusterID, subnet)
+			retErr = err
+		} else {
+			logger.Infof("Successfully delete Calico IPPool %q", poolName)
+		}
+	}
```
tpantelis: We shouldn't log success if not found:

```suggestion
		err := h.client.ProjectcalicoV3().IPPools().Delete(context.TODO(),
			poolName, metav1.DeleteOptions{})
		if err == nil {
		    logger.Infof("Successfully deleted Calico IPPool %q", poolName)
		    continue
		}
		    
		if !apierrors.IsNotFound(err) {
		    retErrors = append(retErrors, errors.Wrapf(err, "error deleting Calico IPPool for ClusterID %q subnet %q", endpoint.Spec.ClusterID, subnet)))
	}
```

tpantelis (changes requested):

yboaron:
> Few observations after checking this PR on OCP with Calico clusters on IBM ROKS,
> 
> * According to Calico recommendation we should use  projectcalico.org/v3 to create IPPools and not crd.projectcalico.org/v1 (see [1] )
> * To enable projectcalico.org/v3  API the Calico API server should be installed in the cluster  [2]
> * Calico API server not installed on OCP clusters on ROKS
> * After installing Calico API server by following the instructions from [2], Calico IPPools needed for Submariner created by this PR as expected.
> 
> I filed [3] issue on Calico project, let's wait for their reply to decide how to proceed with this PR
> 
> [1] [projectcalico/calico#6412](https://github.com/projectcalico/calico/issues/6412) [2] https://docs.tigera.io/calico/latest/operations/install-apiserver#how-to [3] [projectcalico/calico#7845](https://github.com/projectcalico/calico/issues/7845)


Calico folks plan to work on providing a new API  to create IPPools that can be used regardless to the Calico API server , until this API will be available running Calico API server is a prerequisite for automating Calico IPPools creation.

tpantelis (approved):

pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func (h *calicoIPPoolHandler) RemoteEndpointCreated(endpoint *submV1.Endpoint) error {
+	h.remoteEndpoints[endpoint.Name] = endpoint
+	if !h.isGateway.Load() {
+		logger.V(log.TRACE).Info("Ignore RemoteEndpointCreated event (node isn't Gateway)")
+		return nil
+	}
+
+	err := h.createIPPool(endpoint)
```
sridhargaddam, line 87: Will we get an error if Calico API server is not available in the cluster?

If so, we can probably include that in the error message instructing them to install the API server.

yboaron: Yep, we should get an error,  
updated the error msg

sridhargaddam (approved):

PR #2599, opened 2023-07-03
libreswan: Present --dpddelay if --dpdaction is specified

ueno:
These couple of commits are based on the suggestion from @paulwouters some time ago. I'm marking this as RFC as I haven't tested it but just built with "make build".


pkg/cable/libreswan/libreswan.go
```diff
@@ -45,6 +45,7 @@ import (
 const (
 	cableDriverName = "libreswan"
 	whackTimeout    = 5 * time.Second
+	dpdDelay        = 30
```
skitt: Please indicate the units:

```suggestion
	dpdDelay        = 30 // seconds
```

sridhargaddam:
Thank you @ueno for the PR.

skitt (changes requested):
Thanks for this; both changes make sense. I only have one small readability request...

skitt (approved):

pkg/cable/libreswan/libreswan.go
```diff
@@ -460,7 +463,8 @@ func (i *libreswan) serverConnectToEndpoint(connectionName string, endpointInfo
 		"--id", remoteEndpointIdentifier,
 		"--host", "%any",
 		"--client", rightSubnet,
-		"--dpdaction=hold")
+		"--dpdaction=hold",
+		"--dpddelay", strconv.Itoa(dpdDelay))
```
yboaron, line 477: Thanks for this PR @ueno , overall code LGTM.
I have short Q:   do we need to set also **dpdtimeout** in case **dpddelay** is set?  

According to https://libreswan.org/man/ipsec.conf.5.html  : 
`dpddelay

Set the delay (in time units, defaults to seconds) between Dead Peer Detection (IKEv1 [RFC 3706](http://www.ietf.org/rfc/rfc3706.txt)) or IKEv2 Liveness keepalives that are sent for this connection (default 0 seconds). Set to enable checking. **If dpddelay is set, dpdtimeout also needs to be se**t.`

ueno: According to Paul, dpdtimeout is only for IKEv1 and ignored if IKEv2 is used, so if you only use IKEv2 it shouldn't be necessary.

yboaron (approved):

nyechiel:
What's the plan for taking this out of draft? Do we need further testing?

sridhargaddam (approved):

pkg/cable/libreswan/libreswan.go
paulwouters: correct. with ikev1 message ID is random, and you can send many messages to see if you get a response with the same msgid. For IKEv2, msgid is linear and has a window size. Libreswan uses windows size = 1, so you can only have one IKE message outstanding per IKE SA. So if you send a DPD probe (IKEv2 Informational exchange msg), then you cannot any other IKE message until you get a response. Therefor, not getting a response is fatal, and this is tied to the generic IKE message response timeouts and there is no seperate "DPD timeout"

pkg/cable/libreswan/libreswan.go
```diff
@@ -175,7 +176,7 @@ func retrieveActiveConnectionStats() (map[string]int, map[string]int, error) {
 	defer cancel()
 
 	// Retrieve active tunnels from the daemon
-	cmd := exec.CommandContext(ctx, "/usr/libexec/ipsec/whack", "--trafficstatus")
+	cmd := exec.CommandContext(ctx, "/usr/sbin/ipsec", "whack", "--trafficstatus")
```
paulwouters, line 176: this is marginally better, but the best is to use "ipsec" without a path, so it works as long as it is in the path (/sbin or /usr/sbin or /usr/local/sbin or whatever compile time options are used). Using "ipsec <cmd> <options>" is supported specifically to avoid any hardcoded paths. if exec.CommandContext uses regular PATH handling, this is best run without the hardcoded path

pkg/cable/libreswan/libreswan.go
paulwouters: I will fix the man page upstream.

paulwouters (approved):
LGTM but see the comment on using /usr/sbin/ipsec vs ipsec

pkg/cable/libreswan/libreswan.go
ueno: Yeah, I can modify the PR to use `exec.LookPath("ipsec")`, maybe somewhere in init.

nyechiel:
@ueno we are approaching the end of the Submariner 0.16 dev cycle, and going to enter code freeze soon. Can you please take this out of draft? I am assuming this is something we want to merge. Thanks!

sridhargaddam:
All e2e deployments are failing while overriding the network-plugin-syncer component.
The following PR should fix the issue - https://github.com/submariner-io/shipyard/pull/1340

PR #2614, opened 2023-07-18
Consider Pluto status on Libreswan GetConnections

yboaron:
To reduce the downtime of the data path in case the gateway pod is restarted, we start Pluto only when Liberswan connects to the first remote endpoint.
Which means if 0 remote endpoints are defined Pluto is not running.

The GetConnections function returns the connections status retrieved from 'whack --trafficstatus' command and its called periodically regardless of Pluto status.

This PR updates GetConnections also to support the case when Pluto is not running


sridhargaddam (commented):
Please fix the linter issues, otherwise LGTM.

sridhargaddam (approved):

skitt (approved):

tpantelis (approved):

PR #2638, opened 2023-08-03
Use LeasesResourceLock instead of ConfigMapsLeasesResourceLock

tpantelis:
...for leader election. The latter is more expensive and is intended as a temporary migration between `ConfigMaps` and `Leases`. We migrated from `ConfigMaps` a year ago so we should be safe to migrate to `Leases` only.


sridhargaddam:
> We migrated from `ConfigMaps` a year ago so we should be safe to migrate to `Leases` only.

Currently there is a configMap created for the submariner-gateway-lock. Can you elaborate what you mean by migrated from configMaps a year ago?

tpantelis:
> > We migrated from `ConfigMaps` a year ago so we should be safe to migrate to `Leases` only.
> 
> Currently there is a configMap created for the submariner-gateway-lock. Can you elaborate what you mean by migrated from configMaps a year ago?

We migrated from ConfigMaps resource lock to ConfigMapsAndLeases resource lock a year ago when bumping to K8s 0.24.2 b/c it no longer supported just ConfigMaps. ConfigMapsAndLeases creates both a ConfigMap and a Lease for backward compatibility migration.

sridhargaddam (approved):

skitt (approved):

PR #2644, opened 2023-08-08
Change the node_handler log to trace

sridhargaddam:
As the log from areNodesEquivalent is getting printed at regular intervals change it to trace.


tpantelis (approved):

pkg/controllers/datastoresyncer/node_handler.go
```diff
@@ -62,7 +62,7 @@ func (d *DatastoreSyncer) areNodesEquivalent(obj1, obj2 *unstructured.Unstructur
 	existingGlobalIP := obj1.GetAnnotations()[constants.SmGlobalIP]
 	newGlobalIP := obj2.GetAnnotations()[constants.SmGlobalIP]
 
-	logger.V(log.DEBUG).Infof("areNodesEquivalent called for %q, existingGlobalIP %q, newGlobalIP %q",
+	logger.V(log.TRACE).Infof("areNodesEquivalent called for %q, existingGlobalIP %q, newGlobalIP %q",
```
tpantelis, line 65: I don't think think we need this log at all but TRACE is fine.

maayanf24 (approved):

PR #2656, opened 2023-08-23
Upgrade changes for OVN IC

### LINKED ISSUE #193: Support upgrades to the new OVN IC implementation
Update should be supported 
* when submariner moves from old implementation to new
*  when Openshift moves from an non-ic to ic version

vthapar:
When a submariner installation upgrades from legacy non-OVN-IC version to an OVN-IC version, cleanup resources created by legacy installation.

Fixes https://github.com/submariner-io/enhancements/issues/193

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/handler.go
+++ b/pkg/routeagent_driver/handlers/ovn/handler.go
@@ -88,6 +88,8 @@ func (ovn *Handler) GetNetworkPlugins() []string {
 }
 
 func (ovn *Handler) Init() error {
+	ovn.LegacyCleanup()
+
 	err := ovn.initIPtablesChains()
 	if err != nil {
 		return err
--- a/pkg/routeagent_driver/handlers/ovn/vsctl/vsctl.go
+++ b/pkg/routeagent_driver/handlers/ovn/vsctl/vsctl.go
@@ -65,7 +65,7 @@ func AddBridge(bridgeName string) error {
 }
 
 func DelBridge(bridgeName string) error {
-	_, err := vsctlCmd("del-br", bridgeName)
+	_, err := vsctlCmd("--if-exists", "del-br", bridgeName)
 
 	return err
 }


aswinsuryan (dismissed):

pkg/routeagent_driver/handlers/ovn/handler.go
```diff
@@ -88,6 +88,8 @@ func (ovn *Handler) GetNetworkPlugins() []string {
 }
 
 func (ovn *Handler) Init() error {
+	ovn.LegacyCleanup()
```
sridhargaddam, line 91: LegacyCleanup is deleting the `ovn-k8s-sub0` and `br-submariner` interfaces. 
What about the flows like Logical Router policies that are programmed in the `ovn_cluster_router` and `submariner_router` ?

vthapar: @aswinsuryan Can correct if my understanding is wrong, but those exist only on node where NetworkPluginSyncer was running and will be removed as part of Cleanup when NetworkPluginSyncer is "stopped" as part of upgrade process.

This LegacyCleanup is for resources that are created on all nodes as part of RouteAgent and is not covered by NetowrkPluginSyncer's cleanup.

aswinsuryan: ya that is right , the changes done here https://github.com/submariner-io/submariner-operator/pull/2759 , will delete any networkplugin-syncer deployment  if found and that should cleanup the routes added to ovn routers.

tpantelis: @aswinsuryan Are you assuming the operator will initiate the uninstall process for the networkplugin-syncer? If so, this is not (currently) the case - the operator just deletes the networkplugin-syncer deployment.

aswinsuryan: oh ok ,  I was thinking on delete Stop would be called. If that is not the case we will  have to add it here.

tpantelis: We can have the operator initiate the uninstall process. That would save us having to preserve the networkplugin-syncer [uninstall code](https://github.com/submariner-io/submariner/blob/c2cae00ace9c6c50cb1294a94cd36d75f19f207a/pkg/networkplugin-syncer/handlers/ovn/uninstall.go#L31) which is not trivial.

In the operator, should we uninstall the networkplugin-syncer first before re-deploying the other components, ie can the legacy networkplugin-syncer OVN implementation (temporarily) co-exist with the new OVN IC implementation?

aswinsuryan: When I gave  more thought on it , I think calling it from nps may not work too. If the OVN changes from non-IC to IC , nps cannot connect to ovn db the way it does now . Also it is not sure if OVN is going to preserve these flows that submariner have created when it changes from non-ic to ic,  as it is not OVN created flows.

We may have to test it to see the state after an upgrade.

But for a submariner upgrade for Opeshift < 4.14 this would work.

Another way is to use a logic similar to here on startup to delete the flows. Like if the CIDR and priority matches but nethop does not delete the flow.

https://github.com/submariner-io/submariner/blob/devel/pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go#L78

tpantelis: Ok so it sounds like you think we should do whatever legacy NPS cleanup is needed here.

sridhargaddam: > Also it is not sure if OVN is going to preserve these flows that submariner have created when it changes from non-ic to ic, as it is not OVN created flows.

Good point. We have to understand how OVN-IC upgrade happens, whether it will read the etcd and change the state of the system to match it or if it uses the info from the NBDB/SBDB.

> We may have to test it to see the state after an upgrade.

Normally for a regular linux interface, when its deleted the associated routes/ips on that interface are deleted. In case of OVN, when `br-submariner` is deleted, does the associated openflow rules get deleted or remain as stale rules?

> Another way is to use a logic similar to here on startup to delete the flows. Like if the CIDR and priority matches but nethop does not delete the flow.

Are you suggesting that we do this while programming the required flows or as part of init() operation?

aswinsuryan: > Ok so it sounds like you think we should do whatever legacy NPS cleanup is needed here. 

Yes that is right

aswinsuryan: > > Also it is not sure if OVN is going to preserve these flows that submariner have created when it changes from non-ic to ic, as it is not OVN created flows.
> 
> 
> 
> Good point. We have to understand how OVN-IC upgrade happens, whether it will read the etcd and change the state of the system to match it or if it uses the info from the NBDB/SBDB.
> 
> 
> 
> > We may have to test it to see the state after an upgrade.
> 
> 
> 
> Normally for a regular linux interface, when its deleted the associated routes/ips on that interface are deleted. In case of OVN, when `br-submariner` is deleted, does the associated openflow rules get deleted or remain as stale rules?

Ya right it should get deleted when we remove ovn-k8s-sub0
> 
> 
> 
> > Another way is to use a logic similar to here on startup to delete the flows. Like if the CIDR and priority matches but nethop does not delete the flow.
> 
> 
> 
> Are you suggesting that we do this while programming the required flows or as part of init() operation?

As a part of init().
> 
>

vthapar:
@aswinsuryan Delete ip routes or logicalrouteer policies?

aswinsuryan:
> @aswinsuryan Delete ip routes or logicalrouteer policies?

Logical router policies and static routes in Ovn cluster router I mean

pkg/routeagent_driver/handlers/ovn/handler.go
vthapar: > When I gave more thought on it , I think calling it from nps may not work too. If the OVN changes from non-IC to IC , nps cannot connect to ovn db the way it does now . Also it is not sure if OVN is going to preserve these flows that submariner have created when it changes from non-ic to ic, as it is not OVN created flows.
> 
Is that a valid upgrade path? Upgrading from non-ic to ic before upgrading means deployment will be using an unsupported deployment of Submariner [< 0.16 with ic]. Proper way to upgrade should be to upgrade Submariner to 0.16+ before upgrading to IC coz 0.16 is only one that supports IC and non IC both.

> We may have to test it to see the state after an upgrade.
> 
> But for a submariner upgrade for Opeshift < 4.14 this would work.
> 
So, requiring a Submariner upgrade before upgrading OCP should take care of it. I am sure other projects will also have a similar requirement, upgrade them before switching OCP tp 4.14 coz older code will likely be incompatible with OCP 4.14.

> Another way is to use a logic similar to here on startup to delete the flows. Like if the CIDR and priority matches but nethop does not delete the flow.
> 
> https://github.com/submariner-io/submariner/blob/devel/pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go#L78

Will this be still needed if we require submariner upgrade before OCP?

aswinsuryan: > needed if we require submariner upgrade before OCP?

This should solve the cleanup issue if we add the code to remove flows during netwrok plugin uninstall. However the operator needs to discover the CNI  again to move from non-IC to IC. Currently, we do not support on-the-fly change from non-ic mode to IC mode.

aswinsuryan (approved):

sridhargaddam (approved):

skitt (approved):

vthapar:
@aswinsuryan OVN tests failing. Failed to deploy kind with OVN.

```
2023-08-30T13:53:59.2282190Z [36m[13:53:59.227] [dir=submariner]$ exit_error Failed to create clusters using 'kind'.[0m
2023-08-30T13:53:59.2294776Z ERROR: Failed to create clusters using 'kind'.
2023-08-30T13:53:59.2321587Z [36m[13:53:59.231] [dir=submariner; fn=exit_error]$ exit 1[0m
2023-08-30T13:53:59.2346922Z make[1]: *** [/opt/shipyard/Makefile.inc:188: clusters] Error 1
2023-08-30T13:53:59.2347631Z make[1]: Leaving directory '/go/src/github.com/submariner-io/submariner'
```

aswinsuryan:
> @aswinsuryan OVN tests failing. Failed to deploy kind with OVN.
> 
> ```
> 2023-08-30T13:53:59.2282190Z �[36m[13:53:59.227] [dir=submariner]$ exit_error Failed to create clusters using 'kind'.�[0m
> 2023-08-30T13:53:59.2294776Z ERROR: Failed to create clusters using 'kind'.
> 2023-08-30T13:53:59.2321587Z �[36m[13:53:59.231] [dir=submariner; fn=exit_error]$ exit 1�[0m
> 2023-08-30T13:53:59.2346922Z make[1]: *** [/opt/shipyard/Makefile.inc:188: clusters] Error 1
> 2023-08-30T13:53:59.2347631Z make[1]: Leaving directory '/go/src/github.com/submariner-io/submariner'
> ```

This should fix it 

https://github.com/submariner-io/shipyard/pull/1351

PR #2659, opened 2023-08-24
Ignore endpoint events after retrying couple of times

sridhargaddam:
If the endpoint handlers cannot recover from a failure, even after retrying 20 times, it's best to disregard the event. This prevents the logs from being flooded with repetitive errors.

### DIFF
--- a/pkg/event/controller/controller.go
+++ b/pkg/event/controller/controller.go
@@ -54,6 +54,10 @@ type Controller struct {
 	isGatewayNode bool
 }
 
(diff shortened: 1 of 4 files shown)


sridhargaddam:
Some background: In a Calico deployment where the Calico API server was not installed, the route-agent pod logs were completely filled with errors. This PR would help avoid such situations.

tpantelis (approved):

skitt (approved):

PR #2681, opened 2023-09-12
Fix e2e failures seen with OCP OVN-IC deployments

sridhargaddam:
When a client pod attempts to establish a connection with a service in the remote cluster, it has been observed that the response traffic is being discarded at the local Gateway node. The return traffic is getting dropped because of some DROP rules in the FORWARDing chain of the filter table (which seems to have been added recently in the OVN OCP deployments).

Submariner normally has rules to allow such traffic, but with the new architecture designed to support OVN-IC deployments, we no longer use the ovn-k8s-sub0 iface and simply use the ovn-k8s-mp0 interface. This PR updates the firewall rules accordingly to allow the return traffic on the Gateway node.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
@@ -102,9 +102,20 @@ func (ovn *Handler) getForwardingRuleSpecs() ([][]string, error) {
 			"this will be retried")
 	}
 
-	rules := [][]string{
-		{"-i", OVNK8sMgmntIntfName, "-o", ovn.cableRoutingInterface.Name, "-j", "ACCEPT"},
-		{"-i", ovn.cableRoutingInterface.Name, "-o", OVNK8sMgmntIntfName, "-j", "ACCEPT"},
+	// On the Gateway node, the incoming traffic first lands on the br-ex, which includes the physical interface.
+	// The OpenFlow rules on the br-ex subsequently direct Submariner traffic to the local networking stack.
+	// To reroute incoming traffic over the ovn-k8s-mp0 interface, we employ routes in table 149. Before the traffic
+	// hits ovn-k8s-mp0, firewall rules would be processed. Therefore, we include these firewall rules in the FORWARDing
+	// chain to allow such traffic. Similar thing happens for outbound traffic as well, and we use routes in table 150.
+	rules := [][]string{}
+	for _, remoteCIDR := range ovn.getRemoteSubnets().UnsortedList() {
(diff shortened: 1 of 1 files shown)


aswinsuryan (approved):

nyechiel (approved):

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -102,9 +102,20 @@ func (ovn *Handler) getForwardingRuleSpecs() ([][]string, error) {
...
+	// On the Gateway node, the incoming traffic first lands on the br-ex, which includes the physical interface.
+	// The OpenFlow rules on the br-ex subsequently direct Submariner traffic to the local networking stack.
+	// To reroute incoming traffic over the ovn-k8s-mp0 interface, we employ routes in table 149. Before the traffic
+	// hits ovn-k8s-mp0, firewall rules would be processed. Therefore, we include these firewall rules in the FORWARDing
+	// chain to allow such traffic. Similar thing happens for outbound traffic as well, and we use routes in table 150.
+	rules := [][]string{}
+	for _, remoteCIDR := range ovn.getRemoteSubnets().UnsortedList() {
+		rules = append(rules,
+			[]string{
+				"-d", remoteCIDR, "-i", OVNK8sMgmntIntfName, "-o", ovn.cableRoutingInterface.Name, "-j", "ACCEPT",
```
yboaron, line 114: Not deeply familiar with OVN architecture, does using 'remoteCIDR' should work also with GlobalNet ?

sridhargaddam: Yes, it will work with Globalnet as well. In Globalnet deployments, remoteCIDR will point to globalCIDR of remote clusters.

tpantelis (approved):

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
yboaron: Yep, I know that remoteCIDR point to globalCIDR, wasn't sure if we hit these rules before or after GN natting, 
if we hit them before GN natting I guess it should be OK.
Thanks

yboaron (approved):

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
sridhargaddam: Ack. I verified this PR both on Globalnet and non-Globalnet deployments.

PR #2674, opened 2023-09-06
Fix network discovery issue in ovn non-ic deployments

aswinsuryan:
Fix network discovery issue in ovn non-ic deployments

Fixes: https://github.com/submariner-io/submariner/issues/2673

PR #2682, opened 2023-09-12
Refactor the ovn network discovery code

aswinsuryan:
Refactor the ovn network discovery code

This is the follow up PR for the comments in https://github.com/submariner-io/submariner/pull/2674

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/connection.go
+++ b/pkg/routeagent_driver/handlers/ovn/connection.go
@@ -204,17 +204,16 @@ func discoverOvnKubernetesNetwork(ctx context.Context, k8sClientset clientset.In
 }
 
 /*
-
-openshift non-ic: in this case there will get the "app=ovnkube-node" and check for ovnkube-db service,
-if present we will the default Openshift db path
-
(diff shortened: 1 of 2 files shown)


pkg/routeagent_driver/handlers/ovn/connection.go
```diff
@@ -245,11 +247,15 @@ func discoverKindOvnKubernetesNetwork(ctx context.Context, k8sClientSet clientse
 		return discoverKindOvnDBClusterNetwork(ctx, ovnDBPod, k8sClientSet)
 	}
 
-	ovnPod, err := FindPod(ctx, k8sClientSet, "app=ovnkube-node")
+	ovnPod, err := FindPod(ctx, k8sClientSet, ovnPodLabel)
```
sridhargaddam, line 250: We only use the namespace from the ovnPod. So can we use the namespace of ovnDBPod instead?

aswinsuryan: This is for the case where ovnDB pod is nil. if not nil we use discoverKindOvnDBClusterNetwork

sridhargaddam (approved):

tpantelis (approved):

PR #2708, opened 2023-09-21
Fix IPsec whack returning exit status 33

sridhargaddam:
Normally, when the Gateway pod goes down, Submariner retains the existing IPsec entries on the host to ensure data continuity. However, when the pod restarts and the Pluto process initializes, it clears the existing connections from the kernel and data path gets restored as soon as the connections are re-established.

A recent PR[*] aimed to reduce the downtime of the data path by initializing the Pluto process during the first connection request, which is a positive change. However, there was an issue with the API responsible for starting the Pluto process. It was waiting for a mere 5 seconds and was not handling the timeout properly. Consequently, attempting to execute any 'whack' commands while the control socket was absent resulted in an error code of 33 and connections could not be established.

In one of the OCP deployments, it was observed that it took approximately 25 seconds for the control socket to be created when Pluto was initialized. This PR addresses the issue by increasing the monitoring interval to 1 minute and also handles timeout scenarios properly.

[*] https://github.com/submariner-io/submariner/pull/2543

### DIFF
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -616,19 +616,9 @@ func (i *libreswan) runPluto() error {
 		logger.Fatalf("Pluto exited: %v", cmd.Wait())
 	}()
 
-	// Wait up to 5s for the control socket.
-	for i := 0; i < 250; i++ {
-		_, err := os.Stat("/run/pluto/pluto.ctl")
-		if err == nil {
-			break
-		}
-
-		if !os.IsNotExist(err) {
-			logger.Infof("Failed to stat the control socket: %v", err)
-			break
-		}
-
-		time.Sleep(20 * time.Millisecond)
+	err := i.waitForControlSocket()
+	if err != nil {
+		return err
 	}
 
 	if i.debug {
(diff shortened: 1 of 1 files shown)


sridhargaddam:
This PR has to be backported to the following branches: 0.13, 0.14, 0.15 and 0.16

pkg/cable/libreswan/libreswan.go
```diff
@@ -616,28 +616,41 @@ func (i *libreswan) runPluto() error {
...
+		}
+	}
+
+	return nil
+}
+
+func (i *libreswan) waitForControlSocket() error {
+	// Wait for upto a minute for the control socket to be created.
+	const maxAttempts = 600
+	const retryInterval = 100 * time.Millisecond
```
yboaron, line 636: QQ: don't we want to monitor the status at a faster rate? to shorten DP downtime when GW pod is restarted ?

sridhargaddam: I can reduce it to 20 ms like before, but it also means more API calls. Also, looking at the time it normally takes on OCP which is in seconds, 100 ms sounded a reasonable time for me. Hence I used 100 ms.

yboaron: OK, let's leave it 100mSec, we can optimize the interval in the future

yboaron (approved):

skitt (approved):

PR #2709, opened 2023-09-21
Add kubectl to the route-agent image

skitt:
In order to ensure that host path mounts don't interfere with socket creation, the route agent pods need to wait for the node to be ready (which includes CNI readiness) before starting. The simplest way to do that is to ship kubectl in the container image; to ensure it is built from scratch, this adds the minimal command wrapper in a separate module, and builds it for use in the image container.

This will be put to use in the operator; the corresponding change will have full details of the issue being fixed.


tools/kubectl.go
```diff
@@ -0,0 +1,34 @@
...
+	"k8s.io/component-base/cli"
+	"k8s.io/kubectl/pkg/cmd"
+	"k8s.io/kubectl/pkg/cmd/util"
+
+	// Import to initialize client auth plugins.
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+)
+
+func main() {
+	command := cmd.NewDefaultKubectlCommand()
```
sridhargaddam, line 29: Please take a look at my comment on the other PR - https://github.com/submariner-io/submariner-operator/pull/2824#discussion_r1332872651

PR #2710, opened 2023-09-21
Retry IPsec connections when there are any whack errors

sridhargaddam:
In the current code, when 'whack' encounters an error, we only log a warning message without hanlding the error case. As a result, in cases where there is an issue connecting to a remote endpoint, the caller is left unaware that connection attempts have failed, and no retries are initiated. This PR fixes it.


sridhargaddam:
This PR has to be backported to 0.13, 0.14, 0.15 and 0.16 branches.

skitt (approved):

yboaron (approved):

PR #2719, opened 2023-09-21
Allow the route agent to wait for node readiness

skitt:
In order to ensure that host path mounts don't interfere with socket
creation, the route agent pods need to wait for the node to be ready
(which includes CNI readiness) before starting. This adds a variant of
the route agent, invoked using the SUBMARINER_WAITFORNODE environment
variable, to do just that (similar in style to the uninstallation
command).

This will be put to use in the operator; the corresponding change will
have full details of the issue being fixed.

### DIFF
--- a/pkg/routeagent_driver/main.go
+++ b/pkg/routeagent_driver/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/kelseyhightower/envconfig"
 	"github.com/pkg/errors"
@@ -47,10 +48,12 @@ import (
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/mtu"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/ovn"
 	"github.com/submariner-io/submariner/pkg/versions"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/clientcmd"
+	nodeutil "k8s.io/component-helpers/node/util"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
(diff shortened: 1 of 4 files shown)


pkg/routeagent_driver/main.go
```diff
@@ -133,18 +147,6 @@ func main() {
 		logger.Fatalf("Error registering the handlers: %s", err.Error())
 	}
 
-	if env.Uninstall {
-		if err := registry.StopHandlers(true); err != nil {
-			logger.Warningf("Error stopping handlers: %v", err)
-		}
-
```
sridhargaddam, line 140: Why is this code removed?

sridhargaddam (changes requested):

pkg/routeagent_driver/main.go
sridhargaddam: The stopHandlers is where the actual cleanup happens.

skitt: Oops!

sridhargaddam (approved):

pkg/routeagent_driver/main.go
```diff
@@ -211,3 +213,36 @@ func getTCPMssValue(k8sClientSet *kubernetes.Clientset) int {
...
+
+		if err != nil {
+			logger.Error(err, "Error retrieving local node")
+		} else if localNode != nil {
+			for _, condition := range localNode.Status.Conditions {
+				if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
+					logger.Info("Node ready")
+					return
+				}
+			}
```
tpantelis: import "k8s.io/apimachinery/pkg/api/meta"

```suggestion
                         condition = meta.FindStatusCondition(localNode.Status.Conditions, corev1.NodeReady)
			if condition != nil && condition.Status == corev1.ConditionTrue {
			    logger.Info("Node ready")
			    return
			}
```

tpantelis (changes requested):

pkg/routeagent_driver/main.go
skitt: Nice, thanks!

skitt: ... except that node conditions use their own type, so `FindStatusCondition` can’t be used. There is [a `Node` equivalent](https://pkg.go.dev/k8s.io/kubernetes/pkg/controller/util/node#GetNodeCondition) but it’s not designed for use outside k/k and it can’t be imported without resorting to major `replace` shenanigans.

skitt: Ah, there is a usable one in `component-helpers`.

tpantelis: yeah there's still legacy `Condition` types pre-existing the standard **metav1** `Condition` type.

tpantelis (approved):

PR #2720, opened 2023-09-22
Remove the MTU workaround used for OVN load-balancers

sridhargaddam:
With the following update[*], OVN now forwards ICMP packets to OVN Loadbalancers, facilitating end-to-end PMTUD. As a result, the workaround used in the OVN route-agent handler may no longer be necessary.

[*] https://github.com/ovn-org/ovn-kubernetes/pull/3762

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
@@ -141,15 +141,6 @@ func (ovn *Handler) getMSSClampingRuleSpecs() ([][]string, error) {
 			})
 	}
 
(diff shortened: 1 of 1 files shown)


skitt:
Do we know that this fix is deployed on all the OVN-based platforms we care about? (Or will be by the time we release 0.17...)

aswinsuryan (approved):

aswinsuryan:
> Do we know that this fix is deployed on all the OVN-based platforms we care about? (Or will be by the time we release 0.17...)

This fix was reffering to ovn-k8s-sub0 and is not used anymore by the new implementation. So even if the fix does not land we should be fine.

skitt (approved):

dfarrell07:
We want to re-evaluate the OpenShift on OpenStack to public-cloud and then decide if this needs to be backported.

PR #2738, opened 2023-10-03
Fix log message from nodeHandler

sridhargaddam:
When debug is turned on in the Gateway pod, we are logging the globalIP assigned to the node at regular intervals. These repeated log messages do not add any significant value. Hence this PR updates the API to log only when there is a change in the globalIP assigned to the node.

### DIFF
--- a/pkg/controllers/datastoresyncer/node_handler.go
+++ b/pkg/controllers/datastoresyncer/node_handler.go
@@ -21,7 +21,6 @@ package datastoresyncer
 import (
 	"net"
 
-	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/submariner/pkg/globalnet/constants"
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


skitt (approved):

pkg/controllers/datastoresyncer/node_handler.go
```diff
@@ -62,8 +61,10 @@ func (d *DatastoreSyncer) areNodesEquivalent(obj1, obj2 *unstructured.Unstructur
 	existingGlobalIP := obj1.GetAnnotations()[constants.SmGlobalIP]
 	newGlobalIP := obj2.GetAnnotations()[constants.SmGlobalIP]
 
-	logger.V(log.TRACE).Infof("areNodesEquivalent called for %q, existingGlobalIP %q, newGlobalIP %q",
-		obj1.GetName(), existingGlobalIP, newGlobalIP)
+	if existingGlobalIP != newGlobalIP {
+		logger.Infof("areNodesEquivalent called for %q, existingGlobalIP %q, newGlobalIP %q",
```
tpantelis: Now that we're only logging a change, we should reflect that in the message.

```suggestion
		logger.Infof("Global IP for node %q changed from %q to %q", obj1.GetName(), existingGlobalIP, newGlobalIP)
```

tpantelis (changes requested):

tpantelis (approved):

PR #2743, opened 2023-10-07
Use leader election in globalnet controller

tpantelis:
Due to timing of notifications during gateway node transition, it's been observed that there can be a small window of a few seconds where the controllers on the previously active gateway are still running when the controllers on the newly active gateway are started. This results in two instances of the controllers running simultaneously in the cluster which can cause strange behavior and inconsistent results.

To alleviate this scenario, use leader election in the `GatewayMonitor` to ensure that the instance on the newly active gateway doesn't start its controllers until the previous instance has stopped its controllers and released the leader lock. There's a couple things to be aware of:

 - The K8s leader election functionality periodically renews the lease. If it can't be renewed prior to the `RenewDeadline`, it stops. For our usage, we don't have instances concurrently vying for leadership so we really don't need to keep renewing the lease. Ideally we would set the `RenewDeadline` very high to essentially disable it but it needs to be less than the `LeaseDuration` setting which we don't want too high. So we handle renewal failure by simply re-running leader election.

 - When transitioning to a non-gateway or shutting down, the `GatewayMonitor` cancels leader election which causes it to release the lock, which clears the `HolderIdentity` field and sets `LeaseDuration` to one second. This enables the next instance to quickly acquire the lock. However, if the controller instance crashes and doesn't properly release the lock, the next instance will have to await the `LeaseDuration` period before acquiring the lock. So we don't want `LeaseDuration` set too high and we don't want it too low either to give the current instance enough time to complete stopping its controllers. I picked 20 sec for `LeaseDuration` and 15 sec for `RenewDeadline`. So if the current instance takes longer than 20 sec to stop the controllers, it's still possible there could be two instance running concurrently but this scenario should be highly unlikely.


pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -39,23 +41,44 @@ import (
...
-		localSubnets:            set.New(localCIDRs...).UnsortedList(),
+		shuttingDown:            atomic.Bool{},
+		localSubnets:            set.New(config.LocalCIDRs...).UnsortedList(),
 		remoteSubnets:           set.New[string](),
 		remoteEndpointTimeStamp: map[string]metav1.Time{},
+		leaderElectionInfo:      atomic.Pointer[LeaderElectionInfo]{},
+		LeaderElectionConfig:    config.LeaderElectionConfig,
 	}
 
+	if gatewayMonitor.LeaseDuration == 0 {
```
skitt, line 79: Could you add the explanations for the default values as comments here?

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -506,6 +603,28 @@ func (g *gatewayMonitor) markRemoteClusterTraffic(remoteCidr string, addRules bo
 	}
 }
 
+func (l *LeaderElectionInfo) stop() {
+	if l.stopFunc != nil {
+		l.stopFunc()
+	}
+}
+
+func (l *LeaderElectionInfo) awaitStopped(timeout int) {
```
skitt: The timeout would be clearer if it was passed directly as a `time.Duration` (and on `stopControllers`).

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -134,15 +157,17 @@ func (g *gatewayMonitor) Start() error {
 func (g *gatewayMonitor) Stop() {
 	logger.Info("GatewayMonitor stopping")
 
+	g.shuttingDown.Store(true)
+
 	g.baseController.Stop()
 
-	g.stopControllers()
+	g.stopControllers(10)
```
skitt: Please add a comment to explain the value.

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -353,9 +439,10 @@ func (g *gatewayMonitor) startControllers() error {
 	return nil
 }
 
-func (g *gatewayMonitor) stopControllers() {
+func (g *gatewayMonitor) stopControllers(timeout int) {
```
skitt:
```suggestion
func (g *gatewayMonitor) stopControllers(leaderElectionTimeout int) {
```

but that’s not great, because the timeout isn’t used for the election timeout itself, but when stopping the controller; so perhaps `leaderElectionControllerTimeout` but that’s a mouthful.

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -506,6 +603,28 @@ func (g *gatewayMonitor) markRemoteClusterTraffic(remoteCidr string, addRules bo
...
+		return
+	}
+
+	if timeout <= 0 {
+		<-l.stopped
+	} else {
+		select {
+		case <-l.stopped:
+		case <-time.After(time.Duration(timeout) * time.Second):
+			logger.Warning("Timed out waiting for leader election stopped")
```
skitt:
```suggestion
			logger.Warning("Timed out waiting for leader election to stop")
```

pkg/globalnet/controllers/gateway_monitor.go
skitt: The timeout handling here might be better done with a context with a timeout ;-). (Just a thought, I haven’t tried it.)

skitt (changes requested):

pkg/globalnet/controllers/gateway_monitor.go
tpantelis: I changed it to a context.

dfarrell07 (approved):

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -252,11 +290,75 @@ func (g *gatewayMonitor) handleRemovedEndpoint(obj runtime.Object, _ int) bool {
...
 	defer g.controllersMutex.Unlock()
 
-	logger.Infof("On Gateway node - starting controllers")
+	// Since this is called asynchronously when leadership is gained, check that we're still on the gateway node and that we're
+	// not shutting down. Also, we may have regained leadership so ensure the controllers weren't already started.
+	if g.shuttingDown.Load() || !g.isGatewayNode.Load() || len(g.controllers) > 0 {
+		return nil
+	}
+
+	logger.Infof("Starting controllers")
```
sridhargaddam: nit: 
```suggestion
	logger.Info("Starting controllers")
```

sridhargaddam (approved):
Thanks for working on the fix @tpantelis

tpantelis:
@skitt @sridhargaddam So if leadership is lost due to failure after the `RenewDeadline` expires, currently we try to re-acquire leadership but we don't stop the controllers. A failure can really only happen if the node loses contact with the API server and can't update the `Lease` resource. I did this b/c we're only using leader election to prevent concurrent instances running during gateway transition. However, it's possible there could be a gateway transition during the time that leadership is lost and a new instance is able to contact the API server to acquire the lock (after the `LeaseDuration` expires). In this case, we'd have two controller instances running although the first is most likely effectively disabled anyway since most likely it can't contact the API server. However, if it regains contact,  before it observes the gateway has transitioned, it could try to process globalnet resources and causes strange inconsistencies like we've seen. This is an edge case but perhaps we should stop the controllers if leadership is lost due to failure, especially since the controllers are likely offline as well.  WDYT?

sridhargaddam:
> @skitt @sridhargaddam So if leadership is lost due to failure after the `RenewDeadline` expires, currently we try to re-acquire leadership but we don't stop the controllers. A failure can really only happen if the node loses contact with the API server and can't update the `Lease` resource. I did this b/c we're only using leader election to prevent concurrent instances running during gateway transition. However, it's possible there could be a gateway transition during the time that leadership is lost and a new instance is able to contact the API server to acquire the lock (after the `LeaseDuration` expires). In this case, we'd have two controller instances running although the first is most likely effectively disabled anyway since most likely it can't contact the API server. However, if it regains contact, before it observes the gateway has transitioned, it could try to process globalnet resources and causes strange inconsistencies like we've seen. This is an edge case but perhaps we should stop the controllers if leadership is lost due to failure, especially since the controllers are likely offline as well. WDYT?

@tpantelis I believe that if the Lease cannot be updated due to API server issues, we can stop the controllers without deleting the globalnet chains (i.e., clearGlobalnetChains), as this action might disrupt the datapath. 
After the API server connectivity is restored, the initial Gateway can verify if it has transitioned to a non-Gateway node and then proceed to clearGlobalnetChains.

tpantelis:
> @tpantelis I believe that if the Lease cannot be updated due to API server issues, we can stop the controllers without deleting the globalnet chains (i.e., clearGlobalnetChains), as this action might disrupt the datapath. After the API server connectivity is restored, the initial Gateway can verify if it has transitioned to a non-Gateway node and then proceed to clearGlobalnetChains.

OK - makes sense. Pushed a new [commit](https://github.com/submariner-io/submariner/pull/2743/commits/6ae6fc936d9913910caabb9011887f6db651dc07).

sridhargaddam:
@skitt can you please re-review?

skitt (approved):

PR #2752, opened 2023-10-16
On route agent startup, delete invalid directories

skitt:
Until recently, the operator set the route agent up with host volume mounts which would result in creating directories for any mounted socket which didn't exist. The operator now mounts the parent directories, but this doesn't remove directories which shouldn't have been created; to recover from such scenarios, the route agent needs to remove the invalid directories. (It can do this because the parent directories are mounted from the host.)

Depends on https://github.com/submariner-io/submariner-operator/pull/2846

### DIFF
--- a/pkg/routeagent_driver/main.go
+++ b/pkg/routeagent_driver/main.go
@@ -21,6 +21,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/fs"
 	"os"
 	"strconv"
 	"time"
@@ -82,6 +83,9 @@ func main() {
 	logger.Info("Starting submariner-route-agent using the event framework")
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler().Done()
+
+	// Clean up "sockets" created as directories by previous versions
+	removeInvalidSockets()
 
 	var env environment.Specification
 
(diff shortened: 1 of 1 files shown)


pkg/routeagent_driver/main.go
```diff
@@ -83,6 +84,9 @@ func main() {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler().Done()
 
+	// Clean up "sockets" created as directories by previous versions
+	removeInvalidSockets()
```
yboaron, line 88: Q: do we need to run the removeInvalidSockets function every time RA pod is restarted? or we can move it under RA init container code?

pkg/routeagent_driver/main.go
```diff
@@ -249,3 +253,14 @@ func waitForNodeReady(k8sClientSet *kubernetes.Clientset) {
...
 	}
 }
+
+func removeInvalidSockets() {
+	// This can be removed once we stop supporting upgrades from 0.16.0 or older
+	for _, dir := range []string{"/run/openvswitch/db.sock", "/var/run/openvswitch/ovnnb_db.sock", "/var/run/ovn-ic/ovnnb_db.sock"} {
+		info, err := os.Stat(dir)
+		if (err == nil || errors.Is(err, fs.ErrExist)) && info.IsDir() {
+			logger.Infof("Deleting invalid socket %s", dir)
+			os.Remove(dir)
```
yboaron: Do we want to check os.Remove return err value?

pkg/routeagent_driver/main.go
skitt: The init container doesn’t have the required mount volumes — its purpose is to wait until the node state is such that mounts are liable to succeed, so it can’t access anything in the corresponding mounts.

pkg/routeagent_driver/main.go
skitt: Good point, I’ll rework this to log either success or failure.

pkg/routeagent_driver/main.go
skitt: However with the updated volume handling in the operator, the init container shouldn’t be necessary.

sridhargaddam (approved):

yboaron (approved):

PR #2761, opened 2023-10-23
Allow halting on certificate errors

skitt:
When certificate errors are encountered, the fix is usually to restart the affected pod. To allow this to happen automatically, add a configuration setting for the gateway agent.

The setting is disabled by default; it will be enabled by default by the operator.

Depends on https://github.com/submariner-io/admiral/pull/767


skitt:
Let’s wait for https://github.com/submariner-io/admiral/pull/767 instead (if deemed appropriate).

sridhargaddam (approved):

tpantelis (approved):

Issue #2747, opened 2023-10-12
Improvements to the Gateway leader election


sridhargaddam:
While troubleshooting various issues with OCP, the following observations have been noted

The current configuration for Gateway leadership is as follows:
* defaultLeaseDuration = 10 seconds
* defaultRenewDeadline = 5 seconds
* defaultRetryPeriod = 2 seconds

During times of heavy load, when multiple nodes are labeled as gateway nodes, it has been observed that the active Gateway leader sometimes fails to renew its lease within the 5-second interval and is getting timedout.

```
E1009 15:52:02.920492       1 leaderelection.go:327] error retrieving resource lock submariner-operator/submariner-gateway-lock: Get "https://172.30.0.1:443/apis/coordination.k8s.io/v1/namespaces/submariner-operator/leases/submariner-gateway-lock": context deadline exceeded
I1009 15:52:02.920599       1 leaderelection.go:280] failed to renew lease submariner-operator/submariner-gateway-lock: timed out waiting for the condition
[90m2023-10-09T15:52:02.971Z[0m [33mDBG[0m ..ols/record/event.go:298 main                 Event(v1.ObjectReference{Kind:"ConfigMap", Namespace:"submariner-operator", Name:"submariner-gateway-lock", UID:"9a3de507-5553-4341-b55c-01969f1bf699", APIVersion:"v1", ResourceVersion:"33398175", FieldPath:""}): type: 'Normal' reason: 'LeaderElection' compute-1-submariner-gateway stopped leading
[90m2023-10-09T15:52:02.972Z[0m [33mDBG[0m ..ols/record/event.go:298 main                 Event(v1.ObjectReference{Kind:"Lease", Namespace:"submariner-operator", Name:"submariner-gateway-lock", UID:"caed3c1e-f5ee-4dbd-a8a1-e3ce708bf83c", APIVersion:"coordination.k8s.io/v1", ResourceVersion:"33398176", FieldPath:""}): type: 'Normal' reason: 'LeaderElection' compute-1-submariner-gateway stopped leading
[90m2023-10-09T15:52:03.395Z[0m [31mWRN[0m ..ne/syncer/syncer.go:184 GWSyncer             Deleted stale gateway: compute-2, didn't report for 15s
[90m2023-10-09T15:52:03.415Z[0m [32mINF[0m ..ne/syncer/syncer.go:290 GWSyncer             The Gateway entry for "compute-1" has been deleted
[...]

PR #2763, opened 2023-10-24
Improvements to the Gateway leader election

tpantelis:
When the leader lease is lost, we stop the K8s controllers but leave the cable driver in its current state so there's no disruption if/when leadership is re-acquired.

Related to https://github.com/submariner-io/submariner/issues/2747


skitt:
Could you add a reference to https://github.com/submariner-io/submariner/issues/2747 in the last commit message, as “explanation” for the new values?

test/e2e/redundancy/leader_election.go
```diff
@@ -0,0 +1,135 @@
...
+func updateLeaseUpdatePermission(cluster framework.ClusterIndex, ns string, remove bool) {
+	err := util.Update[*rbacv1.Role](context.Background(), resource.ForRole(framework.KubeClients[cluster], ns),
+		&rbacv1.Role{
+			ObjectMeta: metav1.ObjectMeta{Name: "submariner-gateway"},
+		},
+		func(role *rbacv1.Role) (*rbacv1.Role, error) {
+			for i := range role.Rules {
+				if role.Rules[i].APIGroups[0] == coordinationv1.GroupName && role.Rules[i].Resources[0] == "leases" {
+					if remove {
+						role.Rules[i].Verbs, _ = slices.Remove(role.Rules[i].Verbs, "update", func(e string) string {
```
skitt: How about passing in `slices.Remove` or `slices.AppendIfNotPresent` instead of a boolean? Both functions have the same signature.

tpantelis: yeah I could try that.

skitt (approved):

pkg/gateway/gateway.go
```diff
@@ -53,9 +55,11 @@ import (
 )
 
 const (
-	defaultLeaseDuration = 10 * time.Second
-	defaultRenewDeadline = 5 * time.Second
+	defaultLeaseDuration = 15 * time.Second
```
sridhargaddam, line 58: @yboaron to me these values look reasonable and are aligned with what K8s leaderElection code recommends. But if you have any concerns, do let us know.

pkg/gateway/gateway.go
```diff
@@ -258,50 +270,78 @@ func (g *gatewayType) startLeaderElection() error {
...
+	if g.cableHealthChecker != nil {
+		g.cableHealthChecker.Stop()
+	}
+
+	g.cableEngine.Stop()
+
+	logger.Info("Controllers stopped")
+
+	if err := g.gatewayPod.SetHALabels(context.Background(), subv1.HAStatusPassive); err != nil {
+		logger.Warningf("Error updating pod label: %s", err)
```
sridhargaddam:
```suggestion
		logger.Warningf("Error updating pod label to Passive: %s", err)
```

pkg/gateway/gateway.go
```diff
@@ -258,50 +270,78 @@ func (g *gatewayType) startLeaderElection() error {
...
-	go func() {
-		if err := g.gatewayPod.SetHALabels(subv1.HAStatusActive); err != nil {
-			g.fatalError <- errors.Wrap(err, "error updating pod label")
+	g.runAsync(g.leaderComponentsStarted, func() {
+		msgLogged := atomic.Bool{}
+
+		_ = wait.PollUntilContextCancel(ctx, 100*time.Millisecond, true, func(ctx context.Context) (bool, error) {
+			if err := g.gatewayPod.SetHALabels(ctx, subv1.HAStatusActive); err != nil {
+				if msgLogged.CompareAndSwap(false, true) {
+					logger.Warningf("Error updating pod label: %s", err)
```
sridhargaddam:
```suggestion
					logger.Warningf("Error updating pod label to Active: %s", err)
```

sridhargaddam (commented):
Thanks for the PR @tpantelis

sridhargaddam:
@tpantelis the following test seems to be failing in CI jobs.
```
Summarizing 1 Failure:
  [FAIL] Leader election tests when renewal of the leader lease fails [It] should re-acquire the leader lease after the failure is cleared
  /go/pkg/mod/github.com/submariner-io/shipyard@v0.16.0-m4.0.20231024090419-c555677f992b/test/e2e/framework/network_pods.go:196
```

tpantelis:
> the following test seems to be failing in CI jobs.

It fails with globalnet enabled. It tries to run a connectivity test while gateway leader election is lost. It seems the dataplane is somehow affected when globalnet is enabled. Any idea why?

tpantelis:
> > the following test seems to be failing in CI jobs.
> 
> It fails with globalnet enabled. It tries to run a connectivity test while gateway leader election is lost. It seems the dataplane is somehow affected when globalnet is enabled. Any idea why?

Oh it's b/c we have to run a different test with globalnet. I'll look into it. We could also skip this test with globalnet enabled.

tpantelis:
> > > the following test seems to be failing in CI jobs.
> > 
> > 
> > It fails with globalnet enabled. It tries to run a connectivity test while gateway leader election is lost. It seems the dataplane is somehow affected when globalnet is enabled. Any idea why?
> 
> Oh it's b/c we have to run a different test with globalnet. I'll look into it. We could also skip this test with globalnet enabled.

Actually just need to call `subFramework.VerifyDatapathConnectivity`.

yboaron (approved):

PR #2766, opened 2023-10-25
Delete Calico IPPools when Submariner uninstalled

### LINKED ISSUE #2764: Calico IPPools created by Submariner not deleted when Submariner uninstalled
**What happened**:

Calico IPPools created by Submariner (to enable cross-cluster traffic ) are not deleted when Submariner is uninstalled.

Submariner should delete these IPPools.

1. Verify that Calico API-server runs within cluster
2. Deploy Submariner (version > 0.16) 
3. Check for Calico IPPools created by Submariner
4. Uninstall Submariner

**Environment**:
- Install tools:
- Others:

yboaron:
All Calico IPPools created by Submariner should be deleted when Submariner is uninstalled.

This PR updates Calico IPPools handler Uninstall function to delete all IPPools regardless of isGateway field status.

Fixes: https://github.com/submariner-io/submariner/issues/2764

### DIFF
--- a/pkg/routeagent_driver/handlers/calico/ippool_handler.go
+++ b/pkg/routeagent_driver/handlers/calico/ippool_handler.go
@@ -129,10 +129,6 @@ func (h *calicoIPPoolHandler) TransitionToGateway() error {
 }
 
 func (h *calicoIPPoolHandler) Uninstall() error {
-	if !h.isGateway.Load() {
-		return nil
-	}
-
 	logger.Info("Uninstalling Calico IPPools used for Submariner")
 
 	labelSelector := labels.SelectorFromSet(map[string]string{submarinerIPPool: "true"}).String()


pkg/routeagent_driver/handlers/calico/ippool_handler.go
```diff
@@ -129,10 +129,6 @@ func (h *calicoIPPoolHandler) TransitionToGateway() error {
 }
 
 func (h *calicoIPPoolHandler) Uninstall() error {
-	if !h.isGateway.Load() {
```
sridhargaddam, line 132: The [bug report ](https://github.com/submariner-io/submariner/issues/2764) says that Calico IPPools are not deleted as part of Uninstall. 

@yboaron, this PR would ensure that every route-agent tries to delete the Calico IPPools. Even without this fix, the route-agent on the Gateway node should have deleted the Calico IPPools right?

yboaron: @sridhargaddam , also thought that with the original code the RA pod that runs on GW node should delete the Calico IPPools.
 
I think that the root cause is that 'TransitionToGateway' event not sent to Calico handler when we run the RA uninstall daemonset pod.

sridhargaddam: The Calico IPPool handler in route-agent uses the same handler framework as used by the other handlers, so ideally it should be notified of `TransitionToGateway` event. Are you seeing this [bug](https://github.com/submariner-io/submariner/issues/2764) consistently in your setup or was it a one-off case?

yboaron: I hit this bug consistently on my env,  even after I manually [edit RA clusterrole to enable deletecollection ](https://github.com/submariner-io/submariner-operator/pull/2856)

sridhargaddam: Then we should really understand why the routeAgent running on the Gateway node is unable to cleanup the IPPools. This PR might fix the issue by masking the original problem.

yboaron: >The Calico IPPool handler in route-agent uses the same handler framework as used by the other handlers, so ideally it should be notified of TransitionToGateway event

Q: Do you expect that Calico handler (or any other handlers) should be notified of TransitionToGateway event even for RA Uninstall DaemonSet ?

sridhargaddam: As discussed over slack, the TransitionToGateway would be fired during regular Gateway migrations which would update the isGateway status to True. However, during Uninstall,  an all-together new uninstall Pod would be created and this will not be aware of the Gateway status. Good catch Yossi.

sridhargaddam (approved):

skitt (approved):

skitt:
This needs to be backported, doesn’t it?

yboaron:
@skitt @sridhargaddam 
We need also [this PR](https://github.com/submariner-io/submariner-operator/pull/2856) to address [this issue](https://github.com/submariner-io/submariner/issues/2764)

PR #2793, opened 2023-11-06
use existing iptables APIs where possible

yboaron:
This PR updates
- mtuhandler to use UpdateChainRules API
- globalnet to use ChainExists API

### DIFF
--- a/pkg/globalnet/controllers/iptables/iface.go
+++ b/pkg/globalnet/controllers/iptables/iface.go
@@ -99,21 +99,6 @@ func (i *ipTables) RemoveClusterEgressRules(subnet, snatIP, globalNetIPTableMark
 	return nil
 }
 
-func (i *ipTables) ipTableChainExists(table, chain string) (bool, error) {
-	existingChains, err := i.ipt.ListChains(table)
-	if err != nil {
-		return false, errors.Wrapf(err, "error listing chains in IP table %q", table)
-	}
-
-	for _, val := range existingChains {
-		if val == chain {
-			return true, nil
-		}
-	}
-
-	return false, nil
-}
-
 func (i *ipTables) AddIngressRulesForHeadlessSvc(globalIP, ip string, targetType TargetType) error {
 	if globalIP == "" || ip == "" {
 		return fmt.Errorf("globalIP %q or %s IP %q cannot be empty", globalIP, targetType, ip)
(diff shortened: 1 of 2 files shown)


yboaron:
cc @sridhargaddam

pkg/globalnet/controllers/iptables/iface.go
```diff
@@ -100,18 +100,7 @@ func (i *ipTables) RemoveClusterEgressRules(subnet, snatIP, globalNetIPTableMark
...
-	}
-
-	for _, val := range existingChains {
-		if val == chain {
-			return true, nil
-		}
-	}
-
-	return false, nil
+	return (i.ipt.ChainExists(table, chain))
```
sridhargaddam:
```suggestion
	return i.ipt.ChainExists(table, chain)
```

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -309,34 +310,10 @@ func (h *mtuHandler) forceMssClamping(endpoint *submV1.Endpoint) error {
...
-		}
-	}
-
-	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecSource...); err != nil {
-		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecSource, " "))
-	}
-
-	if err := h.ipt.AppendUnique(constants.MangleTable, constants.SmPostRoutingChain, ruleSpecDest...); err != nil {
-		return errors.Wrapf(err, "error appending iptables rule %q", strings.Join(ruleSpecDest, " "))
+	if err := h.ipt.UpdateChainRules(constants.MangleTable, constants.SmPostRoutingChain, rules); err != nil {
```
skitt:
```suggestion
	if err := h.ipt.UpdateChainRules(constants.MangleTable, constants.SmPostRoutingChain, [][]string{ruleSpecSource, ruleSpecDest}); err != nil {
```

and drop the references to `rules` above; or if you prefer to keep the separate variable,

```suggestion
	rules := [][]string{ruleSpecSource, ruleSpecDest}

	if err := h.ipt.UpdateChainRules(constants.MangleTable, constants.SmPostRoutingChain, rules); err != nil {
```

and drop the declaration at the start of the function.

skitt (approved):
A minor tweak if you respin.

pkg/globalnet/controllers/iptables/iface.go
yboaron: I think we need the extra parentheses here,  to avoid  [1] lint error

[1] 
error returned from interface method should be wrapped: sig: func (github.com/submariner-io/submariner/pkg/iptables.Basic).ChainExists(table string, chain string) (bool, error) (wrapcheck)go-golangci-lint

skitt: Urgh, that’s not a great workaround — don’t hide linting errors, disable them with a nolint comment if appropriate or address them.

In this instance, it would be better to drop the `ipTableChainExists` function entirely, since it’s only used once; and wrap the error on return from `GetKubeProxyClusterIPServiceChainName` as is done in other functions.

sridhargaddam (approved):

PR #2802, opened 2023-11-15
Switch from stale app to stale GHA

dfarrell07:
The stale app has been deprecated and replaced with a GitHub Action.

probot/probot.github.io#376
probot/stale#385


skitt (approved):
Note for other reviewers: this uses the default delay before closing stale issues/PRs, 7 days.

dfarrell07:
> Note for other reviewers: this uses the default delay before closing stale issues/PRs, 7 days.

Right, which is the same we had manually configured before.

sridhargaddam (approved):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@b69b346013879cedbf50c69f572cd85439a41936
+        with:
+          days-before-issue-stale: 120
+          days-before-pr-stale: 14
+          exempt-issue-labels: 'confirmed,security'
+          exempt-pr-labels: 'confirmed,security'
+          stale-issue-label: 'stale'
```
tpantelis, line 25: Should we use `wontfix` as before?

dfarrell07: We can choose. Thought "stale" made more sense, was a more accurate description.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+          stale-issue-message: |
+            This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further
```
dfarrell07: I should change this to "This issue..."

.github/workflows/stale.yml
dfarrell07: I'll create "stale" labels as I send these changes to repos. Already created one for this repo. I'll also update the scrub filter accordingly. We can clear the old "wolntfix" ones during the next scrub calls.

tpantelis (approved):

PR #2803, opened 2023-11-15
Extend the event.Handler API to provide common state info to implementations

### LINKED ISSUE #2623: Extend the event.Handler API to provide common contextual info to implementations

Most of the `event.Handler` implementations store information across notifications, eg an `isGatewayNode` flag and a `remoteEndpoints` map, and protect them with a lock. We should store the common information in the events controller and provide it to implementations via a Context interface passed to the notifications methods.

**Why is this needed**:

Eliminates code duplication and locking in the implementations.

tpantelis:
See the individual for details. To summarize:

- Added `HandlerState` that contains global state common to all event handlers, specifically IsOnGateway and `GetRemoteEndpoints` which is provided to the handlers via a `SetState` method.
- Modified all handlers to use the `HandlerState` rather than maintaining such state themselves.
- The event controller now provides synchronization across all event callbacks so handlers don't have to.
- Added unit tests for handlers that didn't have any prior.

Fixes https://github.com/submariner-io/submariner/issues/2623

### DIFF
--- a/pkg/event/controller/controller.go
+++ b/pkg/event/controller/controller.go
@@ -49,7 +49,7 @@ type Controller struct {
 
 	handlers *event.Registry
 
-	syncMutex     *sync.Mutex
+	syncMutex     sync.Mutex
 	hostname      string
 	isGatewayNode bool
 }
@@ -84,9 +84,8 @@ func New(config *Config) (*Controller, error) {
 	}
 
 	ctl := Controller{
-		handlers:  config.Registry,
-		syncMutex: &sync.Mutex{},
-		hostname:  hostname,
+		handlers: config.Registry,
+		hostname: hostname,
 	}
 
 	err = envconfig.Process("submariner", &ctl.env)
--- a/pkg/event/controller/controller.go
+++ b/pkg/event/controller/controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"sync"
+	"sync/atomic"
 
 	"github.com/kelseyhightower/envconfig"
 	"github.com/pkg/errors"
@@ -43,15 +44,40 @@ type specification struct {
 	Namespace string
 }
 
+type handlerStateImpl struct {
+	isOnGateway     atomic.Bool
+	wasOnGateway    bool
+	remoteEndpoints sync.Map
+}
+
+func (s *handlerStateImpl) setIsOnGateway(v bool) {
+	s.isOnGateway.Store(v)
+}
+
+func (s *handlerStateImpl) IsOnGateway() bool {
+	return s.isOnGateway.Load()
+}
+
+func (s *handlerStateImpl) GetRemoteEndpoints() []subv1.Endpoint {
+	var endpoints []subv1.Endpoint
+
+	s.remoteEndpoints.Range(func(_, value any) bool {
+		endpoints = append(endpoints, *value.(*subv1.Endpoint))
+		return true
+	})
+
+	return endpoints
+}
+
 type Controller struct {
 	env             specification
 	resourceWatcher watcher.Interface
 
-	handlers *event.Registry
+	handlers     *event.Registry
+	handlerState handlerStateImpl
 
-	syncMutex     sync.Mutex
-	hostname      string
-	isGatewayNode bool
+	syncMutex sync.Mutex
+	hostname  string
 }
 
 // If the handler cannot recover from a failure, even after retrying for maximum requeue attempts,
@@ -137,6 +163,8 @@ func New(config *Config) (*Controller, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "error creating resource watcher")
 	}
+
+	ctl.handlers.SetHandlerState(&ctl.handlerState)
 
 	return &ctl, nil
 }
--- a/pkg/event/controller/endpoint_created.go
+++ b/pkg/event/controller/endpoint_created.go
@@ -51,24 +51,24 @@ func (c *Controller) handleCreatedEndpoint(obj runtime.Object, requeueCount int)
 }
 
 func (c *Controller) handleCreatedLocalEndpoint(endpoint *smv1.Endpoint) error {
-	if err := c.handlers.LocalEndpointCreated(endpoint); err != nil {
-		return err //nolint:wrapcheck  // Let the caller wrap it
-	}
-
 	if endpoint.Spec.Hostname == c.hostname {
-		// Verify if this node was a GatewayNode already. If not, it just transitioned to Gateway Node.
-		if !c.isGatewayNode {
-			if err := c.handlers.TransitionToGateway(); err != nil {
-				return err //nolint:wrapcheck  // Let the caller wrap it
-			}
-		}
-
-		c.isGatewayNode = true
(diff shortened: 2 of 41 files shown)


aswinsuryan (approved):

pkg/event/controller/controller.go
```diff
@@ -43,15 +44,36 @@ type specification struct {
 	Namespace string
 }
 
+type handlerStateType struct {
```
skitt: I found this odd on first read — `Type` suggests this describes a type, but it’s really the anchor for an implementation of `HandlerState`, so how about `handlerStateImpl`?

pkg/event/controller/endpoint_created.go
```diff
@@ -51,24 +51,24 @@ func (c *Controller) handleCreatedEndpoint(obj runtime.Object, requeueCount int)
 }
 
 func (c *Controller) handleCreatedLocalEndpoint(endpoint *smv1.Endpoint) error {
-	if err := c.handlers.LocalEndpointCreated(endpoint); err != nil {
-		return err //nolint:wrapcheck  // Let the caller wrap it
+	if endpoint.Spec.Hostname == c.hostname {
+		c.handlerState.isOnGateway.Store(true)
```
skitt: Reads go through `IsOnGateway()`, so it feels unbalanced to directly access the atomic here.

pkg/event/handler.go
```diff
@@ -25,10 +25,28 @@ import (
...
+func (c *DefaultHandlerState) GetRemoteEndpoints() []submV1.Endpoint {
+	return nil
+}
+
 type Handler interface {
 	// Init is called once on startup to let the handler initialize any state it needs.
 	Init() error
 
+	// SetHandlerState is called once on startup after Init with the HandlerState that can be used to access global data from event callbacks.
+	SetState(handlerCtx HandlerState)
```
skitt, line 48: Is it a context, or a state?

skitt: Other than the amount of churn involved, is there a reason to have a separate `SetState` method instead of adding the handler state to `Init`?

Currently the handlers are set up when the registry is constructed, and then controller construction sets the shared handler state. But AFAICT nothing would stop a handler being added after that, and then it wouldn’t be given the shared state...

pkg/routeagent_driver/handlers/kubeproxy/node_handler.go
```diff
@@ -73,9 +63,9 @@ func (kp *SyncHandler) populateRemoteVtepIps(vtepIP string, operation Operation)
 	}
 
 	logger.V(log.DEBUG).Infof("populateRemoteVtepIps is called with vtepIP %s, isGatewayNode %t",
-		vtepIP, kp.isGatewayNode)
+		vtepIP, kp.State().IsOnGateway())
 
-	if kp.isGatewayNode {
+	if kp.State().IsOnGateway() {
```
skitt: This suggests a small (pre-existing) race here, but it’s protected by the lock. However to make the correctness more obvious it would be better to store the current return value of `IsOnGateway()` and use that both for logging and decision-marking (so it’s apparent that the branch taken matches the log message).

skitt (changes requested):

pkg/event/controller/controller.go
tpantelis: We could just name it `handlerState` but it's nice to distinguish it from variables of the same name.  I've seen the use of `Type` in K8s code so I borrowed it (I've used it before as well) - it is actually describing the type. `Impl` seems Java-ish but I can change it - I'm fine either way (or name it `handlerState`).

pkg/event/controller/endpoint_created.go
tpantelis: `isOnGateway` is private so it really doesn't matter either way. But to balance it we can either add a private function to wrap `isOnGateway.Store` or not use `IsOnGateway()` internally.

pkg/event/handler.go
tpantelis: > Is it a context, or a state?

I went back and forth on that. I finally went with state b/c context is usually used per function and this state persists.

I thought about passing it via `Init` but that is called before the controller is initialized and it's the controller that owns/implements the state. The registry and controller are completely decoupled. But we could have the registry construct and start the controller - we'd have to have a public method on the controller to get the `HandlerState` but that seemed awkward, unless we have the controller called `Init`. Also, each `Handler` impl would have to explicitly store the state, presumably call `SetState` on the `HandlerBase`, but this seemed awkward to me. That's why I went with this approach.

> But AFAICT nothing would stop a handler being added after that, and then it wouldn’t be given the shared state...

You're right but that would problematic b/c `eventHandlers` in `Registry` is not synchronized. The `Registry` is designed to add all the handlers first before starting the controller. We could force this by passing all the handlers via `NewRegistry` and removing `AddHandlers`. I can do that in a followup.

pkg/routeagent_driver/handlers/kubeproxy/node_handler.go
tpantelis: sure, for readability.

pkg/event/controller/controller.go
skitt: Right, I didn’t bother mentioning `handlerState` because of the variable name conflict! I prefer `Impl` personally, especially as more type parameters start appearing in various places. The naming in Kubernetes seems to be shifting too, at least in discussions I’ve been having recently people talk about `Impl`, and it’s about 50/50 now.

pkg/event/controller/endpoint_created.go
skitt: I agree it’s not important, but a private function would be good (not exposing the atomic for reads — I’d rather keep this encapsulated so we can change the implementation easily).

pkg/event/handler.go
skitt: Right, thanks for the explanation, that makes sense. I like the idea of dropping `AddHandlers`.

tpantelis: > I like the idea of dropping AddHandlers

https://github.com/submariner-io/submariner/pull/2803/commits/57cdf7ef9f5a2993a6e7a10b1bc4ffd1ec2bbb22

skitt:
```suggestion
	SetState(handlerState HandlerState)
```

if you respin this.

pkg/routeagent_driver/handlers/ovn/fake/ovsdb_client.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/ovn-org/libovsdb/cache"
+	libovsdbclient "github.com/ovn-org/libovsdb/client"
+	"github.com/ovn-org/libovsdb/model"
+	"github.com/ovn-org/libovsdb/ovsdb"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
+	"github.com/submariner-io/admiral/pkg/resource"
+	"github.com/submariner-io/admiral/pkg/slices"
+)
+
+type OVSDBClient struct {
```
skitt, line 38: Writing OVSDB test code, that takes us back!

skitt (approved):

PR #2809, opened 2023-11-22
Exclude code from SonarCloud analysis

tpantelis:
- Generated code
- Test support
- External library wrappers


skitt (approved):

sonar-project.properties
```diff
@@ -2,7 +2,7 @@ sonar.projectKey=submariner-io_submariner
 sonar.projectName=submariner
 sonar.organization=submariner-io
 sonar.sources=.
-sonar.exclusions=**/vendor/**,**/*_test.go,**/test/**,**/fake/**,**/client/**
+sonar.exclusions=**/vendor/**,**/*_test.go,**/test/**,**/fake/**,**/client/**,**/natdiscovery/proto/natdiscovery.pb.go,**/netlink/**,**/event/testing/**,**/zz_generated.deepcopy.go,**/ipset/**
```
sridhargaddam, line 5: Is the removal of ipset and netlink packages intentional?

skitt: IIUC, yes, because they are external library wrappers.

sridhargaddam: Not all APIs are passthrough, we have some APIs implemented. Do we want to avoid sonar on them?

tpantelis: `netlink` is a straight pass-through. `ipset` is not our code - we copied it from somewhere. There might be some helper methods we could unit test but I don't think it's necessary. I think the E2E tests it sufficiently, at least the code we use. @sridhargaddam What API are you specifically concerned about?

tpantelis: Actually, FWICT, virtually all helpers, eg validate*, are private.

sridhargaddam: Right, there are few APIs like Create/Add/Del/Destroy etc, but the code is pretty much stable and has not undergone any major changes, so we can skip it. I wanted to confirm that we did not accidentally remove them.

sridhargaddam (approved):

PR #2830, opened 2023-12-11
Don't use extra-toggles for Globalnet GHA config

dfarrell07:
In Shipyard, Globalnet GHA config has a dedicated field. There is no need to pass it with extra-toggles. Doing so makes the config in this repo unnecessary different from other repos and harder to maintain.

This should not change the matrix of tested combinations.


dfarrell07:
Not sure why Libreswan is still explicitly listed...

PR #2831, opened 2023-12-11
Standardize and simplify E2E GHA config 

dfarrell07:
Make the E2E-related GHA config easier to maintain and reason about. See commit messages for details.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -44,7 +44,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


dfarrell07:
I think this will have to go with the K8s version change, as we can't have an all-defaults E2E job in the e2e-full workflow as there's an assumption that will be the name of the default E2E job related to both workflows being in the same file so they can both depend on the images step. This results in our current config duplicating a job, the `E2E (libreswan)` run and the `End to End / E2E` run.

PR #2832, opened 2023-12-11
Eliminate duplicate all-defaults E2E run, add K8s E2E, standardize GHA config

dfarrell07:
Set of changes that results in:

* Eliminate duplication between `E2E (libreswan)` and `End to End / E2E` runs.
* Add E2E for top and bottom of supported K8s versions range, as well as the oldest-working K8s version. The oldest-working coverage currently in the Operator repo will be removed.
* Standardize and simplify GHA config, to make it more like the other repos and therefore easier to maintain and reason about.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -44,7 +44,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cable-driver: ['libreswan', 'wireguard', 'vxlan']
+        cable-driver: ['', 'wireguard', 'vxlan']
         extra-toggles: ['', 'ovn']
         globalnet: ['', 'globalnet']
         exclude:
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

.github/workflows/e2e.yml
```diff
@@ -44,15 +44,24 @@ jobs:
...
+        extra-toggles: ['', 'ovn']
+        globalnet: ['', 'globalnet']
+        # Run most tests against the latest K8s version
+        k8s_version: ['1.28']
         exclude:
           - cable-driver: wireguard
             extra-toggles: ovn
         include:
           - extra-toggles: dual-stack
           - extra-toggles: dual-stack, globalnet
```
skitt, line 57: Doesn’t this need to be changed to match the extraction of Globalnet from `extra-toggles` above?

skitt (approved):

PR #2837, opened 2023-12-15
Add monitor for rp_filter changes

aswinsuryan:
In some of the baremetal setup, the rp_filter was getting reset to strick to loose for vx_submariner interface. This PR monitors the value periodically and resets the value.


aswinsuryan:
It was decided not to add a monitor. As the issue was not observed only in a particular setup and in not any other.

PR #2867, opened 2024-01-06
Update Gateway on local Endpoint public IP update

### LINKED ISSUE #1871: Sometimes gateway object is not updated when using LB mode
**What happened**:
When using Loadbalancer mode of Submariner on AWS Cluster (where the public-ip associated with the LB Service changes), it was seen that local endpoint is properly updated with the correct public-ip, however the public-ip in the Gateway object is not updated. Because of this, `subctl show ...` command displays wrong public-ip to the users.

From gateway object:

```
    haStatus: active
    localEndpoint:
      backend: libreswan
      backend_config:
        natt-discovery-port: "4490"
        preferred-server: "true"
        preferred-server-timestamp: "1654881267"
        udp-port: "4500"
        using-loadbalancer: "true"
      cable_name: submariner-cable-sgaddam-aws-spoke1-10-0-129-37
      cluster_id: sgaddam-aws-spoke1
      healthCheckIP: 10.135.0.1
      hostname: ip-10-0-129-37
      nat_enabled: true
      private_ip: 10.0.129.37
      public_ip: 3.128.47.166                          <--- Old public-ip
      subnets:
      - 172.32.0.0/16
      - 10.132.0.0/14
    statusFailure: ""
    version: devel-f55a4d1
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
[sgaddam@localhost aws-ocp]$

```
Although there is no issue for the actual connection because the local endpoint is properly updated and sync'ed to the Broker, the information displayed to the user as part of gateway object (and subctl show command) could be misleading to the end users.

tpantelis:
When it's detected that the public IP has changed, the local `Endpoint` resource is properly updated however the `Gateway` resource is not and retains the old IP. This is because the `Gateway` syncer is initialized with the local `SubmarinerEndpoint` instance on startup and assumes it never changes.

Introduce an `endpoint.Local` struct that encapsulates the local `EndpointSpec` and allows updating. Share this instance with the various components rather than passing the static `SubmarinerEndpoint`.

See the individual commits for details.

Fixes #1871

### DIFF
--- a/pkg/apis/submariner.io/v1/types.go
+++ b/pkg/apis/submariner.io/v1/types.go
@@ -24,6 +24,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 // +genclient
@@ -391,4 +392,16 @@ type RoutePolicySpec struct {
 
 	// Specifies the remote CIDRs available via the next hop
 	RemoteCIDRs []string `json:"remoteCIDRs"`
+}
+
+var EndpointGVR = schema.GroupVersionResource{
+	Group:    SchemeGroupVersion.Group,
+	Version:  SchemeGroupVersion.Version,
+	Resource: "endpoints",
+}
+
+var ClusterGVR = schema.GroupVersionResource{
+	Group:    SchemeGroupVersion.Group,
+	Version:  SchemeGroupVersion.Version,
+	Resource: "clusters",
 }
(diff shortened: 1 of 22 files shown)


dfarrell07 (approved):

sridhargaddam (approved):

sridhargaddam:
@tpantelis what are your thoughts on backporting this PR to other branches?

tpantelis:
> @tpantelis what are your thoughts on backporting this PR to other branches?

I'm not sure how critical this is. If you think it is then sure.

sridhargaddam:
> > @tpantelis what are your thoughts on backporting this PR to other branches?
> 
> I'm not sure how critical this is. If you think it is then sure.

AFAIU this does not affect any functionality, so we can avoid the backport.

PR #2886, opened 2024-01-25
Packetfilter: add support for namedset

yboaron:
Add support for generic namedSet APIs under packetfilter.


pkg/packetfilter/packetfilter.go
```diff
@@ -138,6 +136,37 @@ type ChainIPHook struct {
...
+	// SetType specifies the named type.
+	SetType string
+	// nftables named set attached to tables
+	Table TableType
+	// SetFamily specifies the protocol family of the IP addresses to be stored in the set.
+	// The default is IPv4.
+	Family SetFamily
+}
+
+type Named interface {
```
tpantelis: I think this should be called `NamedSet` as this is what consumers will store and use continuously.  The struct above is referenced once as input so I suggest `SetConfig` or `SetInfo`.

```suggestion
type NamedSet interface {
```

pkg/packetfilter/packetfilter.go
```diff
@@ -153,6 +182,11 @@ type Driver interface {
 	List(table TableType, chain string) ([]*Rule, error)
 	Append(table TableType, chain string, rule *Rule) error
 	Insert(table TableType, chain string, pos int, rule *Rule) error
+
+	// named Sets
+	NewNamed(set *NamedSet) Named
```
tpantelis:
```suggestion
	NewNamedSet(set *NamedSet) Named
```

pkg/packetfilter/packetfilter.go
```diff
@@ -153,6 +182,11 @@ type Driver interface {
...
+	DestroySet(set string) error
```
tpantelis: We really don't need this method. It's used in [one place](https://github.com/submariner-io/submariner/blob/636b457da21542ec3a7632143562c220c57e4f55/pkg/globalnet/controllers/uninstall.go#L106) to destroy all sets with a name prefix but the consumer can create a `NamedSet` and call `Destroy`. But I think a better alternative is to replace `ListSets` with a general `DestroySets` that takes a name filter:

`DestroySets(nameFilter func(string) bool) error`

Ideally `ListSets` would return a `NamedSet` but the ipset lib doesn't have a way to retrieve a set's params after creation so perhaps it's best not to expose just set names.

pkg/packetfilter/iptables/iptables.go
```diff
@@ -52,7 +54,8 @@ var (
 )
 
 type PacketFilter struct {
-	ipt *iptables.IPTables
+	ipt        *iptables.IPTables
+	IPSetIface ipset.Interface
```
tpantelis: Let's not export this.

```suggestion
	ipSetIface ipset.Interface
```

pkg/packetfilter/iptables/namedset.go
```diff
@@ -0,0 +1,47 @@
...
+	hashFamily := ipset.ProtocolFamilyIPV4
+	if set.Family == packetfilter.SetFamilyV6 {
+		hashFamily = ipset.ProtocolFamilyIPV6
+	}
+
+	return ipset.NewNamed(&ipset.IPSet{
+		Name:       set.Name,
+		SetType:    ipset.HashNet,
+		HashFamily: hashFamily,
+	}, a.IPSetIface)
```
tpantelis: I think we should move the `Named` implementation in ipset here as we expect consumers to use the API in packetfilter. We could move all the ipset code here as well but I think it makes sense to keep that in its own package as we inherited it from somewhere else.

pkg/packetfilter/fake/namedset.go
```diff
@@ -0,0 +1,335 @@
...
+	"strings"
+	"time"
+
+	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
+	"github.com/submariner-io/submariner/pkg/packetfilter"
+	"k8s.io/utils/set"
+)
+
+type setInterface interface {
```
tpantelis: I'm not clear on the purpose of this local interface. This seems to be patterned after `ipset.Interface` but we don't need all these methods.

pkg/packetfilter/fake/namedset.go
```diff
@@ -0,0 +1,335 @@
...
+	TestEntry(entry string, set string) (bool, error)
+	// ListEntries lists all the entries from a named set
+	ListEntries(set string) ([]string, error)
+	// ListSets list all set names from kernel
+	ListSets() ([]string, error)
+}
+
+type namedType struct {
+	nSet  *packetfilter.NamedSet
+	iface setInterface
```
tpantelis: Just use `PacketFilter` directly.

```suggestion
	pfilter *PacketFilter
```

tpantelis (changes requested):

pkg/packetfilter/fake/namedset.go
```diff
@@ -0,0 +1,307 @@
...
+	"strings"
+	"time"
+
+	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
+	"github.com/submariner-io/submariner/pkg/packetfilter"
+	"k8s.io/utils/set"
+)
+
+type namedType struct {
```
tpantelis: Minor thing but more clear.

```suggestion
type namedSet struct {
```

pkg/packetfilter/fake/namedset.go
```diff
@@ -0,0 +1,307 @@
...
+	return n.setInfo.Name
+}
+
+func (n *namedType) Flush() error {
+	n.pfilter.flushSet(n.setInfo.Name)
+	return nil
+}
+
+func (n *namedType) Destroy() error {
+	return errors.Wrap(n.pfilter.destroySet(n.setInfo.Name), "Destroy failed")
```
tpantelis: Minor but we should need to wrap here nor in the methods below.

```suggestion
	return n.pfilter.destroySet(n.setInfo.Name)
```

pkg/packetfilter/fake/namedset.go
```diff
@@ -0,0 +1,307 @@
...
+		return err
+	}
+
+	if i.sets[setName] == nil {
+		return nil
+	}
+
+	delete(i.sets, setName)
+
+	return nil
```
tpantelis: `delete` handles nil

```suggestion
	delete(i.sets, setName)

	return nil
```

pkg/packetfilter/fake/namedset.go
```diff
@@ -0,0 +1,307 @@
...
+	for _, setName := range names {
+		if nameFilter(setName) {
+			err := matchForError(&i.failOnDestroySetMatchers, setName)
+			if err != nil {
+				return err
+			}
+
+			if i.sets[setName] != nil {
+				delete(i.sets, setName)
+			}
```
tpantelis:
```suggestion
			delete(i.sets, setName)
```

pkg/packetfilter/iptables/namedset.go
```diff
@@ -0,0 +1,101 @@
...
+
+package iptables
+
+import (
+	"github.com/pkg/errors"
+	"github.com/submariner-io/submariner/pkg/ipset"
+	"github.com/submariner-io/submariner/pkg/packetfilter"
+)
+
+type namedType struct {
```
tpantelis:
```suggestion
type namedSet struct {
```

tpantelis (changes requested):

tpantelis (approved):

yboaron:
cc @sridhargaddam

pkg/packetfilter/fake/namedset.go
sridhargaddam: Its just a minor thing, but just want to check if you wanted to use namedSet here? (as the conversation is marked as resolved).

pkg/packetfilter/fake/namedset.go
```diff
@@ -0,0 +1,300 @@
...
+	defer i.mutex.Unlock()
+
+	err := matchForError(&i.failOnAddEntryMatchers, entry)
+	if err != nil {
+		return err
+	}
+
+	entries := i.sets[nSet.Name]
+	if entries == nil {
+		return fmt.Errorf("IP set %q does not exist", nSet.Name)
```
sridhargaddam:
```suggestion
		return fmt.Errorf("named set %q does not exist", nSet.Name)
```

pkg/packetfilter/fake/namedset.go
yboaron: > Its just a minor thing, but just want to check if you wanted to use namedSet here? (as the conversation is marked as resolved).

Ooops,   missed it. 
thanks

sridhargaddam (approved):

PR #2924, opened 2024-02-28
Adjustments to the fake packetfilter implementation

tpantelis:

### DIFF
--- a/pkg/packetfilter/fake/packetfilter.go
+++ b/pkg/packetfilter/fake/packetfilter.go
@@ -32,7 +32,7 @@ import (
 type PacketFilter struct {
 	mutex                    sync.Mutex
 	chainRules               map[string]set.Set[string]
-	tableChains              map[packetfilter.TableType]set.Set[string]
+	tableChains              map[uint32]set.Set[string]
 	failOnAppendRuleMatchers []interface{}
 	failOnDeleteRuleMatchers []interface{}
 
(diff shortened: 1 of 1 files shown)


pkg/packetfilter/fake/packetfilter.go
```diff
@@ -67,117 +58,90 @@ func New() *PacketFilter {
...
-	if err != nil {
-		return errors.Wrap(err, "Delete failed")
-	}
-
-	return i.delete(table, chain, string(ruleSpecStr))
+	return i.delete(table, chain, toRuleString(rule))
 }
 
-func fromRuleSpec(spec string) *packetfilter.Rule {
+func fromRuleSting(str string) *packetfilter.Rule {
```
yboaron: nit - fromRuleString

tpantelis: ugh - thanks for catching that.

pkg/packetfilter/fake/packetfilter.go
```diff
@@ -67,117 +58,90 @@ func New() *PacketFilter {
...
 func (i *PacketFilter) DeleteChain(table packetfilter.TableType, chain string) error {
-	i.deleteChain(table, chain)
+	i.deleteChain(uint32(table), chain)
 
 	return nil
 }
 
 func (i *PacketFilter) ClearChain(table packetfilter.TableType, chain string) error {
-	i.clearChain(table, chain)
+	i.mutex.Lock()
```
yboaron, line 89: Overall code LGTM, 
I have one comment, can't we call deleteChain here?

yboaron (approved):

pkg/packetfilter/fake/packetfilter.go
tpantelis: yeah so `DeleteChain` and `ClearChain` are the same in this implementation. I was wondering about that too. It seems `ClearChain` should just remove all the rules associated to the chain? If so then I can change that in a separate PR since I didn't change this functionality with this PR.

PR #2925, opened 2024-02-28
Adjustments to the iptables packetfilter implementation

tpantelis:
Minor adjustments and cleanup. See individual commits for details.

### DIFF
--- a/pkg/packetfilter/fake/packetfilter.go
+++ b/pkg/packetfilter/fake/packetfilter.go
@@ -44,12 +44,12 @@ type PacketFilter struct {
 }
 
 var (
-	iphookChainTypeToTableType = [packetfilter.ChainTypeMAX]packetfilter.TableType{
+	iphookChainTypeToTableType = []packetfilter.TableType{
 		packetfilter.TableTypeFilter,
 		packetfilter.TableTypeRoute,
 		packetfilter.TableTypeNAT,
 	}
-	chainHookToStr = [packetfilter.ChainHookMAX]string{"PREROUTING", "INPUT", "FORWARD", "OUTPUT", "POSTROUTING"}
+	chainHookToStr = []string{"PREROUTING", "INPUT", "FORWARD", "OUTPUT", "POSTROUTING"}
 )
 
 func New() *PacketFilter {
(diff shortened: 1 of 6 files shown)


pkg/packetfilter/fake/packetfilter.go
```diff
@@ -44,12 +44,12 @@ type PacketFilter struct {
 }
 
 var (
-	iphookChainTypeToTableType = [packetfilter.ChainTypeMAX]packetfilter.TableType{
+	iphookChainTypeToTableType = []packetfilter.TableType{
```
yboaron, line 47: maybe we should use map also here (same as your suggested change for pkg/packetfilter/iptables/iptables.go) ?

pkg/packetfilter/iptables/iptables.go
```diff
@@ -327,9 +307,7 @@ func ToRuleSpec(rule *packetfilter.Rule) ([]string, error) {
 		ruleSpec = append(ruleSpec, "--set-mark", rule.MarkValue)
 	}
 
-	logger.Infof("ToRuleSpec: %s", strings.Join(ruleSpec, " "))
```
yboaron, line 330: I think it would be useful to print ruleSpec here for debugging

pkg/packetfilter/fake/packetfilter.go
tpantelis: I didn't change this one b/c I actually remove it with https://github.com/submariner-io/submariner/pull/2924 😄

pkg/packetfilter/iptables/iptables.go
tpantelis: OK - we can leave it in as TRACE - I think as just Info level it might fill logs too much. Or perhaps we should log the spec when something fails...

yboaron: Let's keep it as TRACE , to make sure we didn't miss something in packetfilter rule to iptables rule translation

tpantelis: I pushed commit https://github.com/submariner-io/submariner/pull/2925/commits/a0c44f52a5d9021e1450ad9922269ec7a4a533c6.  I put the TRACE log back in and also added the rule spec string to the error logging.

tpantelis: Also added a `String` method for `Rule`: https://github.com/submariner-io/submariner/pull/2925/commits/21e672b4c73a32d67d9ce62acca782a2ea91fec2.

yboaron (approved):

vthapar (approved):

PR #2933, opened 2024-03-05
Adjust chain rule implementation in fake packetfilter

tpantelis:
The purpose of `ClearChain` is to clear/flush all rules for a table chain. The fake implementation was modified to do this, previously it deleted the entire chain, ie it was equivalent to `DeleteChain`. To facilitate this, the `tableChains` map, which just tracked the existence of chains, was removed. The `chainRules` map can also be used for this purpose.

In addition, inserting/appending a rule now fails if the chain doesn't exist to align with the behavior of the real iptables implementation. This required adjustments to the globalnet unit tests.

### DIFF
--- a/pkg/globalnet/controllers/controllers_suite_test.go
+++ b/pkg/globalnet/controllers/controllers_suite_test.go
@@ -142,6 +142,24 @@ func (t *testDriverBase) afterEach() {
 	t.controller.Stop()
 }
 
+func (t *testDriverBase) initChains() {
+	for _, chain := range []string{
+		constants.SmGlobalnetIngressChain,
+		constants.SmGlobalnetEgressChain,
+		constants.SmGlobalnetEgressChainForPods,
+		constants.SmGlobalnetEgressChainForHeadlessSvcPods,
+		constants.SmGlobalnetEgressChainForHeadlessSvcEPs,
+		constants.SmGlobalnetEgressChainForNamespace,
+		constants.SmGlobalnetEgressChainForCluster,
+		routeAgent.SmPostRoutingChain,
+		constants.SmGlobalnetMarkChain,
+	} {
+		Expect(t.pFilter.CreateChainIfNotExists(packetfilter.TableTypeNAT, &packetfilter.Chain{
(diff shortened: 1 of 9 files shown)


pkg/globalnet/controllers/controllers_suite_test.go
```diff
@@ -142,6 +142,24 @@ func (t *testDriverBase) afterEach() {
 	t.controller.Stop()
 }
 
+func (t *testDriverBase) initChains() {
```
yboaron, line 145: Agree that with this change tests code is aligned with real nftables/iptables behavior.

the new initChains function should be synced with [GN chains creation function](https://github.com/submariner-io/submariner/blob/d405b10dad8dc7b973435ef3ce70163bc6a78f9f/pkg/globalnet/controllers/gateway_monitor.go#L454),  if we can't run 'real' GN chains creation function here, maybe we should at least add a comment saying that this function should be synced with real function

tpantelis: I'm not sure what you mean by synced with but I think you're concern is re: not running the real function. That is in the `gatewayMonitor` which is tested in _gateway_monitor_test.go_. In those tests we do not call `initChains` and also verify that the chains are created by the `gatewayMonitor`.

yboaron: Yep, was concerned that future changes to gateway_monitor createGlobalnetChains  or fake packetfilter implementation (e.g: different implementation for CreateIPHookChainIfNotExists and CreateChainIfNotExists) will break testing.

On the other hand, since we don't expect changes in GN chains, maybe we can leave it as you suggested

tpantelis: yeah if `createGlobalnetChains` is changed, eg add a new chain, then the unit tests would need to be adjusted accordingly but that should be expected.

yboaron (approved):

yboaron:
cc @vthapar

skitt (approved):

PR #2947, opened 2024-03-21
Use proxy vars

### LINKED ISSUE #2913: Support PROXY vars on Gateway Pod
Support configuring Submariner GW pods for proxy access. this includes setting Proxy env vars in gw pods and using those envs to use proxy for gw connections.

**Why is this needed**:
In a OCP4 cluster behind a proxy with a global proxy configured the submariner operator pod gets NO_PROXY, HTTP_PROXY, and HTTPS_PROXY env vars set. The gateway pod does not. There doesn't seem to be a set environment vars in the operator`kind: Submariner` yaml either.

As the gateway pod isn't configured for proxy access or you can't configure it we get a bunch of time out errors.

vthapar:
Use proxy vars in Gateway pods for gw connections if available

Fixes #2913

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/natdiscovery"
 	"github.com/submariner-io/submariner/pkg/types"
 	"github.com/submariner-io/submariner/pkg/versions"
+	"golang.org/x/net/http/httpproxy"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -91,6 +92,10 @@ func main() {
 	logger.FatalOnError(envconfig.Process("submariner", &submSpec), "Error processing env vars")
 
 	logger.Infof("Parsed env variables: %#v", submSpec)
+
+	proxyEnv := httpproxy.FromEnvironment()
+	logger.Infof("Proxy env variables: HTTP_PROXY: %v, HTTPS_PROXY: %v, NO_PROXY: %v",
+		proxyEnv.HTTPProxy, proxyEnv.HTTPSProxy, proxyEnv.NoProxy)
 
 	defer http.StartServer(http.Metrics|http.Profile, submSpec.MetricsPort)()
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -91,6 +92,8 @@ func main() {
 	logger.FatalOnError(envconfig.Process("submariner", &submSpec), "Error processing env vars")
 
 	logger.Infof("Parsed env variables: %#v", submSpec)
+	logger.Infof("Proxy env variables: HTTP_PROXY: %v, HTTPS_PROXY: %v, NO_PROXY: %v", os.Getenv("HTTP_PROXY"),
```
skitt: `ProxyFromEnvironment` also checks the lowercase variants, see https://cs.opensource.google/go/x/net/+/refs/tags/v0.22.0:http/httpproxy/proxy.go;l=89 — `http_proxy`, `https_proxy`, and `no_proxy`. A more reliable way of logging the relevant info would be to call `FromEnvironment()` in `golang.org/x/net/http/httpproxy` and dump the `HTTPProxy` etc. fields in the returned `Config` struct.

skitt (changes requested):

main.go
```diff
@@ -20,6 +20,7 @@ package main
 
 import (
 	"flag"
+	"os"
```
skitt: This shouldn’t be necessary (but Go linting should catch it).

skitt (approved):

main.go
vthapar: Was having some problems with IDE, it kept freezing on trying to save so had couple commits without changes I wanted.

tpantelis (approved):

adibrastegarnia:
How do you set them in pod? I don't think submariner operator is modified to set those enviroment variables? Am I missing something?

skitt:
@adibrastegarnia once https://github.com/submariner-io/submariner-operator/pull/3013 is merged you’ll be able to set them on the operator and have them propagated to managed pods.

dfarrell07:
Moving release note tracking to the user-facing change, https://github.com/submariner-io/subctl/pull/1103.

PR #2956, opened 2024-04-01
Add nftables backend driver to packetfilter

yboaron:
This PR adds nftables implementation to packetfilter, 
after this change iptables still remains as packetfilter default driver.


dfarrell07 (approved):

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+func (r RuleSpec) String() string {
+	return strings.Join(r, " ")
+}
+
+type packetFilter struct {
+	nftables knftables.Interface
+}
+
+func New() (packetfilter.Driver, error) {
+	nft, err := knftables.New(knftables.IPv4Family, submarinerTable)
```
skitt, line 83: `namedset.go` supports IPv6 (at least, appears to), but this hard-codes IPv4; is that intentional?

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating Knftables")
```
skitt:
```suggestion
		return nil, errors.Wrap(err, "error creating knftables")
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+		nftables: nft,
+	}, nil
+}
+
+func (p *packetFilter) ChainExists(_ packetfilter.TableType, chain string) (bool, error) {
+	return p.chainExists(chain)
+}
+
+func (p *packetFilter) chainExists(chain string) (bool, error) {
+	chainsList, err := p.nftables.List(context.Background(), "chains")
```
skitt:
```suggestion
	chainsList, err := p.nftables.List(context.TODO(), "chains")
```

go.mod
```diff
@@ -27,6 +28,7 @@ require (
 	k8s.io/component-helpers v0.29.2
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
 	sigs.k8s.io/controller-runtime v0.17.2
+	sigs.k8s.io/knftables v0.0.15
```
skitt:
```suggestion
	sigs.k8s.io/knftables v0.0.16
```

(released after you submitted this).

skitt (changes requested):

pkg/packetfilter/nftables/nftables.go
yboaron: Yep, we have the same issue also with iptables , [iptables hardcoded to v4](https://github.com/submariner-io/submariner/blob/devel/pkg/packetfilter/iptables/iptables.go#L77) while iptables/namedset supports also [v6](https://github.com/submariner-io/submariner/blob/devel/pkg/packetfilter/iptables/namedset.go#L33)  .

I'll delete https://github.com/submariner-io/submariner/blob/devel/pkg/packetfilter/packetfilter.go#L257 and update the relevant code.

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+
+	for _, existingElem := range nftablesElems {
+		strElems = append(strElems, existingElem.Key[0])
+	}
+
+	return strElems, nil
+}
+
+func (p *packetFilter) DestroySets(nameFilter func(string) bool) error {
+	setsList, err := p.nftables.List(context.Background(), "sets")
```
skitt:
```suggestion
	setsList, err := p.nftables.List(context.TODO(), "sets")
```

skitt (changes requested):
Apologies, I missed another `context.Background()`.

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+package nftables
+
+import (
+	"context"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/log"
+	"github.com/submariner-io/submariner/pkg/packetfilter"
+	"golang.org/x/exp/slices"
```
skitt: Please use plain `"slices"` instead.

skitt (changes requested):

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+
+	err := n.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error deleting entry to set")
+}
+
+func (n *namedSet) ListEntries() ([]string, error) {
+	nftablesElems, err := n.nftables.ListElements(context.TODO(), "set", n.set.Name)
+	if err != nil {
+		return []string{}, errors.Wrap(err, "error listings elements")
```
tpantelis:
```suggestion
		return []string{}, errors.Wrapf(err, "error listing elements for set %q", n.Name())
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+}
+
+func (n *namedSet) Create(_ bool) error {
+	tx := n.nftables.NewTransaction()
+
+	tx.Add(&n.set)
+	err := n.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error creating set")
```
tpantelis: Minor thing but it might be useful to add more context to the error message.

```suggestion
	return errors.Wrapf(err, "error creating set %q", n.Name())
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+	return n.set.Name
+}
+
+func (n *namedSet) Flush() error {
+	tx := n.nftables.NewTransaction()
+
+	tx.Flush(&n.set)
+	err := n.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error flushing set")
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error flushing set %q", n.Name())
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (n *namedSet) Destroy() error {
+	tx := n.nftables.NewTransaction()
+	tx.Delete(&n.set)
+
+	err := n.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error deleting set")
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error deleting set %q", n.Name())
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+	tx := n.nftables.NewTransaction()
+
+	tx.Add(&knftables.Element{
+		Set: n.set.Name,
+		Key: []string{entry},
+	})
+
+	err := n.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error adding entry to set")
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error adding entry %q to set %q", entry, n.Name())
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+	tx := n.nftables.NewTransaction()
+
+	tx.Delete(&knftables.Element{
+		Set: n.set.Name,
+		Key: []string{entry},
+	})
+
+	err := n.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error deleting entry to set")
```
tpantelis, line 105:
```suggestion
	return errors.Wrapf(err, "error deleting entry %q from set %q", entry, n.Name())
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (n *namedSet) Name() string {
+	return n.set.Name
+}
+
+func (n *namedSet) Flush() error {
+	tx := n.nftables.NewTransaction()
+
+	tx.Flush(&n.set)
+	err := n.nftables.Run(context.TODO(), tx)
```
tpantelis, line 68: We should check for not found to be consistent with iptables.

```suggestion
	err := n.nftables.Run(context.TODO(), tx)
	if knftables.IsNotFound(err) {
		return nil
	}
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return errors.Wrap(err, "error flushing set")
+}
+
+func (n *namedSet) Destroy() error {
+	tx := n.nftables.NewTransaction()
+	tx.Delete(&n.set)
+
+	err := n.nftables.Run(context.TODO(), tx)
```
tpantelis, line 77:
```suggestion
	err := n.nftables.Run(context.TODO(), tx)
	if knftables.IsNotFound(err) {
		return nil
	}
```

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+
+			err = p.nftables.Run(context.TODO(), tx)
+			if err != nil && !knftables.IsNotFound(err) {
+				logger.Errorf(err, "Error destroying the set %q", set)
+				retErr = err
+			}
+		}
+	}
+
+	return retErr
```
tpantelis, line 146: To avoid code duplication, we can extract a function to delete a set:

```
func deleteSet(nftables knftables.Interface, name string) error {
	tx := nftables.NewTransaction()
	tx.Delete(&knftables.Set{
		Name: name,
	})

	err := nftables.Run(context.TODO(), tx)
	if knftables.IsNotFound(err) {
		return nil
	}

	return errors.Wrapf(err, "error deleting set %q", name)
}
```

which we can also call from `namedSet.Destroy().`. Also we should aggregate the errors here so we don't lose any in case there's multiple

```suggestion
import utilerrors "k8s.io/apimachinery/pkg/util/errors"
...

	var errs []error

	for _, set := range setsList {
		if nameFilter(set) {
			err = deleteSet(p.nftables, set)
			errs = append(errs, err)
		}
	}

	return utilerrors.NewAggregate(errs)
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	nftChain := &knftables.Chain{
+		Name:     chain.Name,
+		Type:     ptr.To(iphookChainTypeToNftablesType[chain.Type]),
+		Hook:     ptr.To(iphookChainHookToNftablesHook[chain.Hook]),
+		Priority: ptr.To(chainPriority),
+	}
+	tx.Add(nftChain)
+	err = p.nftables.Run(context.TODO(), tx)
```
tpantelis, line 134: We can check for `IsAlreadyExists` so we don't need to separately check `chainExists` above.
```suggestion
	err = p.nftables.Run(context.TODO(), tx)
	if knftables.IsAlreadyExists(err) {
		return nil
	}
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+	tx := p.nftables.NewTransaction()
+	tx.Add(&knftables.Table{
+		Comment: ptr.To("rules for submariner"),
+	})
+
+	tx.Add(&knftables.Chain{
+		Name: chain.Name,
+	})
+
+	err = p.nftables.Run(context.TODO(), tx)
```
tpantelis, line 154: Same here re: `IsAlreadyExists`.

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+	nftChain := &knftables.Chain{
+		Name:     chain.Name,
+		Type:     ptr.To(iphookChainTypeToNftablesType[chain.Type]),
+		Hook:     ptr.To(iphookChainHookToNftablesHook[chain.Hook]),
+		Priority: ptr.To(chainPriority),
+	}
+	tx.Add(nftChain)
+	err = p.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error creating IPHook Chain")
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error creating IPHook Chain %q", chain.Name)
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+		Comment: ptr.To("rules for submariner"),
+	})
+
+	tx.Add(&knftables.Chain{
+		Name: chain.Name,
+	})
+
+	err = p.nftables.Run(context.TODO(), tx)
+
+	return errors.Wrap(err, "error creating chain")
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error creating chain %q", chain.Name)
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+	tx.Delete(&knftables.Chain{
+		Name: chain,
+	})
+
+	err := p.nftables.Run(context.TODO(), tx)
+	if knftables.IsNotFound(err) {
+		return nil
+	}
+
+	return errors.Wrap(err, "error deleting chain")
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error deleting chain %q", chain)
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+	tx.Flush(&knftables.Chain{
+		Name: chain,
+	})
+
+	err := p.nftables.Run(context.TODO(), tx)
+	if knftables.IsNotFound(err) {
+		return nil
+	}
+
+	return errors.Wrap(err, "error clearing chain")
```
tpantelis:
```suggestion
	return errors.Wrapf(err, "error clearing chain %q", chain)
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+	err := p.nftables.Run(context.TODO(), tx)
+	if knftables.IsNotFound(err) {
+		return nil
+	}
+
+	return errors.Wrap(err, "error deleting rule")
+}
+
+func (p *packetFilter) List(_ packetfilter.TableType, chain string) ([]*packetfilter.Rule, error) {
+	existingRules, err := p.nftables.ListRules(context.TODO(), chain)
```
tpantelis, line 244: The docs for `ListRules`  says "_the Rule objects will have their `Comment` and `Handle` fields filled in, but *not* the actual `Rule` field._". I also verified this in the code. I think we need to also set the `Comment` field to the rule spec string on insert/append.

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,511 @@
...
+
+	return errors.Wrap(err, "error clearing chain")
+}
+
+func (p *packetFilter) Delete(_ packetfilter.TableType, chain string, rule *packetfilter.Rule) error {
+	tx := p.nftables.NewTransaction()
+	tx.Delete(&knftables.Rule{
+		Chain: chain,
+		Rule:  ToRuleSpec(rule).String(),
+	})
```
tpantelis, line 218: The docs for `Delete` says the `Rule` field is ignored and you must specify the `Handle`. To find the `Handle`, it seems we need to use `ListRules` and find the desired rule by its `Comment` field (see my other comment re: this).

tpantelis (changes requested):

pkg/packetfilter/nftables/namedset.go
```diff
@@ -0,0 +1,142 @@
...
+			Name: set.Name,
+			Type: setType,
+		},
+		nftables: p.nftables,
+	}
+}
+
+func (n *namedSet) Create(_ bool) error {
+	tx := n.nftables.NewTransaction()
+
```
tpantelis, line 49: Don't we need to add the Table to the tx like we do for chains et al?

```suggestion
        tx.Add(&knftables.Table{
		Comment: ptr.To("rules for submariner"),
	})
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	if exists {
+		return nil
+	}
+
+	tx := p.nftables.NewTransaction()
+	tx.Add(&knftables.Table{
+		Comment: ptr.To("rules for submariner"),
+	})
```
tpantelis: Since this is done in multiple places we should refactor it to a function.

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,535 @@
...
+		return nil
+	}
+
+	return errors.Wrapf(err, "error deleting rule %q from chain %q", *kRule.Comment, chain)
+}
+
+func (p *packetFilter) getNftablesRuleFromList(chain string, rule *packetfilter.Rule) *knftables.Rule {
+	existingRules, err := p.nftables.ListRules(context.TODO(), chain)
+	if err != nil {
+		return nil
```
tpantelis: This should return the error instead of silencing it.

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	for _, existingRule := range existingRules {
+		if ToRuleSpec(rule).String() == *existingRule.Comment {
```
tpantelis: We probably shouldn't assume Comment is non-nil.

```suggestion
		if ToRuleSpec(rule).String() == ptr.Deref(existingRule.Comment, "") {
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	rules := []*packetfilter.Rule{}
+
+	// Note: since currently ListRules returns only `Comment` and `Handle` fields
+	// of Rule object. When creating a new rule `Comment` will be set to RuleSpec.
+
+	for _, existingRule := range existingRules {
+		ruleSpec := strings.Split(*existingRule.Comment, " ")
+		rules = append(rules, FromRuleSpec(ruleSpec))
```
tpantelis:
```suggestion
                if ptr.Deref(existingRule.Comment, "") != "" {
		    ruleSpec := strings.Split(*existingRule.Comment, " ")
		    rules = append(rules, FromRuleSpec(ruleSpec))
		}
```

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,535 @@
...
+	chainsList, err := p.nftables.List(context.TODO(), "chains")
+	if err != nil && !knftables.IsNotFound(err) {
+		return false, errors.Wrap(err, "error listing chains")
+	}
+
+	return slices.Contains(chainsList, chain), nil
+}
+
+func (p *packetFilter) AppendUnique(_ packetfilter.TableType, chain string, rule *packetfilter.Rule) error {
+	return p.insertRuleAtPosition(chain, rule, -1)
```
tpantelis, line 107: The knftables API doesn't ensure uniqueness so perhaps we should implement that here like we did for `PrependUnique`.

tpantelis (changes requested):
I noticed that we're ignoring the `packetfilter.TableType` param in the implementation methods. Also we're passing "submariner" for the table name to `knftables.New`. Since the `knftables.Interface` pertains to a specific table, should we create `knftables.Interface` instances for each `packetfilter.TableType`?

tpantelis:
So knftables has a fake implementation and I've used that to write unit tests. It shares a lot of code with the real implementation, specifically the input validation. With your latest changes, the tests for the Rules methods are passing. Set creation is failing b/c the table isn't being first added as we're doing for chains. If you address this and the other minor comments, I think this PR will be in a good state to merge. Then I can push a PR to add the unit tests and we can address the other things.

pkg/packetfilter/nftables/nftables.go
```diff
@@ -0,0 +1,535 @@
...
+func (p *packetFilter) Append(_ packetfilter.TableType, chain string, rule *packetfilter.Rule) error {
+	return p.insertRuleAtPosition(chain, rule, -1)
+}
+
+func (p *packetFilter) insertRuleAtPosition(chain string, rule *packetfilter.Rule, pos int) error {
+	knftRule := knftables.Rule{
+		Chain:   chain,
+		Rule:    ToRuleSpec(rule).String(),
+		Comment: ptr.To(ToRuleSpec(rule).String()),
+	}
```
tpantelis: We should only call `ToRuleSpec` once. 

```suggestion
        ruleSpec := ToRuleSpec(rule).String()
        
	knftRule := knftables.Rule{
		Chain:   chain,
		Rule:    ruleSpec,
		Comment: ptr.To(ruleSpec),
	}
```

tpantelis (changes requested):

yboaron:
> I noticed that we're ignoring the `packetfilter.TableType` param in the implementation methods. Also we're passing "submariner" for the table name to `knftables.New`. Since the `knftables.Interface` pertains to a specific table, should we create `knftables.Interface` instances for each `packetfilter.TableType`?

Nope, unlike iptables in nftables a single table can have rules of all `packetfilter.TableType` (e.g: nat, filter) , so we'll have all our chains/rules under single table ('submariner').

yboaron:
> So knftables has a fake implementation and I've used that to write unit tests. It shares a lot of code with the real implementation, specifically the input validation. With your latest changes, the tests for the Rules methods are passing. Set creation is failing b/c the table isn't being first added as we're doing for chains. If you address this and the other minor comments, I think this PR will be in a good state to merge. Then I can push a PR to add the unit tests and we can address the other things.

Yep, I added the tables to chain creation b/c I got error when I run some tests with fake nftables  (I changed main.go for both globalnet and routeagent to use nftables instead of iptables, and create the fake nftables  in nftables.go ). 
Didn't get any error in Set creation, but I'll add the table also for Sets

pkg/packetfilter/nftables/nftables.go
yboaron: In case of AppendUnique , p.insertRuleAtPosition function is called with position = -1, which means we'll use tx.Add(&knftRule) without specifying index .
AFAIU, from knftables doc [1] that should ensure uniqueness.

[1]
// Add adds an "nft add" operation to tx, ensuring that obj exists by creating it if it
// did not already exist. (If obj is a Rule, it will be appended to the end of its chain,
// or else added after the Rule indicated by this rule's Index or Handle.)

tpantelis:
> Yep, I added the tables to chain creation b/c I got error when I run some tests with fake nftables (I changed main.go for both globalnet and routeagent to use nftables instead of iptables, and create the fake nftables in nftables.go ). Didn't get any error in Set creation, but I'll add the table also for Sets

I get this error when creating `namedSet.Create` :

```
error creating set "my-set": no such table "ip submariner"
      {
          error: <*errors.withMessage | 0xc00011c660>{
              cause: <*knftables.nftablesError | 0xc000350090>{
                  wrapped: nil,
                  msg: "no such table \"ip submariner\"",
                  errno: 0x2,
              },
              msg: "error creating set \"my-set\"",
          },
          stack: [0x56debf, 0x772c08, 0x730f53, 0x742acd, 0x46d461],
      }
```

In looking at the knftables code, at least the fake impl, internally the table is created once and shared across API calls so I'm sure you didn't see it b/c a create chain operation was issued before set creation. But to be safe, we should do the same with set creation.

pkg/packetfilter/nftables/nftables.go
tpantelis: So that's the docs for the general `tx.Add` operation and create-if-not-exists may not apply to all objects. In the fake implementation, it does apply to chains, sets, maps et al. But it doesn't apply to rules (i tried `AppendUnique` plus I can see it in the code).  I think this is b/c rules are a bit different in that they don't have an persistent unique identifier, like a chain or set that has a name. They have a handle but that's not assigned until it's created.  So `knftables` doesn't have a way to check for an existing rule. In fact, I think this [doc](https://github.com/kubernetes-sigs/knftables?tab=readme-ov-file#missing-apis) confirms it: "_Most IPTables libraries have an API for "add this rule only if it doesn't already exist", but that does not seem as useful in nftables (or at least "in nftables as used by Kubernetes-ish components that aren't just blindly copying over old iptables APIs"), because chains tend to have static rules and dynamic sets/maps, rather than having dynamic rules._".  So we're going to have to ensure uniqueness on append. But we can do this in a followup. I have cycles if you want after I add the unit tests.

yboaron: Ack, yep we can have it in a followup PR.
10x

tpantelis (approved):

PR #2957, opened 2024-04-01
Create a common library for route-agent vxlan driver and vxlan cable driver

### LINKED ISSUE #1393: Create a common library for route-agent vxlan driver and vxlan cable driver
Create a common library for routeagent vxlan driver and vxlan cable driver.
**Why is this needed**:

tpantelis:

Fixes https://github.com/submariner-io/submariner/issues/1393


dfarrell07 (approved):

pkg/vxlan/vxlan.go
```diff
@@ -0,0 +1,271 @@
...
+			LinkIndex: i.link.Index,
+			Src:       srcIP,
+			Dst:       &destCIDRs[j],
+			Gw:        gwIP,
+			Type:      netlink.NDA_DST,
+			Flags:     netlink.NTF_SELF,
+			Priority:  100,
+			Table:     tableID,
+		}
+		err := i.netLink.RouteAdd(route)
```
yboaron: Can't we use netlinkAPI.RouteAddOrReplace here ?

pkg/vxlan/vxlan.go
```diff
@@ -0,0 +1,271 @@
...
+			Gw:        nil,
+			Type:      netlink.NDA_DST,
+			Flags:     netlink.NTF_SELF,
+			Priority:  100,
+			Table:     tableID,
+		}
+
+		err := i.netLink.RouteDel(route)
+		if err != nil {
+			return errors.Wrapf(err, "unable to add the route entry %v", route)
```
yboaron: nit , should be 'delete'

pkg/vxlan/vxlan.go
```diff
@@ -0,0 +1,271 @@
...
+			LinkIndex: i.link.Index,
+			Dst:       &destCIDRs[j],
+			Gw:        nil,
+			Type:      netlink.NDA_DST,
+			Flags:     netlink.NTF_SELF,
+			Priority:  100,
+			Table:     tableID,
+		}
+
+		err := i.netLink.RouteDel(route)
```
yboaron, line 259: Don't we want to return nil if route doesn't exist ? ,  like 
https://github.com/submariner-io/submariner/blob/devel/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go#L48

pkg/vxlan/vxlan.go
tpantelis: yes

pkg/vxlan/vxlan.go
tpantelis: yeah that should be fine here.

yboaron (approved):

PR #2981, opened 2024-04-18
Adjustments to packetfilter markValue

yboaron:
In Submariner we use the netfilter mark field associated with a packet to
handle GlobalNet packets, relevant packets are marked using IPTables
--set-mark flag [1] and we check if the packet is marked using
 --mark flag [2].

In NFTables packet mark field is set or checked using logical OR,AND,XOR
and NOT operations, check [3].

Currently, only '0xC000/0xC000' mark value is used for both IPTables flags,
which means that value and mask are equal.

This PR updates:
A. clients to use packetfilter mark value without mask (e.g: 0xC000)
B. the syntax of NFTables and IPTables mark value rules assuming that
   only mark value is provided.

[1]
--set-mark value[/mask]
Zeroes out the bits given by mask and ORs value into the packet mark.

[2]
--mark value[/mask]
Matches packets with the given unsigned mark value (if a mask is specified,
this is logically ANDed with the mask before the comparison).

[3]
https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes


pkg/packetfilter/nftables/nftables.go
```diff
@@ -334,7 +334,9 @@ func ToRuleSpec(rule *packetfilter.Rule) RuleSpec {
 	}
 
 	if rule.MarkValue != "" && rule.Action != packetfilter.RuleActionMark {
-		ruleSpec = append(ruleSpec, "mark", rule.MarkValue)
+		// syntax for MarkValue '0xc0000/0xc0000': meta mark & 0xc0000 == 0xc0000
+		i := strings.Index(rule.MarkValue, "/")
+		ruleSpec = append(ruleSpec, "meta", "mark", "&", rule.MarkValue[:i], "==", rule.MarkValue[:i])
```
tpantelis: What does the syntax "0xc0000/0xc0000" mean? Should we just assume this implicit syntax of 2 terms separated by a '/'? Since we're only using the first term before the '/' and discarding the second here, is this syntax specific to the iptables implementation? Or should this be:

```suggestion
		ruleSpec = append(ruleSpec, "meta", "mark", "&", rule.MarkValue[:i], "==", rule.MarkValue[i+1:])
```

PS: it might be useful to include an explanation of the intent/changes in the commit/PR description.

pkg/packetfilter/nftables/nftables.go
```diff
@@ -411,7 +414,7 @@ func FromRuleSpec(spec RuleSpec) *packetfilter.Rule {
 				rule.SnatCIDR, i = parseNextTerm(spec, i, noopParse)
 			}
 		case "mark":
-			rule.MarkValue, i = parseNextTerm(spec, i, noopParse)
+			rule.MarkValue, i = parseMark(spec, i)
```
tpantelis, line 422: Perhaps it would be better to check for case "meta" and then check the next term for "mark". Same with "set" above. That would make both consistent simplify `parseMark`.

tpantelis (changes requested):

pkg/packetfilter/nftables/nftables.go
yboaron: Yep, good point.
I updated the commit and PR description.

pkg/packetfilter/iptables/iptables.go
```diff
@@ -341,6 +341,8 @@ func FromRuleSpec(spec []string) *packetfilter.Rule {
 			rule.DPort, i = parseNextTerm(spec, i, noopParse)
 		case "--set-mark":
 			rule.MarkValue, i = parseNextTerm(spec, i, noopParse)
+			j := strings.Index(rule.MarkValue, "/")
+			rule.MarkValue = rule.MarkValue[j+1:]
```
tpantelis: Should we use `rule.MarkValue[:j]` in case it doesn't contain a slash?

pkg/packetfilter/iptables/iptables.go
```diff
@@ -378,7 +378,10 @@ func parseRuleMatch(spec []string, i int, rule *packetfilter.Rule) int {
 	switch spec[i] {
 	case "mark":
 		if i+2 < len(spec) && spec[i+1] == "--mark" {
-			rule.MarkValue = spec[i+2]
+			if j := strings.Index(spec[i+2], "/"); j > -1 {
+				rule.MarkValue = spec[i+2][:j]
+			}
+
```
tpantelis:
```suggestion
                         rule.MarkValue = parseMark(spec[i+2])
```

pkg/packetfilter/nftables/nftables.go
tpantelis: bump

yboaron: We need to parse: 
a. counter meta mark set mark | 0xc000
b. mark & 0xc000 == 0xc0000

I updated 'parseAction' to handle 'counter meta mark set' for mark action, and the case "mark" here should parse mark value for both cases.

I think it is OK.

tpantelis (approved):

yboaron:
cc @aswinsuryan

aswinsuryan (approved):

PR #2982, opened 2024-04-18
Change MSS Clamp IP hook chain type

yboaron:
NetFilter doesn't support post-routing IP hook for route chain type, check [1] ,
(only output IP hook is available).

Moving it to post-routing IP hook for filter chain type is not possible because
POSTROUTING chain isn't supported for IPTables filter table.

Since we run the MTU handler on all the nodes, it should be safe to move MSS
Clamp rules to output IP hook for route chain type.
    
    [ 1 ]
    https://wiki.nftables.org/wiki-nftables/index.php/Netfilter_hooks


yboaron:
cc @sridhargaddam

yboaron:
e2e tests are red, I must have missed something ...

tpantelis (approved):

tpantelis:
Are there any migration concerns with prior` SUBMARINER-POSTROUTING` MSS
clamp rule  entries, ie do they need to be cleaned up?

yboaron:
> Are there any migration concerns with prior` SUBMARINER-POSTROUTING` MSS clamp rule entries, ie do they need to be cleaned up?

If upgrading submariner version x to version x+1 means that we uninstall version x components before installing version x+1 , I guess it should be OK, otherwise, we need to delete the orphan iptablse rules under SUBMARINER-POSTROUTING chain.

@sridhargaddam , any thoughts?

tpantelis:
> > Are there any migration concerns with prior` SUBMARINER-POSTROUTING` MSS clamp rule entries, ie do they need to be cleaned up?
> 
> If upgrading submariner version x to version x+1 means that we uninstall version x components before installing version x+1 , I guess it should be OK, otherwise, we need to delete the orphan iptablse rules under SUBMARINER-POSTROUTING chain.
> 
> @sridhargaddam , any thoughts?

For `subctl upgrade`, we don't uninstall - we just update the operator deployment which will restart the pod. I think a similar thing is done with ACM. We'd have to handle any `SUBMARINER-POSTROUTING` cleanup on pod restart. I'm not sure how critical that is.

yboaron:
I don't think it's critical (we'll have additional iptables rules that also clamp the MSS),
but it's better to handle it as part of this change.

In the meantime, will mark the PR as draft.

yboaron:
Also currently the user can force tcp-mss clamp to a specific value by annotating submariner.io/tcp-clamp-mss=< value >  in GW node this won't work if rules are under SUBMARINER-OUTPUT chain.

PR #2983, opened 2024-04-19
Adjustments to ipset API and implementation

tpantelis:
Various improvements, simplifications and other minor adjustments to the ipset API and implementation.
Also added more unit tests for nearly complete coverage.


dfarrell07 (approved):

yboaron:
overall code LGTM, 

I wonder if we can delete all unused IPSet functions ( we only expose [1] functions in packetfilter) 

[1]
https://github.com/submariner-io/submariner/blob/devel/pkg/packetfilter/packetfilter.go#L273

yboaron (approved):

tpantelis:
> overall code LGTM,
> 
> I wonder if we can delete all unused IPSet functions ( we only expose [1] functions in packetfilter)
> 
> [1] https://github.com/submariner-io/submariner/blob/devel/pkg/packetfilter/packetfilter.go#L273

I removed a couple already but there's a few more that are unused: `DestroyAllSets`, `TestEntry` (this is used in `NamedSet.TestEntry` but that isn't used), `GetVersion`, `AddEntryWithOptions`. I thought about those but wasn't sure if we'd ever need them.

PR #2984, opened 2024-04-19
Fix panic in kubeproxy SyncHandler

tpantelis:
This was observed in an E2E run. In `populateRemoteVtepIps`, `isOnGateway` was true but the `vxlanDevice` was nil. This normally shouldn't happen since the `vxlanDevice` is created on transition to gateway. But in this case, another unrelated event handler (specifically the `MTUHandler`) encountered an error in its `LocalEndpointCreated` callback and thus prevented the shared event controller from proceeding to `TransitionToGateway`. However the `IsOnGateway` flag was kept as true while it retried which led to the inconsistent state in the `kubeproxy` handler. Although the event controller should prevent this, we should at least ensure `vxlanDevice` is non-nil.


dfarrell07 (approved):

tpantelis:
Ping @aswinsuryan @vthapar

aswinsuryan (approved):

PR #2986, opened 2024-04-26
Refactor event controller to use a shared informer

tpantelis:
...and create a resource syncer per event handler. This decouples the handlers from one another and allows them to run in parallel and independently from one another, such that if an error or delay occurs in one handler, it doesn't affect the others.

Depends on https://github.com/submariner-io/admiral/pull/895


tpantelis:
Ping @aswinsuryan @vthapar @dfarrell07

aswinsuryan (approved):

dfarrell07 (approved):

PR #3009, opened 2024-05-17
Globalnet not running in E2E

tpantelis:
The E2E jobs intended to install/run Globalnet are not doing so. This is b/c the E2E GHA needs to specify the `matrix.globalnet` var in the `using` statement. This was broken by https://github.com/submariner-io/submariner/pull/2832/commits/6cd0a39b9114f68da5a8304fbd2e71ce840d5b5a when the `globalnet` option was removed from `extra-toggles`.


aswinsuryan (approved):

dfarrell07 (approved):
Good catch, thanks.

skitt (approved):

PR #3010, opened 2024-05-17
Remove route agent "submariner.io/cniIfaceIp" node annotation

tpantelis:
The route agent discovers the CNI IP and adds the annotation to the node for use by Globalnet. However granting RBAC permission to update nodes poses a potential vulnerability exploitation by attackers. We can eliminate this annotation and the RBAC permission by having Globalnet discover the CNI IP itself.

See commits for details.


pkg/globalnet/controllers/node_controller.go
```diff
@@ -52,21 +52,34 @@ func NewNodeController(config *syncer.ResourceSyncerConfig, pool *ipam.IPPool, n
 		nodeName:                   nodeName,
 	}
 
+	clusterCIDR := ""
+	if len(clusterCIDRs) > 0 {
+		clusterCIDR = clusterCIDRs[0]
+	}
+
+	cniIface, err := cni.Discover(clusterCIDR)
```
aswinsuryan, line 60: It is not just healthchecker that fails but the host networking connectivity too,  so should we continue here?

tpantelis: I'm basically mimicking the behavior from before in that if the CNI IP annotation wasn't set then the controller skipped processing. Here we determine the CNI IP on startup. We don't want to return an error as that would fail the whole startup. We could return and elide the rest of the initialization but the base controller expects a valid `resourceSyncer` instance  to start. We could get around that but I though it's easiest to let the syncer run but skip processing like it did before.

aswinsuryan (approved):

pkg/globalnet/controllers/node_controller.go
```diff
@@ -52,21 +52,34 @@ func NewNodeController(config *syncer.ResourceSyncerConfig, pool *ipam.IPPool, n
 		nodeName:                   nodeName,
 	}
 
+	clusterCIDR := ""
+	if len(clusterCIDRs) > 0 {
```
yboaron, line 56: Don't we want to report an error if  len(clusterCIDRs) isn't bigger than 0?

pkg/globalnet/controllers/node_controller.go
```diff
@@ -52,21 +52,34 @@ func NewNodeController(config *syncer.ResourceSyncerConfig, pool *ipam.IPPool, n
...
+		clusterCIDR = clusterCIDRs[0]
+	}
+
+	cniIface, err := cni.Discover(clusterCIDR)
+	if err == nil {
+		controller.cniIP = cniIface.IPAddress
+
+		logger.Infof("Discovered CNI interface IP %q", controller.cniIP)
+	} else {
+		logger.Errorf(err, "Error obtaining CNI IP address - health check functionality will not work")
```
yboaron, line 66: AFAIK, service discovery also ignores exported services from clusters with health check down

pkg/globalnet/controllers/controllers_suite_test.go
```diff
@@ -61,7 +63,7 @@ const (
 	globalIngressIPName       = "nginx-ingress-ip"
 	kubeProxyIPTableChainName = "KUBE-SVC-Y7DIXXI5PNAUV7FB"
 	serviceName               = "nginx"
-	cniInterfaceIP            = "10.20.30.40"
+	cniInterfaceIP            = "169.254.1.50"
```
yboaron, line 66: Why do we need to change 'cniInterfaceIP' value ? 
 cniInterfaceIP shouldn't be from globalIP cidr range

pkg/cni/cni_iface.go
```diff
@@ -86,56 +81,3 @@ func discover(clusterCIDR string) (*Interface, error) {
 
 	return nil, errors.Errorf("unable to find CNI Interface on the host which has IP from %q", clusterCIDR)
 }
-
-func AnnotateNodeWithCNIInterfaceIP(nodeName string, clientSet kubernetes.Interface, clusterCidr []string) error {
```
yboaron, line 90: I assume that after upgrade we'll have orphaned/useless cni IP address annotation that won't be deleted , 
do we want to handle it ?

tpantelis: We can't remove it b/c then we'd need update permission for nodes which we're removing.  I don't see any way we can handle it other than to add a release note. It's just an annotation so doesn't hurt anything.

pkg/globalnet/controllers/controllers_suite_test.go
tpantelis: I changed it to be in range of `localCIDR` b/c originally I was checking for that in the fake `cni.DiscoverFunc` like the real one does but I later changed `cni.DiscoverFunc` to just verify that the configured local CIDR is passed in. `localCIDR` is also used as the global CIDR which actually doesn't really matter for the tests but logically should be defined separately.

pkg/globalnet/controllers/node_controller.go
tpantelis: yeah we could although the operator would never pass empty cluster CIDRs anyway. I have a follow-up PR to check all `clusterCIDRs` so this check will go away anyway.

pkg/cni/cni_iface.go
yboaron: Yep, adding release note sounds reasonable.

yboaron (approved):

dfarrell07 (approved):

PR #3016, opened 2024-05-23
Enable retry when setting passive label for standby GW pod

yboaron:

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -339,8 +339,15 @@ func (g *gatewayType) onStoppedLeading() {
 
 	logger.Info("Controllers stopped")
 
-	if err := g.gatewayPod.SetHALabels(context.Background(), subv1.HAStatusPassive); err != nil {
-		logger.Warningf("Error updating pod label to passive: %s", err)
+	ctx := context.Background()
+	if err := wait.PollUntilContextCancel(ctx, 100*time.Millisecond, true, func(ctx context.Context) (bool, error) {
+		if g.gatewayPod.SetHALabels(ctx, subv1.HAStatusPassive) != nil {
+			return false, nil //nolint:nilerr // keep trying
+		}
+
+		return true, nil
+	}); err != nil {
+		logger.Errorf(err, "Failed updating pod label to passive: %s", err)
 	}
 
 	err := g.startLeaderElection(context.Background())


pkg/gateway/gateway.go
```diff
@@ -339,9 +339,19 @@ func (g *gatewayType) onStoppedLeading() {
...
 
-	if err := g.gatewayPod.SetHALabels(context.Background(), subv1.HAStatusPassive); err != nil {
-		logger.Warningf("Error updating pod label to passive: %s", err)
-	}
+	msgLogged := atomic.Bool{}
+	ctx := context.Background()
+	_ = wait.PollUntilContextCancel(ctx, 100*time.Millisecond, true, func(ctx context.Context) (bool, error) {
+		if err := g.gatewayPod.SetHALabels(ctx, subv1.HAStatusPassive); err != nil {
+			if msgLogged.CompareAndSwap(false, true) {
+				logger.Warningf("Error updating pod label to passive: %s", err)
```
skitt: This has the potential to end up being confusing: since the operation is retried, a failure would result in a warning in the logs, which might be resolved by a subsequent retry — but nothing in the logs would indicate that. It would be better to only log an error if the pod label can’t be updated at all.

skitt (changes requested):

skitt (approved):

aswinsuryan (approved):

pkg/gateway/gateway.go
```diff
@@ -339,8 +339,15 @@ func (g *gatewayType) onStoppedLeading() {
...
-		logger.Warningf("Error updating pod label to passive: %s", err)
+	ctx := context.Background()
+	if err := wait.PollUntilContextCancel(ctx, 100*time.Millisecond, true, func(ctx context.Context) (bool, error) {
+		if g.gatewayPod.SetHALabels(ctx, subv1.HAStatusPassive) != nil {
+			return false, nil //nolint:nilerr // keep trying
+		}
+
+		return true, nil
+	}); err != nil {
+		logger.Errorf(err, "Failed updating pod label to passive: %s", err)
```
tpantelis, line 350: A few things here... re: prior comment https://github.com/submariner-io/submariner/pull/3016#discussion_r1611798154, we won't get any indication of a failure to update b/c we don't return an error to `PollUntilContextCancel` so it would only return an error if the passed `ctx` is cancelled. However, since we pass the `Background` context which is never cancelled, `PollUntilContextCancel` won't return unless/until `SetHALabels` succeeds. We would expect it to eventually succeed if it's a transient error but, if not, it would silently loop forever. Also if the pod was stopped, it wouldn't gracefully exit.

I propose:

1) Use `retry.OnError` to filter out errors that aren't transient (like NotFound or permission-related) so that it returns the error and we log it 
2) Log the first non-transient error and indicate that it's retrying
3) Use the original `ctx` that was passed to `Run`, ie have `startLeaderElection` pass the `ctx` to `onStoppedLeading`

I'll submit a follow-up PR.

yboaron: @tpantelis 
Can't we replace PollUntilContextCancel with PollUntilContextTimeout ? 
so it won't run forever.

Not sure that I understand the pod stop use case you mentioned.
 
I can take care of that too,

tpantelis: > Can't we replace PollUntilContextCancel with PollUntilContextTimeout ?
so it won't run forever.

But we don't want it to time out on a transient error. Also what time out would you give it... I think it's fine if it loop forever as long as it can be cancelled.

> Not sure that I understand the pod stop use case you mentioned.

Normally when a pod is stopped, it is sent a signal and we set up a `Context` in `main` that is cancelled which we use to stop controllers etc and gracefully shut down. But if the process doesn't stop, I believe K8s does a hard kill. `onStoppedLeading` is executed by the leader election on a separate thread so it shouldn't block graceful shutdown if it's stuck but it's still better if it's cancellable. But let's not worry about that now - I can look into that later.

For now let's address 1 and 2 above. But `retry.OnError` doesn't take a `Context` so we should continue using `PollUntilContextCancel` and check for certain errors that indicate a transient API server issue, like `apierrors.IsTooManyRequests` and `apierrors.IsServerTimeout`, in which case we log a warning once and and return false to retry. We should modify [this code](https://github.com/submariner-io/submariner/blob/37c5d34e4a5968a11107eac8cbf1b0bdbe4ba2a8/pkg/gateway/gateway.go#L284) and refactor to a function and use in both places.

I can make these changes but if you want to that's fine.

tpantelis: I submitted https://github.com/submariner-io/submariner/pull/3018 (since you were off the rest of the day).

PR #3017, opened 2024-05-24
Fix issues with Pinger on relection as leader

### LINKED ISSUE #3008: Invalid healthchecker counters when the leader is relected
**What happened**:
The health checker county er is not getting updated when a gateway node lose leader election and get elected again.
When the 

The counters should have the updated values

Try to simulate a leader re-election

**Environment**:
- Install tools:
- Others:

aswinsuryan:
The pinger  for remote endpoints are started
when a node is relected as a leader, once it
looses it leadership.

Fixes: https://github.com/submariner-io/submariner/issues/3008

### DIFF
--- a/pkg/cableengine/healthchecker/healthchecker.go
+++ b/pkg/cableengine/healthchecker/healthchecker.go
@@ -61,9 +61,8 @@ type Config struct {
 
 type controller struct {
 	sync.RWMutex
-	endpointWatcher watcher.Interface
-	pingers         map[string]PingerInterface
-	config          *Config
+	pingers map[string]PingerInterface
+	config  *Config
 }
 
 var logger = log.Logger{Logger: logf.Log.WithName("HealthChecker")}
@@ -87,13 +86,6 @@ func New(config *Config) (Interface, error) {
 		},
 	}
 
-	var err error
-
-	controller.endpointWatcher, err = watcher.New(config.WatcherConfig)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error creating watcher")
-	}
-
 	return controller, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -116,6 +120,19 @@ func (h *controller) Start(stopCh <-chan struct{}) error {
...
+	if err != nil {
+		return errors.Wrap(err, "error listing endpoints")
+	}
+
+	endpointList := endpoints.Items
+
+	for i := range endpointList {
+		h.endpointCreatedOrUpdated(&endpointList[i], 0)
+	}
+
```
tpantelis: So this gets all `Endpoints` and runs them through `endpointCreatedOrUpdated`. This duplicates what the informer already does so I assume this is to handle subsequent calls to `Start`. While this will reprocess current `Endpoints`, the `endpointWatcher` will not be restarted so it will not handle updates to remote `Endpoints`.

The real problem is that the `controller` is not re-startable and the caller assumes it is. To make it so, we need to create the `endpointWatcher` instance in the `Start` method so it's recreated on each call.

tpantelis (changes requested):

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: thanks for the detailed explanation. Made the changes.

tpantelis:
I submitted https://github.com/submariner-io/submariner/pull/3017/commits/161240975ca0e4b419d50af883851011324ad070 to update the unit tests to test stopping and restarting.

tpantelis (approved):

dfarrell07 (approved):

PR #3019, opened 2024-05-29
Modify globalnet to annotate the local Gateway with the global IP

tpantelis:
..instead of the local `Node`.  Also adjust the  datastore syncer to watch the local `Gateway` for global IP updates.
This will further reduce the RBAC requirements by removing `Node` update permission.

Once a global IP is allocated and annotated to a `Gateway`, it remains for the lifetime of the `Gateway` instance. Same for the ingress rules. On failover to another gateway, the new globalnet controller will reserve the global IPs for all the `Gateway` instances so they're not reused.

See commits for details.


tpantelis:
The globalnet E2E fails b/c the RBAC permissions need to be adjusted.  These are pending in https://github.com/submariner-io/submariner-operator/pull/3091 and https://github.com/submariner-io/submariner-charts/pull/521. However we have a chicken-and-the-egg situation here. I think the best solution is to merge this first with the E2E failures and then test and merge the RBAC PRs, adjusting the changes if necessary.

vthapar:
> The globalnet E2E fails b/c the RBAC permissions need to be adjusted. These are pending in [submariner-io/submariner-operator#3091](https://github.com/submariner-io/submariner-operator/pull/3091) and [submariner-io/submariner-charts#521](https://github.com/submariner-io/submariner-charts/pull/521). However we have a chicken-and-the-egg situation here. I think the best solution is to merge this first with the E2E failures and then test and merge the RBAC PRs, adjusting the changes if necessary.

I think a better solution would be to modify operator PR to just add RBAC for gateways. And once submariner changes are merged, we can create another PR to delete nodes RBAC.

tpantelis:
The wireguard E2E failures are unrelated to this PR - they're occurring on all PRs.

aswinsuryan (approved):

vthapar (approved):

skitt (approved):

dfarrell07 (approved):

PR #3040, opened 2024-06-04
Drop the dependency on wireguard in e2e

skitt:
The wireguard package has dependencies which can't be built on Windows. If the e2e tests end up with a dependency on the wireguard cable driver, that leaks into subctl and prevents it from being built on Windows.


tpantelis (approved):

aswinsuryan (approved):

dfarrell07 (approved):

skitt:
Verification PR: https://github.com/submariner-io/subctl/pull/1147

PR #3069, opened 2024-06-30
Fix GW failover bug for wireguard cable

### LINKED ISSUE #3031: Wireguard gateway failover tests fails in 0.18
**What happened**:Wireguard gateway failover tests fails in 0.18

**What you expected to happen**: The tests should pass 

**How to reproduce it (as minimally and precisely as possible)**: Run e2e in 0.18 branch, the logs are attached.

**Environment**:
- Install tools:
- Others:
[
[submariner-gather.zip](https://github.com/user-attachments/files/15517245/submariner-gather.zip)
](url)

yboaron:
After GW pod failure, the wireguard network interface (named 'submariner')
is created with a new ifindex, meaning all host routes created with LinkIndex
pointing to previous wireguard ifindex are deleted.
    
This PR ensures that for wireguard cable driver, host routes on GW node will
be reconfigured (pointing to updated wireguard ifindex) after GW failover.

Fixes: https://github.com/submariner-io/submariner/issues/3031


skitt:
Reverting https://github.com/submariner-io/submariner/commit/b1edc52299e01f6ccaad9c1c837e7d4b45b8bf91 will break `subctl` builds on Windows. It’s OK to hard-code `"wireguard"` instead of hard-coding the cable driver package.

pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go
```diff
@@ -76,6 +76,11 @@ func (kp *SyncHandler) updateRoutingRulesForCIDRBlock(inputCidrBlock string, ope
 		if !kp.routeCacheGWNode.Has(inputCidrBlock) {
 			kp.routeCacheGWNode.Insert(inputCidrBlock)
 
+			if err := kp.configureRoute(inputCidrBlock, operation, viaGW); err != nil {
+				logger.Errorf(err, "Failed to add route %q for HostNetwork support on the Gateway node: %v",
+					inputCidrBlock, err)
+			}
+		} else if kp.localCableDriver == wireguard.GetName() {
```
skitt: Is this appropriate for all calls to `updateRoutingRulesForCIDRBlock` and `updateRoutingRulesForHostNetworkSupport`, or only for the special case added in `LocalEndpointCreated`?

skitt (changes requested):

yboaron:
> Reverting [b1edc52](https://github.com/submariner-io/submariner/commit/b1edc52299e01f6ccaad9c1c837e7d4b45b8bf91) will break `subctl` builds on Windows. It’s OK to hard-code `"wireguard"` instead of hard-coding the cable driver package.

Hmm, can't see any reference to wireguard in test/e2e/redundancy/gateway_failover.go after reverting both https://github.com/submariner-io/submariner/commit/b1edc52299e01f6ccaad9c1c837e7d4b45b8bf91 and https://github.com/submariner-io/submariner/pull/3069/commits/86813989247e54a153704348fe81973a0d65933a 


Do you think that subctl build for Windows will break even after reverting both commits ?

pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go
yboaron: Fair point, I reverted the changes in route_iface file, 
so, the current change will reconfigure host routes if  we are on GW node and cable driver is wireguard

skitt:
> > Reverting [b1edc52](https://github.com/submariner-io/submariner/commit/b1edc52299e01f6ccaad9c1c837e7d4b45b8bf91) will break `subctl` builds on Windows. It’s OK to hard-code `"wireguard"` instead of hard-coding the cable driver package.
> 
> Hmm, can't see any reference to wireguard in test/e2e/redundancy/gateway_failover.go after reverting both [b1edc52](https://github.com/submariner-io/submariner/commit/b1edc52299e01f6ccaad9c1c837e7d4b45b8bf91) and [8681398](https://github.com/submariner-io/submariner/commit/86813989247e54a153704348fe81973a0d65933a)
> 
> Do you think that subctl build for Windows will break even after reverting both commits ?

As it stands, with the reworked third commit, everything will be fine. And yes, reverting both commits was also OK, I’d missed that.

Given the overall result of the PR, I don’t think it’s worth keeping the two reverts; it would be more readable to have a single commit without all the noise created by the reverts. The change is simple enough: you’re fixing gateway failover on WireGuard, as documented, and removing the now-redundant test skip.

I do think it would be worth keeping the comment that was here: https://github.com/submariner-io/submariner/pull/3069/files#diff-b579142107f4f78ee1a961875b4255ee72cd1407e4246e79aae1333b25b48346L84

skitt (approved):

test/e2e/redundancy/gateway_failover.go
```diff
@@ -78,16 +78,7 @@ func testGatewayPodRestartScenario(f *subFramework.Framework) {
 	gatewayNodes := framework.FindGatewayNodes(primaryCluster)
 	Expect(gatewayNodes).To(HaveLen(1), fmt.Sprintf("Expected only one gateway node on %q", primaryClusterName))
 	framework.By(fmt.Sprintf("Found gateway on node %q on %q", gatewayNodes[0].Name, primaryClusterName))
-
-	submEndpoint := f.AwaitSubmarinerEndpoint(primaryCluster, subFramework.NoopCheckEndpoint)
-
 	// The cable driver name can't be imported from pkg/cable/wireguard to avoid breaking the subctl build on Windows
```
skitt: My previous comment wasn’t clear — I meant this comment should be preserved along with the use of `"wireguard"` in the new code :wink:. So drop it from `gateway_failover.go` and add it to `endpoint_handler.go`.

yboaron: Not sure I understand your point, do you mean if wierguard is imported from pkg/routeagent_driver/handlers/kubeproxy/endpoint_handler.go it will break  subctl build for Windows?

I can see that Wireguard already imported from pkg/routeagent_driver/handlers/kubeproxy/route_iface.go

skitt: Ah, that’s correct, this change only affects the route agent and doesn’t end up in `subctl`, so the comment is irrelevant and should just be dropped entirely.

skitt (approved):

vthapar (approved):

dfarrell07:
We discussed if this should be backported. It likely exists on 0.18, maybe 0.17. It's marked as not required (or really, the test is skipped) on recent branches (>=0.17).

The weird part is why it started showing up, what relevant thing changed.

Might be simplest to backport, not worry about it.

dfarrell07:
Discussing on the scrub, we think this was tech preview in 0.17 so it should be backported there too.

PR #3070, opened 2024-07-03
Ensure exported test code doesn't depend on wireguard

skitt:
The WireGuard packages depend on packages which aren't available on Windows; since the exported test code is used in subctl, if the test packages end up with a dependency on the WireGuard packages, subctl can no longer be built.

This adds depguard configuration to enforce the exclusion.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,15 @@
 ---
 linters-settings:
+  depguard:
+    rules:
+      # Importing WireGuard breaks downstream projects, we mustn't use it in exported packages
+      wireguard:
+        list-mode: lax
+        files:
+          - "**/test/**/*.go"
+        deny:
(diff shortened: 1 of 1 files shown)


yboaron:
Very nice solution for guarding dependency! 
QQ: in case of package dependency failure, the error will be reported in linting/go , right? 


Also , AFAIK, currently only submariner/test/e2e/*  packages are imported by subctl  , but I guess we can deny all files under submariner/test

yboaron (approved):

vthapar (approved):

skitt:
> QQ: in case of package dependency failure, the error will be reported in linting/go , right?

That’s right!

> Also , AFAIK, currently only submariner/test/e2e/* packages are imported by subctl , but I guess we can deny all files under submariner/test

Indeed, nothing imports the external network tests.

PR #3073, opened 2024-07-07
Fix testGatewayPodRestartScenario for ROKS platform

MaxBab:
During execution of testGatewayPodRestartScenario test an active gateway connection tested by using provided gateway hostname.

Currently, the GW node is identified by the nodes which are matching the 'submariner.io/gateway' value.

In ROKS setup, the name of a node from "oc get nodes" with a label of 'submariner.io/gateway' does not match the actual hostname of the GW node.

As a result, the process of locating the active gateway node fails.

In order to fix it, change the process of identifying the active GW node by fetching the "hostname" of the node from the local endpoint of the setup, which provides the correct value for all the platforms.

### DIFF
--- a/test/e2e/redundancy/gateway_failover.go
+++ b/test/e2e/redundancy/gateway_failover.go
@@ -87,7 +87,8 @@ func testGatewayPodRestartScenario(f *subFramework.Framework) {
 
 	framework.By(fmt.Sprintf("Ensuring that the gateway reports as active on %q", primaryClusterName))
 
-	activeGateway := f.AwaitGatewayFullyConnected(primaryCluster, resource.EnsureValidName(gatewayNodes[0].Name))
+	submEndpoint := f.AwaitSubmarinerEndpoint(primaryCluster, subFramework.NoopCheckEndpoint)
+	activeGateway := f.AwaitGatewayFullyConnected(primaryCluster, resource.EnsureValidName(submEndpoint.Spec.Hostname))
 
 	framework.By(fmt.Sprintf("Deleting submariner gateway pod %q", gatewayPod.Name))
 	f.DeletePod(primaryCluster, gatewayPod.Name, framework.TestContext.SubmarinerNamespace)


test/e2e/redundancy/gateway_failover.go
```diff
@@ -87,7 +87,8 @@ func testGatewayPodRestartScenario(f *subFramework.Framework) {
 
 	framework.By(fmt.Sprintf("Ensuring that the gateway reports as active on %q", primaryClusterName))
 
-	activeGateway := f.AwaitGatewayFullyConnected(primaryCluster, resource.EnsureValidName(gatewayNodes[0].Name))
```
yboaron, line 90: Maybe leave a comment here to explain why hostname is retrieved from endpoint instead of gatewayNodes[0].Name

yboaron (approved):

test/e2e/redundancy/gateway_failover.go
```diff
@@ -87,7 +87,8 @@ func testGatewayPodRestartScenario(f *subFramework.Framework) {
...
+	submEndpoint := f.AwaitSubmarinerEndpoint(primaryCluster, subFramework.NoopCheckEndpoint)
```
yboaron, line 90: It is corner case, but it better check that  submEndpoint  is not nil

Expect(submEndpoint).ToNot(BeNil(), "Did not find local endpoint")

tpantelis: It's actually not necessary b/c AwaitSubmarinerEndpoint will itself fail the test if it's not found. Also note that other call sites don't check for nil.

tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
@MaxBab I think this should be backported?

MaxBab:
@dfarrell07 
This should be backported to release-0.18 branch.
I'll do it.

PR #3077, opened 2024-07-11
Add route-agent-healthcheck

aswinsuryan:
Add route-agent-healthcheck


tpantelis:
I know this is still a draft but I took a cursory look at it. It seems you're trying to incorporate route agent ping info into the `Gateway` CR however I don't think we should do this. I think it makes sense to introduce a similar `RouteAgent` CR corresponding to each route agent pod, eg:

```
type RouteAgent struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
	Status            RouteAgentStatus `json:"status"`
}

type RouteAgentStatus struct {
	Version       string       `json:"version"`
	StatusFailure string       `json:"statusFailure"`
	RemoteEndpoints   []RemoteEndpoint `json:"remoteEndpoints"`
}

type RemoteEndpoint struct {
	Status        ConnectionStatus `json:"status"`
	StatusMessage string           `json:"statusMessage"`
	Spec      EndpointSpec     `json:"endpoint"`
	// +optional
	LatencyRTT *LatencyRTTSpec `json:"latencyRTT,omitempty"`
}
```

aswinsuryan:
> I know this is still a draft but I took a cursory look at it. It seems you're trying to incorporate route agent ping info into the `Gateway` CR however I don't think we should do this. I think it makes sense to introduce a similar `RouteAgent` CR corresponding to each route agent pod, eg:
> 
> ```
> type RouteAgent struct {
> 	metav1.TypeMeta   `json:",inline"`
> 	metav1.ObjectMeta `json:"metadata,omitempty"`
> 	Status            RouteAgentStatus `json:"status"`
> }
> 
> type RouteAgentStatus struct {
> 	Version       string       `json:"version"`
> 	StatusFailure string       `json:"statusFailure"`
> 	RemoteEndpoints   []RemoteEndpoint `json:"remoteEndpoints"`
> }
> 
> type RemoteEndpoint struct {
> 	Status        ConnectionStatus `json:"status"`
> 	StatusMessage string           `json:"statusMessage"`
> 	Spec      EndpointSpec     `json:"endpoint"`
> 	// +optional
> 	LatencyRTT *LatencyRTTSpec `json:"latencyRTT,omitempty"`
> }
> ```

Thanks @tpantelis for the input. Yes, I was exploring various thoughts and was thinking of reusing the existing gateway status. This idea looks better, I will work in this direction.

pkg/routeagent_driver/main.go
```diff
@@ -126,9 +131,21 @@ func main() {
 	}
 
 	transitSwitchIP := ovn.NewTransitSwitchIP()
+	submSpec := types.SubmarinerSpecification{}
+	logger.FatalOnError(envconfig.Process("submariner", &submSpec), "Error processing env vars")
 
 	config := &watcher.Config{RestConfig: cfg}
 
+	healthcheckerConfig := &healthchecker.Config{
+		PingInterval:         submSpec.HealthCheckInterval * 60,
```
yboaron, line 140: Maybe we should consider running the nodes pinger less frequently than GW2GW pinger

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,334 @@
...
+		h.localNodeName, metav1.DeleteOptions{})
+	if err != nil {
+		logger.Warningf("Error deleting RouteAgent: %s: %v", h.localNodeName, err)
+	}
+
+	return nil
+}
+
+func (h *controller) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	if h.config.HealthCheckerEnabled && h.State().IsOnGateway() {
```
yboaron: Shouldn't it be 
` !h.config.HealthCheckerEnabled || h.State().IsOnGateway()` ?

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -26,17 +26,11 @@ import (
 	"github.com/submariner-io/admiral/pkg/log"
```
yboaron, line 26: If we decide to run the same pinger logic from all nodes regardless if its GW or not, I think it makes sense to include current GW pinger under routeagent pinger,

I mean we can implement for each node :
 - healthcheck pinger to remote endpoints
 - store status under routeAgentResource CR
 
 The GW component can retrieve pinger status from the corresponding routeAgentResource CR

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,334 @@
...
+		delete(h.pingers, endpoint.Spec.CableName)
+	}
+
+	pingerConfig := pinger.Config{
+		IP:                 endpoint.Spec.HealthCheckIP,
+		MaxPacketLossCount: h.config.MaxPacketLossCount,
+	}
+
+	if h.config.PingInterval != 0 {
+		pingerConfig.Interval = time.Second * time.Duration(h.config.PingInterval)
```
yboaron: Do we want to use the same pinger interval for GW and non GW nodes ?

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: good catch, fixed it.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: There is  multiplier 0f 60s while configuring it in main.go.

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: Should we do it now as the g/w health checker is used by other components too, I did not want to change that.

yboaron: I assume we can do it in a follow-up PR
So, ping interval for non-GW node will be 60*GW interval , right ?

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,334 @@
...
+	"github.com/submariner-io/submariner/pkg/event"
+	"github.com/submariner-io/submariner/pkg/pinger"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var (
+	GatewayUpdateInterval = 60 * time.Second
```
yboaron: I assume it should be RouteAgent instead of Gateway here

pkg/cableengine/healthchecker/healthchecker.go
aswinsuryan: yes that is right

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/pinger/pinger.go
```diff
@@ -46,24 +63,26 @@ var (
...
 	IP                 string
 	Interval           time.Duration
 	Timeout            time.Duration
 	MaxPacketLossCount uint
 }
 
-type pingerInfo struct {
-	sync.Mutex
+type Info struct {
+	mu                 sync.Mutex
```
yboaron: Maybe we should use 'mutex' instead 'mu' ?

pkg/routeagent_driver/main.go
```diff
@@ -213,14 +226,10 @@ func uninstall(registry *event.Registry) {
 	}
 }
 
-func waitForNodeReady(k8sClientSet *kubernetes.Clientset) {
+func waitForNodeReady(localNode *corev1.Node) {
```
yboaron: Won't we be stuck in an infinite loop if node state != ready? B/C we only read node information once on line 115?

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+		p.Stop()
+	}
+
+	h.pingers = map[string]pinger.Interface{}
+
+	close(h.stopCh)
+
+	err := h.client.Delete(context.TODO(),
+		h.localNodeName, metav1.DeleteOptions{})
+	if err != nil {
```
yboaron: Don't we want to check if err != not_found ?

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+	remoteEndpoint map[string]*submarinerv1.Endpoint
+	version        string
+	config         *Config
+	stopCh         chan struct{}
+	client         v1typed.RouteAgentInterface
+}
+
+var logger = log.Logger{Logger: logf.Log.WithName("HealthChecker")}
+
+func NewHealthCheckerHandler(config *Config, client v1typed.RouteAgentInterface, version, nodeName string) event.Handler {
```
yboaron: Should we also implement Uninstall event to make sure that all RouteAgent CRs are deleted ?

pkg/pinger/pinger.go
tpantelis: Why not leave it as sync.Mutex?

pkg/pinger/pinger.go
```diff
@@ -46,24 +63,26 @@ var (
...
+type Config struct {
 	IP                 string
 	Interval           time.Duration
 	Timeout            time.Duration
 	MaxPacketLossCount uint
 }
 
-type pingerInfo struct {
-	sync.Mutex
+type Info struct {
```
tpantelis: This is the private implementation and shouldn't be exported.

```suggestion
type pingerImpl struct {
```

pkg/pinger/statistics_internal_test.go
```diff
@@ -16,14 +16,14 @@ See the License for the specific language governing permissions and
...
-package healthchecker
+package pinger
 
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("Statistics", func() {
+var _ = Describe("statistics", func() {
```
tpantelis: Why was this uncapitalized?

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
tpantelis: They'll get deleted when the namespace is deleted.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+		h.localNodeName, metav1.DeleteOptions{})
+	if err != nil {
+		logger.Warningf("Error deleting RouteAgent: %s: %v", h.localNodeName, err)
+	}
+
+	return nil
+}
+
+func (h *controller) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
+	if !h.config.HealthCheckerEnabled && h.State().IsOnGateway() {
```
tpantelis: Shouldn't this be OR?

```suggestion
	if !h.config.HealthCheckerEnabled || h.State().IsOnGateway() {
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
yboaron: Is there any point in running this controller if h.config.HealthCheckerEnabled is false ?

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil
+	}
+
+	h.remoteEndpoint[endpoint.Name] = endpoint
+
+	return h.processEndpointCreatedOrUpdated(endpoint)
+}
+
+func (h *controller) processEndpointCreatedOrUpdated(endpoint *submarinerv1.Endpoint) error {
+	logger.Infof("Endpoint created: %#v", endpoint)
```
tpantelis:
```suggestion
	logger.Infof("Processing Endpoint: %#v", endpoint)
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+			endpoint.Spec.HealthCheckIP, endpoint.Spec.CableName, endpoint.Name)
+		return nil
+	}
+
+	h.Lock()
+	defer h.Unlock()
+
+	if pingerObject, found := h.pingers[endpoint.Spec.CableName]; found {
+		if pingerObject.GetIP() == endpoint.Spec.HealthCheckIP {
+			logger.Infof("HealthChecker did not change %q ", endpoint.Name)
```
tpantelis: I don't think we need to log this or make it TRACE.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+	newPingerFunc := h.config.NewPinger
+	if newPingerFunc == nil {
+		newPingerFunc = pinger.NewPinger
+	}
+
+	pingerObject := newPingerFunc(pingerConfig)
+	h.pingers[endpoint.Spec.CableName] = pingerObject
+	pingerObject.Start()
+
+	logger.Infof("CableEngine HealthChecker started pinger for CableName: %q with HealthCheckIP %q",
```
tpantelis:
```suggestion
	logger.Infof("HealthChecker started pinger for CableName: %q with HealthCheckIP %q",
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+// TransitionToNonGateway is called once for each transition of the local node from Gateway to a non-Gateway.
+func (h *controller) TransitionToNonGateway() error {
+	if h.config.HealthCheckerEnabled {
+		for i := range h.remoteEndpoint {
+			err := h.processEndpointCreatedOrUpdated(h.remoteEndpoint[i])
+			if err != nil {
+				logger.Warningf("Error processing remote endpoint %s: %v", h.remoteEndpoint[i].Name, err)
+			}
```
tpantelis: Perhaps we should return the error so it's retried?

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+					CableName:     endpoint.Spec.CableName,
+					HealthCheckIP: endpoint.Spec.HealthCheckIP,
+					Hostname:      endpoint.Spec.Hostname,
+					Subnets:       endpoint.Spec.Subnets,
+					PrivateIP:     endpoint.Spec.PrivateIP,
+					PublicIP:      endpoint.Spec.PublicIP,
+					NATEnabled:    endpoint.Spec.NATEnabled,
+					Backend:       endpoint.Spec.Backend,
+					BackendConfig: endpoint.Spec.BackendConfig,
+				},
```
tpantelis:
```suggestion
				Spec: endpoint.Spec,
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+	for _, endpoint := range h.remoteEndpoint {
+		pingerObject, found := h.pingers[endpoint.Spec.CableName]
+		if !found {
+			logger.Warningf("Pinger not found for %q", endpoint.Spec.CableName)
+			continue
+		}
+
+		latencyInfo := pingerObject.GetLatencyInfo()
+		if latencyInfo != nil {
+			connectionStatus := submarinerv1.ConnectionStatus(latencyInfo.ConnectionStatus)
```
tpantelis: These are different enums so we shouldn't treat them as equivalent. We should use a switch statement to translate.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+				LatencyRTT: &submarinerv1.LatencyRTTSpec{
+					Last:    latencyInfo.Spec.Last,
+					Min:     latencyInfo.Spec.Min,
+					Average: latencyInfo.Spec.Average,
+					Max:     latencyInfo.Spec.Max,
+					StdDev:  latencyInfo.Spec.StdDev,
+				},
+			}
+
+			routeAgent.Status.RemoteEndpoints = append(routeAgent.Status.RemoteEndpoints, remoteEndpoint)
```
tpantelis: I think we should add a `RemoteEndpoint` entry even if there's no `LatencyInfo`.  In this case, the `Status` would be `submarinerv1.Connecting`.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+
+func (h *controller) generateRouteAgentObject() *submarinerv1.RouteAgent {
+	return &submarinerv1.RouteAgent{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "submariner.io/v1",
+			Kind:       "RouteAgent",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        h.localNodeName,
+			Annotations: map[string]string{UpdateTimestampAnnotation: strconv.FormatInt(time.Now().UTC().Unix(), 10)},
```
tpantelis: This annotation is not necessary. It's used for the `Gateway` CR for cleaning up stale instances but that doesn't apply here.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+		return
+	}
+
+	if result == util.OperationResultCreated {
+		logger.Infof("RouteAgent does not exist - created: %+v", routeAgent)
+	} else if result == util.OperationResultUpdated {
+		logger.Infof("RouteAgent already exists - updated: %+v", routeAgent)
+	} else {
+		logger.Info("RouteAgent already exists but doesn't need updating")
+	}
```
tpantelis: The above logs should at least be reduced to TRACE level but I don't think they're necessary at all.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+
+			routeAgent.Status.RemoteEndpoints = append(routeAgent.Status.RemoteEndpoints, remoteEndpoint)
+		}
+	}
+
+	// Use CreateOrUpdate to handle the RouteAgent resource
+	result, err := util.CreateOrUpdate(context.TODO(), h.routeAgentResourceInterface(), routeAgent,
+		func(existing *submarinerv1.RouteAgent) (*submarinerv1.RouteAgent, error) {
+			existing.TypeMeta = routeAgent.TypeMeta
```
tpantelis: This is unnecessary.

```suggestion
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+			existing.Status = routeAgent.Status
+
+			if existing.Annotations == nil {
+				existing.Annotations = map[string]string{}
+			}
+
+			existing.Annotations[UpdateTimestampAnnotation] = routeAgent.Annotations[UpdateTimestampAnnotation]
```
tpantelis: As noted elsewhere, the `UpdateTimestampAnnotation` isn't needed.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+	HealthCheckerEnabled bool
+	NewPinger            func(pinger.Config) pinger.Interface
+}
+
+type controller struct {
+	event.HandlerBase
+	sync.RWMutex
+	pingers        map[string]pinger.Interface
+	localNodeName  string
+	remoteEndpoint map[string]*submarinerv1.Endpoint
```
tpantelis: You don't need to track these as `event.HandlerBase` does that already - they're accessible via `State().GetRemoteEndpoints()`.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+type Config struct {
+	PingInterval         uint
+	MaxPacketLossCount   uint
+	HealthCheckerEnabled bool
+	NewPinger            func(pinger.Config) pinger.Interface
+}
+
+type controller struct {
+	event.HandlerBase
+	sync.RWMutex
```
tpantelis: You don't really need a `RWMutex` here.

```suggestion
	syncMutex
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+// TransitionToGateway is called once for each transition of the local node from non-Gateway to a Gateway.
+func (h *controller) TransitionToGateway() error {
+	if h.config.HealthCheckerEnabled {
+		for i := range h.pingers {
+			h.pingers[i].Stop()
+			delete(h.pingers, i)
+		}
+	}
+
+	close(h.stopCh)
```
tpantelis: This will permanently stop the syncer goroutine and it won't get restarted should it transition back to gateway. The syncer goroutine is started in `Init` so is intended to run until shutdown.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+
+	err := h.client.Delete(context.TODO(), h.localNodeName, metav1.DeleteOptions{})
+	if err != nil {
+		logger.Warningf("Error deleting RouteAgent: %s: %v", h.localNodeName, err)
+	}
```
tpantelis: We shouldn't delete the `RouteAgent` here. We could call `syncRouteAgentStatus` instead of waiting for the syncer goroutine.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+
+	delete(h.remoteEndpoint, endpoint.Name)
+
+	return nil
+}
+
+func (h *controller) Init() error {
+	go func() {
+		wait.Until(h.syncRouteAgentStatus, RouteAgentUpdateInterval, h.stopCh)
```
tpantelis: `syncRouteAgentStatus` accesses shared data so we need to lock.

```suggestion
		wait.Until( func() {
		    h.Lock()
	            defer h.Unlock()
	            
		    h.syncRouteAgentStatus()
		}, RouteAgentUpdateInterval, h.stopCh)
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+	}()
+
+	return nil
+}
+
+// TransitionToNonGateway is called once for each transition of the local node from Gateway to a non-Gateway.
+func (h *controller) TransitionToNonGateway() error {
+	if h.config.HealthCheckerEnabled {
```
tpantelis, line 193: Need to lock here.

```suggestion
        h.Lock()
	defer h.Unlock()
	
	if h.config.HealthCheckerEnabled {
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// TransitionToGateway is called once for each transition of the local node from non-Gateway to a Gateway.
+func (h *controller) TransitionToGateway() error {
+	if h.config.HealthCheckerEnabled {
```
tpantelis, line 212: Need to lock here.

```suggestion
         h.Lock()
	defer h.Unlock()
	
	if h.config.HealthCheckerEnabled {
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,331 @@
...
+func (h *controller) syncRouteAgentStatus() {
+	routeAgent := h.generateRouteAgentObject()
+	routeAgent.Status.RemoteEndpoints = []submarinerv1.RemoteEndpoint{}
+
+	for _, endpoint := range h.remoteEndpoint {
+		pingerObject, found := h.pingers[endpoint.Spec.CableName]
+		if !found {
+			logger.Warningf("Pinger not found for %q", endpoint.Spec.CableName)
+			continue
+		}
```
tpantelis: A pinger will be missing if health check is not enabled or we're on the gateway. We should still create a `RemoteEndpoint` entry and set the `StatusMessage`, ie if `!HealthCheckerEnabled` then "_Health check is not enabled_" or if `IsOnGateway` then "_Health check is not performed on gateway nodes"_.  We could leave the `Status` field empty but it's defined as an enum. None of the current  enums apply so we can add `ConnectionNone`.

test/e2e/dataplane/gateway_status.go
```diff
@@ -137,7 +137,7 @@ func verifyGateway(gw *submarinerv1.Gateway, otherCluster string, healthCheckedE
 			}
 
 			if gw.Status.Connections[i].LatencyRTT.StdDev == "" {
-				return false, fmt.Sprintf("Connection for cluster %q is missing StdDev RTT data", otherCluster), nil
+				return false, fmt.Sprintf("Connection for cluster %q is missing stdDev RTT data", otherCluster), nil
```
tpantelis: Did you mean to uncapitalize this?

pkg/routeagent_driver/main.go
tpantelis: yeah this code needs to read the localNode each time so it can't be passed in.

pkg/routeagent_driver/main.go
```diff
@@ -110,8 +112,11 @@ func main() {
 	restMapper, err := util.BuildRestMapper(cfg)
 	logger.FatalOnError(err, "Error building the REST mapper")
 
+	localNode, err := node.GetLocalNode(k8sClientSet)
+	logger.FatalOnError(err, "Error getting information on the local node")
```
tpantelis: Move this after L122

pkg/routeagent_driver/main.go
```diff
@@ -126,9 +131,21 @@ func main() {
...
 
+	healthcheckerConfig := &healthchecker.Config{
+		PingInterval:         submSpec.HealthCheckInterval * 60,
+		MaxPacketLossCount:   submSpec.HealthCheckMaxPacketLossCount,
+		HealthCheckerEnabled: submSpec.HealthCheckEnabled,
+	}
+
+	if err != nil {
+		logger.Errorf(err, "Error creating healthChecker")
+	}
```
tpantelis: This is out of place.

```suggestion
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
tpantelis:
```suggestion
func New(config *Config, client v1typed.RouteAgentInterface, version, nodeName string) event.Handler {
```

tpantelis (changes requested):
The RA health checker needs unit tests 😄

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
tpantelis: yeah this also maintains the `RouteAgent` CR which we always want running. We can still populate the `RemoteEndpoints` even with no connection status info.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/pinger/pinger.go
aswinsuryan: it was a suggestion given my linter , may be when the type names where different , now it is happy so leaving it as sync.Mutex

aswinsuryan:
> The RA health checker needs unit tests 😄

I will add the UTs. :)

pkg/pinger/pinger.go
tpantelis: It's b/c you had changed the type name to "Info" which exports it.

pkg/pinger/pinger.go
```diff
@@ -154,8 +173,8 @@ func (p *pingerInfo) doPing() error {
 
 		// Pinger will mark a connection as an error if the packet loss reaches the threshold
 		if pinger.PacketsSent-pinger.PacketsRecv > int(p.maxPacketLossCount) {
-			p.Lock()
-			defer p.Unlock()
+			p.Mutex.Lock()
```
tpantelis: Don't need to reference `Mutex`.

```suggestion
			p.Lock()
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,305 @@
...
+		p.Stop()
+	}
+
+	h.pingers = map[string]pinger.Interface{}
+
+	close(h.stopCh)
+
+	err := h.client.Delete(context.TODO(),
+		h.localNodeName, metav1.DeleteOptions{})
+	if err != nil && apiError.IsNotFound(err) {
```
tpantelis:
```suggestion
	if err != nil && !apiError.IsNotFound(err) {
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
tpantelis: The `Lock()` needs to be inside the func passed to `Until` - refer to the suggested change above. You want to protect the call to `h.syncRouteAgentStatus()`.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
tpantelis: This wasn't addressed.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	// Use CreateOrUpdate to handle the RouteAgent resource
+	_, err := util.CreateOrUpdate(context.TODO(), h.routeAgentResourceInterface(), routeAgent,
+		func(existing *submarinerv1.RouteAgent) (*submarinerv1.RouteAgent, error) {
+			existing.Status = routeAgent.Status
+
+			if existing.Annotations == nil {
+				existing.Annotations = map[string]string{}
+			}
```
tpantelis: Not needed.

```suggestion
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Oh sorry,  Did an undo it between and it got reverted, fixed now.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: oh I took the comment wrongly, corrected it.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,325 @@
...
+		MaxPacketLossCount: h.config.MaxPacketLossCount,
+	}
+
+	if h.config.PingInterval != 0 {
+		pingerConfig.Interval = time.Second * time.Duration(h.config.PingInterval)
+	}
+
+	if h.config.MaxPacketLossCount != 0 {
+		pingerConfig.MaxPacketLossCount = h.config.MaxPacketLossCount
+	}
```
tpantelis, line 146: This is redundant since you've already set `MaxPacketLossCount` on L138.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,325 @@
...
+	h.Lock()
+	defer h.Unlock()
+
+	if h.config.HealthCheckerEnabled {
+		remoteEndpoints := h.State().GetRemoteEndpoints()
+
+		for i := range remoteEndpoints {
+			err := h.processEndpointCreatedOrUpdated(&remoteEndpoints[i])
+			if err != nil {
+				logger.Warningf("Error processing remote endpoint %s: %v", remoteEndpoints[i].Name, err)
```
tpantelis: I think we should return the error here.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,325 @@
...
+	remoteEndpoints := h.State().GetRemoteEndpoints()
+
+	for i := range remoteEndpoints {
+		pingerObject, found := h.pingers[remoteEndpoints[i].Spec.CableName]
+		if !found {
+			logger.Warningf("Pinger not found for %q", remoteEndpoints[i].Spec.CableName)
+			continue
+		}
+
+		latencyInfo := pingerObject.GetLatencyInfo()
```
tpantelis: There won't be a pinger if `!h.config.HealthCheckerEnabled` or `h.State().IsOnGateway()` so this needs to be after L256.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,325 @@
...
+		remoteEndpoint = submarinerv1.RemoteEndpoint{
+			Status:        connectionStatus,
+			StatusMessage: statusMessage,
+			Spec:          remoteEndpoints[i].Spec,
+			LatencyRTT: &submarinerv1.LatencyRTTSpec{
+				Last:    latencyInfo.Spec.Last,
+				Min:     latencyInfo.Spec.Min,
+				Average: latencyInfo.Spec.Average,
+				Max:     latencyInfo.Spec.Max,
+				StdDev:  latencyInfo.Spec.StdDev,
```
tpantelis: `latencyInfo` could be nil here. I'd suggest initializing the `RemoteEndpoint` above and filling in the fields appropriately:

```
remoteEndpoint := submarinerv1.RemoteEndpoint{
    Status: submarinerv1.Connecting,
    Spec:          remoteEndpoints[i].Spec,
}

if !h.config.HealthCheckerEnabled {
    remoteEndpoint.Status = submarinerv1.ConnectionNone
    remoteEndpoint.StatusMessage = "Health check is not enabled"
} else if h.State().IsOnGateway() {
    remoteEndpoint.Status = submarinerv1.ConnectionNone
    remoteEndpoint.StatusMessage = "Health check is not performed on gateway nodes"
} else if pingerObject, found := h.pingers[remoteEndpoints[i].Spec.CableName]; found {
     latencyInfo := pingerObject.GetLatencyInfo()
     if latencyInfo != nil {
         switch latencyInfo.ConnectionStatus {
             ....
	 }

         remoteEndpoint.LatencyRTT =  &submarinerv1.LatencyRTTSpec {
             ...
         }
     }
} 

routeAgent.Status.RemoteEndpoints = append(routeAgent.Status.RemoteEndpoints, remoteEndpoint)
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,325 @@
...
+		return
+	}
+}
+
+func (h *controller) generateRouteAgentObject() *submarinerv1.RouteAgent {
+	return &submarinerv1.RouteAgent{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "submariner.io/v1",
+			Kind:       "RouteAgent",
+		},
```
tpantelis: This isn't needed.

```suggestion
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,325 @@
...
+	return []string{event.AnyNetworkPlugin}
+}
+
+func (h *controller) GetName() string {
+	return "routeAgent-health-checker"
+}
+
+func (h *controller) syncRouteAgentStatus() {
+	routeAgent := h.generateRouteAgentObject()
+	routeAgent.Status.RemoteEndpoints = []submarinerv1.RemoteEndpoint{}
```
tpantelis: This is already initialized in `generateRouteAgentObject`.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Removed it from L138

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Refactored the code and added a new variable latencyInfo

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,324 @@
...
+			statusMessage = "Health check is not enabled"
+		} else if h.State().IsOnGateway() {
+			connectionStatus = submarinerv1.ConnectionNone
+			statusMessage = "Health check is not performed on gateway nodes"
+		} else {
+			pingerObject, found := h.pingers[remoteEndpoints[i].Spec.CableName]
+			if !found {
+				logger.Warningf("Pinger not found for %q", remoteEndpoints[i].Spec.CableName)
+				continue
+			}
```
tpantelis: I don't think we need a warning here. This could normally happen in the small window between when a `RemoteEndpoint` is added to the cache and the handler is invoked. The next sync period would pick it up. Either way we can still add it with status `submarinerv1.Connecting`.  That's why I initialized `remoteEndpoint` before this block [here](https://github.com/submariner-io/submariner/pull/3077#discussion_r1724161667).  It ensures `Status`  is always set and eliminates the need for the else on L276.

```suggestion
		} else if pingerObject, found := h.pingers[remoteEndpoints[i].Spec.CableName]; found {
			pingerObject, found := h.pingers[remoteEndpoints[i].Spec.CableName]
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
aswinsuryan: Done.

tpantelis (approved):

yboaron (approved):

PR #3080, opened 2024-07-15
Remove TODO in libreswan.Init

tpantelis:
...for "_Check whether the file already exists_". This isn't necessary as os.Create truncates/overwrites an existing file.

### DIFF
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -157,7 +157,6 @@ func (i *libreswan) GetName() string {
 func (i *libreswan) Init() error {
 	// Write the secrets file:
 	// %any %any : PSK "secret"
-	// TODO Check whether the file already exists
 	file, err := os.Create("/etc/ipsec.d/submariner.secrets")
 	if err != nil {
 		return errors.Wrap(err, "error creating the secrets file")


pkg/cable/libreswan/libreswan.go
```diff
@@ -157,7 +157,6 @@ func (i *libreswan) GetName() string {
 func (i *libreswan) Init() error {
 	// Write the secrets file:
 	// %any %any : PSK "secret"
-	// TODO Check whether the file already exists
```
vthapar, line 160: Do we want os.Create to overwrite existing file or use existing? Rest of code implies okay to overwrite, but just want to be sure we don't lose intent of the TODO.

tpantelis: I'm not sure of the original intent of the TODO but I think it could only use existing if it contained the same PSK (which is most of the time) but it seems simpler to just overwrite each time. This code and the TODO has been the same and working for 5 years so I see no compelling reason to try to change it, and thus no reason to keep the TODO (it's flagged by Sonar).

sridhargaddam: os.Create will either create or truncate the file if it already exists. Since we always want to use the latest PSK that is configured as part of the `submariner-gateway` pod, we should be good.

sridhargaddam (approved):

yboaron (approved):

PR #3081, opened 2024-07-15
Remove metrics.RecordNoConnections

tpantelis:
This function is only called by `libreswan.refreshConnectionStatus` to delete all current metrics but really isn't necessary (nor desired) since the metrics are recorded from the stored connections so they're immediately re-added.

Removing `RecordNoConnections` also eliminates a TODO which would become relevant if multiple concurrent cable drivers are ever supported.


pkg/cable/libreswan/libreswan.go
```diff
@@ -243,8 +243,6 @@ func (i *libreswan) refreshConnectionStatus() error {
 		return err
 	}
 
-	cable.RecordNoConnections()
```
yboaron, line 245: IIUC, with this change we rely on [DisconnectFromEndpoint](https://github.com/submariner-io/submariner/blob/devel/pkg/cable/libreswan/libreswan.go#L579) to clear connection metrics.

Could there be a situation where we will miss the DisconnectFromEndpoint event for some reason?

Also I think we need to make sure that [cable.RecordDisconnected](https://github.com/submariner-io/submariner/blob/devel/pkg/cable/libreswan/libreswan.go#L579) function is called (even if DisconnectFromEndpoint returns error)

tpantelis: > IIUC, with this change we rely on [DisconnectFromEndpoint](https://github.com/submariner-io/submariner/blob/devel/pkg/cable/libreswan/libreswan.go#L579) to clear connection metrics.
> 
> Could there be a situation where we will miss the DisconnectFromEndpoint event for some reason?
> 

Not that I know of. The other 2 cable drivers rely on it, ie they don't call `RecordNoConnections` in any place. `DisconnectFromEndpoint` is called when a remote `Endpoint`  is removed so we'd have much bigger problems than a stale metric if the event was missed 😄 

> Also I think we need to make sure that [cable.RecordDisconnected](https://github.com/submariner-io/submariner/blob/devel/pkg/cable/libreswan/libreswan.go#L579) function is called (even if DisconnectFromEndpoint returns error)

If `DisconnectFromEndpoint` returned an error, the caller will keep retrying. This would mean at least some remnants of the connection remains until it succeeds but we can remove the metric earlier in the function to avoid errors.

`RecordNoConnections` isn't ideal b/c it blindly deletes all metrics for the vector. It was added by https://github.com/submariner-io/submariner/pull/942 - it's unclear to me why.  @maayanf24 Do you recall why?

yboaron: > If DisconnectFromEndpoint returned an error, the caller will keep retrying. This would mean at least some remnants of the connection remains until it succeeds but we can remove the metric earlier in the function to avoid errors.

Yep, I missed the event retry part, in that case I think we can be safe with this change.

yboaron (approved):

dfarrell07 (approved):

pkg/cable/libreswan/libreswan.go
maayanf24: I don't recall the details..

PR #3084, opened 2024-07-17
Address underlying gateway failover issue with wireguard

tpantelis:
https://github.com/submariner-io/submariner/pull/3069 addressed this issue with a workaround in the route-agent's `kubeproxy` handler but the underlying root cause is due to a missed delete event in admiral's resource syncer.  The `wireguard` driver sets a “_public-key_” entry in the local `Endpoint`'s  `BackendConfig` map which is used by the remote clusters to establish a connection. This is a random-generated key so is different on each startup. So with `wireguard`, a new local `Endpoint` is always created on startup and the prior one deleted. However, client listeners, specifically the `kubeproxy` handler, may not observe the delete event due to timing which results in a failure of the gateway failover E2E test. This is addressed by https://github.com/submariner-io/admiral/pull/953. So this PR bumps the admiral dependency to pick up the change and removes the workaround by https://github.com/submariner-io/submariner/pull/3069.

Depends on https://github.com/submariner-io/admiral/pull/953

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/projectcalico/api v0.0.0-20230602153125-fb7148692637
 	github.com/prometheus-community/pro-bing v0.4.0
 	github.com/prometheus/client_golang v1.19.1
-	github.com/submariner-io/admiral v0.19.0-m0
+	github.com/submariner-io/admiral v0.19.0-m0.0.20240718120154-11bb4f8f51a2
 	github.com/submariner-io/shipyard v0.19.0-m0
 	github.com/vishvananda/netlink v1.2.1-beta.2
 	golang.org/x/net v0.27.0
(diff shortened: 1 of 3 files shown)


yboaron:
Nice catch!

I can see that Wireguard gates are green even before https://github.com/submariner-io/admiral/pull/953 got merged , could it be that bumping  controller-runtime to v0.18.4 addressed this issue?

yboaron (approved):

maayanf24 (approved):

tpantelis:
> Nice catch!
> 
> I can see that Wireguard gates are green even before [submariner-io/admiral#953](https://github.com/submariner-io/admiral/pull/953) got merged , could it be that bumping controller-runtime to v0.18.4 addressed this issue?

This PR's [commit](https://github.com/submariner-io/submariner/pull/3084/commits/dbcdf2df91ba2b57a29204519f6638bf882f8006) pulls in the commit in https://github.com/submariner-io/admiral/pull/953 so it's verifying the admiral fix.

tpantelis:
> This PR's [commit](https://github.com/submariner-io/submariner/pull/3084/commits/dbcdf2df91ba2b57a29204519f6638bf882f8006) pulls in the commit in [submariner-io/admiral#953](https://github.com/submariner-io/admiral/pull/953) so it's verifying the admiral fix.

https://github.com/submariner-io/admiral/pull/953 is merged now so I've updated the dependency to the merged commit.

PR #3142, opened 2024-08-27
Bump shipyard and address golangci-lint v1.60.3 errors

tpantelis:
Mostly gosec "integer overflow conversion" issues.

Depends on https://github.com/submariner-io/shipyard/pull/1715


skitt (approved):
... but I don’t think the Shipyard bump is necessary.

tpantelis:
> ... but I don’t think the Shipyard bump is necessary.

It is b/c some shipyard e2e APIs changed due to `gosec` strictness.

dfarrell07 (approved):

PR #3144, opened 2024-08-30
Rework integer type selection

skitt:
Following the recent gosec-related int work (to annotate known-safe conversions), this furthers that by:

* using ints instead of uints for healthcheck configuration and tracking
* using uint32 instead of int for IP pool sizes, since their use is tied to uint32 representations of IP addresses
* using fmt.Sprintf instead of manual string concatenation to construct the connectivity verification command


pkg/ipam/ippool.go
```diff
@@ -26,14 +26,15 @@ import (
 	"sync"
```
tpantelis: This file is being moved to admiral : https://github.com/submariner-io/admiral/pull/982. I can make these changes with my PR or you can submit a new one.

pkg/cableengine/healthchecker/pinger.go
```diff
@@ -34,7 +34,7 @@ const (
 )
```
tpantelis: Note that this file is being moved by https://github.com/submariner-io/submariner/pull/3077.

tpantelis (approved):

tpantelis:
The linting failures need the [shipyard golangci-lint bump](https://github.com/submariner-io/shipyard/pull/1726).

PR #3152, opened 2024-09-03
Add UT for route agent health checker

aswinsuryan:
Add UT for route agent health checker

### DIFF
--- a/pkg/routeagent_driver/handlers/healthchecker/health_checker_suite_test.go
+++ b/pkg/routeagent_driver/handlers/healthchecker/health_checker_suite_test.go
@@ -1,0 +1,43 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthchecker_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/submariner-io/admiral/pkg/log/kzerolog"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func init() {
+	kzerolog.AddFlags(nil)
+}
+
+var _ = BeforeSuite(func() {
+	kzerolog.InitK8sLogging()
+	Expect(submarinerv1.AddToScheme(scheme.Scheme)).To(Succeed())
+})
+
+func TestHealthChecker(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "RouteAgent Health Checker Suite")
+}
--- a/pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
+++ b/pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
@@ -1,0 +1,407 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthchecker_test
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	fakeClient "github.com/submariner-io/submariner/pkg/client/clientset/versioned/fake"
+	submarinerv1client "github.com/submariner-io/submariner/pkg/client/clientset/versioned/typed/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/event"
+	eventtesting "github.com/submariner-io/submariner/pkg/event/testing"
+	"github.com/submariner-io/submariner/pkg/pinger"
+	"github.com/submariner-io/submariner/pkg/pinger/fake"
+	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/healthchecker"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	kubeScheme "k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	namespace       = "submariner"
+	remoteClusterID = "west"
+	healthCheckIP1  = "1.1.1.1"
+	healthCheckIP2  = "2.2.2.2"
+	localNodeName   = "nodeName"
+)
+
+var _ = Describe("RouteAgent syncing", func() {
+	t := newTestDriver()
+	It("should create a RouteAgent resource", func() {
+		var routeAgent *submarinerv1.RouteAgent
+		Eventually(func() bool {
+			var err error
+			routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false
+			}
+
+			Expect(err).To(Succeed())
+			return true
+		}).Should(BeTrue())
+		Expect(routeAgent.Status.RemoteEndpoints).To(BeEmpty())
+	})
+
+	When("a remote Endpoint is created/updated/deleted", func() {
+		It("should add/update/delete its RemoteEndpoint information to the RouteAgent resource", func() {
+			endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(healthCheckIP1))
+
+			var routeAgent *submarinerv1.RouteAgent
+			Eventually(func() bool {
+				var err error
+				routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
+					return false
+				}
+
+				Expect(err).To(Succeed())
+				return len(routeAgent.Status.RemoteEndpoints) != 0
+			}).Should(BeTrue())
+
+			remoteEndpoint := routeAgent.Status.RemoteEndpoints[0]
+			Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+
+			endpoint1.Spec.Hostname = "newHostName"
+			t.UpdateEndpoint(endpoint1)
+
+			Eventually(func() string {
+				var err error
+				routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
+					return ""
+				}
+
+				Expect(err).To(Succeed())
+				if len(routeAgent.Status.RemoteEndpoints) != 0 {
+					return routeAgent.Status.RemoteEndpoints[0].Spec.Hostname
+				}
+				return ""
+			}).Should(Equal(endpoint1.Spec.Hostname))
+
+			remoteEndpoint = routeAgent.Status.RemoteEndpoints[0]
+			Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+
+			t.DeleteEndpoint(endpoint1.Name)
+			Eventually(func() bool {
+				var err error
+				routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
+					return false
+				}
+
+				Expect(err).To(Succeed())
+				return len(routeAgent.Status.RemoteEndpoints) == 0
+			}).Should(BeTrue())
+		})
+	})
+})
+
+var _ = Describe("RemoteEndpoint latency info", func() {
+	When("a remote Endpoint is created", func() {
+		t := newTestDriver()
+		It("should start a pinger and correctly update the RemoteEndpoint Status and LatencyInfo", func() {
+			endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(healthCheckIP1))
+			t.pingerMap[healthCheckIP1].AwaitStart()
+
+			latencyInfo := t.newLatencyInfo()
+			t.setLatencyInfo(healthCheckIP1, latencyInfo)
+
+			var routeAgent *submarinerv1.RouteAgent
+			Eventually(func() *submarinerv1.LatencyRTTSpec {
+				var err error
+				routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
+					return nil
+				}
+
+				Expect(err).To(Succeed())
+				if len(routeAgent.Status.RemoteEndpoints) != 0 && routeAgent.Status.RemoteEndpoints[0].LatencyRTT != nil {
+					return routeAgent.Status.RemoteEndpoints[0].LatencyRTT
+				}
+				return nil
+			}).Should(Equal(latencyInfo.Spec))
+
+			remoteEndpoint := routeAgent.Status.RemoteEndpoints[0]
+			Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+			Expect(remoteEndpoint.Status).To(Equal(submarinerv1.Connected))
+		})
+		Context("with no HealthCheckIP", func() {
+			It("should not start a pinger and should set the RemoteEndpoint Status to None", func() {
+				endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(""))
+				t.pingerMap[healthCheckIP1].AwaitNoStart()
+
+				var routeAgent *submarinerv1.RouteAgent
+				Eventually(func() submarinerv1.ConnectionStatus {
+					var err error
+					routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+					if apierrors.IsNotFound(err) {
+						return submarinerv1.Connecting
+					}
+
+					Expect(err).To(Succeed())
+					if err == nil && len(routeAgent.Status.RemoteEndpoints) != 0 {
+						return routeAgent.Status.RemoteEndpoints[0].Status
+					}
+					return submarinerv1.Connecting
+				}).Should(Equal(submarinerv1.ConnectionNone))
+
+				remoteEndpoint := routeAgent.Status.RemoteEndpoints[0]
+				Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+				Expect(remoteEndpoint.Status).To(Equal(submarinerv1.ConnectionNone))
+			})
+		})
+		Context("on the gateway", func() {
+			It("should not start a pinger and should set the RemoteEndpoint Status to None", func() {
+				_ = t.CreateLocalHostEndpoint()
+				endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(healthCheckIP1))
+				t.pingerMap[healthCheckIP1].AwaitNoStart()
+
+				var routeAgent *submarinerv1.RouteAgent
+				Eventually(func() submarinerv1.ConnectionStatus {
+					var err error
+					routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+					if apierrors.IsNotFound(err) {
+						return submarinerv1.Connecting
+					}
+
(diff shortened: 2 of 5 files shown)


pkg/routeagent_driver/handlers/healthchecker/health_checker_suite_test.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+var _ = BeforeSuite(func() {
+	kzerolog.InitK8sLogging()
+	Expect(submarinerv1.AddToScheme(scheme.Scheme)).To(Succeed())
+})
+
+func TestHealthChecker(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "RouteAgent Health Checker")
```
tpantelis:
```suggestion
	RunSpecs(t, "RouteAgent Health Checker Suite")
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,277 @@
...
+func newTestDriver() *testDriver {
+	t := &testDriver{
+		ControllerSupport: eventtesting.NewControllerSupport(),
+	}
+
+	BeforeEach(func() {
+		t.stopCh = make(chan struct{})
+		scheme := runtime.NewScheme()
+		Expect(submarinerv1.AddToScheme(scheme)).To(Succeed())
+		Expect(submarinerv1.AddToScheme(kubeScheme.Scheme)).To(Succeed())
```
tpantelis: This is already done in `BeforeSuite`.

```suggestion
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+func newTestDriver() *testDriver {
+	t := &testDriver{
+		ControllerSupport: eventtesting.NewControllerSupport(),
+	}
+
+	BeforeEach(func() {
+		t.stopCh = make(chan struct{})
+		scheme := runtime.NewScheme()
+		Expect(submarinerv1.AddToScheme(scheme)).To(Succeed())
```
tpantelis: This isn't needed - `scheme` is unused.

```suggestion
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+		dynamicClient := dynamicfake.NewSimpleDynamicClient(kubeScheme.Scheme)
+
+		t.endpoints = dynamicClient.Resource(submarinerv1.SchemeGroupVersion.WithResource("endpoints")).Namespace(namespace)
+		t.client = clientset.SubmarinerV1().RouteAgents(namespace)
+		t.pingerMap = map[string]*fake.Pinger{
+			healthCheckIP1: fake.NewPinger(healthCheckIP1),
+		}
+
+		t.localNodeName = "nodeName"
```
tpantelis: This is really a constant and can be defined as such.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+		config := &healthchecker.Config{
+			PingInterval:             1, // Set interval to 1 second for faster testing
+			MaxPacketLossCount:       1,
+			HealthCheckerEnabled:     true,
+			RouteAgentUpdateInterval: 2 * time.Second,
```
tpantelis: You can make this a lot less to speed up the tests

```suggestion
			RouteAgentUpdateInterval: 100 * time.Millisecond,
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,277 @@
...
+			Expect(pingerCfg.MaxPacketLossCount).To(Equal(config.MaxPacketLossCount))
+
+			p, ok := t.pingerMap[pingerCfg.IP]
+			Expect(ok).To(BeTrue())
+
+			return p
+		}
+
+		t.handler = healthchecker.New(config, t.client, "v1", t.localNodeName)
+		Expect(t.handler.Init()).To(Succeed())
```
tpantelis: You shouldn't call `Init` - ControllerSupport does that.

```suggestion
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,277 @@
...
+	healthCheckIP2  = "2.2.2.2"
+)
+
+var _ = Describe("RouteAgent Health Checker", func() {
+	t := newTestDriver()
+
+	When("syncRouteAgentStatus is triggered", func() {
+		It("should create or update the RouteAgent object and check LatencyInfo", func() {
+			endpoint1 := t.CreateEndpoint(t.newSubmEndpoint())
+			err = t.handler.RemoteEndpointCreated(endpoint1)
```
tpantelis: You shouldn't call `RemoteEndpointCreated` directly - the ControllerSupport will do that when triggered by `Endpoint` creation.

tpantelis (changes requested):

tpantelis:
I think we should structure the test specs as follows:

```
var _ = Describe("RouteAgent syncing", func() {
    It("should create a RouteAgent resource", func() {
         // await the RouteAgent - no RemoteEndpoint info expected
    })

    When("a remote Endpoint is created", func() {
        It("should add its RemoteEndpoint information to the RouteAgent resource", func() {
            // await the RemoteEndpoint and verify the Spec
        })
    })

    When("a remote Endpoint is updated", func() {
        It("should update its RemoteEndpoint information in the RouteAgent resource", ...)

    When("a remote Endpoint is deleted", func() {
        It("should remove its RemoteEndpoint information from the RouteAgent resource", ...)
})

var _ = Describe("RemoteEndpoint latency info", func() {
    When("a remote Endpoint is created", func() {
        It("should start a pinger and correctly update the RemoteEndpoint Status and LatencyInfo", func() {
            // await the RemoteEndpoint and verify the LatencyInfo and Status
        })

        Context("with no HealthCheckIP", func() {
		It("should not start a pinger and should set the RemoteEndpoint Status to None", ...)
	})

        Context("with health check not enabled", func() {
		It("should not start a pinger and should set the RemoteEndpoint Status to None", ...)
	})

        When("on the gateway", func() {
		It("should not start a pinger and should set the RemoteEndpoint Status to None", ...)
                 // Call ControllerSupport.CreateLocalHostEndpoint() before creating the remote EP
	})
    })

    When("a remote Endpoint is updated", func() {
        Context("and the HealthCheckIP was changed", func() {
            It("should stop the pinger and start a new one", ...)
        })

        Context("and the HealthCheckIP did not change", func() {
            It("should not start a new pinger", ...)
        })
    })

    When("a remote Endpoint is deleted", func() {
        It("should stop the pinger", ...)
})
```

pkg/routeagent_driver/handlers/healthchecker/health_checker_suite_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

aswinsuryan:
> I think we should structure the test specs as follows:
> 
> ```
> var _ = Describe("RouteAgent syncing", func() {
>     It("should create a RouteAgent resource", func() {
>          // await the RouteAgent - no RemoteEndpoint info expected
>     })
> 
>     When("a remote Endpoint is created", func() {
>         It("should add its RemoteEndpoint information to the RouteAgent resource", func() {
>             // await the RemoteEndpoint and verify the Spec
>         })
>     })
> 
>     When("a remote Endpoint is updated", func() {
>         It("should update its RemoteEndpoint information in the RouteAgent resource", ...)
> 
>     When("a remote Endpoint is deleted", func() {
>         It("should remove its RemoteEndpoint information from the RouteAgent resource", ...)
> })
> 
> var _ = Describe("RemoteEndpoint latency info", func() {
>     When("a remote Endpoint is created", func() {
>         It("should start a pinger and correctly update the RemoteEndpoint Status and LatencyInfo", func() {
>             // await the RemoteEndpoint and verify the LatencyInfo and Status
>         })
> 
>         Context("with no HealthCheckIP", func() {
> 		It("should not start a pinger and should set the RemoteEndpoint Status to None", ...)
> 	})
> 
>         Context("with health check not enabled", func() {
> 		It("should not start a pinger and should set the RemoteEndpoint Status to None", ...)
> 	})
> 
>         When("on the gateway", func() {
> 		It("should not start a pinger and should set the RemoteEndpoint Status to None", ...)
>                  // Call ControllerSupport.CreateLocalHostEndpoint() before creating the remote EP
> 	})
>     })
> 
>     When("a remote Endpoint is updated", func() {
>         Context("and the HealthCheckIP was changed", func() {
>             It("should stop the pinger and start a new one", ...)
>         })
> 
>         Context("and the HealthCheckIP did not change", func() {
>             It("should not start a new pinger", ...)
>         })
>     })
> 
>     When("a remote Endpoint is deleted", func() {
>         It("should stop the pinger", ...)
> })
> ```

Thanks for the suggestion . I updated the structure .

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+)
+
+var _ = Describe("RouteAgent syncing", func() {
+	t := newTestDriver(true)
+	It("should create a RouteAgent resource", func() {
+		var routeAgent *submarinerv1.RouteAgent
+		Eventually(func() bool {
+			var err error
+			routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
+			return err == nil && routeAgent != nil
```
tpantelis: We should fail on unexpected error.

```suggestion
			routeAgent, err = t.client.Get(context.TODO(), localNodeName, metav1.GetOptions{})
			if apierrors.IsNotFound(err) {
			    return false
			}
			
			Expect(err).To(Succeed())
			return true
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+		}, time.Second*50, time.Millisecond*500).Should(BeTrue())
```
tpantelis: 50 s is way too long as is 500ms interval - the default  is 100ms max - that should be plenty.

```suggestion
		}).Should(BeTrue())
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+		Expect(routeAgent.Name).To(Equal(localNodeName))
+		Expect(routeAgent.Namespace).To(Equal(namespace))
```
tpantelis: No need to verify these since they're used in the `Get` query.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+	When("a remote Endpoint is updated", func() {
+		It("should update its RemoteEndpoint information in the RouteAgent resource", func() {
+			endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(healthCheckIP1))
+			t.pingerMap[healthCheckIP1].AwaitStart()
+			endpoint1.Spec.Hostname = "newHostName"
+			t.UpdateEndpoint(endpoint1)
+
+			pingerObject, found := t.pingerMap[endpoint1.Spec.HealthCheckIP]
+			Expect(found).To(BeTrue())
+			Expect(pingerObject.GetIP()).To(Equal(healthCheckIP1))
```
tpantelis: This test is to just verify the remote endpoint spec. The pinger stuff is verified in another spec.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+
+			var routeAgent *submarinerv1.RouteAgent
```
tpantelis: You should first verify the original EP spec. You really want to do the same as in L64 above but instead of duplicating code, I suggest to combine the two. So L64+ would become:

```
When("a remote Endpoint is created/updated", func() {
    It("should add/update its RemoteEndpoint information to the RouteAgent resource", func() {
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+			Expect(routeAgent.Namespace).To(Equal(namespace))
+
+			remoteEndpoint := routeAgent.Status.RemoteEndpoints[0]
+			Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+		})
+	})
+
+	When("a remote Endpoint is deleted", func() {
+		It("should remove its RemoteEndpoint information from the RouteAgent resource", func() {
+			endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(healthCheckIP1))
```
tpantelis: Similarly as with update, you can combine delete as well:

`When("a remote Endpoint is created/updated/deleted", func() {`

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+		})
+	})
+})
+
+var _ = Describe("RemoteEndpoint latency info", func() {
+	When("a remote Endpoint is created", func() {
+		t := newTestDriver(true)
+		It("should start a pinger and correctly update the RemoteEndpoint Status and LatencyInfo", func() {
+			endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(healthCheckIP1))
+
```
tpantelis, line 95: First verify the pinger was started.

```suggestion
t.pingerMap[healthCheckIP1].AwaitStart()
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+			Expect(routeAgent.Namespace).To(Equal(namespace))
+
+			remoteEndpoint := routeAgent.Status.RemoteEndpoints[0]
+			Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+			Expect(remoteEndpoint.Status).To(Equal(submarinerv1.Connected))
+		})
+		Context("with no HealthCheckIP", func() {
+			It("should not start a pinger and should set the RemoteEndpoint Status to None", func() {
+				endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(""))
+				var routeAgent *submarinerv1.RouteAgent
```
tpantelis:
```suggestion
                                 t.pingerMap[healthCheckIP1].AwaitNoStart()
                                 
				var routeAgent *submarinerv1.RouteAgent
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+				remoteEndpoint := routeAgent.Status.RemoteEndpoints[0]
+				Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+				Expect(remoteEndpoint.Status).To(Equal(submarinerv1.ConnectionNone))
+			})
+		})
+		Context("on the gateway", func() {
+			It("should not start a pinger and should set the RemoteEndpoint Status to None", func() {
+				_ = t.CreateLocalHostEndpoint()
+				endpoint1 := t.CreateEndpoint(t.newSubmEndpoint(healthCheckIP1))
+
```
tpantelis, line 126:
```suggestion
t.pingerMap[healthCheckIP1].AwaitNoStart()
```

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+				remoteEndpoint := routeAgent.Status.RemoteEndpoints[0]
+				Expect(remoteEndpoint.Spec).To(Equal(endpoint1.Spec))
+				Expect(remoteEndpoint.Status).To(Equal(submarinerv1.ConnectionNone))
+			})
+		})
+	})
+
+	When("a remote Endpoint is created", func() {
+		t := newTestDriver(false)
+		Context("with health check not enabled", func() {
```
tpantelis, line 136: This should be within L144.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
```diff
@@ -0,0 +1,409 @@
...
+			defer GinkgoRecover()
+			Expect(pingerCfg.Interval).To(Equal(time.Second *
+				time.Duration(config.PingInterval))) //nolint:gosec // We can safely ignore integer conversion error
+			Expect(pingerCfg.MaxPacketLossCount).To(Equal(config.MaxPacketLossCount))
+
+			p, ok := t.pingerMap[pingerCfg.IP]
+			Expect(ok).To(BeTrue())
+
+			return p
+		}
```
tpantelis, line 292: Instead of passing in `enableHealthCheck`  to `newTestDriver`, it's better to default  `config.HealthCheckerEnabled` to true and have spec(s) set it to false in their own `BeforeEach` and move L357-359 below to `JustBeforeEach`.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/healthchecker/healthchecker_test.go
aswinsuryan: Done.

aswinsuryan:
@tpantelis there is a race condition being reported in the set state method, where do you think we need to fix it ? some of the  test infra ? Or do you think it is the code itself?

WARNING: DATA RACE
Read at 0x00c0000e3860 by goroutine 21:
  github.com/submariner-io/submariner/pkg/event.(*HandlerBase).State()
      /go/src/github.com/submariner-io/submariner/pkg/event/handler.go:116 +0x30c
  github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/healthchecker.(*controller).syncRouteAgentStatus()

tpantelis:
> @tpantelis there is a race condition being reported in the set state method, where do you think we need to fix it ? some of the test infra ? Or do you think it is the code itself?

I pushed https://github.com/submariner-io/submariner/pull/3152/commits/8a30449f7e30afee47b055859e8fcdcb46698619 to address it.

tpantelis (approved):

tpantelis:
I had additional minor adjustments that I just pushed in [another commit](https://github.com/submariner-io/submariner/pull/3152/commits/6198d81f408e55f18f2b162c3c29a8747cd363e7) rather then adding comments (thought it would be easier).

yboaron (approved):

PR #3186, opened 2024-10-10
Fix bug in RouteAgent resource creation on OpenShift

yboaron:
RouteAgent resources creation failed on OpenShift with [1] error.

[1]
the object provided is unrecognized (must be of type RouteAgent):
 Object 'Kind' is missing in
'{"metadata":{"name":"submqe-azure-ks4th-master-1",

### DIFF
--- a/pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
+++ b/pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
@@ -300,6 +300,9 @@ func (h *controller) syncRouteAgentStatus() {
 
 func (h *controller) generateRouteAgentObject() *submarinerv1.RouteAgent {
 	return &submarinerv1.RouteAgent{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "RouteAgent",
+		},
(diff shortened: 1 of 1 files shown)


pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
```diff
@@ -300,6 +300,9 @@ func (h *controller) syncRouteAgentStatus() {
 
 func (h *controller) generateRouteAgentObject() *submarinerv1.RouteAgent {
 	return &submarinerv1.RouteAgent{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "RouteAgent",
+		},
```
tpantelis, line 305: This should not be needed. We saw this issue before which was addressed by https://github.com/submariner-io/submariner/pull/3166.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/healthchecker/healthchecker.go
yboaron: Ack, saw this error in oldest version that didn't include #3166 , will close it

PR #3192, opened 2024-10-17
Adjust OVN TransitSwitchIP initialization

tpantelis:
We've observed E2E failures on route agent restart that was caused by the addition of the `TransitSwitchIP` refactoring. On code inspection, there's a potential timing issue related to initialization of the `TransitSwitchIP`. The `Init` method is called by `NonGatewayRouteHandler.Init` however the `NewNonGatewayRouteController` is started before that in `Handler.Init` so, depending on timing, `nonGatewayRouteCreatedOrUpdated` could get invoked prior to initialization of the `TransitSwitchIP`, in which case it would observe an empty IP.

To avoid this, move the `TransitSwitchIP.Init` call to `Handler.Init` prior to starting the `NewNonGatewayRouteController`.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/handler.go
+++ b/pkg/routeagent_driver/handlers/ovn/handler.go
@@ -51,7 +51,7 @@ type HandlerConfig struct {
 	DynClient       dynamic.Interface
 	WatcherConfig   *watcher.Config
 	NewOVSDBClient  NewOVSDBClientFn
-	TransitSwitchIP TransitSwitchIPGetter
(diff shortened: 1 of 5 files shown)


aswinsuryan (approved):

vthapar (approved):

Jaanki:
@tpantelis Does this need to be backported to 0.18.0 as well?

tpantelis:
> @tpantelis Does this need to be backported to 0.18.0 as well?

No - it was added in 0.19

PR #3202, opened 2024-10-30
Use exec form instead of shell form in docker files

tpantelis:
This is reported by Sonar:

"_Using the shell form instead of the exec form for CMD and ENTRYPOINT instructions in Dockerfiles can lead to several issues. When you use the shell form, the executable runs as a child process to a shell, which does not pass OS signals. This can cause problems when trying to gracefully stop containers because the main process will not receive the signal intended to terminate it. Moreover, the exec form provides more control and predictability over the execution of the command. It does not invoke a command shell, which means it does not have the potential side effects of shell processing._"

### DIFF
--- a/package/Dockerfile.submariner-gateway
+++ b/package/Dockerfile.submariner-gateway
@@ -35,4 +35,4 @@ COPY --from=base /output/gateway /
 
 COPY --from=builder ${SOURCE}/package/submariner.sh ${SOURCE}/package/pluto ${SOURCE}/bin/${TARGETPLATFORM}/submariner-gateway /usr/local/bin/
 
-ENTRYPOINT submariner.sh
+ENTRYPOINT ["submariner.sh"]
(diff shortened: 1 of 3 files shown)


skitt:
This is a false positive — the entrypoints are all shell scripts, so running them using a shell or directly ends up doing the same thing, with the same number of processes (unless `sh` isn’t `bash`; but that’s not the case in our container images). The shell scripts handle signals themselves (that’s what the `trap` lines set up).

tpantelis:
> This is a false positive — the entrypoints are all shell scripts, so running them using a shell or directly ends up doing the same thing, with the same number of processes (unless `sh` isn’t `bash`; but that’s not the case in our container images). The shell scripts handle signals themselves (that’s what the `trap` lines set up).

Ok - I'll mark the Sonar issues as false positives.

PR #3237, opened 2024-12-03
Create await-node-ready binary and add to gateway and route-agent images

tpantelis:
Related to https://github.com/submariner-io/submariner-operator/issues/3274

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ bin/protoc:
 	unzip protoc-$(PROTOC_VERSION)-linux-x86_64.zip 'bin/*' 'include/*'
 	rm -f protoc-$(PROTOC_VERSION)-linux-x86_64.zip
 
-bin/%/submariner-gateway: main.go $(shell find pkg -not \( -path 'pkg/globalnet*' -o -path 'pkg/routeagent*' \)) pkg/natdiscovery/proto/natdiscovery.pb.go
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -75,6 +75,8 @@ bin/%/submariner-route-agent: $(shell find pkg/routeagent_driver)
 bin/%/submariner-globalnet: $(shell find pkg/globalnet)
 	GOARCH=$(call dockertogoarch,$(patsubst bin/linux/%/,%,$(dir $@))) ${SCRIPTS_DIR}/compile.sh $@ ./pkg/globalnet
 
+bin/%/await-node-ready: $(shell find pkg/await_node_ready)
```
skitt, line 78: This also uses `pkg/node` and `pkg/versions`.

```suggestion
bin/%/await-node-ready: $(shell find pkg/await_node_ready pkg/node pkg/versions)
```

skitt (changes requested):

Makefile
tpantelis: Ok but is that needed? Route agent only specifies `$(shell find pkg/routeagent_driver)` but it uses other packages as well. Same with globalnet - `$(shell find pkg/globalnet)`. The executable binary runs fine locally.

skitt: It doesn’t affect the binary that gets built, it only determines when it gets rebuilt. But this approach is inherently brittle, I’ve got a better one — let’s merge this and I’ll fix all the binaries properly.

skitt (approved):

dfarrell07 (approved):

PR #3239, opened 2024-12-04
Construct package dependencies using Go

skitt:
For best results, Make should be told about all the files a given target depends on. For Go programs, this is the program's main Go file, and all non-test Go files in directories corresponding to packages the main Go file depends upon (including transitive dependencies).

This adds a Make macro using go list to retrieve a package's dependencies, and constructs the corresponding list of files to use as target prerequisites.


Makefile
```diff
@@ -66,13 +66,17 @@ bin/protoc:
 	unzip protoc-$(PROTOC_VERSION)-linux-x86_64.zip 'bin/*' 'include/*'
 	rm -f protoc-$(PROTOC_VERSION)-linux-x86_64.zip
 
-bin/%/submariner-gateway: main.go $(shell find pkg -not \( -path 'pkg/globalnet*' -o -path 'pkg/routeagent*' \)) pkg/natdiscovery/proto/natdiscovery.pb.go
+basepkg = github.com/submariner-io/submariner
+pkgdeps = $(shell find $$(go list -json $(1) | jq -r '.Imports[] | select(startswith("$(basepkg)")) | sub("$(basepkg)"; ".")') -name '*.go' -not -name '*_test.go')
+
+# natdiscovery.pb.go must be listed explicitly because it might not exist when Make evaluates pkgdeps
```
dfarrell07, line 65: Why is `natdiscovery.pb.go` special in this way? Could there be anything else that should be explicitly listed in the future?

dfarrell07 (approved):
Very cool! Minor question/docs feedback inline.

Makefile
skitt: It’s special because it’s generated — this isn’t a problem most of the time because Go projects effectively need all their generated source to be in their repo (so it’s published for downstream consumption), but it’s still perfectly legitimate to delete all generated files and expect the build to work. If you start from a “clean” project (with no generated files), `find` etc. won’t find (sic) `natdiscovery.pb.go` since it doesn’t exist, and so it won’t be listed in the prerequisites and the build will end up failing (Make won’t know that it needs to arrange for `natdiscovery.pb.go` to be created before attempting to build targets that rely on its presence).

This is a general rule for Makefiles: any generated file must be explicitly listed as a prerequisite in some way.

tpantelis (approved):

Makefile
```diff
@@ -66,16 +66,20 @@ bin/protoc:
 	unzip protoc-$(PROTOC_VERSION)-linux-x86_64.zip 'bin/*' 'include/*'
 	rm -f protoc-$(PROTOC_VERSION)-linux-x86_64.zip
 
-bin/%/submariner-gateway: main.go $(shell find pkg -not \( -path 'pkg/globalnet*' -o -path 'pkg/routeagent*' -o -path 'pkg/await_node_ready*' \)) pkg/natdiscovery/proto/natdiscovery.pb.go
+basepkg = github.com/submariner-io/submariner
+pkgdeps = $(shell find $$(go list -json $(1) | jq -r '.Imports[] | select(startswith("$(basepkg)")) | sub("$(basepkg)"; ".")') -name '*.go' -not -name '*_test.go')
+
+# natdiscovery.pb.go must be listed explicitly because it might not exist when Make evaluates pkgdeps
+bin/%/submariner-gateway: main.go $(call pkgdeps,.) pkg/natdiscovery/proto/natdiscovery.pb.go
```
sridhargaddam, line 66: Should this be?

```suggestion
bin/%/submariner-gateway: main.go $(call pkgdeps,./pkg/gateway) pkg/natdiscovery/proto/natdiscovery.pb.go
```

skitt: The root of the dependency tree is the package containing the main file, so `.` in this case. However while investigating this comment I realised that `.Imports[]` only contains first-level imports, and the `pkgdeps` macro really needs to look at `.Deps[]` instead…

skitt: Note too how the argument to `pkgdeps` matches the argument to `compile.sh` :wink:

PR #3242, opened 2024-12-05
Remove wait-for-node-ready in GW and RA main

### LINKED ISSUE #3274: Add a “wait for node readiness” binary and image
Given that several binaries now offer “wait for node readiness” variant executions, it seems like it would be useful to move that to a single, separate, much simpler binary and use that in init containers.

tpantelis:
This is now handled by the separate await-node-ready binary.

Fixes https://github.com/submariner-io/submariner-operator/issues/3274


skitt (changes requested):
The commit message and PR description mention GW, but this only touches RA.

tpantelis:
> The commit message and PR description mention GW, but this only touches RA.

It changes the top-level _main.go_ as well.

skitt (approved):

skitt:
> > The commit message and PR description mention GW, but this only touches RA.
> 
> It changes the top-level _main.go_ as well.

So it does, I’m blind :roll_eyes:

skitt:
Ah, what misled me is that I thought this could remove `WaitForLocalNodeReady` entirely.

aswinsuryan (approved):

PR #3267, opened 2025-01-15
Add plural IP fields to EndpointSpec

tpantelis:
...that mirror the singular `HealthCheckIP`, `PrivateIP` and `PublicIP` fields to support dual-stack addresses. The singular fields are deprecated and remain to support backwards compatibility and migration with prior versions. Going forward only the plural fields will be used. Get/Set methods were added to `EndpointSpec` that handle the singular fields. On Get, if the plural field is empty then retrieve the singular field. On Set, if one of the specified IPs is IPv4 then set the corresponding singular field.


pkg/apis/submariner.io/v1/endpoint.go
```diff
@@ -102,3 +103,50 @@ func (ep *EndpointSpec) hasSameBackendConfig(other *EndpointSpec) bool {
...
+
+func (ep *EndpointSpec) SetHealthCheckIPs(ips ...string) {
+	ep.HealthCheckIPs, ep.HealthCheckIP = toIPArrayAndV4String(ips)
+}
+
+func (ep *EndpointSpec) GetPublicIPs() []string {
+	return fromArrayOrString(ep.PublicIPs, ep.PublicIP)
+}
+
+func (ep *EndpointSpec) SetPublicIPs(ips ...string) {
```
yboaron: overall code LGTM,
I have one minor comment,  since some of the fields (e.g Public IP) can be discovered at different timing maybe it makes sense to replace (or add new API) SetxxxIPs with UpdatexxIP (Update should call GetxxxIPs and update it if needed ) instead of having this logic in clients side

tpantelis: Sorry I'm not sure I follow what you mean...

yboaron: I'm sorry, I'll try to explain it by example ... 

Assuming the following use case for dual-stack case: 
A. public IPV4 address successfully discovered on t1
B. public IPV6 address only discovered on t2 

I thought it might be useful to have Update API (that will add  IP if it doesn't exist, we can use it in step B for example).

But on second thought since there is only single client (gateway) that will call this APIs we can have this logic at the gateway side.

yboaron (approved):

pkg/apis/submariner.io/v1/endpoint.go
tpantelis: I see - there may be cases where we need to set IPs atomically or piecemeal. Either way we can add/modify the helper functions as needed. The main point of these helpers is to hide the singular fields.

tpantelis:
I made `PrivateIPs` and `PublicIPs` optional as well to allow an older cluster to create an `Endpoint` with the legacy singular fields.

pkg/apis/submariner.io/v1/endpoint.go
skitt: This highlights that there’s a difference between the way the data is stored and how it’s used. Externally, the API could be `Set…IPv4`/`Set…IPv6` (with corresponding getters); that would make it explicit that the data structure only stores one IPv4 address and one IPv6 address.

(Incidentally, there’s currently nothing stopping a caller from storing two IPv4 addresses or two IPv6 addresses :wink:.)

tpantelis: OK I changed the getters to retrieve the singular IP for an IP family. The family is passed as a parameter as, after looking at usages, I think that's going to be more natural/easier rather than separate functions for each family. The setters add the address if not already present to ensure there's only one per family.

skitt (approved):

PR #3279, opened 2025-01-19
Gateway - use endpoint set private, public and healthcheck IPs

yboaron:
This PR :
- updates gateway code to use endpoint SetHealthCheckIP ,SetPrivateIP and SetPublicIP.
- add placeholders for IPV6 public,private and healthcheck IPs.


pkg/endpoint/local_endpoint.go
```diff
@@ -171,42 +174,67 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
-	publicIP, resolver, err := getPublicIP(submSpec, k8sClient, backendConfig, airGappedDeployment)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not determine public IP")
+	for _, privateIP := range privateIPs {
+		endpointSpec.SetPrivateIP(privateIP)
 	}
 
-	logger.Infof("Obtained local endpoint public IP %q using resolver %q", publicIP, resolver)
+	endpointSpec.CableName = fmt.Sprintf("submariner-cable-%s-%s", submSpec.ClusterID,
+		strings.ReplaceAll(endpointSpec.GetPrivateIP(k8snet.IPv4), ".", "-"))
```
tpantelis, line 186: What if there's no IPv4 private IP? Why don't we just omit the private IP from the cable name?

pkg/endpoint/local_ip.go
```diff
@@ -37,3 +39,15 @@ func GetLocalIPForDestination(dst string) string {
 func GetLocalIP() string {
 	return GetLocalIPForDestination("8.8.8.8")
 }
+
+func GetLocalIPs(ipFamilies []v1.IPFamily) []string {
```
tpantelis: Since there can be at most 2 families:

```suggestion
func GetLocalIPs(ipFamilies [2]v1.IPFamily) []string {
```

pkg/endpoint/public_ip.go
```diff
@@ -63,6 +64,27 @@ func getPublicIPResolvers() string {
...
+	var publicIPs []string
+	var errs []error
+
+	for _, family := range submSpec.GetIPFamilies() {
+		if family == corev1.IPv4Protocol {
+			publicIP, resolver, err := getPublicIP(submSpec, k8sClient, backendConfig, airGapped)
+			if err != nil {
+				errs = append(errs, errors.Wrap(err, "could not determine public IP"))
+			} else {
+				logger.Infof("Obtained local endpoint public IP %q using resolver %q", publicIP, resolver)
```
tpantelis: The `PublicIPWatcher` should call this function as well so logging here would fill up the log over time. I think we should keep using `getPublicIP` but pass in the `IPFamily`. Then the `PublicIPWatcher` would be:

```
for _, family := range p.config.SubmSpec.GetIPFamilies() {
    publicIP, resolver, err := getPublicIP(p.config.SubmSpec, p.config.K8sClient, localEndpointSpec.BackendConfig, false)
    if err != nil {
        logger.Warningf("Could not determine public IP of the gateway node %q for IP family %v: %v", localEndpointSpec.Hostname, family, err)
	return
    }

    if localEndpointSpec.GetPublicIP(family) != publicIP {
        logger.Infof("Public IP for IP family %v changed for the Gateway, updating the local endpoint with public IP %q obtained from resolver %q", family, publicIP, resolver)
        localEndpointSpec.SetPublicIP(publicIP)
    }
}

if err := p.updateLocalEndpoint(localEndpointSpec); err != nil {
    logger.Error(err, "Error updating the public IPs for local endpoint")
}
```

pkg/types/types.go
```diff
@@ -50,3 +51,11 @@ type SubmarinerSpecification struct {
 	HealthCheckMaxPacketLossCount int
 	MetricsPort                   int `default:"32780"`
 }
+
+func (subSpec *SubmarinerSpecification) GetIPFamilies() []v1.IPFamily {
```
tpantelis: I think we should use the `k8snet.IPFamily` enum to be consistent with the `Endpoint` methods to avoid potentially having to translate between the two later.

Also there can be at most 2 families so we can make that explicit.

```suggestion
func (subSpec *SubmarinerSpecification) GetIPFamilies() [2]k8snet.IPFamily {
```

tpantelis (changes requested):

pkg/endpoint/local_endpoint.go
yboaron: Yes, we need to address this issue when we support IPv6/dual-stack, at the moment I didn't want to remove/change the IPv4 address from the cable name, because I'm not sure if that address is being used.

Anyway, even with this PR we still assume that only IPv4 is currently supported

pkg/endpoint/public_ip_watcher.go
```diff
@@ -65,26 +66,39 @@ func (p *PublicIPWatcher) Run(stopCh <-chan struct{}) {
...
 
-		if err := p.updateLocalEndpoint(publicIP); err != nil {
+	for _, publicIP := range publicIPs {
+		if (k8snet.IPFamilyOfString(publicIP) == k8snet.IPv4) && (publicIP != localEndpointSpec.GetPublicIP(k8snet.IPv4)) {
+			publicIpsChanged = true
+		}
+
+		if (k8snet.IPFamilyOfString(publicIP) == k8snet.IPv6) && (publicIP != localEndpointSpec.GetPublicIP(k8snet.IPv6)) {
+			publicIpsChanged = true
+		}
```
tpantelis, line 78: I think this code is simplified if it loops over the families and passes it to `getPublicIP` as I suggested in my code snippet.  This also avoids the `logPublicIP` flag which seems awkward.

tpantelis (changes requested):

pkg/endpoint/local_endpoint.go
```diff
@@ -171,42 +174,67 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
+	publicIPs, err := getPublicIPs(submSpec, k8sClient, backendConfig, airGappedDeployment, true)
+
+	if len(publicIPs) == 0 && err != nil {
```
tpantelis: Why is this checking `len(publicIPs)` here? If an error was returned with non-empty publicIPs then the code would proceed and silence the error. Perhaps the intention is to proceed if some IP was returned but I think it should fail if any errors occurs.

pkg/endpoint/local_endpoint.go
```diff
@@ -171,42 +174,67 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
 		// When globalnet is enabled, HealthCheckIP will be the globalIP assigned to the Active GatewayNode.
 		// In a fresh deployment, globalIP annotation for the node might take few seconds. So we listen on NodeEvents
 		// and update the endpoint HealthCheckIP (to globalIP) in datastoreSyncer at a later stage. This will trigger
 		// the HealthCheck between the clusters.
-		cniIface, err := cni.Discover(submSpec.ClusterCidr)
-		if err != nil {
-			return nil, fmt.Errorf("error getting CNI Interface IP address: %w."+
-				"Please disable the health check if your CNI does not expose a pod IP on the nodes", err)
+		healthcheckIPs, err := getHealthCheckIPs(submSpec)
+		if len(healthcheckIPs) == 0 && err != nil {
```
tpantelis: Same comment as above wrt publicIPs.

tpantelis (changes requested):

pkg/endpoint/local_endpoint.go
```diff
@@ -171,42 +174,67 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
 	}
 
-	logger.Infof("Obtained local endpoint public IP %q using resolver %q", publicIP, resolver)
+	endpointSpec.CableName = fmt.Sprintf("submariner-cable-%s-%s", submSpec.ClusterID,
+		strings.ReplaceAll(endpointSpec.GetPrivateIP(k8snet.IPv4), ".", "-"))
+	publicIPs, err := getPublicIPs(submSpec, k8sClient, backendConfig, airGappedDeployment)
+
+	if len(publicIPs) == 0 {
+		return nil, errors.Wrap(err, "could not determine public IP")
+	}
```
tpantelis: Again this silences the error if there's only one family that had a failure. It should fail if any error occurs. Also we don't need `getPublicIPs` - use `getPublicIP` and pass in the family as should the watcher:

```
    publicIP, resolver, err := getPublicIP(family, submSpec, k8sClient, backendConfig, airGappedDeployment)
    if err != nil {
        return nil, errors.Wrapf(err, "could not determine public IP for IP family %v", family)
    }

    logger.Infof("Obtained local endpoint public IP %q using resolver %q for family %v", publicIP, resolver, family)

    endpointSpec.SetPublicIP(publicIP)
}

```

pkg/endpoint/local_endpoint.go
```diff
@@ -171,42 +174,67 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
 		// In a fresh deployment, globalIP annotation for the node might take few seconds. So we listen on NodeEvents
 		// and update the endpoint HealthCheckIP (to globalIP) in datastoreSyncer at a later stage. This will trigger
 		// the HealthCheck between the clusters.
-		cniIface, err := cni.Discover(submSpec.ClusterCidr)
-		if err != nil {
-			return nil, fmt.Errorf("error getting CNI Interface IP address: %w."+
-				"Please disable the health check if your CNI does not expose a pod IP on the nodes", err)
+		healthcheckIPs, err := getHealthCheckIPs(submSpec)
+		if len(healthcheckIPs) == 0 {
+			return nil, errors.Wrap(err, "could not determine healthcheck IP")
```
tpantelis: Similar to public IPs above (pass in the family):

```
    healthcheckIP, err := getHealthCheckIP(family, submSpec)
    if err != nil {
        return nil, errors.Wrapf(err, "could not determine health check IP for IP family %v", family)
    }

    endpointSpec.SetHealthCheckIP(healthcheckIP)
}
```

pkg/endpoint/local_endpoint.go
```diff
@@ -171,42 +174,67 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
+					"Please disable the health check if your CNI does not expose a pod IP on the nodes", err))
+			} else {
+				healthCheckIPs = append(healthCheckIPs, cniIface.IPAddress)
+			}
+		} // TODO_IPV6: add V6 healthcheck IP
+	}
+
+	return healthCheckIPs, errors.Wrap(k8serrors.NewAggregate(errs), "getHealthCheckIPs failed")
+}
+
```
tpantelis:
```suggestion
func getHealthCheckIP(family k8snet.IPFamily, submSpec *types.SubmarinerSpecification) (string, error) {
	switch family {
	case k8snet.IPv4:
		cniIface, err := cni.Discover(submSpec.ClusterCidr)
		if err != nil {
			return "", errors.Wrapf(err, "error getting CNI Interface IP address."+
				"Please disable the health check if your CNI does not expose a pod IP on the nodes")
		}

		return cniIface.IPAddress, nil
	case k8snet.IPv6:
		// TODO_IPV6: add V6 healthcheck IP
	}

	return "", nil
}
```

pkg/endpoint/local_endpoint.go
```diff
@@ -171,42 +174,67 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
 		Subnets:       localSubnets,
 		Backend:       submSpec.CableDriver,
 		BackendConfig: backendConfig,
 	}
 
-	publicIP, resolver, err := getPublicIP(submSpec, k8sClient, backendConfig, airGappedDeployment)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not determine public IP")
+	for _, privateIP := range privateIPs {
+		endpointSpec.SetPrivateIP(privateIP)
```
tpantelis: Instead if `GetLocalIPs` above, we should simplify and pass in the family to `GetLocalIP`. 

```suggestion
        for _, family := range submSpec.GetIPFamilies() {
            endpointSpec.SetPrivateIP(GetLocalIP(family))
        }
```

pkg/endpoint/public_ip_watcher.go
```diff
@@ -63,28 +64,44 @@ func (p *PublicIPWatcher) Run(stopCh <-chan struct{}) {
...
+				if publicIP != localEndpointSpec.GetPublicIP(k8snet.IPv4) {
+					publicIPsChanged = true
+
+					logger.Infof("Public IPv4 changed for the Gateway, updating the local endpoint with public IP %q obtained from resolver %q",
+						publicIP, resolver)
+				}
+
+				publicIPs = append(publicIPs, publicIP)
+			}
+		} // TODO_IPV6: add getPublicIPV6
```
tpantelis: Again, this is simplified and consistent if we keep the family-specific logic in one place (`getPublicIP`). 

```suggestion
	        publicIP, resolver, err := getPublicIP(family, p.config.SubmSpec, p.config.K8sClient, localEndpointSpec.BackendConfig, false)
		if err != nil {
		    logger.Warningf("Could not determine public IP for family %v of the gateway node %q: %v", family, localEndpointSpec.Hostname, err)
		    continue
		} 
		
		if publicIP != localEndpointSpec.GetPublicIP(family) {
		   publicIPs = append(publicIPs, publicIP)
		   
		   logger.Infof( "Public IP for family %v changed for the Gateway, updating the local endpoint with public IP %q obtained from resolver %q", family, publicIP, resolver)
		}
```

pkg/endpoint/public_ip_watcher.go
```diff
@@ -63,28 +64,44 @@ func (p *PublicIPWatcher) Run(stopCh <-chan struct{}) {
...
 	}
 
-	if localEndpointSpec.PublicIP != publicIP {
-		logger.Infof("Public IP changed for the Gateway, updating the local endpoint with public IP %q obtained from resolver %q",
-			publicIP, resolver)
+	if publicIPsChanged {
+		logger.Infof("Public IPs changed for the Gateway, updating the local endpoint with public IPs %q", publicIPs)
```
tpantelis: We don't need the `publicIPsChanged` flag if `publicIPs` only contains IPs that changed. Also we already logged any changes above.

```suggestion
	if len(publicIPs) > 0 {
```

tpantelis (changes requested):

pkg/endpoint/local_endpoint.go
yboaron: > Again this silences the error if there's only one family that had a failure. It should fail if any error occurs

Hmm, not 100% sure about that, 
So, in case one family failed in a dual-stack case, you think we should fail the gateway and not continue with the one that succeeded, right?

tpantelis: Not entirely sure but currently that's the behavior with just IP4 and would be the behavior with just IPv6. What would/should be the overall behavior in the GW if some aspect(s) of IPv6 failed in dual-stack? Would some features be available but not others (eg maybe we're able get a private IP but not a public one)? Or do we somehow disable IPv6 support across the board but continue with IPv4?

tpantelis (approved):

dfarrell07 (approved):

PR #3281, opened 2025-01-21
Ignore empty or invalid IP in Endpoint Set*IP

tpantelis:


pkg/apis/submariner.io/v1/endpoint.go
```diff
@@ -120,6 +120,9 @@ func getIPFrom(family k8snet.IPFamily, ips []string, ipv4Fallback string) string
 
 func setIP(ips []string, ipv4Fallback, newIP string) ([]string, string) {
 	family := k8snet.IPFamilyOfString(newIP)
+	if family == k8snet.IPFamilyUnknown {
```
skitt, line 131: Wouldn’t it be better to return an error?

tpantelis: Perhaps for an invalid IP but then every caller has to check the error even though it's likely not possible for it to pass an invalid IP.

Handling an empty IP without failing avoids more generic client code from having to first verify non-empty. Eg

```
    publicIP, resolver, err := getPublicIP(family, submSpec, k8sClient, backendConfig, airGappedDeployment)
    ...

    endpointSpec.SetPublicIP(publicIP)
}
```

`getPublicIP` would currently return empty for IPv6 b/c we haven't implemented it yet. If `SetPublicIP` returned an error on empty, we would want to ignore it here or check if non-empty first. Either way we wouldn't care about the return error but we'd still have to handle it to avoid a linter error.

skitt: OK, so if this is about handling empty strings, I think it’s best to be explicit about it:

```
func setIP(ips []string, ipv4Fallback, newIP string) ([]string, string) {
	if newIP == "" {
		return ips, ipv4Fallback
	}

	family := k8snet.IPFamilyOfString(newIP)
	…
```

Handling `IPFamilyUnknown` suggests that it’s about handling invalid addresses, not IPs (at least, to me).

skitt: Or add a comment mentioning that `IPFamilyUnknown` corresponds to empty addresses :wink:.

tpantelis: Checking `IPFamilyUnknown` handles both empty and invalid as neither translates to V4 or V6. I'll add explicit check for empty so that `IPFamilyUnknown` means invalid and log an error message.

yboaron (approved):

skitt (approved):

PR #3296, opened 2025-02-06
Remove AddToScheme call in event controller New

tpantelis:
The global scheme is already set up by the callers and calling `AddToScheme` here can cause intermittent data race failures in unit tests.


dfarrell07 (approved):

dfarrell07 (approved):
Very nice! For the sake of learning, I'm curious how you found this.

skitt (approved):

PR #3311, opened 2025-02-19
Update nat discovery to support IP family

yboaron:
Current code assumes 1:1 mapping between endpoint to nat-discovery and
endpoint to submariner inter-cluster tunnel.

Since we decided that we continue using a single endpoint for each cluster
also for dual-stack, it is necessary to support mapping endpoint up to 2
nat-discovery and inter-cluster tunnels.

Additionally, nat-discovery and submariner-tunnel also need
to support IP family parameter, as both IPv4 and IPv6 should be supported.

This change update nat-discovery to support IP family, to make it clear,
only IPV4 nat-discovery is supported even after this change.


pkg/cableengine/cableengine.go
```diff
@@ -156,13 +157,13 @@ func (i *engine) installCableWithNATInfo(rnat *natdiscovery.NATEndpointInfo) err
 	i.Lock()
 	defer i.Unlock()
 
-	if _, ok := i.natDiscoveryPending[rnat.Endpoint.Spec.CableName]; !ok {
+	if _, ok := i.natDiscoveryPending[rnat.Endpoint.Spec.GetFamilyCableName(rnat.Family)]; !ok {
```
tpantelis: We could extract `GetFamilyCableName` to a var as it's referenced several times.

pkg/cableengine/cableengine.go
```diff
@@ -242,28 +243,28 @@ func (i *engine) InstallCable(endpoint *v1.Endpoint) error {
...
 	if endpoint.Spec.ClusterID == i.localCluster.ID {
 		logger.V(log.DEBUG).Infof("Cables are not added/removed for the local cluster, skipping removal")
 		return nil
 	}
 
-	logger.Infof("Removing Endpoint cable %q", endpoint.Spec.CableName)
+	logger.Infof("Removing Endpoint IP%v cable %q", family, endpoint.Spec.CableName)
 
-	i.natDiscovery.RemoveEndpoint(endpoint.Spec.CableName)
+	i.natDiscovery.RemoveEndpoint(endpoint.Spec.GetFamilyCableName(family))
```
tpantelis: Same here re: extract `GetFamilyCableName` to a var.

pkg/cableengine/cableengine.go
```diff
@@ -274,9 +275,9 @@ func (i *engine) RemoveCable(endpoint *v1.Endpoint) error {
 		return errors.Wrapf(err, "error disconnecting Endpoint cable %q", endpoint.Spec.CableName)
 	}
 
-	delete(i.installedCables, endpoint.Spec.CableName)
+	delete(i.installedCables, endpoint.Spec.GetFamilyCableName(family))
```
tpantelis: We need to be consistent and also index by `GetFamilyCableName` everywhere we access `installedCables` (L269 above and in `installCableWithNATInfo`).

pkg/cableengine/cableengine.go
```diff
@@ -156,13 +157,13 @@ func (i *engine) installCableWithNATInfo(rnat *natdiscovery.NATEndpointInfo) err
...
+	if _, ok := i.natDiscoveryPending[rnat.Endpoint.Spec.GetFamilyCableName(rnat.Family)]; !ok {
 		return nil
 	}
 
-	i.natDiscoveryPending[rnat.Endpoint.Spec.CableName]--
-	if i.natDiscoveryPending[rnat.Endpoint.Spec.CableName] == 0 {
-		delete(i.natDiscoveryPending, rnat.Endpoint.Spec.CableName)
+	i.natDiscoveryPending[rnat.Endpoint.Spec.GetFamilyCableName(rnat.Family)]--
+	if i.natDiscoveryPending[rnat.Endpoint.Spec.GetFamilyCableName(rnat.Family)] == 0 {
+		delete(i.natDiscoveryPending, rnat.Endpoint.Spec.GetFamilyCableName(rnat.Family))
```
tpantelis: The code below where we check for an active connection is going to also have to take into account the family. I think we're going to have to add a `UsingIPFamily` field to the `Connection` struct. But we can address this later when the cable drivers are modified to use the family.

pkg/apis/submariner.io/v1/endpoint.go
```diff
@@ -172,3 +172,15 @@ func (ep *EndpointSpec) GetPrivateIP(family k8snet.IPFamily) string {
 func (ep *EndpointSpec) SetPrivateIP(ip string) {
 	ep.PrivateIPs, ep.PrivateIP = setIP(ep.PrivateIPs, ep.PrivateIP, ip)
 }
+
+func (ep *EndpointSpec) GetIPFamilies() [2]k8snet.IPFamily {
```
tpantelis: This actually returns an array of length 2 with the second element empty: `["4", ""]`. We should return `[]k8snet.IPFamily` since it can have 0, 1, or 2 elements.

We should be able to implement this method for real now and add unit tests since we only populate `Subnets` with IP4.

```
ipFamilies := set.New[k8snet.IPFamily]()

for _, cidr := range ep.Subnets {
    f := k8snet.IPFamilyOfCIDRString(cidr)
    if f != k8snet.IPFamilyUnknown {
        ipFamilies.Insert(f)
    }
}

return ipFamilies.UnsortedList()
```

pkg/apis/submariner.io/v1/endpoint.go
```diff
@@ -172,3 +172,15 @@ func (ep *EndpointSpec) GetPrivateIP(family k8snet.IPFamily) string {
...
+	var ipFamilies [2]k8snet.IPFamily
+	// TODO_IPV6: set ipFamilies according to Subnets content
+	ipFamilies[0] = k8snet.IPv4
+
+	return ipFamilies
+}
+
+func (ep *EndpointSpec) GetFamilyCableName(family k8snet.IPFamily) string {
```
tpantelis, line 181: Please add a unit test for this function so we don't lose any coverage percentage. You can simply check that the string starts with the cable name and ends with the family.

pkg/controllers/tunnel/tunnel.go
```diff
@@ -27,20 +27,39 @@ import (
...
 
 type controller struct {
-	engine cableengine.Engine
+	engine          cableengine.Engine
+	localIPFamilies [2]k8snet.IPFamily
 }
 
 var logger = log.Logger{Logger: logf.Log.WithName("Tunnel")}
 
+func findCommonIPFamilies(local, remote [2]k8snet.IPFamily) []k8snet.IPFamily {
```
tpantelis:
```suggestion
func findCommonIPFamilies(local, remote []k8snet.IPFamily) []k8snet.IPFamily {
```

pkg/controllers/tunnel/tunnel.go
```diff
@@ -27,20 +27,39 @@ import (
...
+	for _, lf := range local {
+		for _, rf := range remote {
+			if lf == rf {
+				common = append(common, lf)
+				break
+			}
+		}
+	}
+
+	return common
```
tpantelis: You can simplify to (import "github.com/submariner-io/admiral/pkg/slices"):

```suggestion
	return slices.Intersect(local, remote, func(e k8snet.IPFamily) k8snet.IPFamily {
		return e
	})
```

pkg/endpoint/local_ip.go
```diff
@@ -36,6 +36,18 @@ func GetLocalIPForDestination(dst string) string {
...
+	switch family {
+	case k8snet.IPv4:
+		return GetLocalIPForDestination(dst)
+	case k8snet.IPv6:
+		// TODO_IPV6: add V6 healthcheck IP
+	case k8snet.IPFamilyUnknown:
+	}
+
+	return ""
+}
```
tpantelis, line 53: We can restructure and simply these functions as follows:

```
func getLocalIPv4ForDestination(dst string) string {
	...
}

	switch family {
	case k8snet.IPv4:
		return getLocalIPv4ForDestination(dst)
	case k8snet.IPv6:
		// TODO_IPV6: add V6 healthcheck IP
	case k8snet.IPFamilyUnknown:
	}

	return ""
}

func GetLocalIP(family k8snet.IPFamily) string {
	return GetLocalIPForDestination("8.8.8.8", family)
}
```

pkg/natdiscovery/request_handle.go
```diff
@@ -64,7 +69,7 @@ func (nd *natDiscovery) handleRequestFromAddress(req *proto.SubmarinerNATDiscove
 		return nd.sendResponseToAddress(&response, addr)
 	}
 
-	if req.GetReceiver().GetEndpointId() != localEndpointSpec.CableName {
+	if req.GetReceiver().GetEndpointId() != localEndpointSpec.GetFamilyCableName(k8snet.IPv4) {
```
tpantelis: Should we assume v4 here?

tpantelis (changes requested):

pkg/apis/submariner.io/v1/endpoint.go
tpantelis: Pushed commit https://github.com/submariner-io/submariner/pull/3311/commits/6d5338cf9b059031032fbc1df8e302db930d8057 to address this comment.

pkg/controllers/tunnel/tunnel.go
tpantelis: Addressed by commit https://github.com/submariner-io/submariner/pull/3311/commits/6d5338cf9b059031032fbc1df8e302db930d8057.

pkg/controllers/tunnel/tunnel.go
tpantelis: Addressed by commit https://github.com/submariner-io/submariner/pull/3311/commits/6d5338cf9b059031032fbc1df8e302db930d8057.

pkg/cableengine/cableengine.go
yboaron: Yep, I prefer to handle cable driver changes in a separate PR , this PR should handle only natd

pkg/natdiscovery/request_handle.go
yboaron: Nope, we should use family.
Changed it

pkg/apis/submariner.io/v1/endpoint.go
yboaron: Done

pkg/cableengine/cableengine.go
```diff
@@ -242,28 +244,28 @@ func (i *engine) InstallCable(endpoint *v1.Endpoint) error {
...
+func (i *engine) RemoveCable(endpoint *v1.Endpoint, family k8snet.IPFamily) error {
 	if endpoint.Spec.ClusterID == i.localCluster.ID {
 		logger.V(log.DEBUG).Infof("Cables are not added/removed for the local cluster, skipping removal")
 		return nil
 	}
 
-	logger.Infof("Removing Endpoint cable %q", endpoint.Spec.CableName)
-
-	i.natDiscovery.RemoveEndpoint(endpoint.Spec.CableName)
+	logger.Infof("Removing Endpoint IP%v cable %q", family, endpoint.Spec.CableName)
```
tpantelis: Minor but I think this reads better.

```suggestion
	logger.Infof("Removing Endpoint IPv%v cable %q", family, endpoint.Spec.CableName)
```

pkg/cableengine/cableengine.go
```diff
@@ -276,7 +278,7 @@ func (i *engine) RemoveCable(endpoint *v1.Endpoint) error {
 
 	delete(i.installedCables, endpoint.Spec.CableName)
 
-	logger.Infof("Successfully removed Endpoint cable %q", endpoint.Spec.CableName)
+	logger.Infof("Successfully removed IP%v Endpoint cable %q", family, endpoint.Spec.CableName)
```
tpantelis:
```suggestion
	logger.Infof("Successfully removed IPv%v Endpoint cable %q", family, endpoint.Spec.CableName)
```

pkg/cableengine/cableengine.go
```diff
@@ -156,13 +157,14 @@ func (i *engine) installCableWithNATInfo(rnat *natdiscovery.NATEndpointInfo) err
...
+	if _, ok := i.natDiscoveryPending[familyCableName]; !ok {
 		return nil
 	}
 
-	i.natDiscoveryPending[rnat.Endpoint.Spec.CableName]--
-	if i.natDiscoveryPending[rnat.Endpoint.Spec.CableName] == 0 {
-		delete(i.natDiscoveryPending, rnat.Endpoint.Spec.CableName)
+	i.natDiscoveryPending[familyCableName]--
+	if i.natDiscoveryPending[familyCableName] == 0 {
+		delete(i.natDiscoveryPending, familyCableName)
```
tpantelis, line 167: I think `installedCables` below will have to index by `familyCableName` as well but we can address that when cable drivers handle the family.

pkg/controllers/tunnel/tunnel.go
```diff
@@ -76,26 +87,42 @@ func (c *controller) handleCreatedOrUpdatedEndpoint(obj runtime.Object, _ int) b
...
-		logger.Errorf(err, "Error installing cable for Endpoint %#v", endpoint)
-		return true
+	commonIPFamilies := findCommonIPFamilies(c.localIPFamilies, endpoint.Spec.GetIPFamilies())
+
+	var errs []error
+
+	for _, family := range commonIPFamilies {
+		err := c.engine.InstallCable(endpoint, family)
+		if err != nil {
+			logger.Errorf(err, "Error installing IP%v cable for Endpoint %#v", family, endpoint)
```
tpantelis:
```suggestion
			logger.Errorf(err, "Error installing IPv%v cable for Endpoint %#v", family, endpoint)
```

pkg/controllers/tunnel/tunnel.go
```diff
@@ -76,26 +87,42 @@ func (c *controller) handleCreatedOrUpdatedEndpoint(obj runtime.Object, _ int) b
...
+
 	logger.V(log.DEBUG).Infof("Tunnel controller processing removed submariner Endpoint object: %#v", endpoint)
 
-	if err := c.engine.RemoveCable(endpoint); err != nil {
-		logger.Errorf(err, "Tunnel controller failed to remove Endpoint cable %#v from the engine", endpoint)
+	var errs []error
+
+	for _, family := range commonIPFamilies {
+		if err := c.engine.RemoveCable(endpoint, family); err != nil {
+			logger.Errorf(err, "Tunnel controller failed to remove Endpoint IP%v cable %#v from the engine", family, endpoint)
```
tpantelis:
```suggestion
			logger.Errorf(err, "Tunnel controller failed to remove Endpoint IPv%v cable %#v from the engine", family, endpoint)
```

pkg/endpoint/local_ip.go
```diff
@@ -36,14 +41,18 @@ func GetLocalIPForDestination(dst string) string {
 	return localAddr.IP.String()
 }
 
-func GetLocalIP(family k8snet.IPFamily) string {
+func GetLocalIPForDest(dst string, family k8snet.IPFamily) string {
```
tpantelis: I think the names `GetLocalIPForDest` and `GetLocalIPForDestination` are a bit ambiguous wrt to their purpose. As I suggested earlier, I think we should name this `GetLocalIPForDestination`. The prior `GetLocalIPForDestination` on L32 is only for IPv4 (currently anyway) so I think we should name it accordingly. Also I don't think we need to expose it publicly, hence `getLocalIPv4ForDestination`.

pkg/natdiscovery/remote_endpoint.go
```diff
@@ -61,19 +62,22 @@ type NATEndpointInfo struct {
 	Endpoint v1.Endpoint
 	UseNAT   bool
 	UseIP    string
+	Family   k8snet.IPFamily
```
tpantelis: To be consistent with the other field names:

```suggestion
	UseFamily   k8snet.IPFamily
```

tpantelis (changes requested):

pkg/natdiscovery/remote_endpoint.go
yboaron: Done

pkg/endpoint/local_ip.go
yboaron: Done

pkg/controllers/tunnel/tunnel.go
yboaron: Done

tpantelis (approved):

vthapar (approved):

vthapar:
@yboaron Does this need backport to 0.20?

yboaron:
> @yboaron Does this need backport to 0.20?

I don't think so,  only refactoring the code to support IP family which is relevant for IPV6,dual-stack use case (planned for 0.21)

PR #3321, opened 2025-02-27
Add unit tests for the wireguard cable driver

tpantelis:
The driver code was modified to use the netlink wrapper `Interface` and to abstract the wireguard client API behind an interface to facilitate testing.

Also simplified, restructured and removed unnecessary code in the driver to increase coverage and eliminate paths that need to be tested.

### DIFF
--- a/pkg/cable/vxlan/vxlan.go
+++ b/pkg/cable/vxlan/vxlan.go
@@ -157,7 +157,7 @@ func (v *vxLan) ConnectToEndpoint(endpointInfo *natdiscovery.NATEndpointInfo) (s
 		return "", fmt.Errorf("failed to parse remote IP %s", endpointInfo.UseIP)
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/cable/vxlan/vxlan.go
```diff
@@ -157,7 +157,7 @@ func (v *vxLan) ConnectToEndpoint(endpointInfo *natdiscovery.NATEndpointInfo) (s
 		return "", fmt.Errorf("failed to parse remote IP %s", endpointInfo.UseIP)
 	}
 
-	allowedIPs := parseSubnets(remoteEndpoint.Spec.Subnets)
+	allowedIPs := remoteEndpoint.Spec.ParseSubnets()
```
yboaron: Since connection will be either V4 or V6, it makes sense to change ParseSubnets to take IPfamily parameter , we can use UseFamily field from  endpointInfo *natdiscovery.NATEndpointInfo

yboaron (approved):

PR #3329, opened 2025-03-03
Add IP family support to Libreswan driver

yboaron:


pkg/cable/libreswan/libreswan_internal_test.go
```diff
@@ -106,8 +121,9 @@ func testConnectToEndpoint() {
 					Subnets:    []string{"20.0.0.0/16"},
 				},
 			},
-			UseIP:  "172.93.2.1",
-			UseNAT: true,
+			UseIP:     "172.93.2.1",
+			UseNAT:    true,
+			UseFamily: k8snet.IPv4,
```
tpantelis, line 130: Can you add test cases for IPv6 as well, both here and in `testDisconnectFromEndpoint` and `testGetConnections`? Or, if you prefer, I can add them in a follow-up PR.

pkg/apis/submariner.io/v1/endpoint.go
```diff
@@ -198,3 +199,19 @@ func (ep *EndpointSpec) ParseSubnets(family k8snet.IPFamily) []net.IPNet {
 
 	return subnets
 }
+
+func (ep *EndpointSpec) ExtractSubnets(family k8snet.IPFamily) []string {
```
tpantelis: So this method internally uses `GetPrivateIP`, which is hidden to the user. Perhaps this is specific to the libreswan driver but, either way, I suggest we make it explicit by naming it `ExtractPrivateIPSubnets` or (better yet I think) generalize it and pass in the IP, ie `ExtractSubnets(forIP string)` (note that we wouldn't need the family passed in since we can determine it from the IP).

pkg/apis/submariner.io/v1/endpoint.go
```diff
@@ -198,3 +199,19 @@ func (ep *EndpointSpec) ParseSubnets(family k8snet.IPFamily) []net.IPNet {
...
+	for _, subnet := range ep.ParseSubnets(family) {
+		subnetStr := subnet.String()
+
+		if !strings.HasPrefix(subnetStr, privatePrefix) {
+			subnets = append(subnets, subnetStr)
+		}
+	}
+
+	return subnets
+}
```
tpantelis, line 217: Please add unit tests for this new method.

pkg/cable/libreswan/libreswan.go
```diff
@@ -173,12 +179,12 @@ func (i *libreswan) Init() error {
...
 // Line format:
-// 006 #3: "submariner-cable-cluster3-172-17-0-8-0-0", type=ESP, add_time=1590508783, inBytes=0, outBytes=0, id='172.17.0.8'
+// 006 #3: "submariner-cable-cluster3-172-17-0-8-v4-0-0", type=ESP, add_time=1590508783, inBytes=0, outBytes=0, id='172.17.0.8'
 // or:
-// 006 #2: "submariner-cable-cluster3-172-17-0-8-0-0"[1] 3.139.75.179, type=ESP, add_time=1617195756, inBytes=0, outBytes=0, \
+// 006 #2: "submariner-cable-cluster3-172-17-0-8-v4-0-0"[1] 3.139.75.179, type=ESP, add_time=1617195756, inBytes=0, outBytes=0, \
 // id='@10.0.63.203-0-0'"
 // .
-var trafficStatusRE = regexp.MustCompile(`.* "([^"]+)"[^,]*, .*inBytes=(\d+), outBytes=(\d+).*`)
+var trafficStatusRE = regexp.MustCompile(`.* "([^"]+-v[46]-[0-1]-[0-1])"[^,]*, .*inBytes=(\d+), outBytes=(\d+).*`)
```
tpantelis, line 187: What about migration, ie do we need to handle parsing connections with the prior format obtained from `whack`? (it seems not...)

tpantelis (changes requested):

pkg/cable/libreswan/libreswan.go
yboaron: Good point, 
I think it should be fine. 

IIUC, [Cablenegine](https://github.com/submariner-io/submariner/blob/devel/pkg/cableengine/cableengine.go#L183) will not find an existing active connection even though it exists (with previous naming convention), it needs to call libreswan to Connect to Endpoint and libreswan driver should apply the same connection with the new connection name.

pkg/cable/libreswan/libreswan_internal_test.go
yboaron: I'll add tests for V6

pkg/apis/submariner.io/v1/endpoint_test.go
```diff
@@ -459,3 +462,42 @@ func testParseSubnets() {
...
+		It("should return other IPv4 subnet", func() {
+			result := spec.ExtractSubnetsExcludeIP("10.16.1.0")
+			Expect(result).To(Equal([]string{ipV4CIDR2}))
+		})
+	})
+
+	Context("when an IPv6 address is provided", func() {
+		It("should return IPv6 CIDR", func() {
+			result := spec.ExtractSubnetsExcludeIP("3002::1234:abcd:ffff:c0a8:101")
+			fmt.Printf("result is %v", result)
```
tpantelis:
```suggestion
```

pkg/apis/submariner.io/v1/endpoint_test.go
```diff
@@ -459,3 +462,42 @@ func testParseSubnets() {
...
+			Expect(result).To(Equal([]string{ipV6CIDR}))
+		})
+	})
+
+	Context("when an IPv6 substring address is provided", func() {
+		It("should return empty list", func() {
+			result := spec.ExtractSubnetsExcludeIP("2002:0:0:1234::")
+			Expect(result).To(BeNil())
```
tpantelis:
```suggestion
			Expect(result).To(BeEmpty())
```

pkg/apis/submariner.io/v1/endpoint_test.go
```diff
@@ -459,3 +462,42 @@ func testParseSubnets() {
...
+func testExtractSubnetsExcludeIP() {
+	var spec *v1.EndpointSpec
+
+	BeforeEach(func() {
+		spec = &v1.EndpointSpec{
+			Subnets: []string{ipV4CIDR, ipV4CIDR2, ipV6CIDR},
+		}
+	})
+
+	Context("when an IPv4 address is provided", func() {
```
tpantelis:
```suggestion
	When("an IPv4 address is provided", func() {
```

pkg/apis/submariner.io/v1/endpoint_test.go
```diff
@@ -459,3 +462,42 @@ func testParseSubnets() {
...
+		})
+	})
+
+	Context("when an IP is not present in any subnet", func() {
+		It("should return both IPv4 subnets", func() {
+			result := spec.ExtractSubnetsExcludeIP("192.168.1.1")
+			Expect(result).To(Equal([]string{ipV4CIDR, ipV4CIDR2}))
+		})
+	})
+}
```
tpantelis, line 507: I suggest we structure the test cases as follows:

```
Context("with two IPv4 subnets and one IPv6 subnet", func() {
    When("an IPv4 address is provided", func() {
        Context("and it's present in an IPv4 subnet", func() {
            It("should return the other IPv4 subnet", func() {
	        result := spec.ExtractSubnetsExcludeIP("10.16.1.0")
	        Expect(result).To(Equal([]string{ipV4CIDR2}))
            })
        })

        Context("and it's not present in any IPv4 subnet", func() {
            It("should return both IPv4 subnets", func() {
                result := spec.ExtractSubnetsExcludeIP("192.168.1.1")
                Expect(result).To(Equal([]string{ipV4CIDR, ipV4CIDR2}))
            })
        })
    })
 
    When("an IPv6 address is provided", func() {
        Context("and it's present in the IPv6 subnet", func() {
            It("should return empty", func() {
	        result := spec.ExtractSubnetsExcludeIP("2002:0:0:1234::")
	        Expect(result).To(BeEmpty())
            })
        })

        Context("and it's not present in the IPv6 subnet", func() {
            It("should return the IPv6 subnet", func() {
                result := spec.ExtractSubnetsExcludeIP("3002::1234:abcd:ffff:c0a8:101")
                Expect(result).To(Equal([]string{ipV6CIDR}))
            })
        })
    })
})
```

pkg/apis/submariner.io/v1/endpoint_test.go
```diff
@@ -459,3 +462,42 @@ func testParseSubnets() {
 		}))
 	})
 }
+
+func testExtractSubnetsExcludeIP() {
```
tpantelis:
```suggestion
func testExtractSubnetsExcludingIP() {
```

pkg/cable/libreswan/libreswan_internal_test.go
```diff
@@ -95,6 +110,8 @@ func testConnectToEndpoint() {
 	t := newTestDriver()
 
 	var natInfo *natdiscovery.NATEndpointInfo
+	var natInfoIPv6 *natdiscovery.NATEndpointInfo
+	var natInfoDualStack *natdiscovery.NATEndpointInfo
```
tpantelis:
```suggestion
	var (
	     natInfo *natdiscovery.NATEndpointInfo
	     var natInfoIPv6 *natdiscovery.NATEndpointInfo
             natInfoDualStack *natdiscovery.NATEndpointInfo
         )
```

tpantelis (changes requested):

tpantelis (approved):

PR #3334, opened 2025-03-06
Refactor NAT discovery unit tests to use the public Interface

tpantelis:
This makes them more "whitebox" and moves the files to the `natdiscovery_test` package to avoid leakage of test code into the production package. This will also make it a bit easier to implement IPv6.
    
Two of the test files, _request_handle_internal_test.go_ and _listener_internal_test.go_, were not moved to the test package as they test specific internal functions which would entail making too much of the production code public.
    
A new function, `NewWithConfig`, was added that allows the unit tests to construct the `natDiscovery` instance with hooks to mock behavior.


skitt (approved):

pkg/natdiscovery/listener.go
```diff
@@ -71,21 +86,23 @@ func (nd *natDiscovery) runListenerV4(stopCh <-chan struct{}) error {
 	return nil
 }
 
-func createServerConnection(port int32) (*net.UDPConn, error) {
-	serverAddress, err := net.ResolveUDPAddr("udp4", ":"+strconv.Itoa(int(port)))
+func createServerConnection(port int32, _ k8snet.IPFamily) (ServerConnection, error) {
```
yboaron, line 89: Does 'createServerConnection' function should be used for both v4 and v6 ?
if the answer is yes, maybe we should update the code to check family value , or at least leave TODO V6 comment

tpantelis: yeah the intention is to use this for v6 as well. It's hardcoded to v4 like it was before. I assume we would just set `network = "udp6"`? My intention wasn't to implement v6 in this PR. We have a task on the epic for that (_Add support for IPv6 Nat Discovery_) and there's already TODOs in the code (eg L59) so we probably don't need another.

yboaron (approved):

PR #3339, opened 2025-03-09
Configure packefilter driver via ConfigMap setting

tpantelis:


yboaron:
Pushed https://github.com/submariner-io/submariner-operator/pull/3391 to update GN RBAC

pkg/globalnet/main.go
```diff
@@ -86,8 +87,14 @@ func main() {
 	dynClient, err := dynamic.NewForConfig(cfg)
 	logger.FatalOnError(err, "Unable to create dynamic client")
 
-	// Set packetfilter driver to iptables. Once nftables is available, we'll check which driver is supported.
-	packetfilter.SetNewDriverFn(iptables.New)
+	// set up signals so we handle the first shutdown signal gracefully
+	ctx := signals.SetupSignalHandler()
+
+	globalConfigMap, err := configmap.Get(ctx, resource.ForConfigMap(k8sClient, spec.Namespace), configmap.Global)
+	logger.FatalOnError(err, "Error retrieving the global ConfigMap")
```
yboaron, line 94: QQ: With Admiral recent change, we should fatal error here only if  we couldn't read resource from k8s-api (if global cm doesn't exist, we should get empty cm) , right ?

pkg/globalnet/main.go
```diff
@@ -86,8 +87,14 @@ func main() {
 	dynClient, err := dynamic.NewForConfig(cfg)
 	logger.FatalOnError(err, "Unable to create dynamic client")
 
-	// Set packetfilter driver to iptables. Once nftables is available, we'll check which driver is supported.
-	packetfilter.SetNewDriverFn(iptables.New)
+	// set up signals so we handle the first shutdown signal gracefully
+	ctx := signals.SetupSignalHandler()
+
+	globalConfigMap, err := configmap.Get(ctx, resource.ForConfigMap(k8sClient, spec.Namespace), configmap.Global)
```
yboaron, line 93: Can we move cm reading into packetfilter/configure function ?

pkg/packetfilter/configure/configure.go
```diff
@@ -0,0 +1,60 @@
...
+
+			useNftables, err = strconv.ParseBool(value)
+			if err != nil {
+				return errors.Wrapf(err, "unable to parse %q from ConfigMap %q", UseNftablesKey, cm.Name)
+			}
+		}
+	}
+
+	if useNftables {
+		logger.Info("Using nftables packet filter")
```
yboaron: "Using nftables driver" ?

pkg/globalnet/main.go
tpantelis: yeah `configmap.Get ` only returns an error in that case.

yboaron:
Should be update shipyard to latest submariner-operator and then point submariner to latest shipyard?  to address the GN e2e failures

tpantelis:
> Should be update shipyard to latest submariner-operator and then point submariner to latest shipyard? to address the GN e2e failures

It's subctl not shipyard that needs to be updated. I pushed https://github.com/submariner-io/subctl/pull/1337

yboaron (approved):

skitt (approved):

PR #3346, opened 2025-03-11
Handle stale remote Endpoints removed in the route-agent health checker

### LINKED ISSUE #3335: Stale remote Endpoints in the RouteAgent resource
In a production environment, it was observed that there were multiple remote `Endpoints` for the same cluster listed in the `RouteAgent` resource. One was current and the others were stale. This message was also observed in the route-agent log:

`[90m2025-03-03T09:52:48.243Z[0m [32mINF[0m ../endpoint_removed.go:73 [1mEventController      Ignoring deleted remote &v1.Endpoint{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"ocp-dr-mz-submariner-cable-ocp-dr-mz-172-17-24-34", GenerateName:"", Namespace:"submariner-operator", SelfLink:"", UID:"b943f82c-37c0-4afc-9cf4-ce069171831e", ResourceVersion:"596521487", Generation:2, CreationTimestamp:time.Date(2025, time.March, 3, 8, 56, 16, 0, time.Local), DeletionTimestamp:<nil>, DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string{"submariner-io/clusterID":"ocp-dr-mz"}, Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ManagedFields:[]v1.ManagedFieldsEntry(nil)}, Spec:v1.EndpointSpec{ClusterID:"ocp-dr-mz", CableName:"submariner-cable-ocp-dr-mz-172-17-24-34", HealthCheckIP:"242.1.255.254", Hostname:"worker6", Subnets:[]string{"242.1.0.0/16"}, PrivateIP:"172.17.24.34", PublicIP:"", NATEnabled:false, Backend:"libreswan", BackendConfig:map[string]string{"natt-discovery-port":"4490", "preferred-server":"false", "udp-port":"4500"}}} since a later endpoint was alreadyprocessed`

for the stale `Endpoints`. The event controller tries to be smart and elide notifying the listener if the delete event occurs after the create event for the new `Endpoint`. This behavior is desired for other listeners but not the health checker. We need to extend the event `Handler` interface to account for this, perhaps add a new method `StaleRemoteEndpointRemoved` or add a flag to the existing `RemoteEndpointRemoved` method.

tpantelis:
See commits for details.

Fixes https://github.com/submariner-io/submariner/issues/3335

### DIFF
--- a/pkg/event/controller/endpoint_removed.go
+++ b/pkg/event/controller/endpoint_removed.go
@@ -67,16 +67,15 @@ func (c *handlerController) handleRemovedLocalEndpoint(endpoint *smv1.Endpoint) 
 }
 
 func (c *handlerController) handleRemovedRemoteEndpoint(endpoint *smv1.Endpoint) error {
+	c.handlerState.remoteEndpoints.Delete(endpoint.Name)
+
 	lastProcessedTime, ok := c.remoteEndpointTimeStamp[endpoint.Spec.ClusterID]
 
 	if ok && lastProcessedTime.After(endpoint.CreationTimestamp.Time) {
-		logger.Infof("Ignoring deleted remote %#v since a later endpoint was already"+
-			"processed", endpoint)
-		return nil
+		return c.handler.StaleRemoteEndpointRemoved(endpoint) //nolint:wrapcheck  // Let the caller wrap it
 	}
 
 	delete(c.remoteEndpointTimeStamp, endpoint.Spec.ClusterID)
-	c.handlerState.remoteEndpoints.Delete(endpoint.Name)
 
 	return c.handler.RemoteEndpointRemoved(endpoint) //nolint:wrapcheck  // Let the caller wrap it
 }
--- a/pkg/event/controller/controller_test.go
+++ b/pkg/event/controller/controller_test.go
@@ -137,15 +137,41 @@ var _ = Describe("Event controller", func() {
 				Spec:       submV1.EndpointSpec{ClusterID: "east"},
 			})
 			t.ensureNoEvents()
+			Expect(t.handler.remoteEndpoints.Load()).To(Equal([]submV1.Endpoint{*latestEndpoint}))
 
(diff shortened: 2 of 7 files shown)


pkg/event/controller/endpoint_removed.go
```diff
@@ -67,16 +67,15 @@ func (c *handlerController) handleRemovedLocalEndpoint(endpoint *smv1.Endpoint)
 }
 
 func (c *handlerController) handleRemovedRemoteEndpoint(endpoint *smv1.Endpoint) error {
+	c.handlerState.remoteEndpoints.Delete(endpoint.Name)
+
 	lastProcessedTime, ok := c.remoteEndpointTimeStamp[endpoint.Spec.ClusterID]
 
 	if ok && lastProcessedTime.After(endpoint.CreationTimestamp.Time) {
```
yboaron, line 74: I'm not very familiar with the endpoints creation/deletion order in case of multiple GW failover use case, could this change affect ep events order for multi GW fail over use case ?

tpantelis: Sorry not sure I follow. Can you elaborate more, perhaps with an example?

tpantelis: Are you referring to moving line `c.handlerState.remoteEndpoints.Delete(endpoint.Name)` up to the top? If so, we have to always remove the deleted endpoint from `remoteEndpoints`, regardless if stale or not. This was a latent bug uncovered by the RA health checker unit tests.

yboaron: > Are you referring to moving line c.handlerState.remoteEndpoints.Delete(endpoint.Name) up to the top?

Yep. got it.

yboaron (approved):

dfarrell07 (approved):

PR #3350, opened 2025-03-13
Packetfilter API: add support for IPv6

tpantelis:
Modified the `DriverFn` and `packetfilter.New` functions to take the IP family.

### DIFF
--- a/pkg/packetfilter/nftables/namedset.go
+++ b/pkg/packetfilter/nftables/namedset.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/submariner-io/submariner/pkg/packetfilter"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	k8snet "k8s.io/utils/net"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/knftables"
 )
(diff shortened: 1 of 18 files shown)


pkg/packetfilter/nftables/nftables.go
```diff
@@ -79,7 +80,11 @@ type packetFilter struct {
 	nftables knftables.Interface
 }
 
-func New() (packetfilter.Driver, error) {
+func New(family k8snet.IPFamily) (packetfilter.Driver, error) {
+	if family != k8snet.IPv4 {
```
yboaron: Maybe we can also enable  IPV6 driver creation for nftables, it should be safe because currently all clients will only create a V4 driver

yboaron (approved):

pkg/packetfilter/nftables/namedset.go
```diff
@@ -35,6 +36,9 @@ type namedSet struct {
 
 func (p *packetFilter) NewNamedSet(set *packetfilter.SetInfo) packetfilter.NamedSet {
 	setType := "ipv4_addr"
+	if p.family == k8snet.IPv6 {
+		setType = "ipv6_addr"
```
tpantelis, line 40: @yboaron I assume this is correct for IPv6? I couldn't find any concrete info/doc to confirm it...

tpantelis:
@yboaron I [updated the PR](https://github.com/submariner-io/submariner/compare/71822a74216a4e8a283e9d23976ddf93a9778be8..54899922041fd3c76337169b2a81432f9d7d24ce) to add verification of the expected IP family of IPs/CIDRs passed via the API (eg Rules) in the fake packet filter implementation and to [remove the Family field from SetInfo](https://github.com/submariner-io/submariner/pull/3350/commits/54899922041fd3c76337169b2a81432f9d7d24ce).

dfarrell07 (approved):

PR #3360, opened 2025-03-20
MTU handler: use shorter names for packet filter sets

### LINKED ISSUE #3359: Submariner with nftables: routeagent pod reports error when configuring MSS clamp rule
**What happened**:

RA pod logs failed to configure [1] nftable rule when Submariner is deployed with nftables.

[1]
"ip saddr @SUBMARINER-LOCALCIDRS ip daddr @SUBMARINER-REMOTECIDRS tcp flags syn / syn,rst counter packets 0 bytes 0 tcp option maxseg size set rt mtu comment "ip saddr @SUBMARINER-LOCALCIDRS ip daddr @SUBMARINER-REMOTECIDRS tcp flags syn / syn,rst counter tcp option maxseg size set rt mtu"

Seems that root cause is nftables rule comment field exceeding max size (128)

**Environment**:
- Install tools:
- Others:

yboaron:
Since nftables driver ListRules function returns only comments and handles, we set the comment value to rulepec string.
However, nftables comment field is limited to 128 characters. To avoid exceeding this limit, we should use shorter names for SUBMARINER-REMOTECIDRS and SUBMARINER-LOCALCIDRS sets.

Fixes: https://github.com/submariner-io/submariner/issues/3359

### DIFF
--- a/pkg/routeagent_driver/constants/constants.go
+++ b/pkg/routeagent_driver/constants/constants.go
@@ -27,8 +27,8 @@ const (
 	InputChain         = "INPUT"
 	ForwardChain       = "FORWARD"
 	MangleTable        = "mangle"
-	RemoteCIDRIPSet    = "SUBMARINER-REMOTECIDRS"
-	LocalCIDRIPSet     = "SUBMARINER-LOCALCIDRS"
+	RemoteCIDRIPSet    = "SUBMARINER-RCIDRS"
+	LocalCIDRIPSet     = "SUBMARINER-LCIDRS"
 
 	RouteAgentInterClusterNetworkTableID = 149
 


pkg/routeagent_driver/constants/constants.go
```diff
@@ -27,8 +27,8 @@ const (
 	InputChain         = "INPUT"
 	ForwardChain       = "FORWARD"
 	MangleTable        = "mangle"
-	RemoteCIDRIPSet    = "SUBMARINER-REMOTECIDRS"
```
yboaron, line 30: @tpantelis , Do you think we should flush/remove also sets with previous names in uninstall (don't leave orphan sets after submariner uninstall) ?

tpantelis:
So the MSS clamp rule just exceeds 128 chars when "stringified". Shortening just those two set names would get it under 128 but it seems like a bandaid. We don't need the full syntax needed by nftables as rendered by `ToRuleSpec`. We can shorten the string format of the rule used in the comment. I have cycles to play around with that.

tpantelis:
I was able to shorten this rule:

```
packetfilter.Rule{
    SrcSetName:  "SUBMARINER-REMOTECIDRS",
    DestSetName: "SUBMARINER-LOCALCIDRS",
    MssValue:    "1234567890",
    Action:      packetfilter.RuleActionMss,
    ClampType:   packetfilter.ToValue,
}
```

from 134 chars to 58 by packing the bytes efficiently.

yboaron:
> So the MSS clamp rule just exceeds 128 chars when "stringified". Shortening just those two set names would get it under 128 but it seems like a bandaid. We don't need the full syntax needed by nftables as rendered by `ToRuleSpec`. We can shorten the string format of the rule used in the comment. I have cycles to play around with that.

Yep, this PR is just bandaid, if you have cycles to fix it, I'll close this PR.

yboaron:
This PR is just bandaid to enable Submariner nftables deployment, we prefer to have generic solution for nftables rules comments length.

PR #3361, opened 2025-03-20
configured MSS clamping rules under GetMSSClampTypes() table

### LINKED ISSUE #3355: routeagent pods  stuck in CrashLoopBackOff when Submariner deployed with nftables
I deployed submariner with nftables running : **make deploy using=nftables**

**What happened**:

routeagent pods  stuck in CrashLoopBackOff

$ kubectl --kubeconfig output/kubeconfigs/kind-config-cluster1 get pods -n submariner-operator 
NAME                                   READY   STATUS             RESTARTS      AGE
netshoot-test-4gp79                    1/1     Running            0             23h
netshoot-test-4wcpr                    1/1     Running            0             23h
netshoot-test-brk4t                    1/1     Running            0             23h
submariner-gateway-g9s9c               0/1     CrashLoopBackOff   6 (73s ago)   11m
submariner-metrics-proxy-rfvz7         1/1     Running            0             11m
submariner-operator-65485cc9dd-j5cwm   1/1     Running            0             25h
submariner-routeagent-82zfv            0/1     CrashLoopBackOff   7 (41s ago)   11m
submariner-routeagent-p6l72            0/1     CrashLoopBackOff   7 (38s ago)   11m
submariner-routeagent-qs9k9            0/1     CrashLoopBackOff   7 (42s ago)   11m

**The following error reported in routeagent pod logs :** 

2025-03-18T12:20:59.386Z INF ..kg/event/registry.go:78 EventRegistry        Event handler "VXLAN cleanup handler" added to registry "routeagent_driver".
2025-03-18T12:20:59.483Z FTL ..gent_driver/main.go:164 main                 Error registering the handlers error="Event handler \"MTU handler\" failed to initialize: error creating IPHookChain chain SUBMARINER-POSTROUTING: error creating \"SUBMARINER-POSTROUTING\" IPHook Chain: /dev/stdin:2:50-90: Error: Chain of type \"route\" is not supported, perhaps kernel support is missing?\nadd chain ip submariner SUBMARINER-POSTROUTING { type route hook postrouting priority -160 ; }\n                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n"

make deploy using=nftables

**Environment**:
- Install tools:
- Others:

yboaron:
MSS Clamping rules should be configured in different IP hooks for iptables
and nftables.

In iptables, they should be configured in mangle/postrouting hook,
while in nftables, they should be placed in filter/postrouging hook.

This PR configured MSS clamping rules under GetMSSClampTypes() hook

Fixes: https://github.com/submariner-io/submariner/issues/3355


yboaron:
cc @sridhargaddam ^^

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -91,12 +91,12 @@ func (h *mtuHandler) Init(_ context.Context) error {
 	}
 
 	if err := h.pFilter.CreateIPHookChainIfNotExists(&packetfilter.ChainIPHook{
-		Name:     constants.SmPostRoutingChain,
-		Type:     packetfilter.ChainTypeRoute,
-		Hook:     packetfilter.ChainHookPostrouting,
+		Name:     constants.SmForwardChain,
+		Type:     packetfilter.ChainTypeFilter,
+		Hook:     packetfilter.ChainHookForward,
```
sridhargaddam: Instead of applying the MSS clamping rules in the `forward` hook, can you try in the `postrouting` hook so that it applies to both the forwarded traffic as well as traffic that is locally generated.

For example:
```
# nft add table ip mangletcpmss
# nft add chain ip mangletcpmss customchain { type filter hook postrouting priority 0 \; }
# nft add rule ip mangletcpmss customchain tcp flags syn tcp option maxseg size set rt mtu
# nft list ruleset
table ip mangletcpmss {
	chain customchain {
		type filter hook postrouting priority filter; policy accept;
		tcp flags syn tcp option maxseg size set rt mtu
	}
}
#
```

yboaron: Done

sridhargaddam (approved):

yboaron:
Well, iptables doesn't support filter/postrouting  hook ... 

[1] 
` running [/usr/sbin/iptables -t filter -I POSTROUTING 1 -j SUBMARINER-POSTROUTING --wait 5]: exit status 1: iptables: No chain/target/match by that name.\n"`

https://upload.wikimedia.org/wikipedia/commons/3/37/Netfilter-packet-flow.svg

pkg/packetfilter/packetfilter.go
```diff
@@ -272,6 +272,10 @@ type NamedSet interface {
 }
 
 type Driver interface {
+	// MSS Clamp rules should be configured in different tables for iptables and nftables,
+	// TableTypeRoute for iptables and TableTypeFilter for nftables
+	GetMSSClampTable() TableType
```
tpantelis: Why don't we define a new `TableType` enum named "_MSSClamp_" instead of adding a new interface method? Then the implementation can internally decide what it should translate to.

yboaron: Also thought to go with this approach first, for this approach we need two new values : 
1. new table type e.g: **TableTypeMssClamp**  , 
    and **AppendUnique , ClearChain , UpdateChainRules** APIs implementation should be updated to support this value 

2. new chain type e.g: **ChainTypeMssClamp**
 and **CreateIPHookChainIfNotExists , DeleteIPHookChain**  APIs implementation should be updated to support this value 

But in second thought, adding new GetMSSClampTable() API seems like a simpler solution to me

pkg/routeagent_driver/handlers/mtu/mtuhandler.go
```diff
@@ -90,9 +92,16 @@ func (h *mtuHandler) Init(_ context.Context) error {
 		return errors.Wrap(err, "error initializing iptables")
 	}
 
+	h.tableType = h.pFilter.GetMSSClampTable()
+	h.chainType = packetfilter.ChainTypeRoute
+
+	if h.tableType == packetfilter.TableTypeFilter {
+		h.chainType = packetfilter.ChainTypeFilter
+	}
```
tpantelis: Here the API tells which table type to use but the user has to figure out the matching chain type. How about we modify `GetMSSClampTable` to also return the chain type:

  `GetMSSClampInfo() (TableType, ChainType)`

yboaron: Sounds good, I'll change it.

pkg/packetfilter/packetfilter.go
```diff
@@ -272,6 +272,10 @@ type NamedSet interface {
 }
 
 type Driver interface {
+	// MSS Clamp rules should be configured in different tables for iptables and nftables,
+	// TableTypeRoute for iptables and TableTypeFilter for nftables
+	GetMSSClampTable() (TableType, ChainType)
```
tpantelis:
```suggestion
	GetMSSClampTypes() (TableType, ChainType)
```

tpantelis (changes requested):

tpantelis (approved):

PR #3363, opened 2025-03-21
Serialize nftable rules in a compressed format

### LINKED ISSUE #3359: Submariner with nftables: routeagent pod reports error when configuring MSS clamp rule

tpantelis:
The `ListRules` function only returns the `Comment` and `Handle` fields so we need to store the rule details in the `Comment`. We were storing the full nft rule spec which can be lengthy and exceed the `Comment` size limit of 128 chars. We can reduce this significantly by serializing the `packetfilter.Rule` fields.

In addition, the rule conversion functions were moved to a separate file to reduce file size.

Fixes https://github.com/submariner-io/submariner/issues/3359


pkg/packetfilter/nftables/rule_conversion.go
```diff
@@ -0,0 +1,229 @@
...
+func mustWriteString(buf *bytes.Buffer, s string) {
+	_, err := buf.WriteString(s + " ")
+	utilruntime.Must(err)
+}
+
+func mustWriteUint32(buf *bytes.Buffer, i uint32) {
+	mustWriteString(buf, strconv.FormatInt(int64(i), 10))
+}
+
+func SerializeRule(rule *packetfilter.Rule) string {
```
yboaron, line 148: Considering the proposed Serialize/Deserialize approach and the potential need to address future backward compatibility, I wonder if it might be simpler only to store compressed ruleSpec string in Comment field. This way, decoding would consistently rely on nft syntax and no need to address potential backward compatibility in future/

tpantelis: You mean use gzip compression?  I played around with that and it ends up being larger in size due to added overhead. Compression really only yields benefit for data sizes much larger than what we're dealing with here. The nft syntax is pretty "wordy" - I don't know what way we could reduce the size more efficiently than the simple serialization approach I implemented. It's probably unlikely we would change the Rule structure in the future in any significant way, if at all,  but, even so,  it should be easy to handle backwards compatibility.

yboaron: Yep, I meant gzip compression.
Also not sure how likely it would be to change packet filter rule in future, maybe we should at least leave a note that Serialize/Deserialize should be updated if the pf rule changed.

pkg/packetfilter/nftables/rule_conversion.go
```diff
@@ -0,0 +1,229 @@
...
+	var buf bytes.Buffer
+
+	_, err := buf.WriteString(serializedVersion)
+	utilruntime.Must(err)
+
+	mustWriteUint32(&buf, uint32(rule.Action))
+	mustWriteUint32(&buf, uint32(rule.ClampType))
+	mustWriteUint32(&buf, uint32(rule.Proto))
+
+	mustWriteString(&buf, rule.SrcSetName)
```
yboaron: Why do we want to panic here ?  don't we want to return an error and let the caller decide how to handle it ?

tpantelis: `buf.WriteString` documents that the return "_err is always nil. If the buffer becomes too large, WriteString will panic with [ErrTooLarge]._". So I didn't want to pollute the code with a lot of unnecessary error checking. The other option is to ignore the return error instead of panicking. I'm fine either way.

We could get an error in `Deserialize` while reading strings and parsing to integers. I did do error checking and error propagation when parsing the serialized version. Not sure how this could happen outside of someone manually entering a rule out-of-band in the submariner table. We could just log the error and not propagate. After successfully reading the serialized version, if a parsing error occurs after that, I panic b/c that means there's serious corruption in nft.

yboaron: if buf.WriteString always returns nil or panic with ErrTooLarge , maybe we should ignore the return error as you suggeseted.

yboaron (approved):

vthapar (approved):

PR #3365, opened 2025-03-25
Add dual-stack support to cable engine health checker

tpantelis:
On remote `Endpoint`, for each supported IP family, start a pinger. Also added IP family param to `GetLatencyInfo`.

### DIFF
--- a/pkg/cableengine/healthchecker/healthchecker.go
+++ b/pkg/cableengine/healthchecker/healthchecker.go
@@ -34,17 +34,18 @@ import (
 
 type Interface interface {
 	Start(stopCh <-chan struct{}) error
-	GetLatencyInfo(endpoint *submarinerv1.EndpointSpec) *pinger.LatencyInfo
+	GetLatencyInfo(endpoint *submarinerv1.EndpointSpec, ipFamily k8snet.IPFamily) *pinger.LatencyInfo
 	Stop()
 }
 
 type Config struct {
-	WatcherConfig      *watcher.Config
-	EndpointNamespace  string
-	ClusterID          string
-	PingInterval       int
-	MaxPacketLossCount int
-	NewPinger          func(pinger.Config) pinger.Interface
+	WatcherConfig       *watcher.Config
+	SupportedIPFamilies []k8snet.IPFamily
+	EndpointNamespace   string
+	ClusterID           string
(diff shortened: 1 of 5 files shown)


pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -123,27 +128,36 @@ func (h *controller) endpointCreatedOrUpdated(obj runtime.Object, _ int) bool {
...
-
 	h.Lock()
 	defer h.Unlock()
 
-	if pingerObject, found := h.pingers[endpointCreated.Spec.CableName]; found {
-		if pingerObject.GetIP() == endpointCreated.Spec.GetHealthCheckIP(k8snet.IPv4) {
-			return false
+	for _, family := range h.config.SupportedIPFamilies {
+		healthCheckIP := endpointCreated.Spec.GetHealthCheckIP(family)
+		if healthCheckIP == "" {
```
yboaron, line 136: I can see that current code also checks if endpointCreated.Spec.CableName == "" , do we need to keep it ?

pkg/cableengine/healthchecker/healthchecker.go
```diff
@@ -172,9 +183,12 @@ func (h *controller) endpointDeleted(obj runtime.Object, _ int) bool {
 	h.Lock()
 	defer h.Unlock()
 
-	if pingerObject, found := h.pingers[endpointDeleted.Spec.CableName]; found {
-		pingerObject.Stop()
-		delete(h.pingers, endpointDeleted.Spec.CableName)
+	for _, family := range h.config.SupportedIPFamilies {
+		cableName := endpointDeleted.Spec.GetFamilyCableName(family)
+		if pingerObject, found := h.pingers[cableName]; found {
```
yboaron: We should be persistent with field name represents endpoint family cable name , we use 'cableName' here and 'endpointName'  in startPinger

pkg/cableengine/healthchecker/healthchecker.go
tpantelis: `CableName` is never empty plus we don't use it directly now.

skitt (approved):

aswinsuryan (approved):

PR #3368, opened 2025-03-26
Change MSS clamp chain name to SUBMARINER-POSTROUTING-MSS

yboaron:
In Submariner nftables implementation, all chains are configured within a single table (named submariner).

We cannot name filter/postrouting chain SUBMARINER-POSTROUTING because nat/postrouting chain with this name already exists.

### DIFF
--- a/pkg/routeagent_driver/constants/constants.go
+++ b/pkg/routeagent_driver/constants/constants.go
@@ -20,15 +20,16 @@ package constants
 
 const (
 	// IPTable chains used by RouteAgent.
-	SmPostRoutingChain = "SUBMARINER-POSTROUTING"
-	SmInputChain       = "SUBMARINER-INPUT"
-	SmForwardChain     = "SUBMARINER-FORWARD"
-	PostRoutingChain   = "POSTROUTING"
-	InputChain         = "INPUT"
-	ForwardChain       = "FORWARD"
-	MangleTable        = "mangle"
-	RemoteCIDRIPSet    = "SUBMARINER-REMOTECIDRS"
-	LocalCIDRIPSet     = "SUBMARINER-LOCALCIDRS"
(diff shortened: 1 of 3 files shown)


tpantelis (approved):

tpantelis:
`iptables v1.8.10 (nf_tables): chain name "SUBMARINER-POSTROUTING-MSSCLAMP" too long (must be under 29 chars)`

tpantelis:
@yboaron I think the real issue here is that `nftables.CreateIPHookChainIfNotExists` does not actually check if the chain already exists like the iptables impl does. Interestingly, I tried manually adding a chain via the CLI multiple times via

`sudo /usr/sbin/nft add chain ip submariner POSTROUTING`

and it succeeded each time. So it ignored that it already existed.

EDIT: correction... the issue occurred creating differing hook chains, ie

```
nft 'add chain ip submariner POSTROUTING { type filter hook postrouting priority 0 ; }'
nft 'add chain ip submariner POSTROUTING { type nat hook postrouting priority 0 ; }'
```

In this case the second one fails with:

```
Error: Could not process rule: File exists
add chain ip submariner POSTROUTING { type nat hook postrouting priority 0 ; }
                        ^^^^^^^^^^^
```

The CLI command `nft --json list chains` returns all of the information,

`{"family": "ip6", "table": "filter", "name": "DOCKER-USER", "handle": 13}}, {"chain": {"family": "ip", "table": "submariner", "name": "POSTROUTING", "handle": 2, "type": "filter", "hook": "postrouting", "prio": 0, "policy": "accept"}`

but the knftables `List` method only returns the chain names so, unfortunately, we can't check if the chain already exists in its entirety. But we can add a check in the fake packetfilter to verify the nftables behavior.

PR #3369, opened 2025-03-27
Implement IPv6 health check IP resolution

tpantelis:
Use the CNI discovery to obtain an IPv6 IPv6 health check IP as is done for IPv4. The `cni.Discover` function was modified to take an additional IP family param.

Also unit tests were added for `cni.Discover`. To facilitate this, a function hook was added that returns the host interfaces. External unit tests now override this hook rather than the `cni.Discover` function itself.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -20,8 +20,8 @@ package types
 
 import (
 	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cidr"
 	k8snet "k8s.io/utils/net"
-	"k8s.io/utils/set"
 )
 
 type SubmarinerCluster struct {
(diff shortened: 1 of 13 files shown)


skitt (approved):

pkg/types/types.go
```diff
@@ -54,10 +54,5 @@ type SubmarinerSpecification struct {
 }
 
 func (subSpec *SubmarinerSpecification) GetIPFamilies() []k8snet.IPFamily {
-	ipFamilies := set.New[k8snet.IPFamily]()
-
-	// TODO_IPV6: set ipFamilies according to ClusterCidr content
-	ipFamilies.Insert(k8snet.IPv4)
-
-	return ipFamilies.UnsortedList()
+	return cidr.ExtractIPFamilies(subSpec.ClusterCidr)
```
yboaron, line 57: I think this change might break dual-stack CI gates , lets wait and see dual-stack tests results

tpantelis: They all passed. They may start failing as more and more functionality is added, if so we can mark them as non-required.

yboaron: Hmm, for a dual-stack gate, I'm not sure how getPublicIP works for IPv6. But yeah, we can address it when the public IP functionality is added.

[1] 
https://github.com/submariner-io/submariner/blob/devel/pkg/endpoint/local_endpoint.go#L189

yboaron (approved):

pkg/types/types.go
tpantelis: `getPublicIP` returns an empty string and no error for IPv6: https://github.com/submariner-io/submariner/blob/959fb14c9b1d497f7cdb00b86046bc25a9b25fcf/pkg/endpoint/public_ip.go#L117. And `endpointSpec.SetPublicIP` ignores an empty string.

PR #3373, opened 2025-04-01
Add public IPv6 resolution

yboaron:
Depend on: https://github.com/submariner-io/shipyard/pull/1900


pkg/endpoint/public_ip.go
```diff
@@ -41,24 +41,48 @@ import (
...
+		`(25[0-5]|2[0-4][0-9]|[0-1]?[0-9][0-9]?))$`)
+)
+
+func getPublicIPResolvers(family k8snet.IPFamily) string {
+	var serverList []string
 
-func getPublicIPResolvers() string {
-	serverList := []string{
-		"api:ip4.seeip.org", "api:ipecho.net/plain", "api:ifconfig.me",
+	serverList = []string{
```
tpantelis: I think this makes it clearer:

```
var serverList []string

switch family {
    case k8snet.IPv4:
        serverList = []string{...}
    case k8snet.IPv6:
        serverList = []string{
            "api:api64.ipify.org", "api:6.icanhazip.com"
        }
}
```

pkg/endpoint/public_ip.go
```diff
@@ -205,44 +235,62 @@ func publicLoadBalancerIP(clientset kubernetes.Interface, namespace, loadBalance
...
-			ip, err = publicDNSIP(clientset, namespace, ingress.Hostname)
-			return err
-
-		default:
-			return errors.Errorf("no IP or Hostname for service LoadBalancer %q Ingress", loadBalancerName)
+		for _, ingress := range service.Status.LoadBalancer.Ingress {
+			switch {
+			case ingress.IP != "":
+				if (family == k8snet.IPv4 && net.ParseIP(ingress.IP).To4() != nil) ||
+					(family == k8snet.IPv6 && net.ParseIP(ingress.IP).To4() == nil) {
```
```suggestion
				if k8snet.IPFamilyOfString(ingress.IP) == family {
```

pkg/endpoint/public_ip.go
```diff
@@ -205,44 +235,62 @@ func publicLoadBalancerIP(clientset kubernetes.Interface, namespace, loadBalance
...
-		sort.Slice(ips, func(i, j int) bool {
-			return bytes.Compare(ips[i], ips[j]) < 0
+	var filteredIPs []net.IP
+
+	for _, ip := range ips {
+		if family == k8snet.IPv4 && ip.To4() != nil {
+			filteredIPs = append(filteredIPs, ip)
+		} else if family == k8snet.IPv6 && ip.To4() == nil {
+			filteredIPs = append(filteredIPs, ip)
+		}
```
```suggestion
		if k8snet.IPFamilyOf(ip) == family {
			filteredIPs = append(filteredIPs, ip)
		}
```

tpantelis (changes requested):

pkg/endpoint/public_ip.go
```diff
@@ -41,24 +41,51 @@ import (
...
+			`(?:[a-f0-9]{1,4}:){1,7}:?[a-f0-9]{1,4}`,
+	)
+)
+
+func getPublicIPResolvers(family k8snet.IPFamily) string {
+	var serverList []string
 
-func getPublicIPResolvers() string {
-	serverList := []string{
-		"api:ip4.seeip.org", "api:ipecho.net/plain", "api:ifconfig.me",
```
skitt, line 57: Why are these being removed?

pkg/endpoint/local_ip.go
```diff
@@ -40,12 +40,24 @@ func getLocalIPv4ForDestination(dst string) string {
...
+}
+
 func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
 	switch family {
 	case k8snet.IPv4:
 		return getLocalIPv4ForDestination(dst)
 	case k8snet.IPv6:
-		// TODO_IPV6: add V6 healthcheck IP
+		result := GetLocalIPv6ForDestination(dst)
+		return result
```
skitt:
```suggestion
		return GetLocalIPv6ForDestination(dst)
```

to follow the same pattern as IPv4.

Or how about

```
	conn, err := net.Dial("udp" + family, "["+dst+"]:53")
	logger.FatalOnError(err, "Error getting local IPv" + family)

	defer conn.Close()

	return conn.LocalAddr().(*net.UDPAddr).IP.String()
}
```

skitt (changes requested):

pkg/endpoint/local_ip.go
```diff
@@ -40,12 +40,24 @@ func getLocalIPv4ForDestination(dst string) string {
 	return localAddr.IP.String()
 }
 
+func GetLocalIPv6ForDestination(dst string) string {
+	conn, err := net.Dial("udp6", "["+dst+"]:53")
```
skitt: This can’t work in tests run using GHAs: https://github.com/actions/runner-images/issues/668

yboaron: @skitt Thanks for the tip, 
I assume we can use another way to find the local IPv6 address, but I'm curious—can we test IPv6 Submariner in GHA environment? does IPv6 NATd and IPv6 IPsec traffic should work between clusters?

pkg/endpoint/public_ip.go
yboaron: oops

pkg/endpoint/local_ip.go
skitt: You can enumerate the local interface addresses using [`net.InterfaceAddrs()`](https://pkg.go.dev/net#InterfaceAddrs) but I’m not sure that’s usable here. I don’t know about IPv6 tests in GHAs, hopefully the limitation is external access only (but I have no idea).

yboaron: Yeah, seems that submariner-gateway pod can't access external network (to resolve public IPv6 address) 

`[90m2025-04-02T15:27:52.978Z[0m [31mFTL[0m ../submariner/main.go:149 [1mmain                 Error creating gateway instance[0m [36merror=[0m[31m[1m"error creating local endpoint object: could not determine public IPv6: Unable to resolve public IP by any of the resolver methods: [\nResolver[\"api:api64.ipify.org\"]: No IPv6 found in: \"52.234.42.222\", \nResolver[\"api:api6.ipify.org\"]: retrieving public IP from https://api6.ipify.org: Get \"https://api6.ipify.org\": dial tcp [2607:f2d8:1:3c::4]:443: connect: network is unreachable]"[0m[0m`

tpantelis: Perhaps we should treat it like air-gapped and set the public IP on the node via the annotation, at least for now.

pkg/endpoint/local_ip.go
```diff
@@ -29,27 +32,40 @@ var familyToDestIP = map[k8snet.IPFamily]string{
...
-	conn, err := net.Dial("udp4", dst+":53")
-	logger.FatalOnError(err, "Error getting local IP")
+func getLocalIPFromRoutes(family string) (string, error) {
+	familyInt := netlink.FAMILY_V4
+	if family == string(k8snet.IPv6) {
+		familyInt = netlink.FAMILY_V6
+	}
 
-	defer conn.Close()
+	routes, err := netlink.RouteList(nil, familyInt)
```
tpantelis: Use the `submariner-io/submariner/pkg/netlink` wrapper so this can be mocked in unit tests.

tpantelis (changes requested):

pkg/endpoint/local_ip.go
```diff
@@ -29,27 +33,42 @@ var familyToDestIP = map[k8snet.IPFamily]string{
 	k8snet.IPv6: "2001:4860:4860::8888",
 }
 
-func getLocalIPv4ForDestination(dst string) string {
-	conn, err := net.Dial("udp4", dst+":53")
-	logger.FatalOnError(err, "Error getting local IP")
+func getLocalIPFromRoutes(family string) (string, error) {
```
tpantelis: Does this work for IPv6? Do we still need this with https://github.com/submariner-io/submariner/pull/3373/commits/24f1b964576c1676179b95425b8b2aa79093bd44?

yboaron: If you meant to GetLocalIP function, yeah we still need getLocalIPFromRoutes fallback, GetLocalIP is called regardless to air-gap status from [1] .

> Does this work for IPv6? 

getLocalIPFromRoutes returns the IP address of the local interface used in the default V4/V6 route, I think it should work also for IPv6, also noticed that dual-stack gate is green , so GetLocalIP for v6 doesn't crash.

[1] 
https://github.com/submariner-io/submariner/blob/devel/pkg/endpoint/local_endpoint.go#L182

yboaron: Hmm, getLocalIPFromRoutes looks fine in e2e dual-stack but it fails in [ local endpoint unit test with dual-stack](https://github.com/submariner-io/submariner/blob/devel/pkg/endpoint/local_endpoint_test.go#L180) , I added getLocalIPFromRoutes  hook to workaround it.

tpantelis: `getLocalIPFromRoutes` uses the netlink API wrapper so, as long as the fake netlink implementation is hooked in and mocked, then it should work. But I can take a look at that in a follow-up. Let's focus on getting the production working and merging this.

pkg/endpoint/public_ip.go
```diff
@@ -68,19 +98,23 @@ func getPublicIP(family k8snet.IPFamily, submSpec *types.SubmarinerSpecification
...
 				config = submSpec.PublicIP
 			} else {
-				config = getPublicIPResolvers()
+				config = getPublicIPResolvers(family)
 			}
 		}
+		// temporary workaround since GHA restricts IPv6 connections to external networks.
+		if family == k8snet.IPv6 {
+			airGapped = true
+		}
```
tpantelis: Instead of forcing air gapped in code, can't we handle this in the GHA? I see there's an AIR_GAPPED defined in the shipyard Makefile.

yboaron: I added **air-gap** to (dual-stack, globalnet) and (dual-stack) gates ,  is that what you meant?
or there's another way to handle it in GHA ?

Also, I think we should update Shipyard to block IPv6 traffic to external net by adding iptables rule (currently it is only implemented for IPv4) , but we can handle it in a follow up PR

tpantelis (approved):

yboaron:
dual-stack air-gap gates fails , because IPv6 address isn't  annotated when Submariner is deployed in air-gap , the following shipyard https://github.com/submariner-io/shipyard/pull/1900 should address this issue

skitt (approved):

tpantelis:
https://github.com/submariner-io/submariner/pull/3373/commits/64b0c715742407198f58db1f8a4a92aafca9930e  bumped protobuf version so _natdiscovery.pb_ needs to be regenerated.

tpantelis:
It seems with air-gap enabled we also can't resolve external images:

`Back-off pulling image \"quay.io/submariner/nettest:devel\": ErrImagePull: failed to pull and unpack image \"quay.io/submariner/nettest:devel\": failed to resolve reference \"quay.io/submariner/nettest:devel\": failed to do request: Head \"[https://quay.io/v2/submariner/nettest/manifests/devel\](https://quay.io/v2/submariner/nettest/manifests/devel/)": dial tcp 184.73.240.191:443: connect: connection refused`

The dual-stack GHA are not required at this point so we can address this later.

tpantelis:
BTW, I see this in the gateway log: "_AIR_GAPPED_DEPLOYMENT is set to false_". **shipyard** needs to pass the air-gapped flag to `subctl join` (https://github.com/submariner-io/shipyard/pull/1908).

Also, the public and private IPs were that same:

```
privateIPs:
  - 172.18.0.7
  - fc00:1234:4444::7
  publicIPs:
  - 172.18.0.7
  - fc00:1234:4444::7
```

I assume this is b/c they both end up being obtained from the same source, ie shipyard determines the public-ip annotation from the node's addresses and `GetLocalIP` determines the private IP from via `net.Dial` or the netlink routes.

yboaron:
> I assume this is b/c they both end up being obtained from the same source, ie shipyard determines the public-ip annotation from the node's addresses and `GetLocalIP` determines the private IP from via `net.Dial` or the netlink routes.

Yep, we annotate node IP (equals private IP)  as public IP in air-gap deployment 

> It seems with air-gap enabled we also can't resolve external images:
> 
> `Back-off pulling image \"quay.io/submariner/nettest:devel\": ErrImagePull: failed to pull and unpack image \"quay.io/submariner/nettest:devel\": failed to resolve reference \"quay.io/submariner/nettest:devel\": failed to do request: Head \"[https://quay.io/v2/submariner/nettest/manifests/devel\](https://quay.io/v2/submariner/nettest/manifests/devel/)": dial tcp 184.73.240.191:443: connect: connection refused`
> 
> The dual-stack GHA are not required at this point so we can address this later.

Correct, external traffic is blocked [using iptables rules](https://github.com/submariner-io/shipyard/blob/3316105681a3194e2f768c0f22871b296eee2266/scripts/shared/lib/clusters_kind#L160) , 

You mentioned that dual-stack deployment succeeds although air-gap wasn't 
set in subctl join, maybe we can add new flag to shipyard that will be used only for annotating public IP. 
dual-stack gates can be deployed with this new flag , instead of air-gap , WDYT ?

tpantelis:
> You mentioned that dual-stack deployment succeeds although air-gap wasn't set in subctl join, maybe we can add new flag to shipyard that will be used only for annotating public IP. dual-stack gates can be deployed with this new flag , instead of air-gap , WDYT ?

I don't think we need a new flag, especially since it may be temporary. Perhaps what we should do is annotate the public IP if DUAL_STACK is enabled. Then the dual-stack gates wouldn't need to enable air-gapped.

EDIT: I did this as part of https://github.com/submariner-io/shipyard/pull/1909 as the dual-stack deployment job was failing.

PR #3381, opened 2025-04-03
Use k8snet.IPFamily in the netlink wrapper methods

tpantelis:
The underlying lib API takes an int for the family so change to `k8snet.IPFamily` in the wrapper for better typing. Using `k8snet.IPFamily` will also simplify implementing dual-stack support in the various components that use the netlink API.

See commits for details.

### DIFF
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -682,5 +682,5 @@ func (i *libreswan) waitForControlSocket() error {
 func (i *libreswan) Cleanup() error {
 	logger.Info("Uninstalling the libreswan cable driver")
 
-	return netlink.DeleteXfrmRules() //nolint:wrapcheck  // No need to wrap this error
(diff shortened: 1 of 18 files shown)


skitt (approved):

pkg/cable/libreswan/libreswan.go
```diff
@@ -682,5 +682,5 @@ func (i *libreswan) waitForControlSocket() error {
 func (i *libreswan) Cleanup() error {
 	logger.Info("Uninstalling the libreswan cable driver")
 
-	return netlink.DeleteXfrmRules() //nolint:wrapcheck  // No need to wrap this error
+	return netlink.DeleteXfrmRules(k8snet.IPFamilyUnknown) //nolint:wrapcheck  // No need to wrap this error
```
yboaron, line 685: Overall code looks good me , but It's a bit strange that FAMILY_UNKNOWN is translated to ALL

tpantelis: There's really no other choice. It can make sense depending on how you look at it.

yboaron (approved):

PR #3391, opened 2025-04-11
Remove IPv4-only restriction for Cluster/Endpoint CIDRs

tpantelis:
This enables processing of IPv6 Endpoints and reveals dual-stack E2E failures/errors that will be addressed in subsequent PRs.

### DIFF
--- a/pkg/endpoint/local_endpoint.go
+++ b/pkg/endpoint/local_endpoint.go
@@ -31,7 +31,6 @@ import (
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/util"
 	submv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
-	"github.com/submariner-io/submariner/pkg/cidr"
 	"github.com/submariner-io/submariner/pkg/cni"
 	"github.com/submariner-io/submariner/pkg/node"
 	"github.com/submariner-io/submariner/pkg/port"
@@ -155,9 +154,8 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification, 
 		localSubnets = submSpec.GlobalCidr
 		globalnetEnabled = true
 	} else {
-		// TODO_IPV6: set localSubnets to submSpec.ServiceCidr + submSpec.ClusterCidr
-		localSubnets = append(localSubnets, cidr.ExtractIPv4Subnets(submSpec.ServiceCidr)...)
-		localSubnets = append(localSubnets, cidr.ExtractIPv4Subnets(submSpec.ClusterCidr)...)
+		localSubnets = append(localSubnets, submSpec.ServiceCidr...)
+		localSubnets = append(localSubnets, submSpec.ClusterCidr...)
 	}
 
 	backendConfig, err := getBackendConfig(gwNode)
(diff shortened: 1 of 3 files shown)


pkg/endpoint/local_endpoint.go
```diff
@@ -155,9 +154,8 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
 		localSubnets = submSpec.GlobalCidr
 		globalnetEnabled = true
 	} else {
-		// TODO_IPV6: set localSubnets to submSpec.ServiceCidr + submSpec.ClusterCidr
-		localSubnets = append(localSubnets, cidr.ExtractIPv4Subnets(submSpec.ServiceCidr)...)
-		localSubnets = append(localSubnets, cidr.ExtractIPv4Subnets(submSpec.ClusterCidr)...)
+		localSubnets = append(localSubnets, submSpec.ServiceCidr...)
```
yboaron, line 157: I used this change when I tested locally  NATd PR, IPv6 tunnel was created but it was in error state , check [1]  (missing routes for healtcheck ? ) ,  do we want to set up IPv6 datapath in dual-stack deployments, although it isn't fully supported (natD, libreswan) ?

[1] 
`$ subctl show all  --kubeconfig output/kubeconfigs/kind-config-cluster2
Cluster "cluster2"
 ✓ Detecting broker(s)
 ✓ No brokers found

 ✓ Showing Connections
GATEWAY           CLUSTER    REMOTE IP           NAT   CABLE DRIVER   SUBNETS                                    STATUS      RTT avg.     
cluster1-worker   cluster1   172.18.0.4          no    libreswan      100.66.0.0/16, fc00:2001::/112, 10.130.0   connected   177.729µs    
cluster1-worker   cluster1   fc00:1234:4444::4   no    libreswan      100.66.0.0/16, fc00:2001::/112, 10.130.0   error       0s           
`

tpantelis: > do we want to set up IPv6 datapath in dual-stack deployments, although it isn't fully supported (natD, libreswan)?

It's in development so I don't think it matters now. It only affects the dual-stack E2E and I think it's useful b/c it lets us see what fails now. Sooner or later we're going to have to "turn on" IPv6 for testing and I think we might as well do it now rather than waiting for all the pieces to be in place.

tpantelis: So the IPv4 path is broken in the RA kubeproxy handler due to processing IPv6 subnets. We should avoid this while IPv6 is in development. I pushed https://github.com/submariner-io/submariner/pull/3391/commits/33b1432c3bf0c0510c87ce40bca5fd289aef4eae to only process IPv4 subnets for now.

BTW - I plan on working on IPv6 support in the kubeproxy handler next.

tpantelis: The GW failover E2E test fails with: 

"_Failed to await Gateway on "cluster1-worker" with status active and connections UP. Gateway "cluster1-worker" is active but cluster "cluster2" is not connected: Status: "error", Message: "Failed to successfully ping the remote endpoint IP \"fc00:1002:0:1::1\""_"

It expects to find a fully connected GW instance but the IPv6 connection failed. We could filter the GW connections by IP family, ie ignore the IPv6 connection. Or let it fail until the IPv6 connection  issue is addressed.

dfarrell07 (approved):

aswinsuryan (approved):

pkg/endpoint/local_endpoint.go
aswinsuryan: I will try this PR too with OVN, I am hitting the error in gateway pod due to missing service and podCIDR and this should solve it

tpantelis:
Included with https://github.com/submariner-io/submariner/pull/3403

PR #3394, opened 2025-04-14
Use slices functions instead of manual equivalents

skitt:
This uses slices.Delete instead of the equivalent append(); this doesn't seem like much but slices.Delete improves GC behaviour. slices.Contains replaces the equivalent loop.


pkg/netlink/fake/netlink.go
```diff
@@ -22,6 +22,7 @@ import (
 	"net"
 	"os"
 	"reflect"
+	slices0 "slices"
```
tpantelis:
```suggestion
	goslices "slices"
```

dfarrell07 (approved):

tpantelis (approved):

PR #3401, opened 2025-04-17
Generalize kubeproxy SyncHandler per IP family

tpantelis:
...so we can have separate SyncHandler instances for IPv4 and IPv6.

See commits for details.


yboaron (approved):

tpantelis:
Included with https://github.com/submariner-io/submariner/pull/3403.

PR #3403, opened 2025-04-21
Extend  kube-proxy IPv6 support

yboaron:
See commits for more details


pkg/netlink/netlink.go
```diff
@@ -92,6 +93,11 @@ var (
 		k8snet.IPv4: allZeroAddress,
 		k8snet.IPv6: allZeroAddressV6,
 	}
+
+	IPFamilyToNetlinkFamily = map[k8snet.IPFamily]int{
+		k8snet.IPv4: unix.AF_INET,
+		k8snet.IPv6: unix.AF_INET6,
+	}
```
tpantelis: There is an existing `ToNetlinkFamily` function. We should use either one or the other - the map here is simpler.

pkg/routeagent_driver/handlers/kubeproxy/kp_packetfilter.go
```diff
@@ -68,6 +69,13 @@ func NewSyncHandler(ipFamily k8snet.IPFamily, localClusterCidr, localServiceCidr
 	pFilter, err := packetfilter.New(ipFamily)
 	utilruntime.Must(err)
 
+	var vtepPrefix interface{}
```
tpantelis: Using the `any` alias for `interface{}` is more readable.

But ideally we should avoid untyped fields. Perhaps we can define the IPv4 version as a string as well:

`VxLANVTepNetworkPrefixIPv4 = "240.0.0.0/8"`

and define a map:

```
var vxLANVTepNetworkPrefix = map[k8snet.IPFamily]string{
    k8snet.IPv4: VxLANVTepNetworkPrefixIPv4,
    k8snet.IPv6: VxLANVTepNetworkPrefixIPv6,
}
```

to avoid the checks for `ipFamily == k8snet.IPv6`.

In `GetVtepIPAddressFrom`, for IPv4 case:

```
_, ulaNet, err := net.ParseCIDR(prefixStr)
...
ipSlice[0] = strconv.Itoa(int(ulaNet.IP.To4()[0]))
```

tpantelis (approved):

tpantelis:
The IPv6 dual-stack unit test case can now be enabled but with some changes - I submitted [Enable kubeproxy dual-stack IPv6 unit test case](https://github.com/submariner-io/submariner/pull/3403/commits/e44f2cf44c255219f25487050ae268188f25cbe2) to get it working.

pkg/vxlan/vxlan.go
```diff
@@ -134,15 +133,47 @@ func isVxlanConfigTheSame(newLink, currentLink netlink.Link) bool {
...
+			return nil, errors.Errorf("expected IPv4 addresses but got ip=%q and prefix=%q", ipAddr, prefixCIDR)
+		}
+
+		return net.IPv4(prefix[0], src[1], src[2], src[3]), nil
+	case k8snet.IPv6:
+		src := ip.To16()
+		prefix := prefixNet.IP.To16()
 
-	return net.ParseIP(strings.Join(ipSlice, ".")), nil
+		if src == nil || prefix == nil || ip.To4() != nil || prefixNet.IP.To4() != nil {
```
tpantelis: Minor thing but the `To4` checks here really aren't necessary since they can't be reached unless `ip` and `prefixNet.IP` are both known to be v6.

```suggestion
		if src == nil || prefix == nil {
```

aswinsuryan (approved):

PR #3404, opened 2025-04-21
Add IsLinkNotFoundError function to the netlink API

tpantelis:
The `netlink` lib defines `LinkNotFoundError` but it does not implement method `Is(error) bool` so it can't be used with
`errors.Is(err, netlink.LinkNotFoundError{})`. So add a helper function to determine if an err indicates link not found.

### DIFF
--- a/pkg/netlink/netlink.go
+++ b/pkg/netlink/netlink.go
@@ -307,9 +307,8 @@ func DeleteIfaceAndAssociatedRoutes(iface string, tableID int) error {
 
 	link, err := n.LinkByName(iface)
 	if err != nil {
-		//nolint:errorlint // netlink.LinkNotFoundError does not implement method Is(error) bool
-		if _, ok := err.(netlink.LinkNotFoundError); !ok {
-			logger.Warningf("Failed to retrieve the vxlan-tunnel interface: %v", err)
+		if !IsLinkNotFoundError(err) {
+			logger.Warningf("Failed to retrieve link %q: %v", iface, err)
 		}
 
 		return nil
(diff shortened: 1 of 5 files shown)


pkg/routeagent_driver/handlers/ovn/host_networking.go
```diff
@@ -115,8 +115,7 @@ func (ovn *Handler) programRulesForRemoteSubnets(subnets []string, ruleFunc func
 
 func (ovn *Handler) getNextHopOnK8sMgmtIntf() (*net.IP, error) {
 	link, err := ovn.netLink.LinkByName(OVNK8sMgmntIntfName)
-
-	if err != nil && !errors.Is(err, netlink.LinkNotFoundError{}) {
```
yboaron, line 120: Don't we want to check if err is not found here ?

tpantelis: Just seeing your question... the `errors.Is(err, netlink.LinkNotFoundError{})` check was actually incorrect and always returned false. But that was actually a good thing b/c, if it wasn't found, we would try to use a nil `link` instance. So I don't think this needs to check for `LinkNotFound`.

pkg/netlink/netlink.go
```diff
@@ -372,3 +371,9 @@ func ToNetlinkFamily(family k8snet.IPFamily) int {
 
 	return netlink.FAMILY_ALL
 }
+
+func IsLinkNotFoundError(err error) bool {
+	//nolint:errorlint // netlink.LinkNotFoundError does not implement method Is(error) bool
+	_, ok := err.(netlink.LinkNotFoundError)
+	return ok || (err != nil && err.Error() == "Link not found")
```
skitt: This doesn’t matter in current code, but to avoid surprises in future, using `errors.As` would provide support for errors wrapping `LinkNotFoundError` too:

```suggestion
	return errors.As(err, &netlink.LinkNotFoundError{}) || (err != nil && err.Error() == "Link not found")
```

skitt (changes requested):

skitt (approved):

dfarrell07 (approved):

PR #3422, opened 2025-05-06
Fix ip6tables binary provisioning in containers

skitt:
The iptables wrapper installer used an alternatives setup which no longer works correctly on Fedora 41, and ended up skipping ip6tables (and even removing it, which is why the earlier link is removed). This updates the alternatives setup to use different names for the ip6tables alternatives, restoring ip6tables features before and after the wrapper is called.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -55,6 +55,7 @@ jobs:
         include:
           - extra-toggles: dual-stack
           - extra-toggles: dual-stack, globalnet
+          - extra-toggles: ipv6-stack
           - extra-toggles: nftables
           - extra-toggles: nftables, ovn
           - extra-toggles: nftables, globalnet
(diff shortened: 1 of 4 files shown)


skitt:
FYI @yboaron

tpantelis (approved):

tpantelis:
IPv6 deployed successfully and the connection between the clusters was established but something else failed. Anyway, this  PR did its job.

yboaron (approved):

yboaron:
> FYI @yboaron

Thanks!

yboaron:
> IPv6 deployed successfully and the connection between the clusters was established but something else failed. Anyway, this PR did its job.

I think shipyard  [connectivity test ](https://github.com/submariner-io/shipyard/blob/ec48fc231c0eef9a12e68e9f23c504765b7f8d00/scripts/shared/lib/deploy_funcs#L80) should be updated to support also IPv6

PR #3428, opened 2025-05-09
Set IPv6 health check IP if globalnet is enabled in GetLocalSpec

tpantelis:
We want to only skip setting it for IPv4.

### DIFF
--- a/pkg/endpoint/local_endpoint.go
+++ b/pkg/endpoint/local_endpoint.go
@@ -193,12 +193,16 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification, 
 		endpointSpec.SetPublicIP(publicIP)
 	}
 
-	if submSpec.HealthCheckEnabled && !globalnetEnabled {
+	if submSpec.HealthCheckEnabled {
 		// When globalnet is enabled, HealthCheckIP will be the globalIP assigned to the Active GatewayNode.
 		// In a fresh deployment, globalIP annotation for the node might take few seconds. So we listen on NodeEvents
 		// and update the endpoint HealthCheckIP (to globalIP) in datastoreSyncer at a later stage. This will trigger
 		// the HealthCheck between the clusters.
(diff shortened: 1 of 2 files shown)


dfarrell07 (approved):

pkg/endpoint/local_endpoint.go
```diff
@@ -193,12 +193,16 @@ func GetLocalSpec(ctx context.Context, submSpec *types.SubmarinerSpecification,
...
 	}
 
-	if submSpec.HealthCheckEnabled && !globalnetEnabled {
+	if submSpec.HealthCheckEnabled {
 		// When globalnet is enabled, HealthCheckIP will be the globalIP assigned to the Active GatewayNode.
 		// In a fresh deployment, globalIP annotation for the node might take few seconds. So we listen on NodeEvents
 		// and update the endpoint HealthCheckIP (to globalIP) in datastoreSyncer at a later stage. This will trigger
 		// the HealthCheck between the clusters.
 		for _, family := range submSpec.GetIPFamilies() {
+			if globalnetEnabled && family == k8snet.IPv4 {
```
yboaron, line 202: Hmm, do we want to set healthcheckIP to be CNI IF IPv6 address for Globalnet,IPv6 ?
it should be some IP from global IPv6 range (when Gloabalnet supports IPv6) , right ?

tpantelis: I don’t know.  We just need to set it to something.  I’m on PTO this week so please submit another change if you don’t think this is sufficient.

vthapar (approved):

pkg/endpoint/local_endpoint.go
tpantelis: @yboaron Are we OK here?

yboaron: > @yboaron Are we OK here?

Let's merge it , we'll address GN IPv6 healthcheck address when we implement GN IPv6

PR #3431, opened 2025-05-13
Dynamically select IP families for e2e dataplane and redundancy tests

yboaron:
Adjust e2e dataplane and redundancy tests to dynamically select which IP family tests to run based on clusters IP families.
This ensures dataplane connectivity tests match the actual supported IP families between clusters.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/prometheus-community/pro-bing v0.7.0
 	github.com/prometheus/client_golang v1.22.0
 	github.com/submariner-io/admiral v0.21.0-m2
-	github.com/submariner-io/shipyard v0.21.0-m2
 	github.com/vishvananda/netlink v1.3.0
 	golang.org/x/net v0.39.0
 	golang.org/x/sys v0.32.0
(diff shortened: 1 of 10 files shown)


aswinsuryan (approved):

yboaron:
Functionality perspective, both IPv4 and IPv6 datapath and redundancy tests appear to run as expected based on the environment. I did notice two test failures (see details below), and we can file issues to track those.

**IPv6 deployment:**
==================

Gateway fail-over tests when running failover tests for IPv6 runs the gateway failover scenario [redundancy]
 STEP: Waiting for the previous submariner endpoint "submariner-gateway-fj6q5" to be removed on "cluster1"
 
 **Root cause :** 
we named the endpoint with IPv4 private IP (which is empty for ipv6 deployment) -  https://github.com/submariner-io/submariner/blob/devel/pkg/endpoint/local_endpoint.go#L184
  
  

**dual-stack deployment:**
=======================

Gateway fail-over tests when running failover tests for IPv6 runs the gateway failover scenario
Failed on 
STEP: Verifying TCP connectivity from non-gateway node on "cluster1" to non-gateway node on "cluster2" @ 05/14/25 14:02:01.997

**Root cause:** 
Addressed by https://github.com/submariner-io/submariner/pull/3435

tpantelis (approved):

PR #3457, opened 2025-06-10
Fix issues in Ipv6 OVN  Kubernetes dual stack

aswinsuryan:
Some of the rules programmed by Ipv4 was getting overwritten by the ipv6 handler ( or vice versa) fixed the issues.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
@@ -36,10 +36,11 @@ import (
 	"github.com/vishvananda/netlink"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8snet "k8s.io/utils/net"
 )
 
 func (ovn *Handler) cleanupGatewayDataplane() error {
-	currentRemoteSubnets, err := ovn.getExistingIPv4RuleSubnets()
+	currentRemoteSubnets, err := ovn.getExistingRuleSubnets()
 	if err != nil {
 		return errors.Wrapf(err, "error reading ip rule list for IPv4")
 	}
@@ -66,7 +67,7 @@ func (ovn *Handler) updateGatewayDataplane() error {
 	ovn.mutex.Lock()
 	defer ovn.mutex.Unlock()
 
-	currentRuleRemotes, err := ovn.getExistingIPv4RuleSubnets()
+	currentRuleRemotes, err := ovn.getExistingRuleSubnets()
 	if err != nil {
 		return errors.Wrapf(err, "error reading ip rule list for IPv4")
 	}
@@ -118,6 +119,7 @@ func (ovn *Handler) getForwardingRuleSpecs() ([]*packetfilter.Rule, error) {
 	// hits ovn-k8s-mp0, firewall rules would be processed. Therefore, we include these firewall rules in the FORWARDing
 	// chain to allow such traffic. Similar thing happens for outbound traffic as well, and we use routes in table 150.
 	rules := []*packetfilter.Rule{}
+
 	for _, remoteCIDR := range ovn.getRemoteSubnets().UnsortedList() {
 		rules = append(rules, &packetfilter.Rule{
 			DestCIDR:     remoteCIDR,
@@ -151,8 +153,7 @@ func (ovn *Handler) getMSSClampingRuleSpecs() ([]*packetfilter.Rule, error) {
 			Action:    packetfilter.RuleActionMss,
 			ClampType: packetfilter.ToValue,
 			MssValue:  strconv.Itoa(MSSFor1500MTU),
-		},
-		)
+		})
 	}
 
 	return rules, nil
@@ -281,6 +282,10 @@ func (ovn *Handler) processEndpointSubnets(add bool, endpoints ...submarinerv1.E
 
 	for i := range endpoints {
 		for _, subnet := range endpoints[i].Spec.Subnets {
+			if k8snet.IPFamilyOfCIDRString(subnet) != ovn.ipFamily {
+				continue
+			}
+
 			if err := ovn.updateNoMasqueradeRules(subnet, add); err != nil {
 				return err
 			}
--- a/pkg/routeagent_driver/handlers/ovn/gateway_route_controller.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_route_controller.go
@@ -33,6 +33,7 @@ type GatewayRouteController struct {
 	remoteSubnets       sets.Set[string]
 	stopCh              chan struct{}
 	mgmtIP              string
+	ipFamily            k8snet.IPFamily
 }
 
 //nolint:gocritic // Ignore hugeParam
@@ -45,6 +46,7 @@ func NewGatewayRouteController(ipFamily k8snet.IPFamily, config watcher.Config, 
 		connectionHandler: connectionHandler,
 		remoteSubnets:     sets.New[string](),
 		stopCh:            make(chan struct{}),
+		ipFamily:          ipFamily,
 	}
 
 	config.ResourceConfigs = []watcher.ResourceConfig{
@@ -120,10 +122,12 @@ func (g *GatewayRouteController) reconcileRemoteSubnets(subMGWRoute *submarinerv
 	}
 
 	for _, subnet := range subMGWRoute.RoutePolicySpec.RemoteCIDRs {
-		if addSubnet {
-			g.remoteSubnets.Insert(subnet)
-		} else {
-			g.remoteSubnets.Delete(subnet)
+		if k8snet.IPFamilyOfCIDRString(subnet) == g.ipFamily {
+			if addSubnet {
+				g.remoteSubnets.Insert(subnet)
+			} else {
+				g.remoteSubnets.Delete(subnet)
+			}
 		}
 	}
 
--- a/pkg/routeagent_driver/handlers/ovn/gateway_route_handler.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_route_handler.go
@@ -25,6 +25,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/util"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cidr"
 	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	"github.com/submariner-io/submariner/pkg/cni"
 	"github.com/submariner-io/submariner/pkg/event"
@@ -71,6 +72,9 @@ func (h *GatewayRouteHandler) GetNetworkPlugins() []string {
 func (h *GatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpoint) error {
 	if h.State().IsOnGateway() {
 		gwr := h.newGatewayRoute(endpoint)
+		if gwr == nil {
+			return nil
+		}
 
 		result, err := util.CreateOrUpdate(context.TODO(), GatewayResourceInterface(h.smClient, endpoint.Namespace),
 			gwr, util.Replace(gwr))
@@ -86,12 +90,14 @@ func (h *GatewayRouteHandler) RemoteEndpointCreated(endpoint *submarinerv1.Endpo
 
 func (h *GatewayRouteHandler) RemoteEndpointRemoved(endpoint *submarinerv1.Endpoint) error {
 	if h.State().IsOnGateway() {
+		routeName := endpoint.Spec.ClusterID + familySuffix(h.ipFamily)
+
 		if err := h.smClient.SubmarinerV1().GatewayRoutes(endpoint.Namespace).Delete(context.TODO(),
-			endpoint.Spec.ClusterID, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
-			return errors.Wrapf(err, "error deleting gatewayRoute %q", endpoint.Name)
+			routeName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+			return errors.Wrapf(err, "error deleting gatewayRoute %q", routeName)
 		}
 
-		logger.Infof("GatewayRoute %s deleted for remote endpoint %s", endpoint.Spec.ClusterID, endpoint.Name)
+		logger.Infof("GatewayRoute %s deleted for remote endpoint %s", routeName, endpoint.Name)
 	}
 
 	return nil
@@ -109,9 +115,12 @@ func (h *GatewayRouteHandler) TransitionToGateway() error {
 		}
 
 		gwr := h.newGatewayRoute(&endpoints[i])
+		if gwr == nil {
+			continue
+		}
 
-		result, err := util.CreateOrUpdate(context.TODO(), GatewayResourceInterface(h.smClient, endpoints[i].Namespace),
-			gwr, util.Replace(gwr))
+		result, err := util.CreateOrUpdate(context.TODO(),
+			GatewayResourceInterface(h.smClient, endpoints[i].Namespace), gwr, util.Replace(gwr))
 		if err != nil {
 			return errors.Wrapf(err, "error creating/updating GatewayRoute")
 		}
@@ -123,12 +132,18 @@ func (h *GatewayRouteHandler) TransitionToGateway() error {
 }
 
 func (h *GatewayRouteHandler) newGatewayRoute(endpoint *submarinerv1.Endpoint) *submarinerv1.GatewayRoute {
+	famCIDRs := cidr.ExtractSubnets(h.ipFamily, endpoint.Spec.Subnets)
+
+	if len(famCIDRs) == 0 {
+		return nil
+	}
+
 	return &submarinerv1.GatewayRoute{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: endpoint.Spec.ClusterID,
+			Name: endpoint.Spec.ClusterID + familySuffix(h.ipFamily),
 		},
 		RoutePolicySpec: submarinerv1.RoutePolicySpec{
-			RemoteCIDRs: endpoint.Spec.Subnets,
+			RemoteCIDRs: famCIDRs,
 			NextHops:    []string{h.nextHopIP},
 		},
 	}
--- a/pkg/routeagent_driver/handlers/ovn/non_gateway_route_controller.go
+++ b/pkg/routeagent_driver/handlers/ovn/non_gateway_route_controller.go
@@ -24,6 +24,7 @@ import (
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	k8snet "k8s.io/utils/net"
 )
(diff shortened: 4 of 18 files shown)


pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -281,6 +290,10 @@ func (ovn *Handler) processEndpointSubnets(add bool, endpoints ...submarinerv1.E
 
 	for i := range endpoints {
 		for _, subnet := range endpoints[i].Spec.Subnets {
+			if k8snet.IsIPv6CIDRString(subnet) != (ovn.ipFamily == k8snet.IPv6) {
```
tpantelis:
```suggestion
			if k8snet.IPFamilyOfCIDRString(subnet) != ovn.ipFamily {
```

pkg/routeagent_driver/handlers/ovn/gateway_route_controller.go
```diff
@@ -120,10 +122,12 @@ func (g *GatewayRouteController) reconcileRemoteSubnets(subMGWRoute *submarinerv
 	}
 
 	for _, subnet := range subMGWRoute.RoutePolicySpec.RemoteCIDRs {
-		if addSubnet {
-			g.remoteSubnets.Insert(subnet)
-		} else {
-			g.remoteSubnets.Delete(subnet)
+		if k8snet.IsIPv6CIDRString(subnet) == (g.ipFamily == k8snet.IPv6) {
```
tpantelis:
```suggestion
		if k8snet. IPFamilyOfCIDRString(subnet) == g.ipFamily {
```

pkg/routeagent_driver/handlers/ovn/gateway_route_handler.go
```diff
@@ -123,12 +131,29 @@ func (h *GatewayRouteHandler) TransitionToGateway() error {
...
+
+	for _, cidr := range endpoint.Spec.Subnets {
+		isV6 := k8snet.IsIPv6CIDRString(cidr)
+		if isV6 && h.ipFamily == k8snet.IPv6 {
+			famCIDRs = append(famCIDRs, cidr)
+		}
+
+		if !isV6 && h.ipFamily == k8snet.IPv4 {
+			famCIDRs = append(famCIDRs, cidr)
+		}
```
tpantelis:
```suggestion
		if k8snet.IPFamilyOfCIDRString(cidr) == h.ipFamily {
			famCIDRs = append(famCIDRs, cidr)
		}
```

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -141,6 +147,10 @@ func (ovn *Handler) getMSSClampingRuleSpecs() ([]*packetfilter.Rule, error) {
 	rules := []*packetfilter.Rule{}
 
 	for _, remoteCIDR := range ovn.getRemoteSubnets().UnsortedList() {
+		if k8snet.IsIPv6CIDRString(remoteCIDR) != (ovn.ipFamily == k8snet.IPv6) {
+			continue
+		}
```
tpantelis: Handled in `getRemoteSubnets`.

```suggestion
```

pkg/routeagent_driver/handlers/ovn/non_gateway_route_controller.go
```diff
@@ -102,26 +105,33 @@ func (g *NonGatewayRouteController) nonGatewayRouteDeleted(obj runtime.Object, _
...
-				g.remoteSubnets.Delete(subnet)
-			}
-		}
+	nextHop := submNonGWRoute.RoutePolicySpec.NextHops[0]
 
-		return g.connectionHandler.reconcileSubOvnLogicalRouterPolicies(g.remoteSubnets, submNonGWRoute.RoutePolicySpec.NextHops[0])
+	isNextHopV6 := k8snet.IsIPv6String(nextHop)
+	if isNextHopV6 != (g.ipFamily == k8snet.IPv6) {
+		return nil
+	}
```
tpantelis:
```suggestion
	if k8snet.IPFamilyOfCIDRString(nextHop) != g.ipFamily {
		return nil
	}
```

pkg/routeagent_driver/handlers/ovn/non_gateway_route_controller.go
```diff
@@ -102,26 +105,33 @@ func (g *NonGatewayRouteController) nonGatewayRouteDeleted(obj runtime.Object, _
...
-	return nil
+	for _, subnet := range submNonGWRoute.RoutePolicySpec.RemoteCIDRs {
+		if addSubnet {
+			g.remoteSubnets.Insert(subnet)
+		} else {
+			g.remoteSubnets.Delete(subnet)
+		}
+	}
+
+	logger.Warningf("The NonGatewayRoute is and subnet is  %v, %v", submNonGWRoute, g.remoteSubnets)
```
tpantelis: This message isn't formatted correctly but why issue a warning here, especially when it didn't before?

pkg/routeagent_driver/handlers/ovn/non_gateway_route_handler.go
```diff
@@ -142,27 +150,49 @@ func (h *NonGatewayRouteHandler) NodeUpdated(node *corev1.Node) error {
...
+
+	for _, cidr := range endpoint.Spec.Subnets {
+		isV6 := k8snet.IsIPv6CIDRString(cidr)
+		if isV6 && h.ipFamily == k8snet.IPv6 {
+			famCIDRs = append(famCIDRs, cidr)
+		}
+
+		if !isV6 && h.ipFamily == k8snet.IPv4 {
+			famCIDRs = append(famCIDRs, cidr)
+		}
```
tpantelis:
```suggestion
		if k8snet.IPFamilyOfCIDRString(cidr) == h.ipFamily {
			famCIDRs = append(famCIDRs, cidr)
		}
```

pkg/routeagent_driver/handlers/ovn/south_rules.go
```diff
@@ -40,6 +40,11 @@ func (ovn *Handler) handleSubnets(remoteSubnets []string, ruleFunc func(rule *ne
 
 	for _, subnetToHandle := range remoteSubnets {
 		for _, localSubnet := range localCIDRs.UnsortedList() {
+			if k8snet.IsIPv6CIDRString(localSubnet) != (ovn.ipFamily == k8snet.IPv6) ||
+				k8snet.IsIPv6CIDRString(subnetToHandle) != (ovn.ipFamily == k8snet.IPv6) {
+				continue
+			}
+
```
tpantelis: You can use `cidr.ExtractSubnets` to build `localCIDRs`  above, in which case this isn't necessary. Also it's not really necessary to check `subnetToHandle` either since all callers have already filtered `remoteSubnets` but, to be safe, you can also use `cidr.ExtractSubnets` on `remoteSubnets`.

```suggestion
```

pkg/routeagent_driver/handlers/ovn/utils.go
```diff
@@ -74,3 +74,12 @@ func jsonToIP(jsonData string) (string, error) {
...
 	return ip.String(), nil
 }
+
+// familySuffix returns "-v4" or "-v6" for CR name uniqueness.
+func familySuffix(f k8snet.IPFamily) string {
+	if f == k8snet.IPv6 {
+		return "-v6"
+	}
+
+	return "-v4"
```
tpantelis: For backwards compatibility, only suffix for IPv6.

```suggestion
	return ""
```

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -118,7 +119,12 @@ func (ovn *Handler) getForwardingRuleSpecs() ([]*packetfilter.Rule, error) {
 	// hits ovn-k8s-mp0, firewall rules would be processed. Therefore, we include these firewall rules in the FORWARDing
 	// chain to allow such traffic. Similar thing happens for outbound traffic as well, and we use routes in table 150.
 	rules := []*packetfilter.Rule{}
+
 	for _, remoteCIDR := range ovn.getRemoteSubnets().UnsortedList() {
+		if k8snet.IsIPv6CIDRString(remoteCIDR) != (ovn.ipFamily == k8snet.IPv6) {
+			continue
+		}
```
tpantelis: Since `getRemoteSubnets` already filters the IP family this check isn't necessary.
```suggestion
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/subnets.go
```diff
@@ -18,14 +18,21 @@ limitations under the License.
...
 func (ovn *Handler) getRemoteSubnets() set.Set[string] {
 	endpointSubnets := set.New[string]()
 
 	endpoints := ovn.State().GetRemoteEndpoints()
 	for i := range endpoints {
 		for _, subnet := range endpoints[i].Spec.Subnets {
+			if k8snet.IsIPv6CIDRString(subnet) != (ovn.ipFamily == k8snet.IPv6) {
+				continue
+			}
+
```
tpantelis: Also delete L36.

```suggestion
			if k8snet.IsIPv6CIDRString(subnet) == ovn.ipFamily {
				endpointSubnets.Insert(subnet)
			}
```

pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
aswinsuryan: Done.

pkg/routeagent_driver/handlers/ovn/gateway_route_handler.go
```diff
@@ -123,12 +131,24 @@ func (h *GatewayRouteHandler) TransitionToGateway() error {
 }
 
 func (h *GatewayRouteHandler) newGatewayRoute(endpoint *submarinerv1.Endpoint) *submarinerv1.GatewayRoute {
+	var famCIDRs []string
+
+	for _, cidr := range endpoint.Spec.Subnets {
+		if k8snet.IPFamilyOfCIDRString(cidr) == h.ipFamily {
+			famCIDRs = append(famCIDRs, cidr)
+		}
+	}
```
tpantelis: We can actually use `cidr.ExtractSubnets` here.

```suggestion
	famCIDRs := cidr.ExtractSubnets(h.ipFamily, endpoint.Spec.Subnets)
```

pkg/routeagent_driver/handlers/ovn/non_gateway_route_handler.go
```diff
@@ -142,27 +150,44 @@ func (h *NonGatewayRouteHandler) NodeUpdated(node *corev1.Node) error {
...
 }
 
 func (h *NonGatewayRouteHandler) newNonGatewayRoute(endpoint *submarinerv1.Endpoint) *submarinerv1.NonGatewayRoute {
+	var famCIDRs []string
+
+	for _, cidr := range endpoint.Spec.Subnets {
+		if k8snet.IPFamilyOfCIDRString(cidr) == h.ipFamily {
+			famCIDRs = append(famCIDRs, cidr)
+		}
+	}
```
tpantelis:
```suggestion
	famCIDRs := cidr.ExtractSubnets(h.ipFamily, endpoint.Spec.Subnets)
```

pkg/routeagent_driver/handlers/ovn/subnets.go
```diff
@@ -18,15 +18,20 @@ limitations under the License.
...
 func (ovn *Handler) getRemoteSubnets() set.Set[string] {
 	endpointSubnets := set.New[string]()
 
 	endpoints := ovn.State().GetRemoteEndpoints()
 	for i := range endpoints {
 		for _, subnet := range endpoints[i].Spec.Subnets {
-			endpointSubnets.Insert(subnet)
+			if k8snet.IPFamilyOfCIDRString(subnet) == ovn.ipFamily {
+				endpointSubnets.Insert(subnet)
+			}
```
tpantelis: We could actually replace L31 - L35 with:

`endpointSubnets.Insert(cidr.ExtractSubnets(ovn.ipFamily, endpoints[i].Spec.Subnets)...)`

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/transit_switch_ip.go
```diff
@@ -77,10 +82,47 @@ func (t *transitSwitchIPImpl) UpdateFrom(node *corev1.Node) (bool, error) {
...
+	}
+
+	// Fallback to legacy plain IP string
+	rawIP, err := jsonToIP(value)
 	if err != nil {
 		return false, errors.Wrapf(err, "error parsing the transit switch IP")
 	}
 
-	return transitSwitchIP != t.value.Swap(transitSwitchIP), nil
+	ip, err := getValidIP(rawIP)
```
tpantelis: `jsonToIP` returns an error if the parsed data isn't a CIDR so calling `getValidIP` doesn't really add any value.

pkg/routeagent_driver/handlers/ovn/transit_switch_ip.go
```diff
@@ -77,10 +82,47 @@ func (t *transitSwitchIPImpl) UpdateFrom(node *corev1.Node) (bool, error) {
...
+	}
+
+	var fam map[string]string
+	if err := json.Unmarshal([]byte(value), &fam); err == nil {
+		var raw string
+		if t.ipFamily == k8snet.IPv6 {
+			raw = fam["ipv6"]
+		} else {
+			raw = fam["ipv4"]
+		}
```
tpantelis: The expected format here here looks the same as in `jsonToIP`, ie a map containing key "ipv4" or "ipv6". So why is `jsonToIP` considered legacy? It seems the only real difference is that the map data may contain an IP or a CIDR. If so then we could just change `jsonToIP` to handle that. Or am I just missing something obvious here....

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/transit_switch_ip.go
aswinsuryan: The intention was to support both of these formats.

k8s.ovn.org/node-transit-switch-port-ifaddr: '{"ipv4":"100.88.0.5/16","ipv6":"fd97::5/64"}'
k8s.ovn.org/node-transit-switch-port-ifaddr: "100.88.0.5"

pkg/routeagent_driver/handlers/ovn/transit_switch_ip.go
aswinsuryan: it is to accept both CIDR and plane IP

pkg/routeagent_driver/handlers/ovn/transit_switch_ip.go
tpantelis: OK but `jsonToIP` also parses it as a map. Your new code is essentially doing the same thing as `jsonToIP`. So neither would handle the format `k8s.ovn.org/node-transit-switch-port-ifaddr: "100.88.0.5"`. Why are we trying to support this format now where we didn't before? Is it for a newer version?

Also if we want to support a different format we should add unit tests for it.

pkg/routeagent_driver/handlers/ovn/transit_switch_ip.go
tpantelis: yes that's what `getValidIP` does but `jsonToIP` only returns a CIDR.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/transit_switch_ip.go
aswinsuryan: ok , I got it wrong these changes were done for two issues (verified with a setup) , 

1) IPV4 next hop was returned for ipv6 
2) tranist switch ip with CIDR was not mathincing in non-gw contorller . 

So  the json format is same for dual stack and single. We can move it to jsonToIP

pkg/routeagent_driver/handlers/ovn/utils.go
```diff
@@ -50,27 +50,41 @@ func getNextHopOnK8sMgmtIntf(family k8snet.IPFamily) (string, error) {
...
+
+		if raw != "" {
+			return getValidIP(raw)
+		}
 	}
 
-	ip, _, err := net.ParseCIDR(ipStr)
-	if err != nil {
-		return "", fmt.Errorf("invalid IP CIDR address: %s", ipStr)
+	return getValidIP(jsonData)
```
tpantelis: So this is handling non-jSON format, ie `"annotation": "1.2.3.4"`. Going back to a prior question, why are we handling this format now - is this a new supported format? Does this mean it's single-stack and guaranteed to be the desired IP version? Or do we need to verify the IP family matches? Can you add a comment describing the expected behavior?

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/utils.go
aswinsuryan: No that is not needed , it was my oversight, I thought a failure was due to incoorect format. Reoved the backup logic.

pkg/routeagent_driver/handlers/ovn/utils.go
```diff
@@ -50,27 +50,43 @@ func getNextHopOnK8sMgmtIntf(family k8snet.IPFamily) (string, error) {
...
+	} else if family == k8snet.IPv4 {
+		raw = fam["ipv4"]
 	} else {
 		return "", errors.New("JSON data does not contain 'ipv4' or 'ipv6' field")
 	}
 
-	ip, _, err := net.ParseCIDR(ipStr)
-	if err != nil {
-		return "", fmt.Errorf("invalid IP CIDR address: %s", ipStr)
+	return getValidIP(raw)
```
tpantelis: L78 isn't accurate now, ie it now reflects the passed-in `family` being invalid which isn't possible. Also I assume the upstream JSON format hasn't changed at all, in which case, `raw` will always be a CIDR so `getValidIP` isn't needed. The only code that really needs to change is between L61-68:

```
ipStr := data["ipv" + family]
if ipStr == "" {
    return "", errors.Errorf("JSON data does not contain 'ipv%s' field", family)
}
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/utils.go
aswinsuryan: refactored , the intention of isValidIP is to remove the mask from the string and make it IP,  like 10.0.0.1/24 to 10.0.0.1. This is for the nonGatewayRouteController check to be accurate when it compares the transit switch IP in the CR to transit switch  IP of the node. The check had issues with dual stack.

tpantelis: But the prior code already did that: https://github.com/submariner-io/submariner/blob/eaa5600b1fc790ab6abfa608542bcb5698e69763/pkg/routeagent_driver/handlers/ovn/utils.go#L70:


`getValidIP` does that and also does `net.ParseIP` seemingly to handle an IP or CIDR which doesn't seem necessary unless the upstream JSON format now allows it, ie "_ipv4:10.0.0.1_" OR "_ipv4:10.0.0.1/24_".

aswinsuryan: https://github.com/submariner-io/submariner/pull/3457/files#diff-dd592014434c54677c8a085119da7d745062cb1ec8b9cc0567752e668808b007R118 this check is failing when we compare with the mask , but working fine if we remove it.

tpantelis (approved):

yboaron (approved):

PR #3470, opened 2025-06-18
Fix incorrect IP returned for private ip for OVN IPV6

aswinsuryan:
With OVN-Kubenetenes CNI for IPV6 private ip instead of global ip of the interface a wrong ip is returned for an. This change checks if the ip is a valid global ip if not tries to find one

### DIFF
--- a/pkg/endpoint/local_endpoint_test.go
+++ b/pkg/endpoint/local_endpoint_test.go
@@ -116,6 +116,12 @@ var _ = Describe("GetLocalSpec", func() {
 
 		Expect(netLink.RouteAdd(&netlink.Route{
 			LinkIndex: 2,
+			Src:       net.ParseIP("fd69::2"),
+			Gw:        net.ParseIP("2001:0:0:0::"),
+		})).To(Succeed())
+
+		Expect(netLink.RouteAdd(&netlink.Route{
+			LinkIndex: 2,
 			Src:       net.ParseIP(ipv6LocalIP),
 			Gw:        net.ParseIP("2001:0:0:0::"),
 		})).To(Succeed())
--- a/pkg/endpoint/local_ip.go
+++ b/pkg/endpoint/local_ip.go
@@ -44,8 +44,16 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
 	}
 
 	for i := range routes {
-		if routes[i].Gw != nil {
-			return routes[i].Src.String(), nil
+		if routes[i].Gw != nil && routes[i].Src != nil {
+			// OVNK configures special internal masquerade IPs,
+			// these IPs are used internally by OVNK infra dataplane, and they should
+			// not be selected as private IPs.
+			ip := routes[i].Src
+			if family == k8snet.IPv6 && !isValidGlobalIPv6(ip) {
+				continue
+			}
+
+			return ip.String(), nil
 		}
 	}
 
@@ -53,12 +61,24 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
 }
 
 func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
-	conn, err := Dial("udp"+string(family), "["+dst+"]:53")
+	conn, err := Dial("udp"+string(family), net.JoinHostPort(dst, "53"))
 	if err == nil {
 		defer conn.Close()
 		localAddr := conn.LocalAddr().(*net.UDPAddr)
 
-		return localAddr.IP.String()
+		if family == k8snet.IPv4 || isValidGlobalIPv6(localAddr.IP) {
+			return localAddr.IP.String()
+		}
+
+		logger.Warningf("IP %q returned from Dial isn't usable - trying to find another IP from the same interface", localAddr.IP)
+
+		// Try to find a valid global-scope IP from the same interface
+		ifaceIP, err := getValidGlobalIPv6FromSameInterface(localAddr.IP)
+		if err == nil {
+			return ifaceIP
+		} else {
+			logger.Warningf("Failed to retrieve valid IPv6 address for %q: %v", localAddr.IP, err)
+		}
 	}
 
 	// connection failed try fallback method
@@ -66,6 +86,74 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
 	logger.FatalOnError(err, fmt.Sprintf("Error getting local IPv%v", family))
 
 	return localIP
+}
+
+func isValidGlobalIPv6(ip net.IP) bool {
+	// fd00::/8 - avoid locally assigned ULA
+	if ip.To16() == nil || ip.IsLoopback() || ip.IsLinkLocalUnicast() ||
+		ip[0] == 0xfd {
+		return false
+	}
+	// Accept fc00::/8 or global unicast (2000::/3)
+	return ip[0] == 0xfc || (ip[0]&0xe0 == 0x20)
+}
+
+func getValidGlobalIPv6FromSameInterface(ip net.IP) (string, error) {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to list interfaces")
+	}
+	var (
+		ownedIfaceName  string
+		ownedIfaceAddrs []net.Addr
+	)
+
+	for _, iface := range interfaces {
+		addrs, err := iface.Addrs()
+		if err != nil {
+			logger.Warningf("Failed to list the address from the interface %v", iface.Name)
+			continue
+		}
+
+		// Confirm this interface owns the original (non-global) IP
+		ownsIP := false
+
+		for _, addr := range addrs {
+			ipNet, ok := addr.(*net.IPNet)
+			if !ok || ipNet.IP.To4() != nil {
+				continue
+			}
+
+			if ipNet.Contains(ip) {
+				ownsIP = true
+				break
+			}
+		}
+
+		if ownsIP {
+			ownedIfaceName = iface.Name
+			ownedIfaceAddrs = addrs
+
+			break
+		}
+	}
+
+	if len(ownedIfaceAddrs) == 0 {
+		return "", fmt.Errorf("no interface for IP %q was found", ip)
+	}
+	// Scan for a valid global IPv6 on the same interface
+	for _, addr := range ownedIfaceAddrs {
+		ipNet, ok := addr.(*net.IPNet)
+		if !ok {
+			continue
+		}
+
+		if isValidGlobalIPv6(ipNet.IP) {
+			return ipNet.IP.String(), nil
+		}
+	}
+
+	return "", fmt.Errorf("no valid global IPv6 found on interface %q for IP %q", ownedIfaceName, ip)
 }
 
 func GetLocalIP(family k8snet.IPFamily) string {


pkg/endpoint/local_ip.go
```diff
@@ -53,12 +53,24 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
 }
 
 func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
```
yboaron, line 63: Do we know what the root cause is? 
I mean why/how do we get this source IPv6 address with ovnk when trying to connect to external DNS server?

pkg/endpoint/local_ip.go
```diff
@@ -53,12 +53,24 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
...
 		localAddr := conn.LocalAddr().(*net.UDPAddr)
+		ip = localAddr.IP
 
-		return localAddr.IP.String()
+		if family == k8snet.IPv4 {
+			return ip.String()
+		}
+	}
+
+	if family == k8snet.IPv6 && !isValidGlobalIPv6(ip) {
```
yboaron: Do we want to run this logic even if Dial("udp"+string(family), net.JoinHostPort(dst, "53")) failed ?

pkg/endpoint/local_ip.go
yboaron: @aswinsuryan , 

A. 
default IPv6 route has the following IP addresses:
inet6 fd69::2/125               scope global
inet6 fc00:1234:4444::9/64      scope global
inet6 fe80::9c2a:83ff:febe:2c4  scope link
both fd69::2/125 and fc00:1234:4444::9/64  are valid , and the kernel selects teh first one.

 it seems that  **fd69::2**  IPv6 address is hardcoded private IPv6 address set by OVN-K  [1] , maybe we need to add code to exclude this specific IP ? 



[1]
https://github.com/openshift/ovn-kubernetes/blob/31d2803e5f38207bf2137ba2ec15a95cee1d0802/go-controller/pkg/config/config.go#L155

aswinsuryan: But the UDP dial will always return this ip right a we need to find a way to get the valid one.

pkg/endpoint/local_ip.go
aswinsuryan: wanted the back up logic to execute so moved out of the if block.

tpantelis: It seems this logic should be in the if block above since if `err != nil` then `ip` would be nil and `getValidGlobalIPv6FromSameInterface` would be a no-op anyway. Also it might be useful to log a message if the returned ip isn't valid (I assume a warning if it shouldn't normally occur).

```
if err == nil {
    ...
    if family == k8snet.IPv4 {
        return localAddr.IP.String()
    }

    if !isValidGlobalIPv6(localAddr.IP) {
        logger.Warningf("IP %q returned from Dial isn't usable - trying to find another IP from the same interface". localAddr.IP)
        ...
    }
}

// connection failed try fallback method
localIP, err := getLocalIPFromRoutes(family)
...
```

pkg/endpoint/local_ip.go
```diff
@@ -53,12 +53,24 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
...
+		// Try to find a valid global-scope IP from the same interface
+		ifaceIP, err := getValidGlobalIPv6FromSameInterface(ip)
+		if err == nil {
+			return ifaceIP
+		}
```
tpantelis, line 81: I think we should log a warning if this fails, ie:

`logger.Warningf("Failed to retrieve IPv6 from interface for %q", localAddr.IP)`

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +80,65 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+func isValidGlobalIPv6(ip net.IP) bool {
+	if ip.To4() != nil || ip.To16() == nil {
+		return false
+	}
+	if ip.IsLoopback() || ip.IsLinkLocalUnicast() {
+		return false
+	}
+	if ip[0] == 0xfd { // fd00::/8 - avoid locally assigned ULA
+		return false
+	}
```
tpantelis: The `ip.To4()` check really isn't necessary. Also we can logically combine these:

```suggestion
	if ip.To16() == nil || ip.IsLoopback() || ip.IsLinkLocalUnicast() ||
	    ip[0] == 0xfd   // fd00::/8 - avoid locally assigned ULA {
		return false
	}
```

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +80,65 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+	// Accept fc00::/8 or global unicast (2000::/3)
+	return ip[0] == 0xfc || (ip[0]&0xe0 == 0x20)
+}
+
+func getValidGlobalIPv6FromSameInterface(ip net.IP) (string, error) {
+	if ip == nil || ip.To4() != nil || ip.To16() == nil {
+		return "", fmt.Errorf("input IP is not a valid IPv6 address: %v", ip)
+	}
+
```
tpantelis: These checks are redundant since the validity of `ip` has already been verified by the caller.

```suggestion
```

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +80,65 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return "", fmt.Errorf("failed to list interfaces: %w", err)
+	}
+
+	for _, iface := range interfaces {
+		addrs, err := iface.Addrs()
+		if err != nil {
```
tpantelis, line 113: I think we should log a warning here.

tpantelis (changes requested):

pkg/endpoint/local_ip.go
aswinsuryan: moved the code to if

pkg/endpoint/local_ip.go
```diff
@@ -53,12 +53,25 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
...
+		if !isValidGlobalIPv6(localAddr.IP) {
+			logger.Warningf("IP %q returned from Dial isn't usable - trying to find another IP from the same interface", localAddr.IP)
+			// Try to find a valid global-scope IP from the same interface
+			ifaceIP, err := getValidGlobalIPv6FromSameInterface(localAddr.IP)
+			if err == nil {
+				return ifaceIP
+			} else {
+				logger.Warningf("Failed to retrieve valid IPv6 address from interface for %q", localAddr.IP)
+			}
+		}
```
tpantelis: I think it needs to return if IPv6 is valid:

```suggestion
		if family == k8snet.IPv4 || isValidGlobalIPv6(localAddr.IP) {
			return localAddr.IP.String()
		}

		logger.Warningf("IP %q returned from Dial isn't usable - trying to find another IP from the same interface", localAddr.IP)
		// Try to find a valid global-scope IP from the same interface
		...
```

pkg/endpoint/local_ip.go
```diff
@@ -53,12 +53,25 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
...
 		localAddr := conn.LocalAddr().(*net.UDPAddr)
 
-		return localAddr.IP.String()
+		if family == k8snet.IPv4 || isValidGlobalIPv6(localAddr.IP) {
+			return localAddr.IP.String()
+		}
+
+		logger.Warningf("IP %q returned from Dial isn't usable - trying to find another IP from the same interface", localAddr.IP)
+
+		// Try to find a valid global-scope IP from the same interface
```
yboaron, line 75: In case we failed to find validIPv6Adresses , we can fallback to getLocalIPFromRoutes, and we can remove getValidGlobalIPv6FromSameInterface , getLocalIPFromRoutes function should be something like : 

```
unc getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
	netlink := netlinkAPI.New()

	routes, err := netlink.RouteList(nil, family)
	if err != nil {
		return "", errors.Wrapf(err, "error listing routes for IPv%v", family)
	}

	for i := range routes {
		if routes[i].Gw != nil && routes[i].Src != nil {
			ipStr := routes[i].Src.String()

			if  family == k8snet.IPv6  && !isValidGlobalIPv6(ipStr) {
				continue
			}

			return ipStr, nil
		}
	}

	return "", errors.Errorf("couldn't find route for family %v", family)
}
```

aswinsuryan: Yes, that makes it robust. Did the changes.

pkg/endpoint/local_ip.go
```diff
@@ -44,21 +44,38 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
...
+		ifaceIP, err := getValidGlobalIPv6FromSameInterface(localAddr.IP)
+		if err == nil {
+			return ifaceIP
+		} else {
+			logger.Warningf("Failed to retrieve valid IPv6 address from interface for %q", localAddr.IP)
+		}
```
tpantelis: From @yboaron [last comment](https://github.com/submariner-io/submariner/pull/3470#discussion_r2157052235), `getValidGlobalIPv6FromSameInterface` is not necessary so this code can be removed. Also, since we're falling thru to `getLocalIPFromRoutes`, we can set `err` on line L70 instead of logging a message:

`err = errors.Errorf("IP %q returned from Dial isn't usable", localAddr.IP)`

and log the error prior to L81 below:

`logger.Warningf("Unable to obtain local IP from Dial: %q. Attempting to obtain local IP%s from routes", err.Error(), family)`

tpantelis (changes requested):

yboaron:
Also, it will be useful to cover this use case in UT ,
You should add 

```
		Expect(netLink.RouteAdd(&netlink.Route{
			LinkIndex: 2,
			Src:       net.ParseIP("fd69::2"),
			Gw:        net.ParseIP("2001:0:0:0::"),
		})).To(Succeed())

```

Before 
[this line ](https://github.com/submariner-io/submariner/blob/devel/pkg/endpoint/local_endpoint_test.go#L117)

pkg/endpoint/local_ip.go
aswinsuryan: but the backup method gets the ip  from the route right ?  getValidGlobalIPv6FromSameInterface  will try to find a valid  ip from the interface whose ip is returned.

pkg/endpoint/local_endpoint_test.go
```diff
@@ -120,6 +120,12 @@ var _ = Describe("GetLocalSpec", func() {
 			Gw:        net.ParseIP("2001:0:0:0::"),
 		})).To(Succeed())
 
+		Expect(netLink.RouteAdd(&netlink.Route{
```
yboaron, line 123: To simulate ovnk use case, I think that you should add the route before 

https://github.com/submariner-io/submariner/blob/0f0696db23adf1666100dc3050d89f4af16fc703/pkg/endpoint/local_endpoint_test.go#L117

pkg/endpoint/local_ip.go
tpantelis: @yboaron ☝️ ?

pkg/endpoint/local_endpoint_test.go
aswinsuryan: Done.

pkg/endpoint/local_ip.go
```diff
@@ -44,21 +44,38 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
 	}
 
 	for i := range routes {
-		if routes[i].Gw != nil {
-			return routes[i].Src.String(), nil
+		if routes[i].Gw != nil && routes[i].Src != nil {
+			ip := routes[i].Src
+			if family == k8snet.IPv6 && !isValidGlobalIPv6(ip) {
```
yboaron, line 52: Maybe we  should add a comment here to explain why  do we check that IPv6 address is valid .

    OVNK configures special internal masquerade IPs,
    these IPs are used internally by OVNK infra dataplane, and they should
    not be selected as private IPs. [1]
 
    [1] https://github.com/ovn-kubernetes/ovn-kubernetes/blob/eea781bd80d4afb29e0d43c81ea5503fbe452d78/docs/design/host-to-node-port-hairpin-trafficflow.md?plain=1#L59

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +85,60 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+		addrs, err := iface.Addrs()
+		if err != nil {
+			logger.Warningf("Failed to list the address from the interface %v", iface.Name)
+			continue
+		}
+
+		// Confirm this interface owns the original (non-global) IP
+		ownsIP := false
+
+		for _, addr := range addrs {
```
yboaron, line 121: Can we have it in single for loop ?

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +85,60 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+			if ipNet, ok := addr.(*net.IPNet); ok && ipNet.Contains(ip) {
```
yboaron: Also we can skip non-IPv6 addresses , right ?

pkg/endpoint/local_ip.go
```diff
@@ -44,21 +44,38 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
...
+		}
+
+		logger.Warningf("IP %q returned from Dial isn't usable - trying to find another IP from the same interface", localAddr.IP)
+
+		// Try to find a valid global-scope IP from the same interface
+		ifaceIP, err := getValidGlobalIPv6FromSameInterface(localAddr.IP)
+		if err == nil {
+			return ifaceIP
+		} else {
+			logger.Warningf("Failed to retrieve valid IPv6 address from interface for %q", localAddr.IP)
```
yboaron: It would be useful to log the interface name if possible

pkg/endpoint/local_ip.go
aswinsuryan: Done.

pkg/endpoint/local_ip.go
aswinsuryan: Done.

pkg/endpoint/local_ip.go
aswinsuryan: But then we miss the global IP if it comes before we identify the interface?

pkg/endpoint/local_ip.go
aswinsuryan: Done.

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +88,64 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+				continue
+			}
+
+			if isValidGlobalIPv6(ipNet.IP) {
+				return ipNet.IP.String(), iface.Name, nil
+			}
+		}
+	}
+
+	return "", "", fmt.Errorf("no valid global IPv6 found on same interface")
```
tpantelis: Returning the interface name really doesn't add any value b/c we only return it if we don't return an error but the caller only uses it for error reporting. Instead of returning it to the caller, we could embed it in the returned error message where appropriate. But in order to do that we first need to find the right interface. Could we do that by first finding the one that "owns" it and then searching for a valid IP?, ie:

```
var (
ownedIfaceName string
ownedIfaceAddrs []net.Addr
)

for _, iface := range interfaces {
    ...
    ownsIP := false
    for _, addr := range addrs {
        ...
    }

    if ownsIP {
        ownedIfaceName = iface.Name
        ownedIfaceAddrs = addrs
        break
    }
}

if len(ownedIfaceAddrs) == 0 {
    return "", fmt.Errorf("no interface for IP %q was found", ip)
}

// Scan for a valid global IPv6 on the same interface
for _, addr := range ownedIfaceAddrs {
    ...
}

return "", fmt.Errorf("no valid global IPv6 found on interface %q for IP %q", ownedIfaceName, ip)
```

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +88,64 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+		}
+
+		if !ownsIP {
+			continue
+		}
+
+		// Scan for a valid global IPv6 on the same interface
+		for _, addr := range addrs {
+			ipNet, ok := addr.(*net.IPNet)
+			if !ok || ipNet.IP.To4() != nil {
```
tpantelis: No need to check `ipNet.IP.To4()` here as `isValidGlobalIPv6` will handle it.

pkg/endpoint/local_ip.go
```diff
@@ -68,6 +88,64 @@ func GetLocalIPForDestination(dst string, family k8snet.IPFamily) string {
...
+		return false
+	}
+	// Accept fc00::/8 or global unicast (2000::/3)
+	return ip[0] == 0xfc || (ip[0]&0xe0 == 0x20)
+}
+
+func getValidGlobalIPv6FromSameInterface(ip net.IP) (string, string, error) {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return "", "", fmt.Errorf("failed to list interfaces: %w", err)
```
tpantelis:
```suggestion
		return "", "", errors.Wrap(err, "failed to list interfaces")
```

pkg/endpoint/local_ip.go
```diff
@@ -44,21 +44,41 @@ func getLocalIPFromRoutes(family k8snet.IPFamily) (string, error) {
...
+		}
+
+		logger.Warningf("IP %q returned from Dial isn't usable - trying to find another IP from the same interface", localAddr.IP)
+
+		// Try to find a valid global-scope IP from the same interface
+		ifaceIP, ifaceName, err := getValidGlobalIPv6FromSameInterface(localAddr.IP)
+		if err == nil {
+			return ifaceIP
+		} else {
+			logger.Warningf("Failed to retrieve valid IPv6 address from interface %q for %q", ifaceName, localAddr.IP)
```
tpantelis: We should log the error message here:

```suggestion
			logger.Warningf("Failed to retrieve valid IPv6 address for %q: %v", ifaceName, localAddr.IP, err)
```

tpantelis (changes requested):

pkg/endpoint/local_ip.go
yboaron: Yeah, in single loop we need to check for every address if it is valid global IPv6 and also one of the addresses = given IP, let's keep in two for loops.

tpantelis (approved):

yboaron (approved):

PR #3475, opened 2025-06-22
Use IPv6 private address for cable name in IPv6 cluster

yboaron:
Currently, endpoint's cable name is generated using IPv4 private IP. This PR updates the logic to fallback to the IPv6 private IP if no IPv4 address is available.


yboaron:
Now, [IPv6 gate](https://github.com/submariner-io/submariner/actions/runs/15807579845/job/44556434658?pr=3475) is  also green

tpantelis (approved):

vthapar (approved):

PR #3499, opened 2025-07-07
Switch to libreswan-minimal in the gateway

skitt:
This reduces the image size (211MiB instead of 234MiB) and drops all the systemd dependencies.

## Summary by CodeRabbit

* **Chores**
  * Updated base container images from Fedora 42 to Fedora 43 for gateway, globalnet, and route-agent services.
  * Reduced gateway runtime footprint by switching to a minimal VPN package and adjusting base runtime dependencies.

### DIFF
--- a/package/Dockerfile.submariner-gateway
+++ b/package/Dockerfile.submariner-gateway
@@ -24,7 +24,7 @@ COPY package/dnf_install /
 # kmod is required so that libreswan can load modules
 RUN /dnf_install -a ${TARGETPLATFORM} -v ${FEDORA_VERSION} -r /output/gateway \
     glibc bash glibc-minimal-langpack coreutils-single \
-    libcurl-minimal iproute libreswan kmod \
+    libcurl-minimal iproute libreswan-minimal kmod \
     openssl nss-tools
 
 FROM scratch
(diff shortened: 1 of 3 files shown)


package/Dockerfile.submariner-gateway
```diff
@@ -24,7 +24,7 @@ COPY package/dnf_install /
 # kmod is required so that libreswan can load modules
 RUN /dnf_install -a ${TARGETPLATFORM} -v ${FEDORA_VERSION} -r /output/gateway \
     glibc bash glibc-minimal-langpack coreutils-single \
-    libcurl-minimal iproute libreswan kmod
+    libcurl-minimal iproute libreswan-minimal kmod nss-tools
```
skitt: The `nss-tools` dependency will be added to the next release of the Libreswan packages, so that part won’t be necessary.

dfarrell07 (approved):

tpantelis (approved):

dfarrell07:
Seeing all the NFTables jobs failing, might be related to the incompatibility work @yboaron was digging into.

skitt:
> Seeing all the NFTables jobs failing, might be related to the incompatibility work @yboaron was digging into.

Yes, this is the same incompatibility between F43 `nftables` and kind’s.

dfarrell07:
We've also switched the Konflux builds to libreswan 5.x, but not min version. Might should consider those at same time, if want to make edits there.

skitt:
> We've also switched the Konflux builds to libreswan 5.x, but not min version. Might should consider those at same time, if want to make edits there.

The minimal package isn’t available in RHEL 10 yet, it should be published in the next point-release.

PR #3544, opened 2025-07-30
Address SNAT-related issues in nftables mode with kube-proxy

### LINKED ISSUE #3541: Inter-cluster datapath is broken on Kind clusters (with KindNet CNI) when kube-proxy and submariner use nftables.
When deploying Submariner with both Submariner and kube-proxy configured to use nftables (via **make deploy using=nftables** for example), datapath connectivity from pods to the remote cluster is broken. 

Pods using HostNetworking (such as periodic ping Submariner runs from non GW nodes to remote GW node) can reach the remote cluster successfully, while regular pods are not.

The root cause appears to be due to two key behaviors affecting source IP preservation:

**[A] KindNet’s Egress SNAT Behavior**

KindNet CNI applies SNAT to egress traffic when the destination IP is outside the local cluster’s pod CIDR. In the context of Submariner, this results in the pod’s original source IP being rewritten to the IP address of the vx-submariner interface.

[1] Is the content of ping packet sent from local pod IP 10.131.1.6 to remote pod IP 10.130.1.1 captured on local GW node and the relevant nftables chain content[2]. 

This breaks Submariner's datapath, as there is no matching IPsec policy for the rewritten source address.

In iptables mode, this behavior could be overridden by inserting higher-priority rules. However, with nftables, all rules in the nat postrouting chain are processed together, and such prioritization is not possible.

Similar issue has been raised upstream in the Kind project [3], the maintainers clarified that Kindnet CNI is intentionally kept minimal and lightweight, and they do not plan to change this behavior unless it violates Kubernetes conformance requirements.

**[B] kube-proxy’s Ingress SNAT in nftables Mode**

kube-proxy also performs SNAT on ingress traffic, rewriting the remote pod’s source IP to a local IP address. Since nftables chains do not support rule priorities in the same way iptables does, Submariner’s rules to accept the traffic are ineffective — kube-proxy’s SNAT is always applied before any decision is made.

As a result, the remote pod’s source IP is replaced with a local CNI IP, breaking source IP preservation.

[1]
Frame 1: 398 bytes on wire (3184 bits), 398 bytes captured (3184 bits)
Linux cooked capture v2
Internet Protocol Version 4, Src: 172.18.0.4, Dst: 172.18.0.7
User Datagram Protocol, Src Port: 49209, Dst Port: 4800
Virtual eXtensible Local Area Network
Ethernet II, Src: 32:33:6d:91:a3:3c (32:33:6d:91:a3:3c), Dst: c2:75:84:d3:d2:d3 (c2:75:84:d3:d2:d3)
Internet Protocol Version 4, Src: 240.18.0.4, Dst: 10.130.1.1
Internet Control Message Protocol

[2]
chain KIND-MASQ-AGENT {
    ip daddr 10.131.0.0/16 xt match "comment" counter packets 490 bytes 41442 return
    xt match "comment" counter packets 12012 bytes 763364 xt target "MASQUERADE"
}

[3]
https://github.com/kubernetes-sigs/kind/issues/3474

yboaron:
This PR introduces two changes to address SNAT-related issues with kube-proxy in nftables mode and KindNet:

Avoiding KindNet Egress SNAT:
Adds support for SelfSNAT in RouteAgent kube-proxy handler to bypass KindNet's egress SNAT, which rewrites pod source IPs and breaks IPSec-based Submariner datapath on gateway nodes.

Avoid Ingress SNAT by kube-proxy:
Add support for SelfSNAT in RouteAgent kube-proxy  handler to avoid kube-proxy ingress SNAT.

Both changes aim to preserve the original pod source IP to maintain correct Submariner datapath behavior.

Fixes: https://github.com/submariner-io/submariner/issues/3541


tpantelis (approved):

yboaron:
cc @aswinsuryan

aswinsuryan (approved):

PR #3567, opened 2025-08-11
Make Konflux gateway run as root for Pluto

dfarrell07:

### DIFF
--- a/package/Dockerfile.submariner-gateway.konflux
+++ b/package/Dockerfile.submariner-gateway.konflux
@@ -62,7 +62,8 @@ RUN chown -Rc submariner:root /usr/local/bin && \
 
 COPY --from=builder ${SOURCE}/LICENSE /licenses/LICENSE
 
-USER submariner
+# Pluto fails with `permission denied (must be superuser)` if running as submariner user
+USER root
 
 ENTRYPOINT ["/usr/local/bin/submariner.sh"]
 


dfarrell07:
This will break EC. If we must do this, we'll need an exception somehow.

Jaanki (approved):

vthapar (approved):

tpantelis (approved):

dfarrell07:
```
        "failed": [
            {
                "name": "RunAsNonRoot",
                "elapsed_time": 0,
                "description": "Checking if container runs as the root user because a container that does not specify a non-root user will fail the automatic certification, and will be subject to a manual review before the container can be approved for publication",
                "help": "Check RunAsNonRoot encountered an error. Please review the preflight.log file for more information.",
                "suggestion": "Indicate a specific USER in the dockerfile or containerfile",
                "knowledgebase_url": "https://access.redhat.com/documentation/en-us/red_hat_software_certification/2024/html-single/red_hat_openshift_software_certification_policy_guide/index#assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction",
                "check_url": "https://access.redhat.com/documentation/en-us/red_hat_software_certification/2024/html-single/red_hat_openshift_software_certification_policy_guide/index#assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction"
            }
        ],
```

https://konflux-ui.apps.kflux-prd-rh02.0fk9.p1.openshiftapps.com/ns/submariner-tenant/applications/submariner-0-21/pipelineruns/submariner-gateway-0-21-on-pull-request-86jb5/logs?task=ecosystem-cert-preflight-checks

I'm a bit surprised this doesn't fail the EC, seems to just be a warning. I think we saw some different behavior with the old/broken version of this check.

PR #3571, opened 2025-08-12
Set default packetfilter driver to nftables

yboaron:
Starting from submariner 0.22, nftables will be the default packetfilter driver.


tpantelis (dismissed):

tpantelis:
What's the status on the failing CI tests?

yboaron:
We need to update shipyard code before merging this PR  or all gates will test only with nftables.

yboaron:
On second thought, we may not need to update Shipyard code, since the ConfigMap is always created [1]. If nftables isn’t set, the ConfigMap will have use-nftables: false, meaning iptables is used as the packet filter. 

I’ll check the gate logs to double-check.


[1] [config map creation](https://github.com/submariner-io/shipyard/blob/2298fd1856ab860a36ceeaaa43f3ac438259e88a/scripts/shared/deploy.sh#L168)

yboaron:
@tpantelis , I checked locally, and we’re fine — iptables is used as packetfilter driver when **nftables** flag is not set in Shipyard. I think it’s safe to merge this PR.

tpantelis (approved):

yboaron:
Convert to draft until we determine how to handle OVNK avoid-SNAT when driver is iptables

yboaron:
Setting the PR back to **Ready for Review** — since Self-SNAT in 0.20 is only activated when a configmap is present (not by default), it’s safe to use nftables as the default packet filter in devel.

dfarrell07 (approved):

aswinsuryan (approved):

PR #3573, opened 2025-08-13
Add CI gates to cover additional nftables,ovn use cases

yboaron:
Added nftables+OVN+Globalnet and nftables+OVN+dual-stack gates.


tpantelis (approved):

pkg/packetfilter/nftables/nftables.go
```diff
@@ -31,12 +31,12 @@ import (
...
 )
 
-const (
-	/* Single table named 'submariner' is used for nftables configuration.*/
-	submarinerTable = "submariner"
-)
-
 var (
+	ipFamilyToNftTableName = map[k8snet.IPFamily]string{
+		k8snet.IPv4: "submariner-ipv4",
```
tpantelis, line 36: Are there any migration concerns here, ie do you think we should keep the IPv4 name the same?

yboaron: We didn’t use nftables in Submariner previously — only iptables.
I think changing nftables table names shouldn’t impact migration.

yboaron:
Let's hold off on merging this PR until we determine how to handle the ingress SNAT issue.

yboaron:
We can merge this PR.  nftables table named **submariner** will be used for 0.20 Self-SNAT, while a separate table named submariner-ipv4 will be used for nftables IPv4 rules, so no conflicts or clashes are expected during upgrade.

tpantelis (approved):

PR #3575, opened 2025-08-17
Introduce avoid-SNAT handler for OVN and add nftables support alongside iptables

yboaron:
This PR introduces an avoid-SNAT handler for OVN that resolves ingress SNAT issues by defining high-priority nftables self-SNAT rules.

It also adds support for running an nftables driver as secondary to iptables, enabling definition of higher-priority nftables rules required to properly handle SNAT with OVN CNI.


yboaron:
The avoid SNAT handler will be pushed only to release 0.20 , we can close this PR

PR #3581, opened 2025-08-21
Implement intra-cluster routing disablement in the OVN handler

tpantelis:
If intra-cluster routing is disabled then elide processing `NonGatewayRoutes`, ie don't instantiate the `NonGatewayRouteController` and `NonGatewayRouteHandler`.

In addition, only program the host networking dataplane if the local node is the active gateway node. So on a passive gateway node, don't add host networking rules/routes for remote `Endpoints`. On transition to the the active gateway, then add the rules/routes for existing `Endpoints`. On transition to passive, remove the rules/routes for existing `Endpoints`.


yboaron:
Overall code LGTM, 

I have one Q:
What’s should be our plan for handling changes to IntraRoutingDisabled in the Submariner CR at runtime? 

For example, changing it from false to true, the current proposed operator code only updates the RouteAgent DaemonSet, but I think it makes sense also to clean up any unnecessary datapath configuration and related CRs (like NonGatewayRoute).

I think we should either handle this directly in the operator code, or document that Submariner should be uninstalled before changing IntraRoutingDisabled. WDYT ?

tpantelis:
> I think we should either handle this directly in the operator code, or document that Submariner should be uninstalled before changing IntraRoutingDisabled. WDYT ?

We can look into handling and if it looks problematic then document uninstall/reinstall.

yboaron (approved):

aswinsuryan (approved):

PR #3582, opened 2025-08-22
Improve close detection and check for truncated messages in natDiscovery.listenerLoop

tpantelis:
The Cursor AI tool identified a couple potential issues:

The code assumes that `length == 0` from `ReadFromUDP` only means the connection is closed. However, UDP can legitimately receive zero-length packets. To be safe, check the returned error for `net.ErrClosed`.

A fixed buffer size of 2048 is used but a UDP message larger than that will be silently truncated which would cause a parsing error. Add a warning re: truncation if the returned length is 2048.

### DIFF
--- a/pkg/natdiscovery/listener.go
+++ b/pkg/natdiscovery/listener.go
@@ -95,17 +95,31 @@ func createServerConnection(port int32, family k8snet.IPFamily) (ServerConnectio
 }
 
 func (nd *natDiscovery) listenerLoop(serverConnection ServerConnection) {
-	buf := make([]byte, 2048)
+	const maxUDPMessageSize = 2048
+
+	buf := make([]byte, maxUDPMessageSize)
 
 	for {
 		length, addr, err := serverConnection.ReadFromUDP(buf)
-		if length == 0 {
-			logger.Info("Stopping NAT listener")
-			return
-		} else if err != nil {
-			logger.Errorf(err, "Error receiving from udp")
-		} else if err := nd.parseAndHandleMessageFromAddress(buf[:length], addr); err != nil {
-			logger.Errorf(err, "Error handling message from address %s:\n%s", addr.String(), hex.Dump(buf[:length]))
+		if err != nil {
+			if errors.Is(err, net.ErrClosed) {
+				logger.Info("Stopping NAT listener")
+				return
+			}
+
+			logger.Error(err, "Error receiving from UDP")
+
+			continue
+		}
+
+		if length > 0 {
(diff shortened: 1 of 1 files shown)


pkg/natdiscovery/listener.go
```diff
@@ -95,17 +94,31 @@ func createServerConnection(port int32, family k8snet.IPFamily) (ServerConnectio
...
+			continue
+		}
+
+		if length > 0 {
+			if err := nd.parseAndHandleMessageFromAddress(buf[:length], addr); err != nil {
+				if length == maxUDPMessageSize {
+					logger.Warningf("UDP message from %s may have been truncated (received %d bytes)", addr.String(), length)
+				}
+
+				logger.Errorf(err, "Error handling message from address %s", addr.String())
```
yboaron: I assume it would be useful to log also the buffer  here :  

logger.Errorf(err, "Error handling message from address %s:\n%s", addr.String(), hex.Dump(buf[:length]))

tpantelis: I wasn't sure that binary data would actually be useful, also it may fill up the log if it occurred repeatedly. But I'll add it back.

yboaron (approved):

pkg/natdiscovery/listener.go
```diff
@@ -95,17 +95,31 @@ func createServerConnection(port int32, family k8snet.IPFamily) (ServerConnectio
...
 	for {
 		length, addr, err := serverConnection.ReadFromUDP(buf)
-		if length == 0 {
-			logger.Info("Stopping NAT listener")
-			return
-		} else if err != nil {
-			logger.Errorf(err, "Error receiving from udp")
-		} else if err := nd.parseAndHandleMessageFromAddress(buf[:length], addr); err != nil {
-			logger.Errorf(err, "Error handling message from address %s:\n%s", addr.String(), hex.Dump(buf[:length]))
+		if err != nil {
```
skitt, line 104: Even when an error is returned, if length is non-nil then the returned data should be processed (see the `ReadFrom` documentation at https://pkg.go.dev/net#PacketConn).

skitt (changes requested):

pkg/natdiscovery/listener.go
tpantelis: Hmm that's seems counter-intuitive but it does say that...

tpantelis: After further reading, this is to handle cases where a partial read occurred before an error was encountered, specifically a read timeout. However, a partial message would likely fail our parsing anyway. Also the prior code checked the returned error first.

skitt (approved):

PR #3589, opened 2025-08-27
Check for fs.ErrExist and fs.ErrNotExist

tpantelis:
...instead of `os.IsExist` and `os.IsNotExist` as advised in the docs.

### DIFF
--- a/pkg/cable/vxlan/vxlan.go
+++ b/pkg/cable/vxlan/vxlan.go
@@ -21,7 +21,6 @@ package vxlan
 import (
 	"fmt"
 	"net"
-	"os"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


pkg/cable/vxlan/vxlan.go
```diff
@@ -120,7 +119,7 @@ func (v *vxLan) createVxlanInterface(port int) error {
 	}
 
 	err = v.netLink.RuleAddIfNotPresent(netlinkAPI.NewTableRule(TableID, k8snet.IPv4))
-	if err != nil && !os.IsExist(err) {
+	if err != nil {
```
skitt, line 122: For other reviewers, removing the check here is fine because `RuleAddIfNotPresent` can’t return `ErrExist`.

skitt (approved):

yboaron (approved):

PR #3605, opened 2025-09-03
Re-install remote Endpoint in cable driver when Subnets change

tpantelis:
We observed a case where the global CIDR changed on re-deployment and thus changed the `Endpoint` `Subnets` but the remote cluster retained the previous IPsec connection with the old subnet b/c the cable engine did not re-install the `Endpoint`. So the `installCableWithNATInfo` method was modified to also check if the `Subnets` changed when determining whether to disconnect/re-connect the `Endpoint`.

### DIFF
--- a/pkg/cableengine/cableengine.go
+++ b/pkg/cableengine/cableengine.go
@@ -27,6 +27,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/log"
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cable"
+	"github.com/submariner-io/submariner/pkg/cidr"
 	submendpoint "github.com/submariner-io/submariner/pkg/endpoint"
 	"github.com/submariner-io/submariner/pkg/natdiscovery"
 	"github.com/submariner-io/submariner/pkg/types"
@@ -152,6 +153,10 @@ func (i *engine) SetupNATDiscovery(natDiscovery natdiscovery.Interface) {
 }
 
 func (i *engine) shouldProceedWithInstallation(familyCableName string) bool {
+	if !i.running {
+		return false
+	}
+
 	if _, ok := i.natDiscoveryPending[familyCableName]; !ok {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


yboaron:
To properly support endpoint subnet updates, we need to ensure that also all other handlers (RouteAgent, Globalnet (IPAM?), etc.) handle this case as well. 

I don’t think that current code fully supports it, addressing it only on the Gateway won’t be sufficient if, for example, RA handlers still configure routing rules for the old subnets.
To validate this use case, we should also enhance our testing to cover datapath after runtime subnet change.

This scenario is only relevant for Globalnet, the only practical case is when the cluster GlobalCIDR is changed at runtime, right ? 

To keep things simpler, maybe we should document that in such cases it’s recommended to reinstall Submariner. WDYT?

tpantelis:
> This scenario is only relevant for Globalnet, the only practical case is when the cluster GlobalCIDR is changed at runtime, right ?
> 
> To keep things simpler, maybe we should document that in such cases it’s recommended to reinstall Submariner. WDYT?

In the case that was observed, the GlobalCIDR did update but user didn't specifically do it. I think they reinstalled the add-on  which does reinstall the Submariner but I think somehow the Endpoint on the broker was temporarily stale. So the user was unaware of this. But I think the underlying issue is why was the Endpoint seemingly stale. If Submariner was uninstalled/reinstalled, the Endpoint should've been first cleaned up in the broker. Also on reinstall, it seems the clusters didn't get the same Global CIDRs (or at least one of them didn't).

This PR attempts to address a subnet change in a remote cluster, not a runtime change in the local cluster (that would require a restart of all pods anyway). Perhaps there would also need to be changes in the RouteAgent as well to handle remote subnet change. This is also probably a band-aid trying to cover up another issue.

I can abandon the subnet-handling changes (although I'd like to keep the unit test refactoring 😄 ).

yboaron:
Yes, the use case is a runtime subnet change in a remote cluster, and if it occurs without Submariner reinstall, it effectively means that local subnets have changed at runtime in the remote cluster.

AFAIK, some of RA handlers support this scenario via RemoteEndpointRemoved and RemoteEndpointCreated events (which would be triggered by reinstalling Submariner on the remote cluster). It makes sense to stick with this approach.

> (although I'd like to keep the unit test refactoring 😄 ).

Sure, let's keep the unit test refactoring

tpantelis:
> AFAIK, some of RA handlers support this scenario via RemoteEndpointRemoved and RemoteEndpointCreated events (which would be triggered by reinstalling Submariner on the remote cluster). It makes sense to stick with this approach.

That is what **should** happen as an `Endpoint` subnet change could only occur on restart of the gateway and it **should** remove the previous `Endpoint`. But when syncing to the broker, I think there might be edge-case timing where a remove and create isn't observed by a remote cluster b/c the `Endpoint` name doesn't actually change.

PR #3606, opened 2025-09-04
Delete nftables tables on Submariner uninstall

### LINKED ISSUE #3592: nftables tables created by Submariner are not removed after Submariner uninstalled
When Submariner is deployed in nftables mode (e.g., **make deploy using=nftables**), the nftables tables created by Submariner [1] are not removed after Submariner is uninstalled.

[1] 

```
table ip submariner-ipv4 {
        comment "rules for submariner"
}

```

yboaron:
Ensure that all nftables tables created by Submariner are properly removed during uninstallation.

Fixes: https://github.com/submariner-io/submariner/issues/3592


dfarrell07 (approved):

pkg/packetfilter/packetfilter.go
```diff
@@ -302,6 +302,9 @@ type Driver interface {
 	// named Sets.
 	NewNamedSet(set *SetInfo) NamedSet
 	DestroySets(nameFilter func(string) bool) error
+
+	// Tables
+	CleanupTable() error
```
tpantelis: Perhaps we can name this a bit more generic, how about `Uninstall`?

pkg/routeagent_driver/main.go
```diff
@@ -165,6 +166,18 @@ func main() {
...
+			if err != nil {
+				logger.Errorf(err, "error creating packetfilter instance for IPv%v", family)
+				continue
+			}
+
+			if err := pFilter.CleanupTable(); err != nil {
+				logger.Errorf(err, "error Cleanup packetfilter table for IPv%v", family)
+			}
+		}
+
```
tpantelis, line 180: Do we need to do this for globalnet as well?

yboaron: We only need to clean up the Submariner tables once, so handling it in RA should be sufficient since RA runs in both GN and non-GN cases.

tpantelis (approved):

PR #3608, opened 2025-09-06
OVN-Ipsec support

aswinsuryan:
OVN-Ipsec support


tpantelis (commented):
I know this is a draft but, after a cursory look...

I think the certificate syncer and controller could be put into a separate package (and with unit tests 😄), ie `pkg/certificate` (or preferably admiral?).  It would have a public interface:

```
type SigningRequestor interface {
    Issue(name string, ips []string) error
    Cleanup() error
}
```

Then we pass `certificate.SigningRequestor` to the cable engine.

PR #3620, opened 2025-09-10
Add force-udp-encaps coverage in tests

yboaron:
Depends on: https://github.com/submariner-io/shipyard/pull/2122


dfarrell07 (approved):

tpantelis (approved):

.github/workflows/e2e.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           - extra-toggles: nftables, ovn
           - extra-toggles: nftables, globalnet
           - extra-toggles: nftables, globalnet, ovn
-          - extra-toggles: nftables, ovn, dual-stack
+          - extra-toggles: nftables, ovn, dual-stack, force-udp-encaps
```
tpantelis, line 64: Should we keep the original `nftables, ovn, dual-stack` test?

yboaron: Enabling force-udp-encaps only impacts the encapsulation of IPSec packets, while rest of the components behave the same, but I assume we can still duplicate the test and add a dedicated one for force-udp-encaps.

PR #3621, opened 2025-09-11
Align Libreswan startup with the current package

skitt:
The systemd unit no longer uses _stackmanager.

This is required before Submariner can switch to Libreswan 5.3.

### DIFF
--- a/package/pluto
+++ b/package/pluto
@@ -6,8 +6,8 @@ set -e
 
 # These are the ExecStartPre lines from the systemd service definition
 /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
-/usr/libexec/ipsec/_stackmanager start
-/usr/sbin/ipsec --checknss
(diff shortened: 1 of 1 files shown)


vthapar (approved):

yboaron:
QQ : Should we expect any errors when running current code on the latest upstream image? IIRC, we’re now using Libreswan5 after moving to Fedora 42.

yboaron (approved):

skitt:
> QQ : Should we expect any errors when running current code on the latest upstream image? IIRC, we’re now using Libreswan5 after moving to Fedora 42.

Right, F42 has Libreswan 5.2; the `_stackmanager` removal happened in [Libreswan 5.3](https://github.com/libreswan/libreswan/releases/tag/v5.3). The new init sequence used here was introduced a while ago though so this works fine on earlier versions of Libreswan.

tpantelis (approved):

PR #3646, opened 2025-09-25
Switch from tcp.GlobalIP to tcp.GlobalServiceIP

skitt:
The GlobalIP constant is supposed to be replaced by GlobalServiceIP.

### DIFF
--- a/test/e2e/redundancy/gateway_failover.go
+++ b/test/e2e/redundancy/gateway_failover.go
@@ -176,7 +176,7 @@ func AwaitNewSubmarinerGatewayFullyConnected(f *subFramework.Framework, cluster 
 
 func defaultEndpointType() tcp.EndpointType {
 	if framework.TestContext.GlobalnetEnabled {
-		return tcp.GlobalIP
+		return tcp.GlobalServiceIP
 	}
 
 	return tcp.PodIP


test/e2e/redundancy/gateway_failover.go
```diff
@@ -176,7 +176,7 @@ func AwaitNewSubmarinerGatewayFullyConnected(f *subFramework.Framework, cluster
 
 func defaultEndpointType() tcp.EndpointType {
 	if framework.TestContext.GlobalnetEnabled {
-		return tcp.GlobalIP
+		return tcp.GlobalServiceIP
```
skitt, line 179: Given the context I would expect `GlobalPodIP`…

tpantelis: I assume the intention is to remove `GlobalIP` in shipyard to address the TODO.

It's a bit non-intuitive  the `EndpointType` for Globalnet tests determines which type of service to create - `GlobalServiceIP` for ClusterIP and `GlobalPodIP` for headless. The default is the former so this is correct.

skitt: Thanks! That is indeed the context; I was poking at the results of a new linter, `iotamixing`, and it doesn’t like the extra `GlobalServiceIP` constant matching `GlobalIP`; since there’s only one usage left I thought it was time to get rid of the TODO.

tpantelis (approved):

yboaron (approved):

PR #3658, opened 2025-10-03
Cert mode tests

tpantelis:


pkg/cable/libreswan/libreswan.go
```diff
@@ -498,13 +497,14 @@ func removeConnectionStanza(confPath, connName string) error {
...
+	confPath := SubmarinerConfPath()
 	endpoint := &endpointInfo.Endpoint
 	leftID := "submariner-client-" + i.localEndpoint.ClusterID
 	left := i.localEndpoint.GetPrivateIP(endpointInfo.UseFamily)
 	right := endpointInfo.UseIP
-	leftSubnets := i.localEndpoint.Subnets
-	rightSubnets := endpoint.Spec.Subnets
+
+	leftSubnets := i.localEndpoint.ExtractSubnetsExcludingIP(endpointInfo.UseIP)
+	rightSubnets := endpoint.Spec.ExtractSubnetsExcludingIP(endpointInfo.UseIP)
```
tpantelis, line 507: @aswinsuryan  Elsewhere it uses `ExtractSubnetsExcludingIP` so I assume we should do the same here?

pkg/cable/libreswan/libreswan.go
```diff
@@ -561,29 +559,28 @@ func (i *libreswan) connectToEndpointCertMode(endpointInfo *natdiscovery.NATEndp
...
 		}
 	}
 
+	i.connections = append(i.connections,
+		subv1.Connection{
+			Endpoint: endpoint.Spec,
+			UsingIP:  endpointInfo.UseIP,
+			UsingNAT: endpointInfo.UseNAT,
+			Status:   subv1.Connected,
+		})
```
tpantelis, line 579: @aswinsuryan Same as PSK mode, I assume we should only add one `Connection` entry.

pkg/cable/libreswan/libreswan.go
```diff
@@ -786,19 +783,6 @@ func (i *libreswan) clientConnectToEndpoint(connectionName string, endpointInfo
 
 // DisconnectFromEndpoint disconnects from the connection to the given endpoint.
 func (i *libreswan) DisconnectFromEndpoint(endpoint *types.SubmarinerEndpoint, family k8snet.IPFamily) error {
-	if i.authMode == AuthModeCert {
-		confPath := submarinerConfPath
-
-		connName := endpoint.Spec.CableName
-		if err := removeConnectionStanza(confPath, connName); err != nil {
```
tpantelis: @aswinsuryan Since multiple IPsec connections are added per subnet pairing, this doesn't work so I refactored to follow PSK mode below.

aswinsuryan: yes this  is right , initially code was adding only once entry with  left and right subnet as list , later I refactored it to reuse `refreshConnectionStatus` and have an entry per subnet pair, and missed this,

pkg/cable/libreswan/libreswan.go
aswinsuryan: +1

aswinsuryan: +1 , this may be required , though the default use case works

pkg/cable/libreswan/libreswan.go
aswinsuryan: +1 , this may be required , though the default use case works

PR #3664, opened 2025-10-09
Use pkutil for loading certificate private key

tpantelis:


pkg/cable/libreswan/certificate_handler.go
```diff
@@ -75,45 +74,79 @@ func (c *CertificateHandler) initNSSDatabase(ctx context.Context) error {
...
-	// Load client private key
-	err := c.loadPrivateKey(ctx, tlsKey, ClientKeyName)
+	// Load client certificate and key using pk12util
+	err := c.loadPrivateKey(ctx, tlsCert, tlsKey, ClientCertName)
 
-	return errors.Wrap(err, "failed to load client private key")
+	return errors.Wrap(err, "failed to load client certificate with key")
 }
 
 func (c *CertificateHandler) loadCertificate(ctx context.Context, certData []byte, nickname, trustFlags string) error {
```
aswinsuryan, line 85: stdin is not working here either seeing an error ,
''' OnSignedFn returned an error for secret "libreswan-asuryana-gcp-1-n42sx-submariner-gw-us-east1-b-ks6s5-asuryana-gcp-1" error="failed to load certificates into NSS database: failed to load CA certificate: failed to load certificate \"subm-ca-cert\": failed to execute certutil: certutil:  unable to open \"-\" for reading (-5950, 2).\n: exit status 255"''

so we need something like 
```
 func (c *CertificateHandler) loadCertificate(ctx context.Context, certData []byte, nickname, trustFlags string) error {
	// Write certificate to temporary file
	certFile, err := os.CreateTemp("", "submariner-ca-*.crt")
	if err != nil {
		return errors.Wrap(err, "failed to create temporary cert file")
	}
	defer os.Remove(certFile.Name())

	if _, err := certFile.Write(certData); err != nil {
		return errors.Wrap(err, "failed to write certificate to temporary file")
	}
	certFile.Close()

	//nolint:gosec // certutil args are from trusted config
	execCmd := exec.CommandContext(ctx, "certutil", "-A", "-d", "sql:"+c.nssDBDir, "-n", nickname, "-t", trustFlags, "-i", certFile.Name(), "-a")
	cmd := command.New(execCmd)

	if err := cmd.Run(); err != nil {
		return errors.Wrapf(err, "failed to load %s", nickname)
	}

	return nil
}
```

pkg/cable/libreswan/certificate_handler.go
```diff
@@ -39,7 +39,6 @@ var certLogger = log.Logger{Logger: logf.Log.WithName("CertHandler")}
 const (
 	CACertName     = "subm-ca-cert"
 	ClientCertName = "subm-client-cert"
```
aswinsuryan, line 41: this needs to be changed in certificate_auth_mode.go line 33 as well , 

`	leftID := ClientCertName`

pkg/cable/libreswan/certificate_handler.go
skitt: `certutil -A` should read from standard input if you drop the `i` option entirely:

```
	execCmd := c.newCertUtilCmd(ctx, "-A", "-n", nickname, "-t", trustFlags, "-a")
```

tpantelis: yeah I get the same error when running from the command line using `-i -`.  Omitting `-i` works.

aswinsuryan (approved):

pkg/cable/libreswan/certificate_handler.go
aswinsuryan: i tested with Openshift as well , it works

skitt (approved):

PR #3670, opened 2025-10-17
Minor adjustments to address issues found by GoLand's static analysis tool (Qodana)

tpantelis:
See commits for details.

### DIFF
--- a/pkg/cable/libreswan/libreswan.go
+++ b/pkg/cable/libreswan/libreswan.go
@@ -217,8 +217,11 @@ func (i *libreswan) Init(ctx context.Context) error {
 		}
 		defer file.Close()
 
-		fmt.Fprintf(file, "%%any %%any : PSK \"%s\"\n", i.secretKey)
+		if _, err = fmt.Fprintf(file, "%%any %%any : PSK \"%s\"\n", i.secretKey); err != nil {
+			return errors.Wrap(err, "error writing the secrets file")
+		}
 	} else if i.authMode == AuthModeCert {
+		i.plutoStarted = true
 		i.connectionFile = &ConnectionFile{Path: SubmarinerConfPath()}
 
 		logger.Infof("Issuing certificate with Private IPs %s Public IPs %s", i.localEndpoint.PrivateIPs, i.localEndpoint.PublicIPs)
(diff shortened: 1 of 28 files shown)


pkg/cable/libreswan/libreswan.go
```diff
@@ -207,18 +207,25 @@ func (i *libreswan) GetName() string {
 func (i *libreswan) Init(ctx context.Context) error {
 	logger.Infof("Initializing libreswan driver with authentication mode: %s", i.authMode)
 
+	var err error
```
skitt: More idiomatically:

```
	if i.authMode == AuthModePSK {
		logger.Info("Setting up PSK authentication")

		// Write the secrets file: %any %any : PSK "secret"
		file, err := os.Create(RootDir + "/etc/ipsec.d/submariner.secrets")
		if err != nil {
			return errors.Wrap(err, "error creating the secrets file")
		}

		defer file.Close()

		if _, err = fmt.Fprintf(file, "%%any %%any : PSK \"%s\"\n", i.secretKey); err != nil {
			return errors.Wrap(err, "error writing the secrets file")
		}
	} else if i.authMode == AuthModeCert {
		i.plutoStarted = true
		i.connectionFile = &ConnectionFile{Path: SubmarinerConfPath()}

		logger.Infof("Issuing certificate with Private IPs %s Public IPs %s", i.localEndpoint.PrivateIPs, i.localEndpoint.PublicIPs)

		sanIPs := make([]string, 0, len(i.localEndpoint.PrivateIPs)+len(i.localEndpoint.PublicIPs))
		sanIPs = append(sanIPs, i.localEndpoint.PrivateIPs...)
		sanIPs = append(sanIPs, i.localEndpoint.PublicIPs...)

		i.certificateHandler = NewCertificateHandler()

		if err := i.signingRequestor.Issue(ctx, "libreswan-"+i.localEndpoint.Hostname, sanIPs, i.certificateHandler.OnSignedCallback); err != nil {
			return errors.Wrap(err, "unable to issue certificate")
		}

		logger.Info("Successfully issued certificate")
	}

	return nil
```

(`if err == nil` is the opposite of what people expect in Go, and can lead to problems down the line.)

skitt (changes requested):

pkg/cable/libreswan/libreswan.go
tpantelis: yeah I was trying to "cheat" and gain extra line coverage 😄

skitt (approved):

pkg/cable/libreswan/libreswan.go
```diff
@@ -490,7 +490,7 @@ func (i *libreswan) bidirectionalConnectToEndpoint(connectionName string, endpoi
 	localEndpointIdentifier := i.localEndpoint.GetPrivateIP(endpointInfo.UseFamily)
 	remoteEndpointIdentifier := endpointInfo.Endpoint.Spec.GetPrivateIP(endpointInfo.UseFamily)
 
-	args := []string{}
+	var args []string
 
 	args = append(args, "--psk", encryptArg)
```
skitt: This is very minor but a number of these declarations could include the first `append`:

```
args := []string{"--psk", encryptArg}
```

dfarrell07 (approved):

PR #3678, opened 2025-10-22
Add agent-executable workflow docs for CVE fixes

dfarrell07:
CVE workflow docs and commit templates for AI agents and human devs.

PR #3695, opened 2025-11-03
Fix CVE in release-0.21

dfarrell07:
See commit message for details.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/submariner-io/submariner
 
-go 1.24.0
+go 1.24.9
 
 require (
 	github.com/cenkalti/backoff/v4 v4.3.0
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/submariner-io/submariner/tools
 
-go 1.24.0
+go 1.24.9
 
 require (
 	github.com/uw-labs/lichen v0.1.7


tpantelis (approved):

skitt (changes requested):
This defines the language level, not the toolchain. The appropriate fix here is either to add a toolchain specifier in `go.mod`, or to ensure that the build toolchain is 1.24.9 or later (in the 1.24 branch) — and that is already the case, the images provide a version of Go that has the fixes.

Thinking about this more, specifying a 1.24.9 isn’t even a correct fix; Shipyard uses 1.25 toolchains, so any version of those would be accepted, even though the fixes are only in 1.25.3 and later on the 1.25 branch.

dfarrell07:
> This defines the language level, not the toolchain. The appropriate fix here is either to add a toolchain specifier in `go.mod`, or to ensure that the build toolchain is 1.24.9 or later (in the 1.24 branch) — and that is already the case, the images provide a version of Go that has the fixes.
> 
> Thinking about this more, specifying a 1.24.9 isn’t even a correct fix; Shipyard uses 1.25 toolchains, so any version of those would be accepted, even though the fixes are only in 1.25.3 and later on the 1.25 branch.

Thanks, I've added https://github.com/submariner-io/submariner/pull/3678/commits/c0afdc1d06e9101cf7a754ac296c1e090a92e223 to the WIP CVE workflow to try to account for this.

PR #3716, opened 2025-11-19
Clean up orphan iptables chains post-migration to nftables

yboaron:
Introduces cleanup code to remove iptables chains when migrating from iptables to nftables packet filter driver.


tpantelis:
Rather than building infrastructure in the packetfilter and nftables code for a one-time, temporary cleanup migration, I think we should try to isolate the cleanup outside of the packetfilter if possible so we can easily remove it later. It looks like the cleanup consists of removing iptable chains - I assume these are well-known from prior versions that used iptables, so perhaps we can just remove those explicitly.

pkg/packetfilter/configure/configure.go
```diff
@@ -42,3 +42,7 @@ func DriverFromGlobalConfig() {
 		packetfilter.SetNewDriverFn(iptables.New)
 	}
 }
+
+func IPtablesCleanupIsNeeded() bool {
```
tpantelis: We  can make this more generic, ideally to query what driver is being used. It would be nice if we could check `packetfilter.GetNewDriverFn() == nftables.New` but Go doesn't allow that. Instead, we can move the [DriverType](https://github.com/submariner-io/submariner/blob/00184da5be3fa0df959292c5208f53bb3e3a9b3e/pkg/packetfilter/configure/configure_suite_test.go#L36) constants to  this file and add a `GetDriverType` function (that's also used by `DriverFromGlobalConfig`).

Then, to check if IP tables cleanup is needed, callers can check

```
if pfconfigure.GetDriverType() == pfconfigure.IPTables {
    return
}

// do cleanup
```

pkg/packetfilter/iptables_cleanup/iptables_cleanup.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for _, item := range regularChains {
+		_ = pIPtables.ClearChain(item.Table, item.Chain)
+
+		if err := pIPtables.DeleteChain(item.Table, item.Chain); err != nil {
+			logger.V(log.DEBUG).Infof("Failed to delete iptables chain %q: %v", item.Chain, err)
+		} else {
+			logger.Infof("Cleaned up iptables chain %q", item.Chain)
+		}
+	}
```
tpantelis: I don't think we really need this function. It's simply a wrapper for `DeleteChain`/`DeleteIPHookChain` - callers can simply call them directly (in fact, the RA doesn't pass any regular chains). I think it's cleaner that way.

On a side note, all call sites first call `ClearChain` before deleting. I know iptables fails if the chain isn't first cleared, not sure about nftables. We could simplify by having Delete* call `ClearChain` internally. But that can be done in another PR.

pkg/routeagent_driver/main.go
```diff
@@ -264,3 +268,96 @@ func removeInvalidSockets() {
...
+		},
+		{
+			Table: packetfilter.TableTypeFilter,
+			Chain: &packetfilter.ChainIPHook{
+				Name:     forwardingSubmarinerFWDChain,
+				Type:     packetfilter.ChainTypeFilter,
+				Hook:     packetfilter.ChainHookForward,
+				Priority: packetfilter.ChainPriorityFirst,
+			},
+		},
```
tpantelis: This is a duplicate of `constants.SmForwardChain`.

```suggestion
```

pkg/routeagent_driver/main.go
```diff
@@ -264,3 +268,96 @@ func removeInvalidSockets() {
...
+		{
+			Table: packetfilter.TableTypeFilter,
+			Chain: &packetfilter.ChainIPHook{
+				Name:     forwardingSubmarinerMSSClampChain,
```
```suggestion
				Name:     ovn.ForwardingSubmarinerMSSClampChain,
```

tpantelis (changes requested):

pkg/routeagent_driver/main.go
```diff
@@ -264,3 +268,96 @@ func removeInvalidSockets() {
...
+				Hook:     packetfilter.ChainHookPostrouting,
+				Priority: packetfilter.ChainPriorityFirst,
+			},
+		},
+	}
+
+	// Execute cleanup for each IP family in the cluster
+	for _, family := range cidr.ExtractIPFamilies(clusterCidr) {
+		iptables_cleanup.CleanupChains(ipHookChains, nil, family)
+	}
```
tpantelis, line 363: The MTU handler also has 2 [IP sets](https://github.com/submariner-io/submariner/blob/00184da5be3fa0df959292c5208f53bb3e3a9b3e/pkg/routeagent_driver/handlers/mtu/mtuhandler.go#L255).  Do we need to delete those as well?

tpantelis:
This handles migration from iptables to nftables but, theoretically, the reverse could happen as well, if someone had to switch back to iptables for some reason. Not sure if that is a realistic scenario that we should handle.

yboaron:
> This handles migration from iptables to nftables but, theoretically, the reverse could happen as well, if someone had to switch back to iptables for some reason. Not sure if that is a realistic scenario that we should handle.

Hmm, I don't think we should support nftables to iptables migration, as iptables is in maintenance mode and no longer the preferred packet filtering framework

pkg/routeagent_driver/main.go
yboaron: We can remove orphaned IP Sets as well (btw, GlobalNet also creates IPSets), but because we already delete orphaned iptables rules, any leftover IP Sets will no longer be pointed by any rules , so, their presence should not impact datapath. maybe we can handle it in another PR

tpantelis (approved):

vthapar (approved):

PR #3727, opened 2025-11-25
Refactor RouteAgent chain constants to chains package

tpantelis:
...and add functions to instantiate `ChainIPHooks` for reuse.

### DIFF
--- a/pkg/globalnet/controllers/gateway_monitor.go
+++ b/pkg/globalnet/controllers/gateway_monitor.go
@@ -34,6 +34,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/cni"
 	"github.com/submariner-io/submariner/pkg/event"
 	"github.com/submariner-io/submariner/pkg/event/controller"
+	"github.com/submariner-io/submariner/pkg/globalnet/chains"
 	"github.com/submariner-io/submariner/pkg/globalnet/constants"
 	gnpacketfilter "github.com/submariner-io/submariner/pkg/globalnet/controllers/packetfilter"
 	"github.com/submariner-io/submariner/pkg/globalnet/metrics"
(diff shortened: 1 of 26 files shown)


skitt (approved):

pkg/globalnet/controllers/uninstall.go
```diff
@@ -33,7 +33,7 @@ import (
 	"github.com/submariner-io/submariner/pkg/globalnet/constants"
 	pfiface "github.com/submariner-io/submariner/pkg/globalnet/controllers/packetfilter"
 	"github.com/submariner-io/submariner/pkg/packetfilter"
-	routeAgent "github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
+	routeAgent "github.com/submariner-io/submariner/pkg/routeagent_driver/chains"
```
yboaron: Can we import it as 'chains' ( "github.com/submariner-io/submariner/pkg/routeagent_driver/chains") in all files, for consistency?

pkg/globalnet/controllers/gateway_monitor.go
```diff
@@ -516,12 +516,7 @@ func (g *gatewayMonitor) createGlobalnetChains() error {
 			Hook:     packetfilter.ChainHookPrerouting,
 			Priority: packetfilter.ChainPriorityFirst,
```
yboaron: Maybe we should also consider updating **constants.SmGlobalnetIngressChain** GN IPHook ,

https://github.com/submariner-io/submariner/blob/devel/pkg/globalnet/controllers/gateway_monitor.go#L512

pkg/globalnet/controllers/uninstall.go
tpantelis: Renamed to `routeagentchains` so as not to conflict with GN chains.

pkg/globalnet/controllers/gateway_monitor.go
tpantelis: Refactored GN chains as well.

yboaron (approved):

PR #3728, opened 2025-11-25
Bump OVN-K

aswinsuryan:
Bump OVN-K


skitt (approved):

tpantelis (commented):
@aswinsuryan Can you outline/describe the changes associated with the bump in the commit message? Eg why do we no longer call `libovsdbops.CreateOrUpdateLogicalRouterPolicyWithPredicate`, `DeleteLogicalRouterStaticRoutesWithPredicate`, etc?

pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go
```diff
@@ -156,3 +176,219 @@ func buildLRPsFromSubnets(family k8snet.IPFamily, subnetsToAdd []string, nextHop
...
+		lrp.UUID = existingPolicies[0].UUID
+
+		updateOps, err := c.nbdb.Where(lrp).Update(lrp)
+		if err != nil {
+			return errors.Wrapf(err, "failed to create update ops for policy %q", lrp.UUID)
+		}
+
+		ops = append(ops, updateOps...)
+	} else {
+		lrp.UUID = "new_lrp"
```
tpantelis, line 222: This needs to be unique (same on L272). Internally OVSDB code uses [buildNamedUUID](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/575f5cc04674a6f90beda620621697a320eeedb4/go-controller/pkg/libovsdb/ops/named_uuid.go#L26) tho we can't access that. We could duplicate the code but not ideal. I don't know how important it is to follow _OVSDB rfc 7047 section 5.1_. We can also just use `NewUUID` from `k8s.io/apimachinery/pkg/util/uuid`.

pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go
```diff
@@ -84,6 +96,15 @@ func buildLRSRsFromSubnets(subnetsToAdd []string, nextHop string) []*nbdb.Logica
 }
 
 func (c *ConnectionHandler) reconcileSubOvnLogicalRouterPolicies(remoteSubnets sets.Set[string], nextHop string) error {
+	router, err := c.findLogicalRouter(OVNClusterRouter)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find ovn cluster router %q", OVNClusterRouter)
+	}
+
+	if router.UUID == "" {
```
tpantelis, line 104: Is this just defensive or is it actually possible for this to be empty (and have you observed it)?

pkg/routeagent_driver/handlers/ovn/fake/ovsdb_client.go
```diff
@@ -279,3 +279,39 @@ func (c *modelConditionalAPI) Delete() ([]ovsdb.Operation, error) {
 
 	return []ovsdb.Operation{}, nil
 }
+
+func (c *modelConditionalAPI) Update(m model.Model, _ ...any) ([]ovsdb.Operation, error) {
+	c.client.mutex.Lock()
+	defer c.client.mutex.Unlock()
+
+	uuid := getUUID(m)
+	if uuid != "" {
```
tpantelis, line 295: Perhaps we should fail if empty (ie `Expect(uuid).NotTo(BeEmpty())`)? 

I assume the real implementation also looks up by UUID? Perhaps we should also modify `Delete` to do so.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go
tpantelis: I added commit https://github.com/submariner-io/submariner/pull/3728/commits/980b262206267e1459a4f852a1d9d9eb9b51ca4e to adjust the relevant unit test so it now fails due to non-unique UUIDs. It didn't fail before due to timing and luck.

pkg/routeagent_driver/handlers/ovn/fake/ovsdb_client.go
```diff
@@ -279,3 +286,39 @@ func (c *modelConditionalAPI) Delete() ([]ovsdb.Operation, error) {
...
+
+		for i, item := range slice {
+			if getUUID(item) == uuid {
+				c.client.models[t][i] = m
+				return []ovsdb.Operation{}, nil
+			}
+		}
+	}
+
+	c.client.models[reflect.TypeOf(m)] = append(c.client.models[reflect.TypeOf(m)], m)
```
tpantelis, line 307: Since this function is Update I don't think we should add if not present.

dfarrell07:
I think we still need to do this, but we need someone else to drive. I don't have cycles at the moment but maybe I can make some skill to help with this when I do. @coderabbitai make an issue to track this and assign me.

PR #3810, opened 2026-02-02
Add P/Z builds for 0.19

dfarrell07:

## Summary by CodeRabbit

* **Chores**
  * Added support for IBM Power Systems (ppc64le) architecture across gateway and route-agent components with corresponding package dependencies and build configurations.
  * Added support for IBM Z Systems (s390x) architecture for route-agent with updated package repositories and build pipelines.
  * Removed legacy EUS repository configurations for s390x architecture.

### DIFF
--- a/.rpm-lockfiles/gateway/rpms.in.yaml
+++ b/.rpm-lockfiles/gateway/rpms.in.yaml
@@ -13,4 +13,5 @@ packages:
 arches:
   - x86_64
   - aarch64
+  - ppc64le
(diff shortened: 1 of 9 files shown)


dfarrell07:
All 4 arches are building for all components. I also updated the activation-key in the Konflux cluster per new docs I've added in https://github.com/submariner-io/submariner/pull/3811.

tpantelis (approved):

PR #3821, opened 2026-02-09
Workaround nfatbles incompatibility with fedroa43

yboaron:


skitt:
Great investigation, thanks! So this will be fixed in the next release of kind images since https://github.com/kubernetes-sigs/kind/pull/4103 was merged there.

yboaron:
So, Kindnet with kube-proxy (nftables) gates are green with the workaround,
OVNK gates still failing because ovnk pod fail to configure SNAT exclusion rule (e2e tests fails)

After upgrading submariner image to F43:
* submariner use nftables v1.1.3
* kube-proxy uses v1.0.6 nft  getting segfaults ERR when trying to read v1.1.3  
* OVNK also uses v1.0.6 nft - getting the same ERR

This workaround upgrades KIND nodes and patches kube-proxy to use host's nftables    
  v1.1.6, but OVN-K pods still use v1.0.6 nft because we pin to this
   [OVN-K version](https://github.com/submariner-io/shipyard/blob/c09bab892860496de80592f42dc0104b188f7155/scripts/shared/lib/clusters_kind#L305)  in Shipyard, so , we need to update also ovnk image in Shipyard.


I wonder if we might hit backward compatibility issue , when Submariner with Fedora 43 based images (nft v1.1.3) may be incompatible with clusters running older OVN-K  versions that use nftables v1.0.6.  WDYT?

yboaron:
Regarding nftables version used in OCP/ovnk, 
OCP 4.20 ,OCP 4.21  and current OCP 4.22 (nightly) use Fedroa42 (nft 1.1.1-3) which should be compatible with Submariner Fedora43 based image

dfarrell07:
Yossi tells me we can close this as he's working on fixes in other PRs. Thanks for digging into it!

PR #3851, opened 2026-02-19
Fix route agent to watch RouteAgentComponent config

tpantelis:
The route agent was incorrectly watching `GatewayComponent` configuration changes. This updates it to watch the correct `RouteAgentComponent` configuration for its own config updates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected signal monitoring to target the appropriate component for graceful shutdown handling.


skitt (approved):
This should be backported to 0.23 and 0.22.

PR #3881, opened 2026-03-09
Fix deletion of non-Submariner OVN routes and policies during reconciliation

### LINKED ISSUE #3880: [OVN] Route-agent deletes OVN-Kubernetes managed routes when remote cluster joins
When a second cluster joins a Submariner network using OVN-Kubernetes CNI, the route-agent incorrectly deletes the local cluster subnet route from `ovn_cluster_router`, breaking pod connectivity to the API server.

**What happened**:
Local cluster subnet route got deleted, breaking pod connectivity within the node.

Local cluster subnet route exists (`10.42.0.0/24 → 10.42.0.2 src-ip`)

1. Deploy two clusters with OVN-Kubernetes
2. Deploy Submariner broker on cluster A
3. Join cluster A to Submariner
4. Join cluster B to Submariner
5. On cluster A, check routes: `ovn-nbctl lr-route-list ovn_cluster_router`

**Environment**:
- Submariner version: 0.22
- CNI: OVN-Kubernetes
- Platform: MicroShift

pacevedom:
The staleLRSRPred predicate in reconcileOvnLogicalRouterStaticRoutes was deleting routes based only on nexthop and prefix, without checking if the route was created by Submariner. This caused OVN-Kubernetes managed routes (like the local cluster subnet route) to be deleted when a remote cluster joined, breaking pod connectivity. Add a check for the "submariner" key in ExternalIDs to ensure only Submariner-owned routes are considered for deletion.
The same fix also applies to ovn policies.

Closes https://github.com/submariner-io/submariner/issues/3880

## Summary by CodeRabbit

* **Bug Fixes**
  * Reconciliation now only removes routes and policies created by the Submariner component, preserving non-Submariner static routes and policies that share the same gateway or priority.

* **Tests**
  * Added tests verifying non-Submariner routes and policies persist through reconciliation and gateway removal.
  * Improved test utilities to more reliably and repeatedly assert the presence of expected OVN models.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/fake/ovsdb_client.go
+++ b/pkg/routeagent_driver/handlers/ovn/fake/ovsdb_client.go
@@ -189,11 +189,39 @@ func (c *OVSDBClient) AwaitNoModel(m any) {
 	}).Should(BeFalse(), "OVSBD model exists: %s", resource.ToJSON(m))
 }
 
+func (c *OVSDBClient) EnsureModel(m any) {
+	Consistently(func(g Gomega) {
+		found, existing := c.hasModel(m)
+		g.Expect(found).To(BeTrue(), "OVSDB model not found: %s. Actual models: %v", resource.ToJSON(m), existing)
+	}).Should(Succeed())
+}
+
 func (c *OVSDBClient) EnsureNoModel(m any) {
 	Consistently(func() bool {
 		found, _ := c.hasModel(m)
 		return found
 	}).Should(BeFalse(), "OVSBD model exists: %s", resource.ToJSON(m))
+}
+
+func (c *OVSDBClient) GetModel(m any) any {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	for _, o := range c.models[reflect.TypeOf(m)] {
+		switch t := m.(type) {
+		case *nbdb.LogicalRouterPolicy:
+			if strings.Contains(o.(*nbdb.LogicalRouterPolicy).Match, t.Match) &&
+				reflect.DeepEqual(o.(*nbdb.LogicalRouterPolicy).Nexthop, t.Nexthop) {
+				return o
+			}
+		case *nbdb.LogicalRouterStaticRoute:
+			if o.(*nbdb.LogicalRouterStaticRoute).IPPrefix == t.IPPrefix {
+				return o
+			}
+		}
+	}
+
+	return nil
 }
 
 func (c *OVSDBClient) EnsureNoModelsOfType(m any) {
--- a/pkg/routeagent_driver/handlers/ovn/handler_test.go
+++ b/pkg/routeagent_driver/handlers/ovn/handler_test.go
@@ -464,6 +464,140 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
 			})
 		})
 	})
+
+	When("a non-Submariner route exists with the same nexthop", func() {
+		It("should not delete the non-Submariner route during reconciliation", func() {
+			client := t.dynClient.Resource(submarinerv1.SchemeGroupVersion.WithResource("gatewayroutes")).Namespace(testing.Namespace)
+
+			// Create a route that simulates an OVN-K managed route (no submariner external_id)
+			// This uses the same nexthop as Submariner but has a different prefix (local cluster subnet)
+			ovnkRoute := &nbdb.LogicalRouterStaticRoute{
+				IPPrefix:    t.clusterCIDR,
+				Nexthop:     t.OVNK8sMgmntIntCIDR[t.ipFamily].IP.String(),
+				ExternalIDs: map[string]string{},
+			}
+
+			_, err := t.ovsdbClient.Create(ovnkRoute)
+			Expect(err).To(Succeed())
+
+			// Create a GatewayRoute which will trigger reconciliation
+			gwRoute := &submarinerv1.GatewayRoute{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-gateway-route-preserve",
+				},
+				RoutePolicySpec: submarinerv1.RoutePolicySpec{
+					NextHops:    []string{ovnkRoute.Nexthop},
+					RemoteCIDRs: ipFamilySubnets,
+				},
+			}
+
+			test.CreateResource(client, gwRoute)
+
+			// Wait for Submariner routes to be reconciled.
+			for _, cidr := range gwRoute.RoutePolicySpec.RemoteCIDRs {
+				t.ovsdbClient.AwaitModel(&nbdb.LogicalRouterStaticRoute{
+					IPPrefix: cidr,
+				})
+			}
+
+			// Verify the OVN-K route still exists and is untagged
+			t.ovsdbClient.EnsureModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
+			})
+			retrievedRoute := t.ovsdbClient.GetModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
+			}).(*nbdb.LogicalRouterStaticRoute)
+			Expect(retrievedRoute.ExternalIDs).ToNot(HaveKey(ovn.SubmarinerExternalIDKey),
+				"OVN-K route should not be tagged with submariner")
+
+			Expect(client.Delete(context.Background(), gwRoute.Name, metav1.DeleteOptions{})).To(Succeed())
+
+			// Check the OVN-K route still exists and remains untagged after removing the gateway
+			t.ovsdbClient.EnsureModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
+			})
+			retrievedRoute = t.ovsdbClient.GetModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
+			}).(*nbdb.LogicalRouterStaticRoute)
+			Expect(retrievedRoute.ExternalIDs).ToNot(HaveKey(ovn.SubmarinerExternalIDKey),
+				"OVN-K route should not be tagged with submariner after cleanup")
+		})
+	})
+
+	When("a non-Submariner policy exists with the same priority", func() {
+		It("should not delete the non-Submariner policy during reconciliation", func() {
+			client := t.dynClient.Resource(submarinerv1.SchemeGroupVersion.WithResource("gatewayroutes")).Namespace(testing.Namespace)
+
+			// Determine priority and match field based on IP family
+			priority := 20000
+			ipMatchField := "ip4.dst"
+
+			if t.ipFamily == k8snet.IPv6 {
+				priority = 20100
+				ipMatchField = "ip6.dst"
+			}
+
+			// Create a policy that simulates an OVN-K managed policy (no submariner external_id)
+			// This uses the same priority as Submariner but has a different match
+			ovnkPolicy := &nbdb.LogicalRouterPolicy{
+				Priority:    priority,
+				Match:       ipMatchField + " == " + t.clusterCIDR,
+				Action:      "reroute",
+				Nexthop:     ptr.To(t.OVNK8sMgmntIntCIDR[t.ipFamily].IP.String()),
+				ExternalIDs: map[string]string{}, // No submariner tag
+			}
+
+			_, err := t.ovsdbClient.Create(ovnkPolicy)
+			Expect(err).To(Succeed())
+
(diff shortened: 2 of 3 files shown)


pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go
```diff
@@ -41,6 +41,11 @@ func (c *ConnectionHandler) reconcileOvnLogicalRouterStaticRoutes(remoteSubnets
 	nextHop string,
 ) error {
 	staleLRSRPred := func(item *nbdb.LogicalRouterStaticRoute) bool {
+		// Only manage routes that Submariner created, ignore rest of the system's routes.
+		if _, ok := item.ExternalIDs["submariner"]; !ok {
```
tpantelis: Let's make this a constant now that we're referencing it in 2 locations, ie:

`const submarinerExternalIDKey = "submariner"`

pkg/routeagent_driver/handlers/ovn/handler_test.go
```diff
@@ -462,6 +462,54 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
...
+				Nexthop:     t.OVNK8sMgmntIntCIDR[t.ipFamily].IP.String(),
+				ExternalIDs: map[string]string{},
+			}
+
+			_, err := t.ovsdbClient.Create(ovnkRoute)
+			Expect(err).To(Succeed())
+
+			t.ovsdbClient.AwaitModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: "10.42.0.0/24",
+			})
```
tpantelis: We don't need to await the model here since the test just created it above.


```suggestion
```

pkg/routeagent_driver/handlers/ovn/handler_test.go
```diff
@@ -462,6 +462,54 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
...
+
+			// Create a GatewayRoute which will trigger reconciliation
+			gwRoute := &submarinerv1.GatewayRoute{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-gateway-route-preserve",
+				},
+				RoutePolicySpec: submarinerv1.RoutePolicySpec{
+					NextHops:    []string{t.OVNK8sMgmntIntCIDR[t.ipFamily].IP.String()},
```
tpantelis:
```suggestion
					NextHops:    []string{ovnkRoute.Nexthop},
```

pkg/routeagent_driver/handlers/ovn/handler_test.go
```diff
@@ -462,6 +462,54 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
...
+			for _, cidr := range gwRoute.RoutePolicySpec.RemoteCIDRs {
+				t.ovsdbClient.AwaitModel(&nbdb.LogicalRouterStaticRoute{
+					IPPrefix: cidr,
+				})
+			}
+
+			// Verify the OVN-K route still exists
+			t.ovsdbClient.AwaitModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: "10.42.0.0/24",
+			})
```
tpantelis: We already know the model previously existed so we really want to ensure it still exists. I suggest adding an `EnsureModel` function:

```
func (c *OVSDBClient) EnsureModel(m any) {
	Consistently(func(g Gomega) {
		found, existing := c.hasModel(m)
		g.Expect(found).To(BeTrue(), "OVSBD model not found: %s. Actual models: %v", resource.ToJSON(m), existing)
	}).Should(Succeed())
}
```


```suggestion
			t.ovsdbClient.EnsureModel(&nbdb.LogicalRouterStaticRoute{
				IPPrefix: ovnkRoute.IPPrefix,
			})
```

pkg/routeagent_driver/handlers/ovn/handler_test.go
```diff
@@ -462,6 +462,54 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
...
+
+			Expect(client.Delete(context.Background(), gwRoute.Name, metav1.DeleteOptions{})).To(Succeed())
```
tpantelis, line 513: Just deleting the `gwRoute` doesn't really serve any purpose (perhaps a copy/paste remnant?). To be complete, we can verify the OVN-K route still exists via `EnsureModel`, as above.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/handler_test.go
pacevedom: Done

pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go
pacevedom: Done

pkg/routeagent_driver/handlers/ovn/handler_test.go
pacevedom: Done

pkg/routeagent_driver/handlers/ovn/handler_test.go
pacevedom: Done

pkg/routeagent_driver/handlers/ovn/handler_test.go
pacevedom: Ah, this is just the cleanup part to remove the gateway to stop reconciliations. It is true that its not strictly needed because the client is recreated for each test, but I added it for consistency.

tpantelis: yeah deleting it as cleanup isn't necessary but it's useful when  coupled with verification as I mentioned above.

pkg/routeagent_driver/handlers/ovn/handler_test.go
```diff
@@ -462,6 +462,50 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
...
+	When("a non-Submariner route exists with the same nexthop", func() {
+		It("should not delete the non-Submariner route during reconciliation", func() {
+			client := t.dynClient.Resource(submarinerv1.SchemeGroupVersion.WithResource("gatewayroutes")).Namespace(testing.Namespace)
+
+			// Create a route that simulates an OVN-K managed route (no submariner external_id)
+			// This uses the same nexthop as Submariner but has a different prefix (local cluster subnet)
+			ovnkRoute := &nbdb.LogicalRouterStaticRoute{
+				IPPrefix:    "10.42.0.0/24",
+				Nexthop:     t.OVNK8sMgmntIntCIDR[t.ipFamily].IP.String(),
+				ExternalIDs: map[string]string{},
```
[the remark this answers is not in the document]
tpantelis: Technically, CodeRabbit is correct here, ie this function is called for IPv4 and IPv6.  We could use `t.clusterCIDR` for `IPPrefix` for consistency, even though it doesn't really matter.

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/handler_test.go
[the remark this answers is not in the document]
pacevedom: Done

pkg/routeagent_driver/handlers/ovn/handler_test.go
pacevedom: Done

tpantelis:
Should we apply the same logic for deleting non-static logical router policies as noted by CodeRabbit in [this nitpick comment](https://github.com/submariner-io/submariner/pull/3881#pullrequestreview-3920599750)?

pacevedom:
> Should we apply the same logic for deleting non-static logical router policies as noted by CodeRabbit in [this nitpick comment](https://github.com/submariner-io/submariner/pull/3881#pullrequestreview-3920599750)?

Added the changes in a new commit. It is protected by priority for each policy, but it could still happen (although less likely than the routes).

tpantelis:
> Added the changes in a new commit. 

That's fine for now but let's squash them before merging.

Also, please address Code Rabbit's comments, if needed.

pkg/routeagent_driver/handlers/ovn/handler_test.go
```diff
@@ -462,6 +462,117 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
...
+			// Verify the OVN-K route still exists
+			t.ovsdbClient.EnsureModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
+			})
+
+			Expect(client.Delete(context.Background(), gwRoute.Name, metav1.DeleteOptions{})).To(Succeed())
+
+			// Check the OVN-K route still exists even after removing the gateway
+			t.ovsdbClient.EnsureModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
```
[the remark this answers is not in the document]
pacevedom: Done

pkg/routeagent_driver/handlers/ovn/ovn_logical_routes.go
```diff
@@ -32,15 +32,21 @@ import (
...
 )
 
 func (c *ConnectionHandler) reconcileOvnLogicalRouterStaticRoutes(remoteSubnets sets.Set[string],
 	nextHop string,
 ) error {
 	staleLRSRPred := func(item *nbdb.LogicalRouterStaticRoute) bool {
+		// Only manage routes that Submariner created, ignore rest of the system's routes.
+		if _, ok := item.ExternalIDs[submarinerExternalIDKey]; !ok {
+			return false
+		}
```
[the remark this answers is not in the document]
pacevedom: The upsert predicate collision requires overlapping CIDRs between clusters, which is:
- Detected and logged at handler.go:201-206
- Unsupported without Globalnet (would break routing)
- Not exercised by the test - OVN-K route uses clusterCIDR while Submariner routes use remoteSubnets (different prefixes, so upsert predicate never matches)

Adding ownership checks to upsert predicates would only protect against a misconfiguration that already breaks Submariner.

pacevedom:
> > Added the changes in a new commit.
> 
> That's fine for now but let's squash them before merging.
> 
> Also, please address Code Rabbit's comments, if needed.

Ready again.

pkg/routeagent_driver/handlers/ovn/fake/ovsdb_client.go
```diff
@@ -189,13 +189,41 @@ func (c *OVSDBClient) AwaitNoModel(m any) {
...
+}
+
 func (c *OVSDBClient) EnsureNoModel(m any) {
 	Consistently(func() bool {
 		found, _ := c.hasModel(m)
 		return found
 	}).Should(BeFalse(), "OVSBD model exists: %s", resource.ToJSON(m))
 }
 
+func (c *OVSDBClient) GetModel(m any) any {
```
tpantelis, line 206: This is essentially a duplicate of `hasModel`. How about we remove `hasModel` and change call sites to `GetModel` and check for nil?

pkg/routeagent_driver/handlers/ovn/handler_test.go
```diff
@@ -462,6 +462,135 @@ func (t *handlerTestDriver) testGatewayRoute(ipFamilySubnets []string, nonIPFami
...
+			}
+
+			// Verify the OVN-K route still exists and is untagged
+			t.ovsdbClient.EnsureModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
+			})
+			retrievedRoute := t.ovsdbClient.GetModel(&nbdb.LogicalRouterStaticRoute{
+				IPPrefix: ovnkRoute.IPPrefix,
+			}).(*nbdb.LogicalRouterStaticRoute)
+			Expect(retrievedRoute.ExternalIDs).ToNot(HaveKey("submariner"), "OVN-K route should not be tagged with submariner")
```
tpantelis: Let's make `submarinerExternalIDKey` public and use that here (and elsewhere).


```suggestion
			Expect(retrievedRoute.ExternalIDs).ToNot(HaveKey(ovn.SubmarinerExternalIDKey), "OVN-K route should not be tagged with submariner")
```

tpantelis (changes requested):

pkg/routeagent_driver/handlers/ovn/fake/ovsdb_client.go
tpantelis: Actually hasModel also returned the existing ones if not found.

tpantelis:
@coderabbitai approve

tpantelis (approved):

aswinsuryan (approved):

PR #3890, opened 2026-03-12
Add cleanup function to clusterfiles.Get

tpantelis:
The `Get` function now returns a cleanup function that removes temporary directories created for `secret://` and `configmap://` URLs. For `file://`URLs, the cleanup function is a no-op since no temporary files are created.

This prevents disk space leaks from accumulating temporary directories, especially in tests and when pods restart frequently.

Also handle `file://` URLs before path validation to fix rejection of single-component paths like `file:///hosts`. The validation that splits the path into container and object is only needed for `secret://` and `configmap://` schemes.

Additionally, reject malformed `file://` URLs with a host component (e.g., `file://tmp/cert`). The documented format is `file:///<path>` with three slashes, which produces an empty Host. URLs with two slashes are parsed as having a host, leading to incorrect paths.

Also fix `ConfigMap` error message to avoid logging sensitive data. The `%#v` format verb would dump the entire `ConfigMap` structure including `Data` and `BinaryData` fields. Use only the name and namespace instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Temporary files and certificate artifacts are now reliably cleaned up after retrieval, preventing resource leaks.

* **Breaking Change**
  * Retrieval API now returns an additional cleanup callback; callers must invoke or defer it to release temporary resources (local-file cases provide a no-op cleanup).

* **Tests**
  * Tests updated to exercise and verify cleanup behavior and ensure temporary resources are removed.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/connection.go
+++ b/pkg/routeagent_driver/handlers/ovn/connection.go
@@ -167,26 +167,29 @@ func (c *ConnectionHandler) createLibovsdbClient(ctx context.Context, dbModel mo
 	return client, nil
 }
 
-func getFile(ctx context.Context, k8sClientset clientset.Interface, url string) (string, error) {
-	file, err := clusterfiles.Get(ctx, k8sClientset, url)
-	return file, errors.Wrapf(err, "error getting config file for %q", url)
+func getFile(ctx context.Context, k8sClientset clientset.Interface, url string) (string, func(), error) {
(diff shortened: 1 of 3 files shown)


tpantelis:
@coderabbitai approve

skitt:
I suggest #3894 instead.

tpantelis:
> I suggest #3894 instead.

Good idea!

PR #3894, opened 2026-03-13
Return OVN connection data directly

skitt:
Instead of storing data in temporary files (which then need to be cleaned up), return the relevant data directly.

## Summary by CodeRabbit

* **Refactor**
  * Switched TLS/certificate handling to in-memory byte data and removed local disk I/O for cluster files. CA bytes are appended to the trust pool. File URI handling now validates hosts and returns clearer errors. Cluster file retrieval returns raw bytes instead of file paths.

* **Tests**
  * Updated tests to assert returned byte content directly for secret, configmap, and file schemes.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/connection.go
+++ b/pkg/routeagent_driver/handlers/ovn/connection.go
@@ -23,7 +23,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"os"
 	"strings"
 
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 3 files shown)


pkg/util/clusterfiles/cluster_files.go
```diff
@@ -36,86 +36,64 @@ var logger = log.Logger{Logger: logf.Log.WithName("ClusterFiles")}
...
-
 	switch parsedURL.Scheme {
 	case "file":
-		return parsedURL.Path, nil
+		data, err := os.ReadFile(parsedURL.Path)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error reading file %q", parsedURL.Path)
+		}
+
+		return data, nil
```
tpantelis, line 67: Note: In my PR, Code Rabbit flagged this unrelated issue https://github.com/submariner-io/submariner/pull/3890#discussion_r2927698243 and then a follow-up https://github.com/submariner-io/submariner/pull/3890#discussion_r2927846718.

pkg/routeagent_driver/handlers/ovn/connection.go
tpantelis: For consistency...


```suggestion
	tlsConfig, err := getOVNTLSConfig(pkData, certData, caData)
```

pkg/routeagent_driver/handlers/ovn/connection.go
```diff
@@ -167,7 +163,7 @@ func (c *ConnectionHandler) createLibovsdbClient(ctx context.Context, dbModel mo
 	return client, nil
 }
 
-func getFile(ctx context.Context, k8sClientset clientset.Interface, url string) (string, error) {
+func getFile(ctx context.Context, k8sClientset clientset.Interface, url string) ([]byte, error) {
```
tpantelis: For naming consistency...


```suggestion
func getData(ctx context.Context, k8sClientset clientset.Interface, url string) ([]byte, error) {
```

pkg/util/clusterfiles/cluster_files.go
skitt: Good point, thanks, I’ve fixed that.

tpantelis (approved):

PR #3898, opened 2026-03-13
Fix verify-packages.sh to dynamically read repo filenames

dfarrell07:
The script was hardcoded to fetch "submariner-rhel-10.repo" from git, which silently failed (|| true) when verifying older branches that use submariner-rhel-9.repo. This caused the script to verify packages against the wrong RHEL version.

Changes:
- Extract repo filename from rpms.in.yaml for each component
- Remove || true to fail fast when repo file is missing
- Makes the script version-agnostic (works with RHEL 9 and 10)

Verified:
- release-0.22 (RHEL 9) correctly shows rhel-9 repos
- release-0.23 (RHEL 10) correctly shows rhel-10 repos

## Summary by CodeRabbit

* **Chores**
  * Updated package verification to dynamically determine and fetch repository files per component based on configuration rather than a fixed default.
  * Added a preflight check to ensure the required CLI tool is available before running verification, improving reliability and clearer failures.

### DIFF
--- a/.rpm-lockfiles/verify-packages.sh
+++ b/.rpm-lockfiles/verify-packages.sh
@@ -52,7 +52,12 @@ if [[ -n "$BRANCH" ]]; then
     for comp in gateway route-agent globalnet; do
         mkdir -p "$LOCKFILES_DIR/$comp"
         git show "$GIT_REF:.rpm-lockfiles/$comp/rpms.in.yaml" > "$LOCKFILES_DIR/$comp/rpms.in.yaml" 2>/dev/null || continue
-        git show "$GIT_REF:.rpm-lockfiles/$comp/submariner-rhel-10.repo" > "$LOCKFILES_DIR/$comp/submariner-rhel-10.repo" 2>/dev/null || true
+
+        # Extract repo filename from rpms.in.yaml
+        repo_file=$(grep -A1 "repofiles:" "$LOCKFILES_DIR/$comp/rpms.in.yaml" | grep "^ *-" | sed 's/.*- //' | head -1)
+        if [[ -n "$repo_file" ]]; then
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

.rpm-lockfiles/verify-packages.sh
```diff
@@ -52,7 +52,12 @@ if [[ -n "$BRANCH" ]]; then
     for comp in gateway route-agent globalnet; do
         mkdir -p "$LOCKFILES_DIR/$comp"
         git show "$GIT_REF:.rpm-lockfiles/$comp/rpms.in.yaml" > "$LOCKFILES_DIR/$comp/rpms.in.yaml" 2>/dev/null || continue
-        git show "$GIT_REF:.rpm-lockfiles/$comp/submariner-rhel-10.repo" > "$LOCKFILES_DIR/$comp/submariner-rhel-10.repo" 2>/dev/null || true
+
+        # Extract repo filename from rpms.in.yaml
+        repo_file=$(grep -A1 "repofiles:" "$LOCKFILES_DIR/$comp/rpms.in.yaml" | grep "^ *-" | sed 's/.*- //' | head -1)
```
skitt, line 57: Can we rely on `yq` here?


```suggestion
        repo_file=$(yq '.contentOrigin.repofiles[0]' "$LOCKFILES_DIR/$comp/rpms.in.yaml")
```

skitt (changes requested):

.rpm-lockfiles/verify-packages.sh
dfarrell07: Ok, changed to yq in commit on top.

skitt (approved):

PR #3929, opened 2026-03-25
Retry indefinitely on Gateway pod HA status patch errors

tpantelis:
The previous approach of categorizing errors as "retryable" was fragile and incomplete. In a production env, it was observed that the `Patch` operation failed with error

  "_etcdserver: request timed out, possibly due to previous leader failure_"

However, this didn't trigger a retry attempt b/c it was likely either an `InternalError` (500) or `ServiceUnavailable` (503) which `IsTransientErr()` doesn't check for.

This commit simplifies the logic to retry on all `Patch` errors, relying on the caller's context to either time out or cancel the operation.

Additionally, switched from fixed 100ms polling to exponential backoff (100ms → 5s cap) to reduce API server load while maintaining fast recovery for transient errors. Retries continue indefinitely until the patch succeeds or the context is cancelled/times out.

Consolidated all `SetHALabels` calls in _gateway.go_ to use `updateGatewayHAStatus`, centralizing error handling in one place.

This handles all transient errors robustly: etcd timeouts, conflicts, network errors, and any other temporary failures.

## Summary by CodeRabbit

* **Refactor**
  * Retry behavior for HA label updates moved to capped exponential backoff, improving reliability and avoiding tight retry loops.
  * Shutdown and fatal-error paths unified to consistently attempt HA label updates and produce clearer error logs (no longer suppressing cancellation errors).

* **Tests**
  * Tests updated to be context-aware with timeouts and consolidated error scenarios to validate backoff, retries, and eventual recovery.

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
-	"github.com/submariner-io/admiral/pkg/resource"
 	submV1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -80,7 +79,11 @@ func NewGatewayPod(ctx context.Context, k8sClient kubernetes.Interface) (*Gatewa
 	return gp, nil
 }
 
-var patchFormat = fmt.Sprintf(`{"metadata": {"labels": {%q: "%%s", %q: "%%s"}}}`, GatewayNodeLabel, GatewayStatusLabel)
+var (
+	BackOffCap = 5 * time.Second
+
+	patchFormat = fmt.Sprintf(`{"metadata": {"labels": {%q: "%%s", %q: "%%s"}}}`, GatewayNodeLabel, GatewayStatusLabel)
+)
 
 func (gp *GatewayPod) SetHALabels(ctx context.Context, status submV1.HAStatus) error {
 	logger.Infof("Updating Gateway pod HA status to %q", status)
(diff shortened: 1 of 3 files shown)


pkg/pod/pod_test.go
```diff
@@ -196,38 +194,38 @@ func testSetHALabels() {
...
 
 			t.client.Fake.PrependReactor("patch", "pods", func(action k8stesting.Action) (bool, runtime.Object, error) {
-				if attemptCount < len(errorMessages) {
-					err := apierrors.NewTooManyRequests(errorMessages[attemptCount], 1)
-					attemptCount++
-
-					return true, nil, err
+				// Fail for the first 1 second, then succeed
+				if time.Since(startTime) < time.Second {
+					return true, nil, errors.New("mock error")
```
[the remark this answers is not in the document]
tpantelis: The backoff `Cap` of 500 ms  causes `ExponentialBackoffWithContext` to reset `Steps` to 0 after 500 ms which causes it to return. Plus running the test shows the retry warning logged twice  and code coverage report shows all lines covered.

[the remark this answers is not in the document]
tpantelis: Please resolve and approve.

tpantelis:
@coderabbitai approve

tpantelis:
@coderabbitai approve

pkg/pod/pod.go
```diff
@@ -80,38 +79,64 @@ func NewGatewayPod(ctx context.Context, k8sClient kubernetes.Interface) (*Gatewa
...
 			}
 
-			return false, errors.Wrapf(err, "error patching own pod %q in namespace %q with %s", gp.name, gp.namespace, patch)
+			if err == nil {
+				return true, nil
+			}
+
+			if lastError == nil {
+				logRetryWarning(err)
+			}
```
skitt: I’m not sure it matters in practice, but if different errors occur, only the first and last one will be logged — it might be better to log if `err` isn’t the same as `lastError` (and possibly remove the log further down).

pkg/pod/pod_test.go
```diff
@@ -57,7 +59,7 @@ func testNewGatewayPod() {
 
 			Expect(err).NotTo(HaveOccurred())
 			Expect(gw).ToNot(BeNil())
-			t.verifyPodLabels(submarinerv1.HAStatusPassive)
+			t.verifyPodLabels(context.TODO(), submarinerv1.HAStatusPassive)
```
skitt: How about using the Gingko context here and in line 58?

skitt (changes requested):

pkg/pod/pod.go
tpantelis: That's the way it was originally but most likely it would be the same error repeated and it wouldn't indicate that the error keeps recurring. We're blocking indefinitely so I think it's better that we periodically log at least the last error rather than only logging the first error and then silence after that.

tpantelis: I changed it to also log if `err` differs from `lastError` so we get both behaviors.

skitt (approved):

PR #3954, opened 2026-04-07
Refactor HostInterface to use net.IP instead of string

tpantelis:
Change the `HostInterface.Addr` field from string to `net.IP` since callers only use the IP and not the subnet mask. This also eliminates redundant CIDR parsing across multiple callers and improves type safety.

## Summary by CodeRabbit

* **Refactor**
  * Host interface addresses now use native IP values, resulting in more consistent and reliable network-address handling across the product.

* **Tests**
  * Test fixtures updated to reflect the new address representation, improving test consistency and reducing parsing-related noise.

### DIFF
--- a/pkg/cni/cni_iface.go
+++ b/pkg/cni/cni_iface.go
@@ -35,7 +35,7 @@ type Interface struct {
 
 type HostInterface struct {
 	Name string
-	Addr string
+	Addr net.IP
 }
 
 var logger = log.Logger{Logger: logf.Log.WithName("CNI")}
@@ -55,7 +55,12 @@ var HostInterfaces = func() ([]HostInterface, error) {
 		}
 
 		for _, a := range addrs {
-			hostInterfaces = append(hostInterfaces, HostInterface{Name: netInterfaces[i].Name, Addr: a.String()})
+			ipNet, ok := a.(*net.IPNet)
+			if !ok {
+				continue
+			}
+
+			hostInterfaces = append(hostInterfaces, HostInterface{Name: netInterfaces[i].Name, Addr: ipNet.IP})
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/cidr/iputil.go
```diff
@@ -25,12 +25,27 @@ import (
...
+// MustParse parses a CIDR string and panics if it fails.
+// This is intended for use with hardcoded CIDR strings in tests and
+// configuration where parsing failure indicates a programming error.
+// Unlike net.ParseCIDR, this preserves the actual IP address from the CIDR string
+// in the returned IPNet.IP field, rather than the network base address.
+func MustParse(cidr string) *net.IPNet {
+	ip, ipNet, err := net.ParseCIDR(cidr)
+	utilruntime.Must(errors.Wrapf(err, "unable to parse CIDR %q", cidr))
+
+	ipNet.IP = ip
```
skitt: This bothers me; repurposing `net.IPNet` to store a host address and subnet size combo isn’t great IMO (even though the kernel does something similar, which is the whole reason `net.Interface.Addrs()` returns a string combining a host address with a subnet size).

Looking at the callers, I don’t see any actual use of the subnet size part of the address, outside of test code. Given that, wouldn’t it be cleaner to use `net.IP` instead?

skitt (changes requested):

pkg/cidr/iputil.go
skitt: I went and implemented this, LMK if it’s OK for you.

tpantelis: yeah you're right - callers only use the IP and not the subnet part. Then we don't need `cidr.MustParse`.

(BTW - this is an example where we can't just rely on AI code review agents as Code Rabbit approved it as is. 😄 )

skitt (approved):

pkg/cni/cni_iface.go
```diff
@@ -55,7 +55,12 @@ var HostInterfaces = func() ([]HostInterface, error) {
 		}
 
 		for _, a := range addrs {
-			hostInterfaces = append(hostInterfaces, HostInterface{Name: netInterfaces[i].Name, Addr: a.String()})
+			ipNet, ok := a.(*net.IPNet)
+			if !ok {
+				continue
+			}
+
+			hostInterfaces = append(hostInterfaces, HostInterface{Name: netInterfaces[i].Name, Addr: ipNet.IP})
```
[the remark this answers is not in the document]
tpantelis: This code only runs on Linux deployments which only returns `*net.IPNet`. Also the previous code using strings only handled `net.IPNet` as well (ie callers used `net.ParseCIDR`) so there's no regression.

pkg/cidr/iputil.go
tpantelis: I did it as well.

tpantelis: Sorry it looks like I overwrote your changes - I didn't realized you had pushed a commit.

tpantelis:
@coderabbitai approve

PR #3966, opened 2026-04-12
Fix VXLAN cleanup during changing cable driver

yboaron:
When changing cable driver from libreswan to vxlan, a race condition causes the route agent to delete the newly created vxlan-tunnel interface.

Actual timeline during cable driver change (from libreswan to vxlan):
1. New GW pod restarts with new cable driver (vxlan)
2. Cable engine creates vxlan-tunnel interface
3. Datastore syncer creates new vxlan endpoint
4. Datastore syncer deletes old libreswan endpoint (although it is being called before step 3, async)
5. Route agent receives LocalEndpointRemoved event
6. Route agent calls TransitionToNonGateway
7. VXLAN cleanup handler deletes vxlan-tunnel interface

The issue is that the old endpoint deletion (step 4) happens asynchronously after the new vxlan interface is already created, triggering cleanup that removes the wrong interface.

This PR checks the removed endpoint's cable driver in the VxLAN cleanup handler and only run cleanup if the removed endpoint was using vxlan.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VXLAN tunnel cleanup to better coordinate state when local endpoints are removed, preventing incorrect teardown.
  * Guarded gateway→non-gateway transitions so VXLAN route/interface removal is skipped when the most recently removed endpoint indicates a non‑VXLAN backend.
  * Added a callback to record the last-removed local endpoint for use during cleanup decisions.


pkg/routeagent_driver/cabledriver/vxlan_tunnel_cleanup_handler.go
```diff
@@ -31,6 +33,8 @@ import (
 
 type vxlanCleanup struct {
 	event.HandlerBase
+	mutex               sync.Mutex
```
tpantelis: This isn't necessary as the event controller ensures single-thread access.

tpantelis (changes requested):

pkg/routeagent_driver/cabledriver/vxlan_tunnel_cleanup_handler.go
```diff
@@ -48,10 +52,34 @@ func (h *vxlanCleanup) GetName() string {
...
 	errv4 := netlink.DeleteIfaceAndAssociatedRoutes(vxlan.GetVxlanInterfaceName(k8snet.IPv4), vxlan.TableID, k8snet.IPv4)
 
 	return errors.Join(errv6, errv4)
 }
+
+func (h *vxlanCleanup) LocalEndpointRemoved(endpoint *submv1.Endpoint) error {
+	h.mutex.Lock()
+	defer h.mutex.Unlock()
+
+	h.lastRemovedEndpoint = endpoint
```
tpantelis, line 74: So this is called by the event controller right before it calls `TransitionToNonGateway` in the same code path. Storing `lastRemovedEndpoint` here would be unnecessary if the endpoint was passed to `TransitionToNonGateway`. I'll look into that for a future PR.

tpantelis (approved):

pkg/routeagent_driver/cabledriver/vxlan_tunnel_cleanup_handler.go
yboaron: Yep, it would be nicer to get the removed endpoint as parameter in TransitionToNonGateway

PR #3973, opened 2026-04-15
Fix submarinerManagedLabel cleanup in restore for Calico handler

tpantelis:
The `submarinerManagedLabel` was written to `installation.Annotations` but restore removed it from `installation.Labels`, leaving the annotation behind. Now correctly delete from `Annotations`.

Added test coverage for verifying annotations are properly set and removed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Switched encapsulation checks to use the Tigera operator Installation API, ensuring default encapsulation is treated as "None" and improving reliability during install/uninstall flows.
  * ROKS restore/uninstall now correctly updates and removes Submariner-related annotations from the cluster installation to avoid stale config.

* **Refactor**
  * Annotation identifiers made public for clearer integration across components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/submariner-io/submariner/pull/3973)

### DIFF
--- a/pkg/routeagent_driver/handlers/calico/calico_suite_test.go
+++ b/pkg/routeagent_driver/handlers/calico/calico_suite_test.go
@@ -63,7 +63,7 @@ func newTestDriver() *testDriver {
 				CalicoNetwork: &tigerav1.CalicoNetworkSpec{
 					IPPools: []tigerav1.IPPool{
 						{
-							Encapsulation: tigerav1.EncapsulationIPIPCrossSubnet,
+							Encapsulation: tigerav1.EncapsulationNone,
 						},
 					},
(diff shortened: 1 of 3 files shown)


yboaron:
LGTM, however, this logic was originally implemented to address a Calico-specific issue on IBM ROKS, IIRC,  the solution had two pieces: the Submariner side and the ROKS side, ROKS relying on specific annotations and labels set by Submariner.

I know that this change only address removal of orphan label, but to be on the safe side,  maybe we should verify that Submariner on ROKS remains functional with this PR. 

@Tamas-Biro1 @mihivagyok, could you please help here?

skitt (approved):

PR #3979, opened 2026-04-16
Fix IPv6 protocol conflict in nftables rules

### LINKED ISSUE #3978: Dual-stack (IPv4/IPv6) Submariner v0.23.1 Packetfilter Protocol Conflict Issue & IPv6-only Support Inquiry
1.In a dual-stack (IPv4 + IPv6) Kubernetes environment with Cilium (this issue also occurs in Calico dual-stack deployments), when running Submariner with subctl version v0.23.1, the packetfilter module of Submariner encounters a protocol type conflict when adding iptables/nftables rules (as shown in the attached screenshot). Is this a known bug? If it is not a bug, could you please guide me on how to fully resolve this issue?

2.Additionally, I would like to ask: Are there any development plans to support Submariner running in a single-stack IPv6-only environment in the future?

Image

utafrali:
Fixes #3978

Was hitting an issue where IPv6 rules broke due to the generic RuleActionSelfSNAT entry being applied to all IP families. It needs per-family handling, so removed it from the shared map. Cleaned up the indentation while I was at it.

Added IPv6 tests to verify the protocol headers work correctly - UDP gets ip6 nexthdr, ICMP becomes ICMPv6, and source address selection handles IPv6 properly.

## Summary by CodeRabbit

* **New Features**
  * Added IPv6 support for packet-filter rules with IPv6-specific protocol matching and address-family syntax.

* **Bug Fixes**
  * Fixed self-SNAT rendering so SNAT rules emit correct IPv4/IPv6 address-family clauses.

* **Tests**
  * Added IPv6-focused tests verifying UDP, ICMPv6, and self-SNAT rules render with correct IPv6 syntax.

* **Chores**
  * CI updated to include a dual-stack nftables test combination.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -60,6 +60,7 @@ jobs:
           - extra-toggles: dual-stack, globalnet
           - extra-toggles: ipv6-stack
           - extra-toggles: nftables
+          - extra-toggles: nftables, dual-stack
           - extra-toggles: nftables, ovn
           - extra-toggles: nftables, globalnet
           - extra-toggles: nftables, globalnet, ovn
--- a/pkg/packetfilter/nftables/nftables.go
+++ b/pkg/packetfilter/nftables/nftables.go
@@ -58,13 +58,12 @@ var (
 	}
 
 	ruleActionToStr = map[packetfilter.RuleAction][]string{
-		packetfilter.RuleActionAccept:   {"accept"},
-		packetfilter.RuleActionMss:      {"tcp", "option", "maxseg"},
-		packetfilter.RuleActionMark:     {"meta", "mark"},
-		packetfilter.RuleActionSNAT:     {"snat"},
-		packetfilter.RuleActionDNAT:     {"dnat"},
-		packetfilter.RuleActionJump:     {"jump"},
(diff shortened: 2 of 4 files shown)


utafrali:
Done, pushed the fix.

tpantelis (changes requested):
Thanks for the PR!

The DCO job failed - you need to sign your commit.

Also, let's add `- extra-toggles: nftables, dual-stack` to the E2E workflow matrix in the following file so this is tested in CI. 

https://github.com/submariner-io/submariner/blob/e8b7fadb9ece3212cbd6483db96cb659454bb739/.github/workflows/e2e.yml#L62-L64

Also, please address Code Rabbit's [nitpick comments](https://github.com/submariner-io/submariner/pull/3979#pullrequestreview-4118353843).

utafrali:
Done, pushed the fix.

tpantelis:
> Done, pushed the fix.

I see you addressed one nitpick comment but, more important is to modify the E2E test matrix as I outlined.

Also your commit still needs to be signed (`git commit -s`).

Also, when addressing review comments, we amend the original commit rather than creating a new one to keep the git history clean.

utafrali:
Done, pushed the fix.

tpantelis (approved):

PR #4080, opened 2026-06-28
Add input validation for Libreswan endpoint configuration

yboaron:
Add validation for cable_name and subnet fields to ensure they contain only safe 
characters before using them in IPsec configuration. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stricter validation for endpoint names, subnets, and IP addresses before connection setup.
  * Invalid inputs now fail fast with clear errors, helping prevent malformed connection configurations.
  * Improved safeguards to reduce the risk of malformed or injected VPN configuration during connection creation.


tpantelis (approved):

tpantelis:
Could you add unit tests to cover the new code, [here](https://github.com/submariner-io/submariner/blob/5f185be526895de792371ae6a6b81b3464760bec/pkg/cable/libreswan/libreswan_test.go#L311) looks like a good place?

PR #4119, opened 2026-08-01
Fix AmneziaWG gateway startup failure

### LINKED ISSUE #4118: Failed to configure AmneziaWG device
These 2 jobs:

https://github.com/submariner-io/submariner/actions/runs/30710975231/job/91398797024?pr=4109#step:4:2063
https://github.com/submariner-io/submariner/actions/runs/30710975231/job/91398797005?pr=4109#step:4:2044

failed with error:

`2026-08-01T17:55:36.912Z FTL ../submariner/main.go:175 main                 Error running the gateway error="error starting the cable engine: error creating the cable driver: failed to configure AmneziaWG device: read: wguser: errno=-22"`

pujitha24:
Motivation:
The AmneziaWG cable driver's default obfuscation options never
configure successfully against the vendored
amnezia-vpn/amneziawg-go v1.0.4 userspace device: E2E jobs failed
immediately after gateway startup with "failed to configure
AmneziaWG device: read: wguser: errno=-22".

Approach:
The vendored device only implements a subset of the AmneziaWG
protocol: magic headers (h1-h4) are single uint32 values, not
"min-max" ranges, and there is no s3 (cookie) / s4 (transport)
padding support at all. The driver's shipped defaults sent both an
"s3"/"s4" key the device rejects outright and h1-h4 as ranges, so
the very first UAPI "set" operation always failed with EINVAL,
regardless of any cableDriverOptions override.

Remove the s3/s4 options (now rejected as an unknown option instead
of a silent runtime EINVAL if ever set) and change h1-h4 defaults to
single uint32 values, matching what the real device parses.
Simplify header validation to match: drop the range parsing/overlap
checks and replace them with a plain "no two magic headers may
share a value" check, mirroring the device's own uniqueness check.

Validation:
Reproduced the failure and confirmed the fix by driving the actual
vendored github.com/amnezia-vpn/amneziawg-go device (via its
tun/tuntest in-memory TUN, no root or kernel device needed) with
the driver's real UAPI wire output: the original defaults failed
with "IPC error -22: invalid UAPI device key: s3", a range-format
h1 failed with "IPC error -22: ... invalid syntax", and the new
defaults succeeded with no error. Added protocol_test.go so this is
checked against the real device going forward, instead of only the
fakeClient the rest of this package's tests use (which accepts any
string and did not catch this).

Ran, all succeeding:
  GOOS=linux GOARCH=amd64 go build ./...
  GOOS=linux GOARCH=amd64 go vet ./pkg/cable/amneziawg/...
  GOOS=linux GOARCH=amd64 go test -c ./pkg/cable/amneziawg/...
  GOOS=linux GOARCH=arm64 go build ./pkg/cable/amneziawg/...
Could not execute `go test` or `make unit` in this sandbox: this
repo requires Linux, the sandbox is macOS, and
github.com/vishvananda/netlink does not build on darwin at all -
confirmed pre-existing and unrelated by reproducing the identical
failure on an unmodified checkout.

Impact: this only affects AmneziaWG gateway startup. Every fresh
AmneziaWG cable driver instance failed to start, with or without a
cableDriverOptions override; no data-plane traffic or already
established tunnel is affected.

Report: https://github.com/submariner-io/submariner/issues/4118
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #4118

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated AmneziaWG obfuscation settings to match the supported driver format.
  * Replaced header ranges with fixed numeric values and removed unsupported S3/S4 options.
  * Improved validation to reject duplicate header values and invalid configurations.
* **Tests**
  * Added coverage confirming generated default configurations are accepted by the protocol parser.

### DIFF
--- a/pkg/cable/amneziawg/amneziawg_test.go
+++ b/pkg/cable/amneziawg/amneziawg_test.go
@@ -158,10 +158,8 @@ func testNewDriver() {
 		Expect(device.Jmax).To(Equal(200))
 		Expect(device.S1).To(Equal(45))
 		Expect(device.S2).To(Equal(60))
-		Expect(device.S3).To(Equal(35))
-		Expect(device.S4).To(Equal(12))
-		Expect(device.H1).To(Equal("200000000-280000000"))
-		Expect(device.H2).To(Equal("400000000-480000000"))
+		Expect(device.H1).To(Equal("240000000"))
+		Expect(device.H2).To(Equal("440000000"))
 		Expect(device.I1).To(Equal("<r 40>"))
 		Expect(device.Peers).To(BeEmpty(), "ReplacePeers should clear the pre-seeded peer")
 
@@ -171,7 +169,7 @@ func testNewDriver() {
 
 	When("cable driver options are set", func() {
 		BeforeEach(func() {
-			os.Setenv(cable.CableDriverOptionsEnv, `{"jc":"3","h1":"10-20","i1":"<b 0x64>"}`)
+			os.Setenv(cable.CableDriverOptionsEnv, `{"jc":"3","h1":"300000000","i1":"<b 0x64>"}`)
 		})
 
 		AfterEach(func() {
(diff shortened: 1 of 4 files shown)


tpantelis:
Superseded by https://github.com/submariner-io/submariner/pull/4120

PR #4131, opened 2026-08-12
Fix get nexthop IP address for OVN mgmt interface

### LINKED ISSUE #4121: Gateway health checks sometimes fail after ROKS node shutdown/reboot
# Gateway health checks sometimes fail after ROKS node shutdown/reboot

## Symptom
After ROKS node shutdown/reboot in OVN-K Interconnect environment:
- Gateway CR: `status=error` (health checks fail)
- RouteAgent CR: `status=connected` (proves datapath works)

## Root Cause
Table 150 route configured with **network address** (.0) instead of gateway IP.

**Cluster1 gateway node (after reboot) - BROKEN:**
```
# ip route show table 150
default via 172.22.0.0 dev ovn-k8s-mp0
```

**Cluster2 gateway node (not rebooted) - WORKING:**
```
# ip route show table 150
default via 172.17.0.1 dev ovn-k8s-mp0
```

**Difference:** Cluster1 uses network address `172.22.0.0` (ending in `.0`), while Cluster2 uses valid gateway IP `172.17.0.1`.

Main routing table on affected node shows the correct gateway exists:
```
172.22.0.0/17 via 172.22.1.1 dev ovn-k8s-mp0
```
The gateway IP `172.22.1.1` should have been used for table 150, not the network address `172.22.0.0`.

## Suspected Code
`pkg/routeagent_driver/handlers/ovn/host_networking.go` function `getNextHopOnK8sMgmtIntf()`

Appears to use first-match-wins without validating gateway is host address (not network address `.0`).

## Workaround
```bash
ip route del default table 150
ip route add default via 172.22.1.1 dev ovn-k8s-mp0 table 150
```

## Environment
- Platform: ROKS (IBM Cloud)
- CNI: OVN-Kubernetes Interconnect
- Trigger: Cloud-managed node shutdown/reboot

---
Diagnostic data available if needed.

yboaron:
The getNextHopOnK8sMgmtIntf() function would return Dst.IP when no explicit gateway was set on a route. However, Dst.IP is the network address (the base of the CIDR), not a valid host IP.

This PR adds validation to skip routes where Dst.IP equals the network address (IP & Mask == IP). This ensures only valid host IPs are used as gateways.

Also added test case to verify network addresses are rejected and only valid host addresses are used as gateways.

Fixes: #4121

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed route selection to avoid using a network address as a gateway when no explicit gateway is configured.
  * Preserved valid gateway-less host routes during host-network reconciliation.
  * Ensured routes with valid gateways are preferred during route selection.
  * Improved host-network routing reliability when multiple routes are available.

* **Tests**
  * Added coverage for network-address routes, host routes, and gateway selection scenarios.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/handler_test.go
+++ b/pkg/routeagent_driver/handlers/ovn/handler_test.go
@@ -818,6 +818,155 @@ func (t *handlerTestDriver) testOVNMgmtInterfaceAddressChange() {
 		t.netLink.AwaitGwRoutes(0, constants.RouteAgentInterClusterNetworkTableID, t.OVNK8sMgmntIntGw)
 		t.netLink.AwaitGwRoutes(0, constants.RouteAgentHostNetworkTableID, t.OVNK8sMgmntIntGw)
 	})
+
+	It("should skip network address routes and use valid gateway when route order changes", func(ctx context.Context) {
+		// This reproduces issue #4121 where after ROKS node reboot, route ordering changes
+		// and Submariner picks the wrong route (network address instead of valid gateway)
+		//
+		// Scenario:
+		// - Multiple routes exist for the same destination (cluster CIDR)
+		// - One route has Gw=nil (would use Dst.IP = network address like 171.0.1.0)
+		// - Another route has valid Gw=171.0.1.1
+		// - After reboot, the bad route (Gw=nil) comes FIRST in iteration order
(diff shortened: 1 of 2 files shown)


yboaron:
cc @tpantelis

tpantelis (approved):

PR #4189, opened 2026-09-09
Add egress self-SNAT workaround for OVN-K local gateway mode

yboaron:
Problem:
When OVN-K is configured in local gateway mode, Submariner gateway
health checks fail, causing tunnel connections to remain in an error state.

The failure path is:
1. OVN-K's nftables rule SNATs the health check ICMP packet to an IP outside
   the local CIDR before packet encapsulation in IPsec.
2. The remote gateway receives the health check packet but cannot route the
   reply back through the tunnel due to the unroutable source IP.
3. Health check replies never arrive, leaving the tunnel in an error state.

Fix:
Add a nftables SNAT rule (SUBMARINER-SELF-SNAT chain) in the route-agent's
OVN dataplane handler. Executing before OVN-K's SNAT chains, this rule
creates a conntrack entry that preserves the original source IP of health
check packets. Once established, OVN-K's SNAT rules cannot override the
source IP, allowing health check packets to reach the remote gateway with
a routable source address.

The workaround is automatically enabled when OVN-K local gateway mode is
detected via node annotation. 
It is a no-op in shared gateway mode where the problem does not occur.

If needed, the workaround can be explicitly disabled via the
submariner-global ConfigMap (requires route-agent pod restart):

  kubectl patch configmap submariner-global -n submariner-operator \
    --type merge -p '{"data":{"disable-ovn-selfsnat":"true"}}'

This workaround serves as a temporary mitigation until the underlying
OVN-K issue is resolved upstream (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6811).

-->

## Summary by CodeRabbit

* **New Features**
  * Added optional OVN self-SNAT support for gateway nodes, controlled through the GlobalConfigMap.
  * Supports both IPv4 and IPv6 traffic with the appropriate remote CIDR sets.
  * Uses local cluster CIDRs when generating self-SNAT rules.

* **Bug Fixes**
  * Self-SNAT rules are limited to gateway nodes and omitted when disabled or unconfigured.
  * Uninstallation now removes associated self-SNAT networking rules.

### DIFF
--- a/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
+++ b/pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
@@ -26,15 +26,18 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/global"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/slices"
 	"github.com/submariner-io/admiral/pkg/util"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cidr"
 	nodeutil "github.com/submariner-io/submariner/pkg/node"
 	"github.com/submariner-io/submariner/pkg/packetfilter"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/chains"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
+	"github.com/submariner-io/submariner/pkg/routeagent_driver/handlers/mtu"
 	"github.com/vishvananda/netlink"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


pkg/routeagent_driver/handlers/ovn/gateway_dataplane.go
```diff
@@ -319,3 +330,74 @@ func (ovn *Handler) nodeResourceInterface() resource.Interface[*corev1.Node] {
...
 	}
 }
+
+const enableOVNSelfSNATKey = "enable-ovn-selfsnat"
+
+func (ovn *Handler) isOVNSelfSNATEnabled() bool {
+	enableSelfSNAT := false
+
+	if ovn.GlobalConfigMap != nil {
+		if value, ok := ovn.GlobalConfigMap.Data[enableOVNSelfSNATKey]; ok {
```
tpantelis: This should use admiral's `global.Get` rather than propagating and accessing the  global `ConfigMao` directly. `main` sets up the global config via `global.Init`.

tpantelis (changes requested):

yboaron:
@coderabbitai resolve
     
  The self-SNAT workaround is intentionally nftables-only and does not need
  an iptables implementation.
     
  The reason this workaround is needed at all is specific to how nftables
  applies rules: unlike iptables, which stops processing at the first ACCEPT
  rule, nftables continues evaluating all rules in a hook regardless of
  whether a higher-priority rule already accepted the packet. This means
  OVN-K's SNAT rules still fire on packets that Submariner's higher-priority
  rules have already accepted, causing unintended SNAT on health-check
  traffic. The self-SNAT rule is the workaround for this nftables behaviour.
  
  In the iptables case this problem does not arise — iptables stops at the
  first matching ACCEPT, so OVN-K's SNAT rules never fire on packets already
  accepted by Submariner. Silently no-oping for iptables is therefore the
  correct behaviour.

tpantelis (approved):

PR #4190, opened 2026-09-09
Allow LB IP resolution in air-gapped deployments

### LINKED ISSUE #4108: Enable LoadBalancer Public IP Resolution in Air-Gapped Mode
Submariner currently blocks all LoadBalancer public IP resolution when airGapped: true is set, even when the LoadBalancer has a direct IP that doesn't require external DNS query.
    
This may break legitimate air-gapped deployments using  MetalLB LB provider,
in this case LoadBalancers assign IPs directly but are still blocked.
  
Expected Behavior : 
  
LoadBalancer resolution should work in air-gapped mode when ingress.IP is set directly:
  
  status:
    loadBalancer:
      ingress:
      - ip: 10.48.120.52  
  
  Only block when ingress.Hostname is set
  

Note: manually annotating gateway nodes with the LoadBalancer VIP and restarting gateway pods can currently be used to workaround this issue
  
```
  Step 1: Get the LoadBalancer IP:
  oc get svc submariner-gateway -n submariner-operator \
    -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
  # Output: 10.48.120.52
  
  Step 2: Annotate the gateway node with the static IP:
  oc annotate node <gateway-node> \
    gateway.submariner.io/public-ip=ipv4:10.48.120.52 --overwrite
  
  Step 3: Restart the gateway pod to pick up the annotation:
  oc delete pod -n submariner-operator -l app=submariner-gateway

```

yboaron:
In air-gapped mode, LoadBalancer public IP resolution is currently blocked entirely to prevent DNS lookups, which may not be available in such environments. This is overly restrictive: when the LB provider sets a direct IP address in the service's status.loadBalancer.ingress[].ip field, no DNS lookup is needed.

Providers like MetalLB assign IP addresses directly to LoadBalancer services rather than hostnames, yet their IP addresses cannot be resolved in air-gapped mode due to this restriction.

Add support for LoadBalancer public IP resolution in air-gapped mode when status.loadBalancer.ingress[].ip is populated. Ingress entries that carry only a hostname are still skipped, since resolving a hostname requires DNS. If both ip and hostname are present in the ingress status, the direct IP is used.

Closes: #4108

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved public IP resolution in air-gapped environments by checking available IPv4 and IPv6 sources before LoadBalancer ingress.
  - Direct IP addresses from LoadBalancer ingress are now detected without requiring DNS resolution.
  - Hostname-only LoadBalancer entries no longer trigger unavailable DNS lookups or unnecessary retries in air-gapped environments.
  - Mixed ingress configurations now use available direct IP addresses while avoiding hostname resolution when DNS is unavailable.

### DIFF
--- a/pkg/endpoint/public_ip.go
+++ b/pkg/endpoint/public_ip.go
@@ -35,6 +35,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/resource"
 	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/types"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
@@ -120,14 +121,10 @@ func GetPublicIP(ctx context.Context, family k8snet.IPFamily, submSpec *types.Su
 		}
 
 		if airGapped {
-			ip, resolver, err := invokeResolvers(ctx, family, k8sClient, submSpec.Namespace, config, func(method string) bool {
-				return method == v1.IPv4 || method == v1.IPv6
-			})
+			ip, resolver, err := resolvePublicIPAirGapped(ctx, family, k8sClient, submSpec.Namespace, config)
 			if err != nil {
 				logger.Errorf(err, "Unable to resolve public IPv%s in an air-gapped deployment using %q - using empty value",
 					family, config)
-
-				return "", "", nil
 			}
 
 			return ip, resolver, nil
@@ -140,8 +137,48 @@ func GetPublicIP(ctx context.Context, family k8snet.IPFamily, submSpec *types.Su
 	return "", "", nil
 }
 
+func resolvePublicIPAirGapped(ctx context.Context, family k8snet.IPFamily, k8sClient kubernetes.Interface, namespace, config string,
+) (string, string, error) {
+	var errs []error
+
+	ip, resolver, err := invokeResolvers(ctx, family, k8sClient, namespace, config, func(method string) bool {
+		return method == v1.IPv4 || method == v1.IPv6
+	})
+	if ip != "" {
+		return ip, resolver, nil
+	}
+
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	for r := range strings.SplitSeq(config, ",") {
+		method, param, parseErr := parseResolver(r)
+		if parseErr != nil {
+			errs = append(errs, parseErr)
+			continue
+		}
+
+		if method != v1.LoadBalancer {
+			continue
+		}
+
+		ip, lbErr := publicLoadBalancerDirectIP(ctx, family, k8sClient, namespace, param)
+		if lbErr == nil && ip != "" {
+			return ip, r, nil
+		}
+
+		if lbErr != nil {
+			errs = append(errs, errors.Wrapf(lbErr, "\nResolver[%q]", r))
+		}
+	}
+
+	return "", "", goerrors.Join(errs...)
+}
+
 func invokeResolvers(ctx context.Context,
-	family k8snet.IPFamily, k8sClient kubernetes.Interface, namespace, config string, useResolver func(string) bool,
+	family k8snet.IPFamily, k8sClient kubernetes.Interface, namespace, config string,
+	useResolver func(string) bool,
 ) (string, string, error) {
 	resolvers := strings.Split(config, ",")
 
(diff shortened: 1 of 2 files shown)


pkg/endpoint/public_ip.go
```diff
@@ -140,8 +130,40 @@ func GetPublicIP(ctx context.Context, family k8snet.IPFamily, submSpec *types.Su
...
+		return method == v1.IPv4 || method == v1.IPv6
+	})
+	if ip != "" {
+		return ip, resolver, nil
+	}
+
+	// Try lb: resolvers using direct IP only — skips hostname DNS resolution.
+	for r := range strings.SplitSeq(config, ",") {
+		method, param, parseErr := parseResolver(r)
+		if parseErr != nil || method != v1.LoadBalancer {
```
tpantelis: It seems we should return `parseErr`, but, if not. at least log it.

pkg/endpoint/public_ip.go
```diff
@@ -249,25 +286,47 @@ func publicLoadBalancerIP(ctx context.Context, family k8snet.IPFamily, clientset
...
+			}
+
+			if ip != "" {
+				resolvedIP = ip
+				return true, nil
+			}
+		}
+
+		// Hostname-only service with no DNS resolver (air-gapped): stop retrying immediately.
+		if hasHostname && hostnameResolver == nil {
```
tpantelis: `hasHostname` is only set to true if `hostnameResolver == nil`. Since hostnameResolver is passed in, it seems we should just skip the loop on L300 entirely if `hostnameResolver == nil`, ie

```
if hostnameResolver == nil {
    logger.Warningf("Skipping hostname ingress %q for service %q: ...)
    return true, nil
}

for _, ingress := range service.Status.LoadBalancer.Ingress {
    if ingress.Hostname == "" {
	    continue
	}

    ip, err := hostnameResolver(ctx, family, clientset, namespace, ingress.Hostname)
    ...
}
```

pkg/endpoint/public_ip.go
```diff
@@ -140,8 +130,40 @@ func GetPublicIP(ctx context.Context, family k8snet.IPFamily, submSpec *types.Su
...
+
+	// Try lb: resolvers using direct IP only — skips hostname DNS resolution.
+	for r := range strings.SplitSeq(config, ",") {
+		method, param, parseErr := parseResolver(r)
+		if parseErr != nil || method != v1.LoadBalancer {
+			continue
+		}
+
+		ip, lbErr := publicLoadBalancerDirectIP(ctx, family, k8sClient, namespace, param)
+		if lbErr == nil && ip != "" {
```
tpantelis, line 167: Same as above wrt `lbErr`.

pkg/endpoint/public_ip.go
```diff
@@ -140,8 +130,40 @@ func GetPublicIP(ctx context.Context, family k8snet.IPFamily, submSpec *types.Su
...
 
+func resolvePublicIPAirGapped(ctx context.Context, family k8snet.IPFamily, k8sClient kubernetes.Interface, namespace, config string,
+) (string, string, error) {
+	// Try ipv4:/ipv6: resolvers first — no network access required.
+	ip, resolver, err := invokeResolvers(ctx, family, k8sClient, namespace, config, func(method string) bool {
+		return method == v1.IPv4 || method == v1.IPv6
+	})
+	if ip != "" {
+		return ip, resolver, nil
+	}
```
tpantelis, line 149: We're ignoring the `err` at this point and falling thru to check direct IP only below. Can you explain the logic?

pkg/endpoint/public_ip.go
```diff
@@ -140,8 +130,40 @@ func GetPublicIP(ctx context.Context, family k8snet.IPFamily, submSpec *types.Su
 	return "", "", nil
 }
 
+func resolvePublicIPAirGapped(ctx context.Context, family k8snet.IPFamily, k8sClient kubernetes.Interface, namespace, config string,
+) (string, string, error) {
```
tpantelis, line 141: The function signature declares returns `error` but only `nil` is ever returned. We should either remove `error` or some paths should return an error.

tpantelis (changes requested):

pkg/endpoint/public_ip.go
```diff
@@ -140,8 +137,48 @@ func GetPublicIP(ctx context.Context, family k8snet.IPFamily, submSpec *types.Su
...
+
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	for r := range strings.SplitSeq(config, ",") {
+		method, param, parseErr := parseResolver(r)
+		if parseErr != nil {
+			errs = append(errs, parseErr)
+			continue
```
tpantelis, line 159: If a subsequent resolver succeeds then a prior parse error is silently dropped. So, in the end, an IP is obtained but the user doesn't know about the malformed entry. This pattern also exists with `invokeResolvers` so it's not new. This is an edge case worth noting but doesn't need to be addressed with this PR.

tpantelis (approved):

PR #4191, opened 2026-09-10
Pin release-images action to a reviewed commit

dfarrell07:
#4143 merged before this fix could land — recreating it here.

Granting `id-token: write` lets the `release-images` action mint Sigstore OIDC tokens and access the Quay credentials passed to it. Referencing shipyard's mutable `release-0.19` branch let an unreviewed change to that action obtain the same access (flagged by coderabbitai on #4138, the equivalent fix for release-0.24). Pin to `release-0.19`'s current commit instead.

https://claude.ai/code/session_01RmkKzKd3ZRYRNtUwv45mZN

## Summary by CodeRabbit

* **Chores**
  * Updated the release workflow to use a fixed version of the image release action, improving release consistency and reliability.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
           fetch-depth: 0
 
       - name: Build and release new images
-        uses: submariner-io/shipyard/gh-actions/release-images@release-0.19
+        uses: submariner-io/shipyard/gh-actions/release-images@6156e54b88aad8342859250e1e80856ebd5d069b  # release-0.19
         with:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}


skitt (changes requested):
CodeRabbit’s complaint only makes sense for third-party actions we don’t control, but that’s not the case here. Any change to Shipyard would be reviewed by the project, and anyone that can move the tag on the action can also directly manipulate this project.

dfarrell07:
@skitt That's a fair point on the threat model — for a project-owned action, moving the tag and pushing directly to the consuming repo require the same level of access. However, the identical SHA pin was included in #4138 (for release-0.24) and merged. If the project position is to revert the pin here, should #4138 also be reverted to use the branch ref? Happy to either revert this PR or keep it consistent with what landed in #4138, whichever you prefer.

skitt:
The question is what the balance is in terms of overall effort between tracking the branch and pinning to hashes. We’ve established that there’s no security benefit; but tracking the branch means that any breakage in Shipyard immediately affects all dependents, but equivalently, any fix in Shipyard immediately benefits all dependents. Pinning to a hash means that any fix needs a bump PR in the affected project.

Let’s revert and go back to branches. I’ll argue with CodeRabbitAI if necessary.

tpantelis:
> Let’s revert and go back to branches. I’ll argue with CodeRabbitAI if necessary.

I agree.

